PR #40, opened 2015-11-03
avoiding mention of key, since we don't explain it or use it

sethladd:
Trying to keep things simple.


sethladd:
cc @krisgiesing

krisgiesing:
LGTM

sethladd:
thanks for the quick review!

PR #47, opened 2015-11-05
normalize the docs nav menus

devoncarew:
- make the mav menu and the first page table of contents agree
- make the css for the first page and the rest of the pages the same

@sethladd

### DIFF
--- a/index.md
+++ b/index.md
@@ -1,17 +1,20 @@
 ---
-layout: default
+layout: page
+title: Flutter
 ---
-Flutter
-=======
-
 Flutter is a new way to build high-performance, cross-platform mobile apps.
-Flutter is optimized for today's, and tomorrow's, mobile devices. We are
-focused on low-latency input and high frame rates on Android and iOS.
+Flutter is optimized for today's, and tomorrow's, mobile devices. We are focused
+on low-latency input and high frame rates on Android and iOS.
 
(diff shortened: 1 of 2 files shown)


index.md
```diff
@@ -1,17 +1,20 @@
...
 Flutter is a new way to build high-performance, cross-platform mobile apps.
-Flutter is optimized for today's, and tomorrow's, mobile devices. We are
-focused on low-latency input and high frame rates on Android and iOS.
+Flutter is optimized for today's, and tomorrow's, mobile devices. We are focused
+on low-latency input and high frame rates on Android and iOS.
 
-Next steps:
+## Next steps
 
+ - [FAQ](faq)
```
sethladd: Not sure that FAQ is a next step. :)

sethladd:
Thanks! Since we're taking a look at this, not sure FAQ should be the first item in the top menu.

index.md
devoncarew: It looked odd to me that this list was out of sync with the mav menu. I could not find a way to re-order the nav items - I had wanted to move FAQ to near the end of the menu.

I'll push this down the list, or move to the see also section.

devoncarew:
Updated!

devoncarew:
(updated FAQ position; will land)

PR #57, opened 2015-11-06
be clear we are tech preview

sethladd:


sethladd:
cc @Hixie

Hixie:
"Technology preview" implies that it was hidden from view then shown. How about "is an early-stage open-source project. We are still missing core features like accessibility, text input, localisation, ..."?

sethladd:
> implies that it was hidden from view then shown

Interesting. That's not the semantics that Google uses for "tech preview". But, if it's confusing, we'll change it.

sethladd:
ptal thanks!

abarth:
LGTM

PR #60, opened 2015-11-06
Fix markdown in FAQ

sethladd:


sethladd:
cc @Hixie

sethladd:
@abarth fixed this.

PR #65, opened 2015-11-06
be more clear in homepage

sethladd:


sethladd:
TBR

index.md
```diff
@@ -2,15 +2,16 @@
 layout: page
 title: Flutter
 ---
-Flutter is a new way to build high-performance, cross-platform mobile apps.
-Flutter is optimized for today's, and tomorrow's, mobile devices. We are focused
-on low-latency input and high frame rates on Android and iOS.
+
+Flutter is a new project to help developers
+build high-performance, high-fidelity,
+cross-platform mobile apps.
```
abarth: This text exists in several places.  Do we want to update it everywhere?

https://github.com/flutter/engine/blob/master/README.md
https://github.com/flutter/flutter/blob/master/README.md
https://github.com/flutter/engine/blob/master/sky/packages/sky/README.md

sethladd: Ah, good point!

I was trying to simplify the tag line here. Open to suggestions. (e.g. we should add "iOS and Android" back in), but would like to try to keep it down to one sentence.

I'm happy to update everywhere if you like the new phrasing with the one addition.

abarth: SGTM

Issue #42, opened 2015-11-04
List pre-reqs up top in Getting Started


sethladd:
e.g. do I need a phone? What kind, and version(s)? What OS and what version(s)? Any cables? Do I need other software like XCode?

maybe even show a picture to illustrate the setup.

abarth:
Yeah, good point.  At this point you need an Android phone, a Mac or Linux host.  A USB cable to connect them.

sethladd:
Happy to make the addition. Will send PR

FlutterIssues:
This issue was moved to flutter/flutter#1448

PR #75, opened 2015-11-08
Added Prereqs to the getting started page

Explouad:
This solves #42.

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -5,8 +5,13 @@ nav_title: Getting Started
 permalink: /getting-started/
 ---
 
-Flutter currently supports developers on Mac and Linux.
-Windows support is in the works.
+## Prereqs
+
+- An Android phone.
+- Mac or Linux host computer. (Windows support is in the works.)
+- A USB cable to connect the two.
+
+---
 
 To get started, we need to set up the Dart SDK.
 Flutter requires Dart SDK 1.12.2 or greater.


Explouad:
I signed the CLA.

abarth:
Thanks for the patch.  I think we want to be more specific about exactly what kind of phone (e.g., exactly what versions of Android do we support, do we support Android derivatives like Fire phone or Cyanogen), exactly what versions of Linux and Mac we support, etc.  Also, hopefully we can improve the visual presentation of the information so that it feels like an organic part of the document instead of a header added to the beginning.

Explouad:
That makes sense. I do not have this information, but if it could be given
to me (or I could be directed to it, either one) I will happily change the
commit to reflect it. Also, the section seems organic (minus the line
separator), what do you suggest?

abarth:
> I do not have this information, but if it could be given to me (or I could be directed to it, either one) I will happily change the commit to reflect it.

Some of it is that we haven't figured it all out yet.  This (very recent) patch is us working out which versions of Android we actually support: https://github.com/flutter/flutter/pull/52.  We need to so some similar experimentation with the host machines.

For Mac, our stuff seems to work with Yosemite.  Not sure about older or newer versions.  For Linux, there appears to be a minimum glibc that we need, but we haven't chased down exactly what it is or which Linux distributions have it.

We need to chase all these things down in order to really fix #42.

Explouad:
Understood. We'll just keep this open and as the info comes in we'll update accordingly.

sethladd:
Hi! Thanks for the contribution.

A lot has changed since you made this PR. I'm going to close this, because we're doing some housecleaning and moving issues around, but if you'd like to take another crack at it and addressing comments, please feel free to open a new PR.

PR #77, opened 2015-11-09
Small tweaks to Getting Started

sethladd:
More headers to pace through the tutorial. Changed a few "we's" to "you". Clarify where the new starter app will be created. Changed header markup to `##` (making it a bit easier to type, we don't need to match the length of the text now).


sethladd:
cc @abarth are you OK with the markup change?

abarth:
LGTM

sethladd:
thanks for the quick review!

PR #82, opened 2015-11-12
Don't link directly into engine repo

sethladd:


sethladd:
@abarth I assume that engine isn't the primary entry point for most folks?

We could link into flutter/flutter or just github.com/flutter

index.md
```diff
@@ -23,5 +23,5 @@ today. We hope you try it out and send us
 ## See also
 
  - [Architecture diagram](https://docs.google.com/presentation/d/1cw7A4HbvM_Abv320rVgPVGiUP2msVs7tfGbkgdrTy0I/edit?usp=sharing)
- - [github.com/flutter](https://github.com/flutter/engine/)
+ - [github.com/flutter](https://github.com/flutter)
```
abarth: We should probably link to `https://github.com/flutter/flutter`.  That's the main repo.  It has the issue tracker and the code that most folks will be interested in.

sethladd: can do!

sethladd:
ptal

abarth:
LGTM

PR #91, opened 2015-11-20
Add missing parameter to itemBuilder callback

Bema:
The ItemBuilder function-type alias specifies three arguments (BuildContext context, T item, int index). The proposed change adds the missing parameter which was causing the warning bellow and preventing app from running.

```
The argument type '(BuildContext, Product) → dynamic' cannot be assigned to the parameter type 'ItemBuilder'
```

### DIFF
--- a/tutorial.md
+++ b/tutorial.md
@@ -469,7 +469,7 @@ class _ShoppingListState extends State<ShoppingList> {
       body: new MaterialList<Product>(
         type: MaterialListType.oneLineWithAvatar,
         items: config.products,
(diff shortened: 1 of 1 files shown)


Bema:
I signed it!

abarth:
LGTM

The bot is turning red for the same reason it didn't detect this problem: it's using an old version of the Flutter package.  Let's land this patch and then we can work on fixing the bot.

Thanks!!

Bema:
Thanks @abarth!

PR #96, opened 2015-11-20
Switch Travis to using the alpha version of Flutter

abarth:
Previously, we were using the version on pub.dartlang.org.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,5 @@ language: dart
 dart:
   - dev
 sudo: false
+before_install: ./tool/before_install.sh
 script: ./tool/travis.sh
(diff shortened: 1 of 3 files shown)


tool/travis.sh
```diff
@@ -3,6 +3,8 @@
 # Fast fail the script on failures.
 set -e
 
+(cd tool; git clone https://github.com/flutter/flutter.git)
```
sethladd: do we want to point to alpha, since that's what our users are installing? They will want examples to match their location copy.

abarth:
@sethladd Ready for review, PTAL

abarth:
TBR @sethladd

Issue #87, opened 2015-11-13
Better explanations about installation of adb for linux users


a14n:
On _Ubuntu 14.04 (64b TLS)_ the _adb_ provided via _apt_ is too old.
So I followed the link provided in [getting started](http://flutter.io/getting-started/#setting-up-your-android-device) and downloaded the SDK Tools but it's not obvious what should be done next.

It could be worth to mention somewhere :

* launch `tools/android`
* install (only?) _Android SDK Platfrom-tools_
* add `platform-tools/` to your PATH
* check that `adb devices -l` lists your device

On this last item I got an misleading error _adb: No such file or directory_ (or something like that). I figured out that it was because I needed to install _lib32stdc++6_ package (Actually there's a section _Troubleshooting Ubuntu_ on the [Installing the Stand-alone SDK Tools](https://developer.android.com/sdk/installing/index.html?pkg=tools) that mentions this package but I did not read it because it was under a section _Here are the steps to install Java_ ) . It would also be nice to mention that issue too.

sethladd:
Thanks for the report!

sethladd:
If you'd like to submit a patch for review, that would be appreciated. Since you recently went through the flow, I assume you can document all the steps?

a14n:
Fixed by #103

PR #103, opened 2015-11-25
More explanations on adb install

a14n:


a14n:
Related to #87

getting-started.md
```diff
@@ -100,6 +106,8 @@ Jelly Bean, v16, 4.1.x or later.
  - Using a USB cable, plug your phone into your computer. If prompted on your
    device, authorize your computer to access your device.
 
+ - Check that _abd_ lists your device with `adb devices -l`.
```
abarth: _abd_ <--- typo ?

a14n: fixed. thanks :)

abarth:
It looks like we were using `adb` rather than _adb_ to name the tool previously.  We should be consistent about how we name the tool.  Maybe we should use `adb` in the new text as well?  We also use `this style` to refer to text in the Android UI whereas you're using _this style_ for some of the new text.  We should also be consistent with this sort of information...  I think _this style_ is better for the UI strings, so maybe we should change "Android debugging" and similar uses too.

The important thing is to be consistent throughout the document.  (Also, please squash your commits so I can merge them cleanly into master.)

Thanks!!

a14n:
Actually simple backticks mean inline code format. So I'd prefer _Android SDK Manager_ rather than `Android SDK Manager`.

But I agree that consistancy is more important so I've squashed my PR with your recommandations.

getting-started.md
```diff
@@ -91,6 +91,12 @@ Jelly Bean, v16, 4.1.x or later.
    - Linux: `sudo apt-get install android-tools-adb`
      - If the version of `adb` provided by your Linux distribution is too old,
        you might need to [install the Android SDK manually](https://developer.android.com/sdk/installing/index.html?pkg=tools).
+       Note that `adb` needs 32-bit libs (on ubuntu 64-bit, make sure to install
```
abarth: s/ubuntu/Ubuntu/

getting-started.md
```diff
@@ -91,6 +91,12 @@ Jelly Bean, v16, 4.1.x or later.
...
+       those libs listed in _Troubleshooting Ubuntu_ section).
```
abarth: s/libs/libraries/

abarth:
LGTM

a14n:
Let me wake up :)
Thanks!

abarth:
No worries.  I just made the trivial edits after merging your patch.  :)

PR #112, opened 2015-12-09
Add documentation for the gesture arena

abarth:


abarth:
@Hixie

gestures.md
```diff
@@ -0,0 +1,102 @@
...
+   location for a long period of time.
+- Vertical drag
+ - `onVerticalDragStart` A pointer has contacted the screen and might begin to
+   move vertically.
+ - `onVerticalDragUpdate` A pointer that is in contact with the screen and
+   moving vertically has moved again in the vertical direction.
+ - `onVerticalDragEnd` A pointer that was previously in contact with the screen
+   and moving vertically is no longer in contact with the screen and was moving
+   at a specific velocity when it stopped contacting the screen.
+ - Horizontal drag
```
krisgiesing: Was this intended to be at the same indent level as Vertical drag?

abarth: Yes.  Will fix.

gestures.md
```diff
@@ -0,0 +1,102 @@
...
+gesture wins using the following rules:
+
+- At any time, a recognizer can declare defeat and leave the arena.  If there's
+  only one recognizer left in the arena, that recognizer is the winner.
+
+- At any time, a recognizer can declare victory, which causes it to win and all
+  the remaining recognizers to lose.
+
+In the case of disambiguating horizontal and vertical dragging, both recognizers
+enter the arena when the receive the pointer down event.  The recognizers
```
krisgiesing: typo: "the receive" -> "they receive"

krisgiesing:
Nice, LGTM

Hixie:
I'll look in a second but see also
https://github.com/flutter/engine/blob/master/sky/specs/gestures.md
https://github.com/flutter/flutter/issues/700

gestures.md
```diff
@@ -0,0 +1,102 @@
...
+
+Pointers represent raw data about the user's interaction with the device's
+screen.  There are four types of pointer events:
+
+- `PointerDownEvent` The pointer has contacted the screen at a particular
+   location.
+- `PointerMoveEvent` The pointer has moved from one location on the screen to
+   another.
+- `PointerUpEvent` The pointer has stopped contacting the screen.
+- `PointerCancelEvent` The framework is no longer delivering events for this
```
Hixie: The cancel one is not quite accurate. It's more that the system has decided to stop telling the framework about the pointer. Nothing in the framework itself generates the event, even in the case of a widget being taken out of the flow in some way and thus no longer getting events for the pointer.

gestures.md
```diff
@@ -0,0 +1,102 @@
...
+each recognizer join the _gesture arena_. The gesture arena determines which
+gesture wins using the following rules:
+
+- At any time, a recognizer can declare defeat and leave the arena.  If there's
+  only one recognizer left in the arena, that recognizer is the winner.
+
+- At any time, a recognizer can declare victory, which causes it to win and all
+  the remaining recognizers to lose.
+
+In the case of disambiguating horizontal and vertical dragging, both recognizers
```
Hixie: prefix with "For example: "

gestures.md
```diff
@@ -0,0 +1,102 @@
...
+screen, the framework disambiguates which gesture the user intends by having
+each recognizer join the _gesture arena_. The gesture arena determines which
+gesture wins using the following rules:
+
+- At any time, a recognizer can declare defeat and leave the arena.  If there's
+  only one recognizer left in the arena, that recognizer is the winner.
+
+- At any time, a recognizer can declare victory, which causes it to win and all
+  the remaining recognizers to lose.
+
```
Hixie: We should probably talk about holding the arena open, and how it gets swept when the pointer goes up, and how a single gesture can be in multiple arenas since it can track multiple pointers...

gestures.md
abarth: "The system has lost track of this pointer and will not generate more events for it." ?  What text do you recommend here?

gestures.md
abarth: Yeah, maybe in another round...

gestures.md
abarth: "Input from this pointer is no longer directed towards this app."

Hixie: Either of those are fine.

PR #120, opened 2015-12-10
Mention 64-bit for linux

sethladd:
There's probably more we can do, but this looks like the simplest change for now.


sethladd:
cc @abarth

abarth:
We only support 64 bit mac too, but I think that's implied nowadays for macs.

abarth:
LGTM

PR #121, opened 2015-12-11
Update tutorial after TextDecoration change

abarth:
This constant is now in the TextDecoration class.


abarth:
@Hixie 

I'll land this after pushing to the alpha branch.

Hixie:
LGTM

This needs an engine rev, too.

abarth:
> This needs an engine rev, too.

Did that part already.

PR #132, opened 2016-01-11
Various fixes to the debugging page.

Hixie:


abarth:
LGTM

abarth:
[info] Unused import at example/debugging_md_1.dart, line 2.

Hixie:
that unused import warning makes no sense to me. There's two dart blocks. One imports dart:developer and then calls debugger(), the other imports material and then uses MaterialApp.

abarth:
I suspect the problem is that the script that generates the dart files also imports material.dart, which is unused in that block.  There's an annotation you can use to suppress the tool.

Hixie:
Ohhh, yeah, that'll be it. Ok, I'll skip it for that block.

Hixie:
Note for people confused about the StatisticsOverlay rename in this patch: I'm about to check in that rename in the engine and framework (tomorrow). It's just the docs being a little bit ahead of reality for once.

PR #134, opened 2016-01-11
remove unused file, update README

sethladd:
We're no longer hosting on github pages.

Also, only build if pushed to master.


sethladd:
cc @abarth

abarth:
LGTM

PR #135, opened 2016-01-15
update project name in pubspec

sethladd:


sethladd:
TBR

Hixie:
lgtm. this doesn't actually make any difference right?

sethladd:
not from what I can tell. I think it's here to help with some testing (there's a step in travis that extracts code to analyze)

PR #137, opened 2016-01-20
Add documentation about the new animation system

abarth:


qchong:
It would be good to link to a sample Flutter app that demonstrates how animations work. I'd also like to see example code snippets in this doc showing how to create an `Animation` and how to attach it to a widget via an `AnimationListener`.

abarth:
That makes a lot of sense.  Do you think we should include that in this patch or should we add those in a later patch?

qchong:
Thanks Adam, a later patch is fine. I've also opened an issue for the code sample request (https://github.com/flutter/flutter/issues/1322).

abarth:
Thanks.

PR #154, opened 2016-02-02
update init to create

devoncarew:
Rename `init` to create`; fix https://github.com/flutter/flutter/issues/1521. Thanks for the report @pietergreyling!


sethladd:
lgtm and thanks!

PR #162, opened 2016-02-15
Dart SDK is no longer required, we predownload it

### LINKED ISSUE #1886: Improve the Flutter SDK archive page
Page URL: https://flutter.io/docs/development/tools/sdk/archive.html
Page source: https://github.com/flutter/website/tree/master/src/docs/development/tools/sdk/archive.md

-  [ ] **Add a description of each of the channels.** As requested by @Hixie (and reported by @sfshaza2): the page should contain a description of the channels.

    The channels are well documented in the [Flutter's channels][] wiki page, which is referenced at the start of the Archive page:

    > The Stable channel contains the most stable Flutter builds. See [Flutter’s channels][] for details.

    Is this good enough? If not, then I'd rather not have duplicate information. I'd then suggest moving at least the initial part of the wiki page text into the Archive page. WDYT @Hixie?

    Also, you mentioned that you might prefer an interface similar to the [Dart SDK Archive][] page: with **Version** and **OS** drop downs? Or were you only giving that page as an example w.r.t. the channel descriptions?

- [ ] list sha256sums for download files (originally requested by @cbracken in #1955)

- [x] **Question** is the **Ref** value really useful for users? @cbracken @Hixie

Overall, maybe we should just carry over the [Dart SDK Archive][] design and implementation? Thoughts?

cc @kwalrath 

[Dart SDK Archive]: https://www.dartlang.org/tools/sdk/archive
[Flutter's channels]: https://github.com/flutter/flutter/wiki/Flutter-build-release-channels#flutters-channels

Hixie:
Also fixes https://github.com/flutter/flutter/issues/1884, https://github.com/flutter/flutter/issues/1885, https://github.com/flutter/flutter/issues/1886.

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -6,42 +6,22 @@ permalink: /getting-started/
 ---
 
 Flutter currently supports developers on Mac and Linux (64-bit).
-Windows support is planned.
+Windows support is planned but currently incomplete.
 
-## Setting up your environment
+## Dependencies
 
-### Downloading and installing the Dart SDK
+Flutter depends on the following tools being available in your environment. These are commonly already available on Mac and Linux.
 
-To get started, you need to install the Dart SDK.
-Flutter requires Dart SDK 1.12.2 or greater.
-
-To install the [Dart SDK](https://www.dartlang.org/downloads/):
-
-- Mac:
-  - `brew tap dart-lang/dart && brew install dart`, or, if you don't use
-    homebrew, grab the [latest stable channel build](https://www.dartlang.org/downloads/archive/)
-- Linux: See [www.dartlang.org/downloads/linux.html](https://www.dartlang.org/downloads/linux.html)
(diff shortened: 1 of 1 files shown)


Hixie:
@abarth 

We shouldn't land this until we roll the alpha branch.

getting-started.md
```diff
@@ -53,14 +33,13 @@ the `flutter` command.)
 ## Configuring your PATH
 
 After you clone the Flutter repo, set the PATH so you can
-use our scripts and tools.
+use our scripts and tools.
```
abarth: Extra trailing space.

abarth:
LGTM

getting-started.md
Hixie: i'll fix it in post.

Hixie:
woops, I didn't mean to land this. sigh.

PR #163, opened 2016-02-15
Revert "Dart SDK is no longer required, we predownload it"

Hixie:
This reverts commit bad8ce2db0cdd9996fc609323933d1a58b1894d6.

We should only land this once it's rolled into alpha.


Hixie:
TBR @abarth

PR #166, opened 2016-02-16
Explain that flutter start starts on all devices.

eseidelGoogle:
@devoncarew


devoncarew:
in master (not alpha), `flutter start` ==> `flutter run` and `flutter list` ==> `flutter devices`. Don't know if you want to make that change now or wait until master move to alpha.

lgtm

eseidelGoogle:
Going to wait, since this is the safer choice (and will still run on alpha).

PR #174, opened 2016-02-19
docs: Update screenshot for demo app to reflect slow mode banner.

qchong:


qchong:
@sethladd ptal

sethladd:
Please run 'optipng' on this to shrink it. There are other similar tools, which might work also work fine.

PR #178, opened 2016-02-23
docs: Added new 'Flutter Setup' doc with installation instructions

qchong:
This updates includes the following changes: 

* Added a Setup doc, replacing the existing setup content in the Getting Started doc.
* Added new iOS screenshot for the Getting Started app.
* Added ToC links to the top of the Getting Started page.


qchong:
Needs review.

getting-started.md
```diff
@@ -5,156 +5,93 @@ nav_title: Getting Started
 permalink: /getting-started/
 ---
 
-Flutter currently supports developers on Mac and Linux (64-bit).
-Windows support is planned but currently incomplete.
+This quickstart guide describes how to create and run your first Flutter app on iOS or Android.
 
-## Dependencies
+[Before you begin](#before_you_begin)<br>
```
Hixie: This is not really what `<br>` is supposed to be used for. Maybe a bulletted list would be better?

getting-started.md
```diff
@@ -5,156 +5,93 @@ nav_title: Getting Started
...
- - Enable developer mode on your device by visiting `Settings > About phone` and
-   tapping the `Build number` field seven times.
+In the project directory, the code for your app is in `my_app/lib/main.dart`.
 
- - Enable `Android debugging` in `Settings > Developer options`.
+## 2. Run your Flutter app
 
- - Using a USB cable, plug your phone into your computer. If prompted on your
-   device, authorize your computer to access your device.
+Use the `flutter run` command to run your Flutter app on all connected
```
Hixie: See https://github.com/flutter/flutter/issues/2106

getting-started.md
```diff
@@ -5,156 +5,93 @@ nav_title: Getting Started
...
-To run your app, use:
 
-```
-$ flutter run
-```
+If everything works, you should see your starter app on your Android device or iOS Simulator.
 
-If everything works, you should see your starter app
-on your Android device.
+<table>
```
Hixie: this isn't a table, it's just two images side by side. Can we make the <figure> here inline-block?

setup.md
```diff
@@ -0,0 +1,130 @@
+---
```
Hixie: How does this relate to https://flutter.io/getting-started/ ?

cc @abarth who was working on a new version of that

qchong: This doc change is based on the new content shared by Adam offline.

getting-started.md
qchong: Fixed.

setup.md
Hixie: Ok. I think we should make sure to only have one set of instructions on the site at once, so please delete the old instructions when adding these. Ideally, put these in place of the old ones, so that the URL doesn't change, since we have links to it in various places.

setup.md
```diff
@@ -0,0 +1,130 @@
...
+* **Tools**: Flutter depends on these command-line tools being available on your environment.
+  * `bash`, `mkdir`, `rm`, `git`, `curl`, `unzip`
+
+## iOS setup ##
+
+To develop Flutter apps for iOS, you need a Mac with Xcode 7.2 or newer:
+
+1. Install Xcode 7.2 or newer (via [web download](https://developer.apple.com/xcode/) or
+the [Mac App Store](https://itunes.apple.com/us/app/xcode/id497799835)).
+2.  Make sure the Xcode EULA is signed by either opening Xcode once and confirming or
```
Hixie: different space indents here

setup.md
```diff
@@ -0,0 +1,130 @@
...
+This guide describes how to set up your development environment to run Flutter apps on
+iOS or Android.
+
+[System requirements](#system_requirements)<br>
+[iOS setup](#ios_setup)<br>
+[Android setup](#android_setup)<br>
+[Get the Flutter SDK](#get_the_flutter_sdk)<br>
+[Atom editor](#atom_editor)
+
+## System requirements ##
```
Hixie: you use `## foo ##` here but other files (including the other one in this patch) only use `## foo`.

setup.md
```diff
@@ -0,0 +1,130 @@
...
+
+To install and run Flutter, your development environment must meet these minimum requirements.
+
+* **Operating Systems**: Mac or Linux (64-bit). Windows support is planned.
+* **Tools**: Flutter depends on these command-line tools being available on your environment.
```
Hixie: colon, not period

setup.md
```diff
@@ -0,0 +1,130 @@
...
+
+[System requirements](#system_requirements)<br>
+[iOS setup](#ios_setup)<br>
+[Android setup](#android_setup)<br>
+[Get the Flutter SDK](#get_the_flutter_sdk)<br>
+[Atom editor](#atom_editor)
+
+## System requirements ##
+
+To install and run Flutter, your development environment must meet these minimum requirements.
```
Hixie: colon, not period, since you're saying "these".

setup.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Deploy to iOS devices ###
+
+To deploy your Flutter app to a physical iOS device, you’ll need some additional tools:
+
+1. Install [homebrew](http://brew.sh/).
+2. Open the terminal and run these command to install the toools for deploying Flutter apps to
+iOS devices.
+
+    <pre>
```
Hixie: use ``` here instead of <pre>

setup.md
```diff
@@ -0,0 +1,130 @@
...
+    $ brew install ideviceinstaller ios-deploy
+    </pre>
+
+## Android setup ##
+
+To develop Flutter apps for Android, you can use either a Mac or a Linux (64-bit) machine:
+
+1. Install [Android Studio](https//developer.android.com/sdk/index.html).
+2. From Android Studio, install the latest Android SDK and Android SDK Platform-Tools, as described
+[here](https://developer.android.com/sdk/installing/adding-packages.html).
```
Hixie: accessibility guidelines suggest avoiding using "here" as link text; maybe describe the target instead ("`as described [on the Android site]`" or something)

setup.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Set up your Android device
+
+To prepare to run and test your Flutter app on Android, you'll need an Android device running
+Android 4.1 (API level 16) or higher.
+
+1. Enable developer mode on your device by visiting **`Settings > About phone`** and
+   tapping the **`Build number`** property seven times.
```
Hixie: s/property/menu item/

Hixie: s/property/line/ maybe?

setup.md
```diff
@@ -0,0 +1,130 @@
...
+2. Enable **`Android debugging`** in **`Settings > Developer options`**.
+3. Using a USB cable, plug your phone into your computer. If prompted on your
+   device, authorize your computer to access your device.
```
Hixie: on this line you indent the list item text and on some other lines you don't; either is fine by me but we should be consistent throughout the file

setup.md
```diff
@@ -0,0 +1,130 @@
...
+
+By default, Flutter uses the version of the Android SDK where your `adb` tool is based. If
+you want Flutter to use a different installation of the Android SDK, you must set the
+`ANDROID_HOME` environment variable to that specific installation directory.
+
+## Get the Flutter SDK ##
+
+To get Flutter, use `git` to clone the repository and then add the `flutter` tool to your path:
+
+<pre>
```
Hixie: ``` vs `<pre>`

setup.md
```diff
@@ -0,0 +1,130 @@
...
+Before creating your Flutter project in Atom, you’ll need to perform these pre-flight tasks:
+
+1. In the terminal window, run the `atom` command to launch the editor.
+  ```
+  $ atom
+  ```
+2. Open **Packages > Flutter > Package Settings**, and set the **FLUTTER_ROOT** field to the
+root directory where you installed the Flutter SDK.
+3. Open **Packages > Dart > Package Settings**, and set the **Dart SDK Location** field to the
+`/bin/cache/dart-sdk` directory in your Flutter SDK’s root folder.
```
Hixie: `.../bin/cache/dart-sdk` or `bin/cache/dart-sdk`. With a leading slash it's an absolute path from the root.

Hixie:
lgtm

getting-started.md
qchong: fixed.

setup.md
qchong: fixed.

setup.md
qchong: fixed.

setup.md
qchong: fixed on line 19.

setup.md
qchong: The ``` renders each line of code as a separate code block, which is not what we want here. To make the multi-line code block appear correctly, I've used <pre> as a workaround.

setup.md
qchong: fixed.

setup.md
qchong: fixed.

setup.md
qchong: Please see my previous comment.

setup.md
qchong: Fixed to use the latter.

setup.md
qchong: fixed to remove the indent.

setup.md
qchong: fixed to remove the indents.

setup.md
qchong: Yup, I've already removed the old instructions from Getting Started. I'll open an Issue to go through the other docs and fix any old references pointing to the Getting Started for installation-related instructions. Going forward, we should be consistent in pointing to the 'Flutter Setup' page as the central location for installation/configuration instructions.

Hixie: Are we planning on reusing /getting-started/ for something? Unless there's a good reason to change, I really think we should keep the URLs consistent.

setup.md
Hixie: Can you elaborate? When I try them here, \`\`\` seems to do the right thing:

   ```
   $ brew tap flutter/flutter
   ```

   <pre>
   $ brew tap flutter/flutter
   </pre>

getting-started.md
```diff
@@ -4,157 +4,94 @@ title: Getting Started with Flutter
...
- - Check that `adb` lists your device with `adb devices -l`.
 
-By default, Flutter uses the version of the Android SDK where your `adb` tool is based. If
-you want Flutter to use a different installation of the Android SDK, you must set the
-`ANDROID_HOME` environment variable to that specific installation directory.
+If everything works, you should see your starter app on your Android device or iOS Simulator.
 
-## Running a Flutter application on Android
+<div id="starter-app-screenshots">
+  <div class="box2">
```
Hixie: can't the classes go on the `<figure>` elements? I don't really understand why we need the divs at all here.

getting-started.md
```diff
@@ -4,157 +4,94 @@ title: Getting Started with Flutter
...
-## Creating your first sample app
-
-You can use the `flutter` command to create a starter project.
+To create a starter project, open a terminal and run the `flutter` command .
 
 Here is an example:
 
 ```
-$ flutter create -o my_app
+$ flutter create -o myapp
```
Hixie: iirc you can remove the -o now

PR #180, opened 2016-02-24
Make more obvious that devs don't need both iOS and Android installed…

sethladd:
… to get started

Didn't want people to think they needed to install the iOS bits to keep following this guide step by step.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -5,6 +5,11 @@ nav_title: Flutter Setup
 permalink: /setup/
 ---
 
+(Please remember, Flutter is in _technology preview_. Thanks
+for trying it out, and please don't hesitate to
+[file issues or bugs](https://github.com/flutter/flutter/issues/new)
+for us.)
+
 This guide describes how to set up your development environment to run Flutter apps on
 iOS or Android.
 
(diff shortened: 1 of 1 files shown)


sethladd:
cc @abarth

setup.md
```diff
@@ -5,6 +5,11 @@ nav_title: Flutter Setup
 permalink: /setup/
 ---
 
+(Please remember, Flutter is in _technology preview_. Thanks
+for trying it out, and please don't hesitate to
+[file issues or bugs](https://github.com/flutter/flutter/issues/new)
+for us.)
```
abarth: Should we have this note on every page?  Maybe we should incorporate it into the banner at the top of the page somehow.

abarth:
LGTM but you probably want the thumbs up from @qchong

setup.md
sethladd: good call. I'll add it to our top banner, right after landing this (and I'll remove this note here)

sethladd:
@qchong ptal !

setup.md
qchong: +1 to a banner label (or icon) to indicate the Tech Preview status of Flutter.

qchong:
lgtm, thanks Seth!

PR #181, opened 2016-02-24
Fix broken link

krisgiesing:


qchong:
Duplicates fix in flutter/website#182 which has been committed.

PR #191, opened 2016-03-01
docs: Updated the instructions for installing Atom and Flutter package

qchong:
@HansMuller

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -119,25 +119,31 @@ you want Flutter to use a different installation of the Android SDK, you must se
 We recommend using [Atom](https://atom.io/) for editing, running, and debugging Flutter apps.
 However, using our command-line tools, you can use any editor to develop Flutter applications.
 
-Download Atom from [atom.io](https://atom.io/) and then install the
-[Flutter package](https://atom.io/packages/flutter). If you need help installing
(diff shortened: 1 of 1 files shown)


setup.md
```diff
@@ -119,22 +119,28 @@ you want Flutter to use a different installation of the Android SDK, you must se
...
 
-Download Atom from [atom.io](https://atom.io/) and then install the
-[Flutter package](https://atom.io/packages/flutter). If you need help installing
-packages, consult the [Atom documentation](https://atom.io/docs/v1.3.2/using-atom-atom-packages)
-or [email us](flutter-dev@googlegroups.com).
+To install Atom:
+
+1. Download Atom from the [atom.io](https://atom.io/) site.
+2. Extract the Atom executable from the downloaded file.
+
```
HansMuller: I think you still need to say "install the flutter package".

HansMuller: Oops, I see, the package install step is under "Configure Atom...". So maybe the "if you need help installing..." item should go below?

HansMuller:
LGTM

setup.md
qchong: Thanks, I've made the change you suggested and re-pushed.

PR #193, opened 2016-03-03
animations tutorial

sethladd:
closes https://github.com/flutter/flutter/issues/1731


sethladd:
cc @jimbeveridge @qchong ptal , thanks!

sethladd:
fyi @abarth this PR fixes the use of Icons/IconData

animations-tutorial.md
```diff
@@ -0,0 +1,589 @@
...
+[`Animation`](http://docs.flutter.io/flutter/animation/Animation-class.html)
+objects that produce values over a certain duration. Widgets can either
+incorporate these animations in their `build`
+functions directly by reading their current value and listening to their
+state changes or they can use the animations as the basis of more elaborate
+animations that they pass along to other widgets.
+
+There are two separate animation libraries for Flutter. This documentation
+covers the [flutter/animation](http://docs.flutter.io/flutter/animation/animation-library.html) library, which is intended for animations in
+user interfaces that are built on StatefulComponent and StatelessComponent.
```
a14n: Links to http://docs.flutter.io/flutter/widgets/StatefulComponent-class.html and http://docs.flutter.io/flutter/widgets/StatelessComponent-class.html ?

animations-tutorial.md
```diff
@@ -0,0 +1,589 @@
...
+
+const kLogoUrl = "https://raw.githubusercontent.com/dart-lang/logos/master/logos_and_wordmarks/dart-logo.png";
+
+class LogoApp extends StatefulComponent {
+  LogoAppState createState() => new LogoAppState();
+}
+
+class LogoAppState extends State&lt;LogoApp&gt; {
+  Animation&lt;double&gt; animation;
+<b>
```
a14n: `<b>` should be on the previous line.

animations-tutorial.md
```diff
@@ -0,0 +1,589 @@
...
+            width: <b>animation.value</b>,
+            child: new NetworkImage(src: kLogoUrl)));
+  }
+}
+
+void main() {
+  runApp(new LogoApp());
+}
+</code></pre>
+
```
a14n: Starting from here the rendering done (visible at https://github.com/flutter/website/blob/animation-doc/animations-tutorial.md) is without markdown evaluation. Seems to be due to the use of `<code></code>`.

animations-tutorial.md
```diff
@@ -0,0 +1,589 @@
...
+</pre>
+
+If you look at the logs after running this change, you’ll see lines like this as
+the animation progresses:
+
+```
+01-28 10:30:58.268 23497 23524 I flutter : AnimationStatus.forward
+...
+01-28 10:30:59.760 23497 23524 I flutter : AnimationStatus.completed
+```
```
a14n: `forward` to `completed` takes around 1.5s but the duration is set to 2000ms. Why ?

tutorial.md
```diff
@@ -106,9 +106,9 @@ class MyToolBar extends StatelessComponent {
       ),
       child: new Row(
         children: [
-          new IconButton(icon: 'navigation/menu'),
+          new IconButton(icon: Icons.menu),
```
a14n: You should also add `uses-material-design: true` to `flutter.yaml` and remove the sections `material-design-icons`

tutorial.md
```diff
@@ -219,15 +219,15 @@ class TutorialHome extends StatelessComponent {
...
         center: new Text('Example title'),
-        right: [ new IconButton(icon: 'action/search') ]
+        right: [ new IconButton(icon: Icons.search) ]
       ),
       body: new Center(
         child: new Text('Hello, world!')
       ),
       floatingActionButton: new FloatingActionButton(
-        child: new Icon(icon: 'content/add')
+        child: new Icon(icon: Icons.search)
```
a14n: `Icons.add` ?

animations-tutorial.md
sethladd: done

animations-tutorial.md
sethladd: Changed to markdown backticks and added some comments to signal what is new.

animations-tutorial.md
sethladd: Looks fixed now.

animations-tutorial.md
sethladd: cc @jimbeveridge for help

tutorial.md
sethladd: Thank you! done.

tutorial.md
sethladd: done

animations-tutorial.md
jimbeveridge: Just a comment - I spent quite a while looking at the problem that we can have syntax highlighting or HTML, but not both. My conclusion after visually considering what I was asking of the reader was that the HTML boldface was more important than the code markup. It's inconsistent, but it's a lot easier to read, particularly in code examples where there are two or three new sections added. I don't think the decision should be made by looking at this markdown, because the intended result is purely visual.

sethladd: I appreciate that bolding is easier to read, it is nicer visually, but when I turned on our syntax checker/analyzer for these snippets, I found some typos and bugs. We rely on this system to help us update our docs as we update Flutter. So for now, maintainability and testing is critical. As things settle down and the APIs stop churning so much, we can take a look at ways to make the snippets more visually distinctive over maintainability.

animations-tutorial.md
```diff
@@ -0,0 +1,639 @@
...
+  runApp(new LogoApp());
+}
+```
+
+Because `addListener` calls `setState`, every time the animation generates a new
+number, the current frame will be marked dirty, which in turn will force `build`
+to be called again. In `build`, the container changes size because its height
+and width now use `animation.value` instead of hardcoding 300.0.
+
+With these short changes, you’ve created your first animation in Flutter!
```
jimbeveridge: I think that the Dart Language Tricks section was removed? The use of dot dot in examples was specifically identified by Adam as one of the stumbling blocks that people encountered in the codelab this week and it needs to be discussed. If you feel strongly that you don't want it in this doc, then please link another page or a stackoverflow post.

sethladd: Oops, that was accidentally removed and not put back. Thanks for catching that!!

animations-tutorial.md
jimbeveridge: Good catch. The latest code doesn't print the timestamps, so let's match our output to the latest code:

```
I/flutter (21456): AnimationStatus.forward
I/flutter (21456): AnimationStatus.completed
```

jimbeveridge:
LGTM after latest comments are addressed. Thanks Seth!

animations-tutorial.md
sethladd: Done

animations-tutorial.md
sethladd: I ended up removing the .. syntax. One less thing to teach.

animations-tutorial.md
```diff
@@ -0,0 +1,638 @@
+---
+layout: page
+title: Animations Tutorial
+permalink: /animations-tutorial/
+---
+
+Welcome to Flutter's Animations Tutorial. You may also be interested
```
qchong: s/Animations Tutorial/animations tutorial/

No need for uppercase.

animations-tutorial.md
```diff
@@ -0,0 +1,638 @@
...
+in our [Animations Overview](/animations/).
```
qchong: This sentence can be tightened: e.g.
"Before starting this tutorial, we encourage you to read the Animations overview documentation."

jimbeveridge: Speaking from experience, the Overview document is horribly confusing for beginners. Please do not encourage people to read it first.

abarth: Maybe we should just remove the existing overview in favor of this document?

animations-tutorial.md
```diff
@@ -0,0 +1,638 @@
...
+
+## Introduction
```
qchong: Hmm, scanning through the rest of this doc, it doesn't fit the 'tutorial' category. As it reads right now, this is more of a 'developer guide' (i.e. description of feature + instructions on how to use the APIs). A tutorial usually has some clear learning goals (i.e. "what you'll be able to build after following this tutorial") and a set of steps for building that type of app or adding that functionality.

My strong preference is that we do the following:

1. The API descriptions below in this doc be moved into the API docs for the respective classes (Animation, AnimationController, Tween, etc.) or the description for the [animation package](http://docs.flutter.io/flutter/animation/Animation-class.html). 

2. We revise the current [Animations Overview](/animations/) doc to be a proper developer guide. That way, we can move the information about specific animation techniques and common operations (monitoring the progress of animation, etc.) into that animations dev guide. I think it would also be nice to add some actual animated gifs in that doc to demonstrate how thse APIs work.

animations-tutorial.md
krisgiesing: FWIW, I still find .. incredibly hard to parse mentally.

animations-tutorial.md
krisgiesing: Probably so.

qchong:
Hi Seth, I think we should hold off from merging/publishing this, since the material presented does not actually fit the tutorial format. Instead, some of this content should go into the appropriate API docs, or be merged into the current animations overview. Please feel free to reassign this work to me.

jimbeveridge:
@qchong I would like to discuss this with you in person before we make any decisions.

sethladd:
@qchong thanks for the review! You may also want to review the original Animation doc. There may be candidates in there for shuffling, merging, etc.

sethladd:
I am going to close this PR, but happy to help publish it when we're ready.

PR #195, opened 2016-03-06
Update tutorial.md

a14n:
Use `uses-material-design: true`

### DIFF
--- a/tutorial.md
+++ b/tutorial.md
@@ -148,9 +148,7 @@ Be sure to add the new resources to your `flutter.yaml` file.
 
 ```yaml
 name: my_app
-material-design-icons:
-  - name: action/search
-  - name: navigation/menu
+uses-material-design: true
 ```
 
 Many widgets need to be inside of a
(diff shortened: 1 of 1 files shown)


tutorial.md
```diff
@@ -148,9 +148,7 @@ Be sure to add the new resources to your `flutter.yaml` file.
```
Hixie: The line before should also be updated, otherwise it's confusing. But actually, I think `flutter create` sets this already, so we might be able to remove this whole part, or at least just replace it with mentioning the role that `uses-material-design` has.

a14n: Indeed. I'll open a new PR to remove this part.

a14n: Hmm. Actually `flutter create` is not mentionned in this tutorial. So I think a section about `flutter.yaml` should remain until `flutter create` appears here.

Hixie:
Please note that the convention for review is to wait until someone has explicitly written LGTM on the PR before check-in. Not a big deal here but I just thought I'd let you know. We should probably document this better.

Hixie:
Oh never mind, I see you just closed this PR to open a new one. That works too.

PR #199, opened 2016-03-13
[rename fixit] Update web site to use the new names

abarth:
Related to https://github.com/flutter/flutter/issues/2536

### DIFF
--- a/animations.md
+++ b/animations.md
@@ -35,13 +35,13 @@ rebuild with the new value of the animation.
 
 This pattern is so common that there are two widgets that help widgets rebuild
 when animations change value:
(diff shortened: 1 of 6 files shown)


abarth:
@qchong @Hixie

abarth:
Technically, we should wait on merging this pull request until we update the alpha branch.

Hixie:
LGTM. Agreed that we should wait until we update alpha; we should probably bake trunk on Monday before we do that though.

abarth:
There's a problem that Travis is using an older analyzer that doesn't know how we're describing dart:ui.  I guess I can fix that in another patch before landing this one.

PR #200, opened 2016-03-14
Switch Travis to use the Dart SDK from flutter

abarth:
This ensures that we use the same version of the Dart SDK that we're using in
flutter.git.


abarth:
@Hixie Ready for review.

Hixie:
lgtm

PR #208, opened 2016-03-23
Update animation overview

qchong:


qchong:
@abarth @Hixie @jimbeveridge @sethladd ptal. The updated animations guide builds on content from the original Animations overview doc and this previous PR (https://github.com/flutter/website/pull/193).

qchong:
Not sure why the PR picked up several other commits from my branch. For reviewers, please ignore the other commits and focus on this one (https://github.com/flutter/website/pull/208/commits/42f439ae63087b6e32071a8869c6057b5f927f06).

animations.md
```diff
@@ -1,311 +1,377 @@
...
+  * <strong>Value</strong>: The animation object’s typed value represents a data point within a
+  range (also called an <em>interpolation</em>). You can define an animation class to generate
+  values based on a given functional mapping, such as a linear, curve, or step-wise function.
+  * <strong>Status</strong>: The animation object’s current status (such as <em>forward</em>,
+  <em>backward</em>, <em>completed</em>, and <em>dismissed</em>), as listed by the constants of
+  the
+  <code>[AnimationStatus](http://docs.flutter.io/flutter/material/AnimationStatus-class.html)</code>
+  class.
+
+Typically, the animation’s value and status changes over time as it runs. You
```
jimbeveridge: s/changes/change/. Then the antecedent of it needs clarification.

animations.md
```diff
@@ -1,311 +1,377 @@
...
+
+Animations in Flutter are encapsulated as
+<code>[Animation](http://docs.flutter.io/flutter/animation/Animation-class.html)</code>
+objects that contain these key properties:
+
+  * <strong>Value</strong>: The animation object’s typed value represents a data point within a
+  range (also called an <em>interpolation</em>). You can define an animation class to generate
+  values based on a given functional mapping, such as a linear, curve, or step-wise function.
+  * <strong>Status</strong>: The animation object’s current status (such as <em>forward</em>,
+  <em>backward</em>, <em>completed</em>, and <em>dismissed</em>), as listed by the constants of
```
jimbeveridge: Why are you diving into the details like "constants of the code" when this is the overview?

animations.md
```diff
@@ -1,311 +1,377 @@
...
+and <code>[StatefulWidget](http://docs.flutter.io/flutter/widgets/StatefulWidget-class.html)</code>
+classes.
+  * <code>[flutter\_sprites/flutter\_sprites](http://docs.flutter.io/flutter_sprites/flutter_sprites/flutter_sprites-library.html)</code>:
+  Provides APIs for game development.
+
+The rest of this documentation covers how to use the APIs from the
+<code>[flutter/animation](http://docs.flutter.io/flutter/animation/animation-library.html)</code>
+library.
+
+Animations in Flutter are encapsulated as
```
jimbeveridge: You never say what Animations are. The reader has no context to understand why you are telling them about Value and AnimationStatus.

animations.md
```diff
@@ -1,311 +1,377 @@
...
-### Attaching animatables to animations
-
-Passing an `Animation<double>` (the new parent) to an `Animatable`'s
-`animate()` method creates a new `Animation` subclass that acts like
-the `Animatable` but is driven from the given parent.
+You can use animations to make the user experience on your mobile app feel more
+intuitive, dynamic, and polished. Animations are especially useful when the
+screen changes state, such as when content loads or when new actions become
+available. This how-to guide shows some common ways you can add animation
+effects to your apps using Flutter’s animation system.
```
jimbeveridge: animation system perhaps should be animation classes.

animations.md
```diff
@@ -1,311 +1,377 @@
...
+  * <strong>Status</strong>: The animation object’s current status (such as <em>forward</em>,
+  <em>backward</em>, <em>completed</em>, and <em>dismissed</em>), as listed by the constants of
+  the
+  <code>[AnimationStatus](http://docs.flutter.io/flutter/material/AnimationStatus-class.html)</code>
+  class.
+
+Typically, the animation’s value and status changes over time as it runs. You
+can attach an animation object to a widget or listen for changes to the
+animation object. Based on changes to the animation object’s properties, the
+framework can modify the way your widget appears and rebuild the widget tree to
```
jimbeveridge: the framework doesn't do this. Your build function does this. This is the fundamental point of my original introduction - the animation classes don't do any animation.

animations.md
```diff
@@ -1,311 +1,377 @@
...
+produce visual transitions such as sliding, scaling, rotation, height or width
+changes, and fading.
+
+This flexible model allows you to define interpolations for spatial
```
jimbeveridge: You talk about the interpolations, but you never mention the build functions. The reader is not given any information as to why an interpolation is useful.

animations.md
```diff
@@ -1,311 +1,377 @@
...
+<pre class=prettyprint>
+final AnimationController controller =
+    new AnimationController(duration: const Duration(milliseconds: 500));
+</pre>
+
+#### [CurvedAnimation](http://docs.flutter.io/flutter/material/CurvedAnimation-class.html)
+
+Use this class to apply a non-linear curve to an animation object it wraps. By
+applying different curve types, you can visually modify the spatial movement
+pattern for the wrapped animation object. The
```
jimbeveridge: A curve (usually) modifies the output of the AnimationController, adjusting the the graph. This may or may not have anything at all to do with position.

animations.md
```diff
@@ -1,311 +1,377 @@
...
+
+
+<pre class=prettyprint>
+final AnimationController controller =
+    new AnimationController(duration: const Duration(milliseconds: 500));
+final curve = new CurvedAnimation(
+    parent: controller,
+    curve: Curves.easeIn
+);
+</pre>
```
jimbeveridge: You have removed all of the critical information on required imports, where the initializations fit, and how these classes relate to build functions.

animations.md
```diff
@@ -1,311 +1,377 @@
...
+<code>[end](http://docs.flutter.io/flutter/material/Tween/end.html)</code> value. To get the
+interpolated value for a given animation, call the
+<code>[evaluate()](http://docs.flutter.io/flutter/animation/Tween/evaluate.html)</code> method
+on the <code>[Tween](http://docs.flutter.io/flutter/material/Tween-class.html)</code> object. You
+can override the <code>[lerp()](http://docs.flutter.io/flutter/material/Tween/lerp.html)</code>
+(or <em>linear interpolation</em>) method for the
+<code>[Tween](http://docs.flutter.io/flutter/material/Tween-class.html)</code> object, which
+returns a value based on a given animation clock time.
+
+Use the subclasses of <code>[Tween](http://docs.flutter.io/flutter/material/Tween-class.html)</code>
```
jimbeveridge: You are giving a list to memorize without the reader having seen anything that works yet. How is the reader expected to have any context to understand what you are describing? The behavior of a Tween makes no sense at all unless you see the pieces put together (This is based on experience interviewing other developers to write the original document. This isn't simply my opinion.)

Hixie:
On a purely stylistic note, I'd really much rather we used backticks and triple-backticks in the markdown files, instead of HTML tags. The reason is that we are likely to want to render these files using Flutter at some point, and we don't support HTML tags in Flutter, only pure markdown. (Supporting HTML tags is a huge can of worms that we would really like to avoid forever if possible.)

Hixie:
@qchong Is it ok if we close the PR for now, since it's no longer awaiting review, and then reopen it when it's ready again?

qchong:
@Hixie Yes, please go ahead and close this. I'm in the process of creating a separate PR with updates.

Hixie:
cool thanks!

PR #214, opened 2016-04-05
update links to the generated docs

devoncarew:
Update links to the generated documentation (which, there are a lot of them).

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -16,7 +16,7 @@
           <a class="page-link" href="{{ page.url | prepend: site.baseurl }}">{{ page.nav_title }}</a>
           {% endif %}
         {% endfor %}
-        <a class="page-link" href="http://docs.flutter.io/flutter/">API</a>
+        <a class="page-link" href="http://docs.flutter.io/">API</a>
         <form action="/search/" class="nav-searchbox" id="cse-search-box" style="display: inline">
           <input type="hidden" name="cx" value="007067728241810524621:gm6vraqlc8c">
           <input type="hidden" name="ie" value="UTF-8">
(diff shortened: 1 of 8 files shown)


abarth:
@sethladd @Hixie

abarth:
Looking at docs.flutter.io, it's slightly confusing that flutter_test and material are listed as peers but the way you include them in your app is different.  For example, to include flutter_test, I need to edit my pubspec.yaml file whereas I can include material without additional ceremony.

From the bug, it looks like you all discussed this approach and decided to go this direction, which is fine with me.

devoncarew:
Some things we can tweak here:

- we can use `/// <nodoc>` to hide individual libraries
- we can not doc entire packages (there are a few ways to exclude them)
- we could extend dartdoc to know about library categories (introducing a new dartdoc metadata tag)

@keertip

abarth:
Maybe other folks have opinions, but what might work well is if the left navigation bar had two level of hierarchy:

```
flutter
  material
  widgets
flutter_test
  flutter_test
```

With the top-level being packages and the second level being libraries within the package.

abarth:
I didn't mean to derail your patch to fix the broken links on the web site....  Maybe it would be better to generate the current page at docs.flutter.io/flutter and have the root page redirect there for the time being.  That keeps the URLs more sensible so that `package:flutter/material.dart` has the URL `http://docs.flutter.io/flutter/material/material-library.html` rather than `http://docs.flutter.io/material/material-library.html`

devoncarew:
Yeah, I think with that change we would have to update any links.

devoncarew:
re-swizzled where we upload: https://codereview.chromium.org/1863813002

PR #215, opened 2016-04-06
Allow some more => expressions

Hixie:
Derived from discussion in https://github.com/flutter/flutter/pull/3130

@abarth, @HansMuller


Hixie:
I suspect travis won't like this, because of the `@override`s

abarth:
LGTM

Hixie:
I've `<!--skip-->`ped the entire code block, there were tons of complaints because this is a code snippet out of context.

PR #219, opened 2016-04-13
rename tutorial

sethladd:
Closes https://github.com/flutter/flutter/issues/3309

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -14,7 +14,7 @@
   	},
     {
       "source" : "/tutorial/",
-      "destination" : "/widget-intro/",
+      "destination" : "/widgets-intro/",
       "type" : 301
     }
   ]
--- a/firebase.json
+++ b/firebase.json
@@ -14,7 +14,7 @@
   	},
     {
       "source" : "/tutorial/",
-      "destination" : "/introducing-flutters-widget-framework/",
+      "destination" : "/widget-intro/",
       "type" : 301
     }
   ]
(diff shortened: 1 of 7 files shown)


sethladd:
cc @qchong

firebase.json
```diff
@@ -11,6 +11,11 @@
   	  "source" : "/widgets/",
   	  "destination" : "/basic-widgets/",
   	  "type" : 301
-  	}
+  	},
+    {
+      "source" : "/tutorial/",
+      "destination" : "/introducing-flutters-widget-framework/",
```
Hixie: that's a bit of a mouthful if we're sending it around... Could we have a shorter URL?

Hixie: e.g. "widgets-intro"

introducing-flutters-widget-framework.md
```diff
@@ -1,7 +1,7 @@
 ---
 layout: page
-title: Flutter Tutorial
-permalink: /tutorial/
+title: Introducing Flutter's Widget Framework
```
Hixie: I think I'd prefer "Introduction to..."

Hixie:
LGTM

firebase.json
sethladd: Trying to optimize for SEO, I'm assuming people are cutting and pasting, or searching and clicking, and not manually typing these paths.

Hixie: I wouldn't worry too much about SEO. If the page is useful, it'll rise in the rankings. The URL isn't going to make any noticeable difference to the rankings in practice.

sethladd: Renamed.

PR #221, opened 2016-04-19
show rootBundle in action

sethladd:
Closes https://github.com/flutter/flutter/issues/3410

### DIFF
--- a/assets-and-images.md
+++ b/assets-and-images.md
@@ -57,7 +57,7 @@ different locales or regions, reading directions, etc.
 
 ## Working with assets
 
-Your application can access its assets through as
+Your application can access its assets through an
 [AssetBundle](http://docs.flutter.io/flutter/services/AssetBundle-class.html).
 The two main methods on an asset bundle allow you to load a string
 (`loadString`) or an image (`loadImage`) out of the bundle, given a logical key.
--- a/assets-and-images.md
+++ b/assets-and-images.md
@@ -4,10 +4,12 @@ title: Assets and Images
 permalink: /assets-and-images/
 ---
 
-Flutter applications include both code and assets (sometimes called resources).
-Assets can include static data, configuration files, or anything else an
-application needs to function; in particular, assets include icons and other
-images displayed in the UI.
+Flutter applications can include both code and _assets_
+(sometimes called resources). An asset is a file that is bundled
+and deployed with your application, and is accessible at runtime.
+Common types of assets include static data
+(for example, JSON files), configuration files, icons,
+and images.
 
 ## Specifying assets
 
@@ -21,9 +23,11 @@ A typical `flutter.yaml` file might include something like the following:
       - assets/background.png
 
 The `assets` section specifies files that should be included with the
-application. Flutter places them in a special archive called the _asset bundle_,
-which applications can read from at runtime. It lists explicit paths (relative
-to the `flutter.yaml` file) where the asset files are located.
+application. Each asset is identified by an explicit path (relative
+to the `flutter.yaml` file) where the asset file is located.
+
+During a build, Flutter places assets into a special archive called
+the _asset bundle_, which applications can read from at runtime.
 
 ### Asset variants
 
@@ -53,17 +57,30 @@ different locales or regions, reading directions, etc.
 
 ## Working with assets
 
-Your application can access its assets through the
-[AssetBundle](http://docs.flutter.io/flutter/services/AssetBundle-class.html)
-interface. The two main methods on an asset bundle allow you to load a string
+Your application can access its assets through as
+[AssetBundle](http://docs.flutter.io/flutter/services/AssetBundle-class.html).
+The two main methods on an asset bundle allow you to load a string
 (`loadString`) or an image (`loadImage`) out of the bundle, given a logical key.
 The logical key maps to the path to the asset specified in the `flutter.yaml`
-file at build time. The application's asset bundle is accessible through the
-[DefaultAssetBundle](http://docs.flutter.io/flutter/widgets/DefaultAssetBundle/of.html)
-widget's `of` method.
+file at build time.
 
-For example, the dark background from the example above could be loaded in a
-widget's `build` method like so:
+Each Flutter application has a
+[rootBundle](http://docs.flutter.io/flutter/services/rootBundle.html)
+for easy access to the main asset bundle. You can use the `rootBundle`
+from `package:flutter/services.dart`
+to easily access configuration file assets or assets that don't need to be
+aware of a `BuildContext`.
+
+For example, use the `rootBundle` to easily
+load a JSON file asset.
+
+    String assetContents = await rootBundle.loadString('assets/config.json');
+
+You can also access a context-aware `AssetBundle` with
+[DefaultAssetBundle](http://docs.flutter.io/flutter/widgets/DefaultAssetBundle/of.html).
+
+For example, the dark background from the asset declarations above
+can be loaded in a widget's `build` method like so:
 
     Widget build(BuildContext context) {
       ...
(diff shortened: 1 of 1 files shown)


sethladd:
cc @krisgiesing @qchong

assets-and-images.md
```diff
@@ -53,17 +57,30 @@ different locales or regions, reading directions, etc.
 
 ## Working with assets
 
-Your application can access its assets through the
-[AssetBundle](http://docs.flutter.io/flutter/services/AssetBundle-class.html)
-interface. The two main methods on an asset bundle allow you to load a string
+Your application can access its assets through as
```
Hixie: "as" should probably be "an"

sethladd: thanks. updated in the next commit.

assets-and-images.md
```diff
@@ -53,17 +57,30 @@ different locales or regions, reading directions, etc.
...
+from `package:flutter/services.dart`
+to easily access configuration file assets or assets that don't need to be
+aware of a `BuildContext`.
+
+For example, use the `rootBundle` to easily
+load a JSON file asset.
+
+    String assetContents = await rootBundle.loadString('assets/config.json');
+
+You can also access a context-aware `AssetBundle` with
```
Hixie: i would reorder this so that you're introduced to the way to get assets out of bundles in the context of a widget tree first, and then the rootBundle is exposed to you as the way to do it when you have no widget tree.

Otherwise, people will get to the rootBundle and just always use that, and DefaultAssetBundle won't work.

Hixie:
cc @yjbanov for the travis failure

Hixie:
LGTM

assets-and-images.md
sethladd: That's fair, but I don't understand why I want an asset bundle associated with a context. Nor do I know how to associate a bundle with a context. So it's hard for me to sell the longer `DefaultAssetBundle.of(context).loadString(...)`

Hixie: The easy case to sell is for loadImage, because then it gets you the right resolution image for free.

assets-and-images.md
```diff
@@ -4,10 +4,12 @@ title: Assets and Images
...
 
-Flutter applications include both code and assets (sometimes called resources).
-Assets can include static data, configuration files, or anything else an
-application needs to function; in particular, assets include icons and other
-images displayed in the UI.
+Flutter applications can include both code and _assets_
+(sometimes called resources). An asset is a file that is bundled
+and deployed with your application, and is accessible at runtime.
+Common types of assets include static data
+(for example, JSON files), configuration files, icons,
```
qchong: Are audio files also assets?

assets-and-images.md
```diff
@@ -53,17 +57,30 @@ different locales or regions, reading directions, etc.
 
 ## Working with assets
 
-Your application can access its assets through the
-[AssetBundle](http://docs.flutter.io/flutter/services/AssetBundle-class.html)
-interface. The two main methods on an asset bundle allow you to load a string
+Your application can access its assets through an
+[AssetBundle](http://docs.flutter.io/flutter/services/AssetBundle-class.html).
```
qchong: Suggest: "through an [`AssetBundle`](http://docs.flutter.io/flutter/services/AssetBundle-class.html) object."

Please use code fonts when referring to classes, methods, and constants.

assets-and-images.md
```diff
@@ -53,17 +57,30 @@ different locales or regions, reading directions, etc.
...
+The two main methods on an asset bundle allow you to load a string
 (`loadString`) or an image (`loadImage`) out of the bundle, given a logical key.
```
qchong: Does the extracted string represent the content of the data/config file? If so, please say this explicitly. What happens when the loadString() method is called on a image or icon key, or if loadImage() is called on a string-only data file?

assets-and-images.md
```diff
@@ -53,17 +57,30 @@ different locales or regions, reading directions, etc.
...
 The logical key maps to the path to the asset specified in the `flutter.yaml`
-file at build time. The application's asset bundle is accessible through the
-[DefaultAssetBundle](http://docs.flutter.io/flutter/widgets/DefaultAssetBundle/of.html)
-widget's `of` method.
+file at build time.
+
+Each Flutter application has a
```
qchong: s/application/app/

In general, "app" is the preferred term when referring to mobile applications and is shorter to read. Please make this fix in the rest of the doc.

assets-and-images.md
```diff
@@ -53,17 +57,30 @@ different locales or regions, reading directions, etc.
...
+[rootBundle](http://docs.flutter.io/flutter/services/rootBundle.html)
```
qchong: `rootBundle` object. Is there a reason why this classname starts with a lowercase 'r'?

assets-and-images.md
krisgiesing: Yes, although this document doesn't talk much about how audio is handled... Do we have a doc for audio?

assets-and-images.md
```diff
@@ -53,17 +57,30 @@ different locales or regions, reading directions, etc.
...
+for easy access to the main asset bundle. You can use the `rootBundle`
```
krisgiesing: As far as I know, rootBundle is the *only* way to access the main asset bundle...

assets-and-images.md
```diff
@@ -53,17 +57,30 @@ different locales or regions, reading directions, etc.
...
+load a JSON file asset.
+
+    String assetContents = await rootBundle.loadString('assets/config.json');
+
+You can also access a context-aware `AssetBundle` with
+[DefaultAssetBundle](http://docs.flutter.io/flutter/widgets/DefaultAssetBundle/of.html).
 
-For example, the dark background from the example above could be loaded in a
-widget's `build` method like so:
+For example, the dark background from the asset declarations above
```
qchong: Revise to make this sentence less passive:
"For example, your app can load the dark background from the asset declarations above .."

assets-and-images.md
sethladd: done.

assets-and-images.md
sethladd: I used the term object.

assets-and-images.md
sethladd: done.

assets-and-images.md
sethladd: I'm going to omit audio, because I hope the statement above (a file that is bundled and deployed with your app) is sufficient. Otherwise, it's going to be a very long list :)

assets-and-images.md
Hixie: There's nothing magical about `rootBundle`. If you call `takeRootBundle` before the `rootBundle` property is initialised, then you can use mojo to read the main asset bundle yourself.

assets-and-images.md
sethladd: Tried to make this a little clearer.

assets-and-images.md
sethladd: Bad things happen? :) Not sure what else to say here, I think "string/text asset" and "image asset" are pretty clear. But suggestions welcome.

assets-and-images.md
sethladd: done.

assets-and-images.md
sethladd: I used "main asset bundle" colloquially. I don't know what the formal definition of that is. It's not clear to me how asset bundles are organized (is it a tree?) or associated with widgets or widget trees.

krisgiesing: If you call `takeRootBundle` before `rootBundle` is initialized, then `rootBundle` will initialize to null, right? Then all kinds of bad things will probably happen elsewhere in the framework.

Hixie: As of recently it'll initialise to a network asset bundle. But yes. However, nothing requires you to use the rest of the framework.

PR #222, opened 2016-04-20
docs: Update docs to reflect hiding of routes per PR# 3431.

qchong:
@Hixie ptal


Hixie:
not sure what's up with the travis failure, but modulo that, lgtm

PR #224, opened 2016-04-21
Add navigation + style updates to Flutter.io site

qchong:
Fixes https://github.com/flutter/flutter/issues/3414 by adding an expandable sidenav.

@sethladd ptal


sethladd:
Thanks for working on this and getting it working. Next time, before merge, please wait for a review and LGTM.

basic-widgets.md
```diff
@@ -1,9 +1,15 @@
 ---
 layout: page
-title: Basic Widgets Overview
+title: Widgets in Flutter
```
sethladd: This page isn't a comprehensive doc on widgets, so I think it's important that we set expectations that it's just a basic overview. There's a LOT more to widgets than this doc. Can we consider a title that sets the correct expectation? The previous title was chosen after discussing the scope of the doc.

PR #226, opened 2016-04-21
docs: Reduced size of the 'external link' icon.

qchong:
Fixes https://github.com/flutter/flutter/issues/3477


sethladd:
Do we have a way to stage changes like this? I can help set that up if you like.

sethladd:
lgtm

PR #228, opened 2016-04-27
add info about disabling analytics

devoncarew:
- add info about disabling analytics (fix https://github.com/flutter/flutter/issues/3537)

@Hixie @sethladd @qchong

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -41,6 +41,25 @@ itself. Subsequent runs should be much faster.
 Once you have installed any missing dependencies, run the `flutter doctor` command again to
 verify that you’ve set everything up correctly.
 
+## Google Analytics
+
+The Flutter tool uses Google Analytics to anonymously report feature usage statistics
+and basic crash reports. This data is used to help improve Flutter tools over time.
+In order to disable reporting, type:
+
+```
+flutter config --no-analytics
+```
+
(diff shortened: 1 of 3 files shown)


sethladd:
lgtm (modular travis being sad)

Thanks!

devoncarew:
Looks like I fixed a reference for `master` and travis is analyzing against `alpha`.

setup.md
```diff
@@ -41,6 +41,25 @@ itself. Subsequent runs should be much faster.
...
+```
+
+And to display the current state of analytics reporting:
+
+```
+flutter config
+```
+
+See Google's privacy policy:
+[www.google.com/intl/en/policies/privacy](https://www.google.com/intl/en/policies/privacy/).
```
Hixie: Man, you really want to put a dot at the end of that URL!

setup.md
```diff
@@ -41,6 +41,25 @@ itself. Subsequent runs should be much faster.
...
+
```
Hixie: Maybe explicitly state that the very first run, and any runs involving `flutter config`, are not logged.

setup.md
devoncarew: But it's totally not in the hyperlink here :) - people will end up in the right place

PR #229, opened 2016-04-28
default jekyll build location, default to staging server on non-maste…

sethladd:
…r merges, add firebase rules

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
-language: dart
-dart:
-  - dev
+language: node_js
+node_js:
+  - "5.1"
 sudo: false
 before_install: ./tool/before_install.sh
 script: ./tool/travis.sh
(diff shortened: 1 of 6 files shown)


sethladd:
cc @devoncarew

devoncarew:
lgtm

I'm having trouble parsing the travis failure. We may need to change our test dep to `any`? https://github.com/flutter/website/blob/master/pubspec.yaml#L11

sethladd:
Yeah, pub has a bug where it won't tell you if the reason it can't download a package is because of an SDK constraint. I suspect the real issue is a SDK constraint mismatch between test and 1.16-dev.

devoncarew:
Yeah, I just repro'ed it locally after upgrading to `1.17.0-dev.1.0`.

sethladd:
Ah, thanks. Except, our travis script is trying to use Flutter's version of Dart.

devoncarew:
We hit issues with `test` every time the sdk version revs.

devoncarew:
Should build green now w/ https://github.com/flutter/website/pull/230.

PR #230, opened 2016-04-28
Update .travis.yml

devoncarew:
Build with the stable version of the dart sdk.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: dart
 dart:
-  - dev
+  - stable
 sudo: false
 before_install: ./tool/before_install.sh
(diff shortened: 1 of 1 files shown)


sethladd:
lgtm  (but see also https://github.com/flutter/website/blob/master/tool/travis.sh#L8 as I think that's what we really want?)

devoncarew:
This gets us past the initial problem w/ pub get, the dev sdk, and `test`.

sethladd:
Thanks! As discussed over VC, we will remove the "language: dart" from this project, so we can always ensure/enforce that we're using Flutter's vendored Dart SDK.

PR #234, opened 2016-05-03
Fix CSS to set font family for headers to 'Roboto' 

qchong:
Fixes https://github.com/flutter/flutter/issues/3590

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
-language: dart
-dart:
-  - stable
+language: node_js
+node_js:
+  - "5.1"
 sudo: false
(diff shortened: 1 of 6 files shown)


devoncarew:
lgtm

What do we get from the lavish-bootstrap css? This changed lots of things about the website - like fonts and spacing. We're all the changes intentional? If no, perhaps we can extract out the specific css for the left nav, and just include that.

qchong:
The lavish-bootstrap.css comes from [here](http://www.lavishbootstrap.com/). It's just an easier way to customize our website elements. I'll look into whether we can slim down the CSS file, but at a later time.

PR #241, opened 2016-05-19
Switched order of 'Get Started' and 'Setup' in nav sidebar.

qchong:
Fixes https://github.com/flutter/flutter/issues/4035

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.firebaserc
 .DS_Store
 _site
 .sass-cache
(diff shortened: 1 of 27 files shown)


sethladd:
Thanks! There's a lot of commits in here. I wonder if we can help you branch off of an updated version of master, so the only commits in here will be your changes (and any merges since).

stevenroose:
The changes also seem to include more than just what the title says.

sethladd:
> The changes also seem to include more than just what the title says.

We'll take a look at this PR and see if we can squash and rebase. Thanks for the head's up.

qchong:
Abandoning this pull request to retry with a squashed commit list.

PR #243, opened 2016-05-27
Update the testing document with recent changes.

Hixie:


Hixie:
@abarth 

This should fix the build failure that was preventing (I think) the image I added from showing.

abarth:
LGTM

PR #251, opened 2016-06-17
docs: Add howto guide on adding platform and third-party services in Flutter

qchong:
Fixes https://github.com/flutter/flutter/issues/4503.
Stage preview: https://flutter-io-staging-69a8a.firebaseapp.com/platform-services/

@abarth ptal.


abarth:
s/native services/platform services/

What's relevant is that the services are offered by the underlying platform.

abarth:
`hello_services_files.png` this screenshot is very likely to become out-of-date quickly.  Do we have a plan for maintaining it?  If not, we should re-work things so they won't become out of date so easily.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+This guide describes how to connect your Flutter app to platform-specific or
+third-party services that are available on iOS and Android devices.
+
+* TOC
+{:toc}
+
+## Introduction
+
+Mobile devices offer many capabilities to your app which can be accessed through
+platform-specific native APIs, or APIs from third-party vendors. Flutter offers several ways
```
abarth: I'd strike the word native here.  It doesn't add anything to this sentence.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+for you to integrate the functionality of these of these platform services to extend your Flutter
+app.
+
+### First-party services for Flutter {#first-party-services-for-flutter}
+
+For frequently used services, the Flutter SDK provides APIs to access first-party services that
+are supported across platforms. You can call these APIs directly in your Dart code. These
+first-party services are provided through the `SKY_SERVICES` library. The following list
```
abarth: `sky_services` (lower case)

s/library/package/

Also, this statement is a bit misleading.  Although its possible to use the mojoms directly, almost everyone should use the more ergonomic APIs in `package:flutter/services` instead.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+summarizes these first-party services.
+
++ [`activity_mojom`](http://docs.flutter.io/flutter/activity_mojom/activity_mojom-library.html)
```
abarth: I'd remove this one.  We're probably going to remove it.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
++ [`path_provider_mojom`](http://docs.flutter.io/flutter/path_provider_mojom/path_provider_mojom-library.html)
++ [`pointer_mojom`](http://docs.flutter.io/flutter/pointer_mojom/pointer_mojom-library.html)
++ [`raw_keyboard_mojom`](http://docs.flutter.io/flutter/raw_keyboard_mojom/raw_keyboard_mojom-library.html)
++ [`semantics_mojom`](http://docs.flutter.io/flutter/semantics_mojom/semantics_mojom-library.html)
++ [`system_chrome_mojom`](http://docs.flutter.io/flutter/system_chrome_mojom/system_chrome_mojom-library.html)
++ [`system_sound_mojom`](http://docs.flutter.io/flutter/system_sound_mojom/system_sound_mojom-library.html)
+
+### Platform and Third-party Services for Flutter {#platform-and-third-party-services-for-flutter}
+
+To access APIs that are native to the iOS or Android mobile platform, Flutter provides the
```
abarth: Please strike the word native.  Instead, we should say "To access services that are provided by the iOS or Android platforms, ..."

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+`FlutterView` interface. The `FlutterView` interface allows you to embed a bi-directional
```
abarth: This is backwards.  FlutterView is the interface your Java or Objective-C code uses to access the Flutter portion of your app.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+asynchronous messaging component in your platform code to send and receive native service calls
```
abarth: Again, the word native is meaningless here.  Please remove the word native everywhere in this document.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+in a transparent way. This flexible approach lets you create cross-platform mobile apps that can
```
abarth: What does "transparent" mean in this context?

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+connect to platform-specific and third-party services on iOS and Android devices, while sharing a
+common user interface (UI) codebase in Dart.
```
abarth: I'd strike the words "in Dart".

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+
+You can build your own integrations with platform-specific services and third-party SDKs. Ideally,
+the services that you invoke using the `FlutterView` interface should return non-graphical,
+data-only information. Examples include:
```
abarth: I'd strike this sentence about "ideally".  That's not the ideal.  That's just a limitation of our current implementation.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+
+*   Geolocation
+*   Contacts
+*   File system
+*   Network information
+*   Sensors
+
+You can also use the `FlutterView` interface to invoke APIs from third-party SDKs that run on iOS
+and Android devices.
```
abarth: In all these places that you refer to `FlutterView`, I think it would be more accurate to refer to HostMessages: http://docs.flutter.io/flutter/services/HostMessages-class.html  That's the interface that you actually use to call out to the platform-specific part of your applciation.  `FlutterView` is the interface that goes the other way, from the platform-specific part of your application to the cross-platform UI.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+
+To use the `FlutterView` interface, you must add a platform-specific Flutter component into the
+platform’s view hierarchy. For example, on iOS, you add a `FlutterViewController` object to
+interact with a `UIView`.  Similarly, on Android, you add a `FlutterView` object to your app’s
+`android.view.View` hierarchy. Your app can then communicate between the Dart and iOS or
```
abarth: The word Dart here isn't parallel.  If you want to say Dart, then the parallels on iOS are Objective-C and Swift and on Android is Java.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+Android-specific parts of the app by using a simple message-passing interface.
+
+### Message passing {#message-passing}
+
+Flutter offers a simple message-passing interface between your platform-specific code and the
+Dart code. The message passing interface is asynchronous, bi-directional, and supports optional
```
abarth: I'd replace Dart here with UI

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+reply routing. Messages consist of strings encoded in JSON format, and these is no facility for
+transferring handles to other objects along with messages. You are responsible for encoding and
```
abarth: I'd strike the part of this sentence starting with "and".  That's implied by the JSON format.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+decoding messages sent through this mechanism.
```
abarth: That's not entirely true (and likely to become less true over time).  For example, the HostMessges interface has sendJSON (http://docs.flutter.io/flutter/services/HostMessages/sendJSON.html) that does the encoding for you.  It's likely we'll provide similar helpers on the Java and Objective-C sides.  I'd strike this sentence given that it's unlikely to be true in the future.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Get the `hello_services` sample {#the-hello_services-sample}
```
abarth: It's an "example" rather than a "sample".  I'm not sure what the difference between those words is, but it's in a directory called "examples" rather than a directory called "samples".

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+
+The Flutter SDK includes a reference sample project called
+[`hello_services`](https://github.com/flutter/flutter/tree/master/examples/hello_services)
+that demonstrates how to connect to the geolocation service within an iOS or Android application
+by using the `FlutterView` interface.
+
+The `hello_services` project comes with helper code to handle the passing of
+JSON-formatted messages to invoke platform services. If your app needs access to functionality
+provided by platform-specific iOS or Android services, we strongly encourage you to
```
abarth: I don't think we "strongly encourage" that.  That's certainly one option, but we're likely to create a better onboarding experience in the future (e.g., via `flutter create`).  Instead, I would say that you "can" do that.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+sample project as the starter code to build your own Flutter app.
+
+The following diagram shows how the `hello_services` app uses the `FlutterView` interface to
+send and receive messages to platform and third-party services on iOS and Android.
+
+ ![hello_services architecture](/images/hello_services.png)
+
+**Fig 1.** Architecture of the `hello_services` sample.
+
+The Dart, iOS, and Android-specific components are related as follows:
```
abarth: Again, "Dart" isn't parallel with "iOS" and "Android".  The languages aren't as important, so I'd replace "Dart" with something descriptive about the function of that part of your app instead.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+
+* In the Dart code, the sample app imports the
+[`flutter/services.dart`](http://docs.flutter.io/flutter/services/services-library.html)
+package. This package defines the
+[`HostMessages`](http://docs.flutter.io/flutter/services/HostMessages-class.html) class,
+which provides a bridge to send and listen for JSON-encoded messages exchanged with a platform
```
abarth: send and listen are not parallel.  Instead, consider "send and receive".

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+service on a <em>host</em> app (that is, the iOS or Android-specific parts of your app). The
```
abarth: This sentence is confused about whether the platform-specific part of your app is part of your app or a separate thing.  I'd recommend using constructions that imply that these are all pieces of a coherent app.  For example, "... messages are exchanged with a service in the host portion of your app (that is, the part of your app that is specific to iOS or Android)"

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+<code>hello_services</code> app uses the static methods of the
```
abarth: Why `<code>` here and backticks in other places?

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+[`HostMessages`](http://docs.flutter.io/flutter/services/HostMessages-class.html) class
+in the [`State`](http://docs.flutter.io/flutter/material/State-class.html) object of a
+stateful widget to make services requests, then updates the UI accordingly.
+
+* On iOS, the sample app imports the Flutter framework library (`Flutter.h`) as a
+dependency. The library defines the <code>FlutterViewController</code> class, which is a
```
abarth: The word "dependency" doesn't mean anything here.  Is there some other way to import things?  I'd strike the part of the sentence starting with "as a".

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+[`UIViewController`](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIViewController_Class/)
+that provides a bridge from the
+[`UIKit`](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIKit_Framework/)
+view hierarchy to the Flutter app. The `hello_services` app sets an instance of
```
abarth: Again, I'd recommend constructions that imply that the whole thing is one app with various parts.  For example, "... view hierarchy to the Flutter part of the app".

platform-services.md
abarth: Also, why `<code>` here instead of backticks as in other places?

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+[`FlutterViewController`](https://github.com/flutter/engine/blob/master/sky/shell/platform/ios/framework/Headers/FlutterViewController.h)
+as the root view controller of its [`UIWindow`](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIWindow_Class/).
+
+* On Android, the sample imports the `flutter.jar` file as a dependency. This JAR
```
abarth: This bullet is not parallel with the previous bullet.  `flutter.jar` is the build artifact, not the interface, but `Flutter.h` is the interface, not the build artifact (which is `Flutter.framework`).

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+defines the [`FlutterView`](https://github.com/flutter/engine/blob/master/sky/shell/platform/android/io/flutter/view/FlutterView.java)
+class, which is an [`android.view.View`](https://developer.android.com/reference/android/view/View.html)
+subclass that provides a bridge from the Android view hierarchy to the Flutter app. The
```
abarth: Again, I'd prefer a construction that implies that it's all part of one app.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+`hello_services` app creates an instance of [`FlutterView`](https://github.com/flutter/engine/blob/master/sky/shell/platform/android/io/flutter/view/FlutterView.java)
+in its main activity.
```
abarth: To be parallel, I'd stay something about setting it as the content view of the activity.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+
+### App structure {#app-structure}
+
+The `hello_services` sample app has the following project file structure.
+
+ ![hello_services sample files](/images/hello_services_files.png)
+
+**Fig. 2.** Project file structure for the `hello_services` sample.
+
+To make platform-specific service calls, you may need to modify these key files and folders in the `hello_services` project:
```
abarth: Do you mean "to create additional platform-specific services" ?  To make a call, you need only invoke one of the existing interfaces.

Also, I'd strike the word "key".  It doesn't meany anything in this content.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+file handles message passing between iOS and the Flutter app, while the [`LocationProvider.m`](https://github.com/flutter/flutter/blob/master/examples/hello_services/ios/HelloServices/LocationProvider.m)
+file implements the code that queries the iOS system location API
+([`CLLocationManager`](https://developer.apple.com/library/ios/documentation/CoreLocation/Reference/CLLocationManager_Class/)).
+
+* `ios/Flutter/Generated.xcconfig`: Contains the configuration settings to build your iOS app. See
+the `hello_services` [README](https://github.com/flutter/flutter/blob/master/examples/hello_services/README.md)
+to learn more about the properties you can configure.
+
+* [`lib/main.dart`](https://github.com/flutter/flutter/blob/master/examples/hello_services/lib/main.dart):
+The main entry point for your Flutter app. This source file is written in Dart syntax. The main
```
abarth: Again, "the Flutter part of your app".  Also, this section is confused about whether you're talking about `hello_services` or the developer's app.  Here you talk about "your app", but you're referencing hello_services specifically, which doesn't belong to the person reading this document.

"This source file is written in Dart syntax."  I'd strike this sentence.  If you really want to keep it, you should make it parallel to the other constructions and talk about the "Dart source files for the cross-platform part of your app".

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+function uses the [`HostMessages`](http://docs.flutter.io/flutter/services/HostMessages-class.html)class,
+which provides a high-level abstraction for passing messages to “hosts” of platform-specific
+services.
+
+## Add a service to your app {#add-a-service-to-your-app}
+
+In your Dart code (`lib/main.dart` file):
+
+1. Define a method to send a request to your target service (e.g. `_getLocation()`). You must encode
+the request as a JSON message. You’ll typically invoke this method in a
```
abarth: Not necessarily.  You can use the `sendJSON` function.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+[`State`](http://docs.flutter.io/flutter/material/State-class.html) object, when the user performs
+an operation on the UI.
+2. Define a messageHandler to process the messages returned by the target service
+(e.g. `handleGetRandom()`), then add this handler to the
+[`HostMessages`](http://docs.flutter.io/flutter/services/HostMessages-class.html) object.
```
abarth: This step is for a completely different thing: exposing services from the cross-platform part of your app to the platform-specific part of your app.  I'd recommend removing this from this document.  It's just confusing the issue.  We might want to remove it from the example as well.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+
+In your iOS code:
+
+1. In Xcode, open the `HelloServices` iOS project.
+2. Implement a service provider (e.g. `LocationProvider`). The provider class must conform to the
+`FlutterMessageListener` protocol (specifically, you must implement the
+`(NSString*)didReceiveString:(NSString*)` function). If your app wants to connect asynchronously to
+the service, use the [`FlutterAsyncMessageListener`](https://github.com/flutter/engine/blob/master/sky/shell/platform/ios/framework/Headers/FlutterAsyncMessageListener.h)
```
abarth: First you say that it must conform to FlutterMessageListener but then you say that FlutterAsyncMessageListener is also an option.  That's contradictory.

Also, it's unclear what "your app" means in this context.  Do you mean the cross-platform part of your app?  In that case, the sentence is incorrect because the cross-platform part of your app (i.e., the part written in Dart that implements the UI) can't tell whether the platform-specific part of your app uses FlutterMessageListener or FlutterAsyncMessageListener.  In both cases, the interaction between the two is asynchronous.

Instead, it's you should use FlutterMessageListener if (1) the message does not require a reply or (2) you can reply to the message synchronously.  FlutterAsyncMessageListener is needed only when (1) the message requires a reply and (2) the code that creates the reply is asynchronous.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+class. Your provider is where you put the business logic for your service calls.
+3. In your [`AppDelegate`](https://github.com/flutter/flutter/blob/master/examples/hello_services/ios/HelloServices/AppDelegate.m),
+initialize the service provider you defined earlier then add a message listener to that provider.
+
+In your Android code:
+
+1. In your main activity (e.g.
+[`ExampleActivity.java`](https://github.com/flutter/flutter/blob/master/examples/hello_services/android/app/src/main/java/com/example/flutter/ExampleActivity.java)),
+add a message listener to listen for your service calls.
+2. Add the code to marshal and unmarshal the JSON messages for your service calls.
```
abarth: This same thing is required in the iOS part of your app but you don't mention it there.  We should either mention it in both places or in neither place.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+3. Add code to handle the business logic for your service calls.
```
abarth: Again, this step doesn't exist in the iOS instructions.  The two sets of instructions should be exactly parallel because they're describing a very similar sequence of steps in two different settings.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+`FlutterMessageListener` protocol (specifically, you must implement the
+`(NSString*)didReceiveString:(NSString*)` function). If your app wants to connect asynchronously to
+the service, use the [`FlutterAsyncMessageListener`](https://github.com/flutter/engine/blob/master/sky/shell/platform/ios/framework/Headers/FlutterAsyncMessageListener.h)
+class. Your provider is where you put the business logic for your service calls.
+3. In your [`AppDelegate`](https://github.com/flutter/flutter/blob/master/examples/hello_services/ios/HelloServices/AppDelegate.m),
+initialize the service provider you defined earlier then add a message listener to that provider.
+
+In your Android code:
+
+1. In your main activity (e.g.
```
abarth: Why did you tell me to use Xcode when editing the iOS portion of the app but you did not specify what editor to use for the equivalent Android step?  Either we should recommend Android Studio here or we should not recommend Xcode above.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+[`ExampleActivity.java`](https://github.com/flutter/flutter/blob/master/examples/hello_services/android/app/src/main/java/com/example/flutter/ExampleActivity.java)),
+add a message listener to listen for your service calls.
+2. Add the code to marshal and unmarshal the JSON messages for your service calls.
+3. Add code to handle the business logic for your service calls.
+4. In the AndroidManifest.xml, make sure to add permissions for your target service
+(e.g. [`ACCESS_FINE_LOCATION`](https://developer.android.com/reference/android/Manifest.permission.html#ACCESS_FINE_LOCATION)).
```
abarth: In the iOS instructions, we talk about the sync versus async interfaces.  That same distinction exists for Android as well.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Build your app {#build-your-app}
+
+If you are using the `hello_services` sample as your starting point, you should build your app
+directly from the platform-specific tools.  Specifically, on iOS, you build your app using
```
abarth: That statement is true regardless of whether you're using the hello_services example as your starting point.

platform-services.md
```diff
@@ -0,0 +1,215 @@
...
+Xcode and, on Android, you  build your app using
+[Android Studio](https://developer.android.com/studio/index.html) (recommended) or
```
abarth: Why do we recommend Android Studio over the gradle command line?  It seems like either is equally valid and we shouldn't take a position on that topic.

platform-services.md
qchong: removed

qchong:
Staged preview:
[https://flutter-io-staging-69a8a.firebaseapp.com/platform-services/](https://flutter-io-staging-69a8a.firebaseapp.com/platform-services/)

@abarth Thanks for your previous review comments. I believe I've resolved all the issues you raised. Could you ptal at the latest text in this PR update.?

@jason-simmons Adding you as a doc reviewer. Ptal, thanks!

PR #263, opened 2016-08-18
docs: Update FAQ to indicate Flutter does not support tablets.

qchong:
@sethladd ptal.

Fixes https://github.com/flutter/flutter/issues/5480


sethladd:
lgtm!

PR #264, opened 2016-08-22
Use trailing coma and dartfmt on snippets

a14n:


sethladd:
thanks! Should we add a step in Travis to enforce we've formatted the examples? I'd be cool with it.

a14n:
I just added a step in travis to check that the examples are correctly formatted.

sethladd:
Nice!! Thank you :)

lgtm

PR #265, opened 2016-08-22
doc: Update code sample for File Read/Write Howto Guide.

qchong:
Fixes https://github.com/flutter/flutter/issues/5518

Staged preview:
https://flutter-io-staging-69a8a.firebaseapp.com/reading-writing-files/

### DIFF
--- a/reading-writing-files.md
+++ b/reading-writing-files.md
@@ -19,41 +19,94 @@ access commonly used locations on the device's filesystem. The class currently s
 
 Once your Flutter app has a reference to a file location, you can use the [dart:io](https://api.dartlang.org/stable/1.18.1/dart-io/dart-io-library.html) APIs to perform read/write operations to the filesystem. For more information about working on files and directories using Dart, see this [overview](https://www.dartlang.org/articles/dart-vm/io) and [these examples](https://www.dartlang.org/dart-vm/dart-by-example#files-directories-and-symlinks).
 
-## Writing a string to a file
-Use the File object’s [writeAsString()](https://api.dartlang.org/stable/1.18.1/dart-io/File/writeAsString.html) method to write a string to a file. After writing the string, the method closes the file.
+## Example of reading and writing to a file
+
+The following example shows how to modify the default app created by `flutter create`, so that
+the app persists the number of button presses across relaunches.
 
 ```dart
 import 'dart:io';
 import 'dart:async';
 import 'package:flutter/services.dart';
-import 'package:flutter/widgets.dart';
+import 'package:flutter/material.dart';
 
-Future<Null> main() async {
-  runApp(null); // this line is required to initialize the shell
-  String dir = (await PathProvider.getApplicationDocumentsDirectory()).path;
-  String filename = "$dir/foo.txt";
-  File file = new File(filename);
-  await file.writeAsString('some content');
-}
-```
-
-## Reading a file as a string
-Use the File object's [`readAsString()`](https://api.dartlang.org/stable/1.18.1/dart-core/Resource/readAsString.html) method to read a file as a string, then display it on the screen.
-
-```dart
-import 'dart:io';
-import 'dart:async';
-import 'package:flutter/services.dart';
-import 'package:flutter/widgets.dart';
-
-Future<Null> main() async {
-  runApp(null); // this line is required to initialize the shell
-  String dir = (await PathProvider.getApplicationDocumentsDirectory()).path;
-  String filename = "$dir/foo.txt";
-  File file = new File(filename);
-  String contents = await file.readAsString();
+void main() {
   runApp(
-    new Center(child: new Text('Contents: $contents'))
+    new MaterialApp(
+      title: 'Flutter Demo',
+      theme: new ThemeData(
+        primarySwatch: Colors.blue,
+      ),
+      home: new FlutterDemo(),
+    ),
   );
+}
+
+class FlutterDemo extends StatefulWidget {
+  FlutterDemo({Key key}) : super(key: key);
+
+  @override
+  _FlutterDemoState createState() => new _FlutterDemoState();
+}
+
+class _FlutterDemoState extends State<FlutterDemo> {
+  int _counter;
+
+  @override
+  void initState() {
+    super.initState();
+    _readCounter().then((int value) {
+      setState(() {
+        _counter = value;
+      });
+    });
+  }
+
+  Future<File> _getLocalFile() async {
+    String dir = (
+      await PathProvider.getApplicationDocumentsDirectory()
+    ).path;                // get the path to the document directory
+    String filename = "$dir/counter.txt";
+    return new File(filename);
+  }
+
+  Future<int> _readCounter() async {
+    try {
+      File file = await _getLocalFile();
+      String contents = await file.readAsString();
+                            // read the variable as a string from the file
+      return int.parse(contents);
+    } on FileSystemException {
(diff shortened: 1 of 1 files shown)


qchong:
@collinjackson, @sethladd  ptal

reading-writing-files.md
```diff
@@ -19,41 +19,91 @@ access commonly used locations on the device's filesystem. The class currently s
...
-  String dir = (await PathProvider.getApplicationDocumentsDirectory()).path;
-  String filename = "$dir/foo.txt";
-  File file = new File(filename);
-  await file.writeAsString('some content');
-}
-```
-
-## Reading a file as a string
-Use the File object's [`readAsString()`](https://api.dartlang.org/stable/1.18.1/dart-core/Resource/readAsString.html) method to read a file as a string, then display it on the screen.
+The following example shows how to modify the default app created by `flutter create`, so that the number of button presses are peristsed across relaunches of the app.
```
sethladd: typo on peristsed

reading-writing-files.md
```diff
@@ -19,41 +19,91 @@ access commonly used locations on the device's filesystem. The class currently s
...
+        _counter = 0;
+      });
+    }
+  }
+
+  void _incrementCounter() {
+    setState(() {
+      _counter++;
+      _getLocalFile().then((File file) {
+        file.writeAsStringSync("$_counter");
```
sethladd: maybe add a comment here to draw the eye? Something like `// Here we write the variable as a string to the file`

reading-writing-files.md
```diff
@@ -19,41 +19,91 @@ access commonly used locations on the device's filesystem. The class currently s
...
+      setState(() {
+        _counter = 0;
+      });
+    }
+  }
+
+  void _incrementCounter() {
+    setState(() {
+      _counter++;
+      _getLocalFile().then((File file) {
```
sethladd: consider using `await` instead of `then()`

reading-writing-files.md
sethladd: Curious, what happens if this write fails, inside of a setState() ?

Do we want to do synchronous file I/O instead setState() ? cc @Hixie

reading-writing-files.md
```diff
@@ -19,41 +19,91 @@ access commonly used locations on the device's filesystem. The class currently s
...
+  _FlutterDemoState createState() => new _FlutterDemoState();
+}
+
+class _FlutterDemoState extends State<FlutterDemo> {
+  int _counter;
+
+  @override
+  void initState() {
+    super.initState();
+    _readCounter();
```
sethladd: The Future from _readCounter() is dropped on the flow.

Is there an expectation that state is all setup when initState is done? cc @Hixie

reading-writing-files.md
sethladd: Talking to @Hixie we would like to make this an asynchronous call, and move it outside of the setState() function. We'd like to train developers that they should do async I/O.

reading-writing-files.md
sethladd: Talked to @Hixie and we can do async setup in initState().

reading-writing-files.md
collinjackson: Per our discussion, doing it this way avoids the need to have _incrementCounter be Future<Null> function, and avoids needing to create a local variable to increment the counter if you want to make sure things happen in the right sequence.

collinjackson: Agree that the call should be async and I made that change already in https://github.com/flutter/flutter/issues/5518

reading-writing-files.md
collinjackson: The version I have in https://github.com/flutter/flutter/issues/5518 makes it more clear that there's async setup in initState going on.

reading-writing-files.md
```diff
@@ -19,41 +19,94 @@ access commonly used locations on the device's filesystem. The class currently s
...
+        _counter = value;
+      });
+    });
+  }
+
+  Future<File> _getLocalFile() async {
+    String dir = (
+      await PathProvider.getApplicationDocumentsDirectory()
+    ).path;                // get the path to the document directory
+    String filename = "$dir/counter.txt";
```
Hixie: flutter style uses single quotes rather than double quotes.
You can inline the filename here into the constructor call.
Also, by convention we usually use "final" on variables that won't change, like "dir" here

reading-writing-files.md
```diff
@@ -19,41 +19,94 @@ access commonly used locations on the device's filesystem. The class currently s
...
+    }
+  }
+
+  void _incrementCounter() {
+    setState(() {
+      _counter++;
+      _getLocalFile().then((File file) {
+        file.writeAsString("$_counter");
+                            // write the variable as a string to the file
+      });
```
Hixie: why is the writing to disk in the setState() ?

reading-writing-files.md
```diff
@@ -19,41 +19,94 @@ access commonly used locations on the device's filesystem. The class currently s
...
+
+  @override
+  Widget build(BuildContext context) {
+    return new Scaffold(
+      appBar: new AppBar(
+        title: new Text('Flutter Demo'),
+      ),
+      body: new Center(
+        child: new Text('Button tapped $_counter time${
+          _counter == 1 ? '' : 's' }.'),
```
Hixie: it would probably look better to have the string literal all on one line (with the `(` and `)` on the previous and next line)

collinjackson: This is copy pasted from https://github.com/flutter/flutter/blob/master/packages/flutter_tools/templates/create/lib/main.dart.tmpl, we should change it there too.

reading-writing-files.md
collinjackson: The idea is to ensure that the write happens after counter is incremented. Otherwise it's nondeterministic which number gets written, right?

You could increment the counter in a local variable, but I was thinking the ++ syntax is more familiar. That said, I don't feel particular strongly about it.

reading-writing-files.md
sethladd: For examples on the website, I'd defer to whatever the formatter is doing. Our developers will use the formatter that we ship, and our examples should match their expectations if they copy and paste into their projects.

reading-writing-files.md
collinjackson: I talked to Hixie about this and realized that we had a conceptual disconnect because I didn't realize setState is synchronous. Given that


Another option is to store the result of _getLocalFile to a variable somewhere. Like this:

```
void _incrementCounter() {
    setState(() {
      _counter++;
    });
    _getLocalFile().then((File file) => file.writeAsString("$_counter"));
}
```

or if you don't think Future<Null> is confusing

```
Future<Null> _incrementCounter() async {
    setState(() {
      _counter++;
    });
    (await _getLocalFile()).writeAsString("$_counter");
}
```

sethladd: I prefer returning the Future. I'd like our developers to get into the habit of understanding that if something inside their function does an async/Future thing, the only way for the caller to know if that finished or errored is to handle the returned Future.

collinjackson: @sethladd would you prefer that we add an additional await as well?

```
Future<Null> _incrementCounter() async {
  setState(() {
    _counter++;
  });
  await (await _getLocalFile()).writeAsString("$_counter");
}
```

sethladd: oh, good question!

What if we did:

```
await _getLocalFile()..writeAsString(...);
```

PR #266, opened 2016-08-24
update dates in FAQ

sethladd:
cc @qchong @Hixie


faq.md
```diff
@@ -345,9 +345,12 @@ interested in contributing code, you can start by reading our
 
 ## Should I build my next production app with Flutter?
 
-As of November 2015, Flutter is still being developed heavily and is not yet at
-1.0. While lower levels of the system are changing less, we envision changing
-parts of the system based on early adopter feedback.
+Flutter is still being developed and is not yet at
+1.0. While lower levels of the system are stabilizing, we envision changing
```
qchong: Suggest replacing "we envision changing" with "we continue to improve".

sethladd: cool, will do

sethladd:
Will fix travis after review of text is complete.

faq.md
```diff
@@ -345,9 +345,12 @@ interested in contributing code, you can start by reading our
...
 
-As of November 2015, Flutter is still being developed heavily and is not yet at
-1.0. While lower levels of the system are changing less, we envision changing
-parts of the system based on early adopter feedback.
+Flutter is still being developed and is not yet at
+1.0. While lower levels of the system are stabilizing, we continue to improve
+parts of the system based on user feedback.
+
+Flutter is used inside of Google, but those apps are not yet deployed to external
+users.
```
Hixie: I feel this new text doesn't really explicitly answer the question it is associated with. We should either add a sentence saying something like "So... it's up to you", or, rephrase the question somewhat. I'd lean towards answering the question though.

Hixie:
LGTM, but travis is angry.

faq.md
sethladd: Fair enough.

Also, thanks for bringing this back to life. Totally forgot about it :(

PR #268, opened 2016-08-25
output expected formatting on travis

a14n:
Provides more infos on formatting problems (see flutter/flutter#5575)


sethladd:
lgtm thank you!

PR #273, opened 2016-09-12
Update docs to use master rather than alpha

abarth:
Folks who use alpha often find that the branch is stale and would prefer to be
on master.


abarth:
@eseidelGoogle @qchong @sethladd

Hixie:
LGTM

PR #280, opened 2016-09-27
docs: Move Text Input guide to 'Howtos' section.

qchong:


text-input.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: page
-title: Text Input
+title: Handling Text Input
```
sethladd: You used "Handle Text Input" above.

qchong: That's by design. I'm shortening the titles in the sidenav for easier scanning.

PR #281, opened 2016-09-29
State where the flutter.yaml file is located

sethladd:


sethladd:
cc @qchong

assets-and-images.md
```diff
@@ -19,10 +19,10 @@ and images.
...
 ## Specifying assets
 
-Flutter uses the `flutter.yaml` file to identify assets required by an
-app, as shown in the
-[Introduction to Flutter's Widget Framework](/widgets-intro/).
-A typical `flutter.yaml` file might include something like the following:
+Flutter uses the `flutter.yaml` file, located at the root of your
+project, to identify assets required by an app.
+
+Here is an example of declaring two assets inside `flutter.yaml`:
```
qchong: Suggest revising: "Here is an example of how you might declare two assets..."

sethladd: Are there other way to declare two assets? (difference between can and might?)

I could also say: "Here is an example of how you can declare two assets" which might be a little bit better?

qchong: There are, in fact, two ways the assets might be declared.

assets:
  - assets/my_icon.png
  - assets/background.png

vs.

assets:
  - assets/background.png
  - assets/my_icon.png

So, "might" is more appropriate in this case.

sethladd: Oh, commenting on the order. Yeah, we need to state order isn't important.

sethladd: ptal

qchong:
LGTM, thanks for the update!

PR #282, opened 2016-09-29
Stucture fix

qchong:


sethladd:
Looks like this was merged in?

qchong:
Abandoning this PR since it was merged earlier.

PR #283, opened 2016-09-29
docs: Add new 'Animating widgets' how-to guide.

qchong:
Staged preview:
- [Animating widgets](https://flutter-io-staging-69a8a.firebaseapp.com/animating-widgets/) how-to guide (new)
- [Animations in Flutter](https://flutter-io-staging-69a8a.firebaseapp.com/animations/) developer guide (updated)

@abarth @HansMuller ptal.

### DIFF
--- a/_data/sidebars/home_sidebar.yml
+++ b/_data/sidebars/home_sidebar.yml
@@ -17,7 +17,7 @@ entries:
   - title: Flutter Examples
     external_url: https://github.com/flutter/flutter/tree/master/examples
     output: web
-    
+
   - title: New to Flutter
     output: web
     folderitems:
@@ -96,6 +96,10 @@ entries:
 
     - title: Handle Text Input
       url: /text-input/
+      output: web
+
+    - title: Animate Widgets
+      url: /animating-widgets/
       output: web
 
   - title: Helpful Resources
--- a/animating-widgets.md
+++ b/animating-widgets.md
@@ -1,0 +1,293 @@
+---
+layout: page
+title: Animating Widgets
+sidebar: home_sidebar
+permalink: /animating-widgets/
+---
+
+This guide shows some common ways you can add animation
+effects to your apps using Flutter’s animation system.
+
+* TOC
+{:toc}
+
+## Introduction
+
+You can use animations to make the user experience on your mobile app feel more
+intuitive, dynamic, and polished. Animations are especially useful when the
+screen changes state, such as when content loads or when new actions become
+available.
+
+The [`flutter/animation`](https://docs.flutter.io/flutter/animation/animation-library.html)
+package in the Flutter SDK provides APIs to create animation effects in user
+interfaces (UIs) that are built on the
+[`StatelessWidget`](https://docs.flutter.io/flutter/widgets/StatelessWidget-class.html)
+and [`StatefulWidget`](https://docs.flutter.io/flutter/widgets/StatefulWidget-class.html)
+classes.
+
+Animations in Flutter are encapsulated as
+[`Animation`](https://docs.flutter.io/flutter/animation/Animation-class.html)
+objects that contain these key properties:
+
++ **Value:** The animation object’s typed value represents a data point within
+a range (also called an interpolation). You can define an animation class to
+generate values based on a given functional mapping, such as a linear, curve,
+or step-wise function.
++ **Status:** The animation object’s current status (such as forward,
+backward, completed, and dismissed), as listed by the
+[`AnimationStatus`](https://docs.flutter.io/flutter/animation/AnimationStatus-class.html)
+constants.
+
+To create an animation effect in your app, you need to write code to do the
+following:
+
+1. Attach an animation object to a widget or a listener to detect changes to the
+animation object.
+2. Specify how the the animation is *interpolated* (that is, how the animation
+objects’s value and status changes over
+time) as it runs.
+
+Based on the animation object’s properties at a specific frame, Flutter
+dynamically modifies your widget visual appearence, and rebuilds the widget
+tree to produce visual transitions such as sliding, scaling, rotation, height
+or width changes, and fading.
+
+Flutter's flexible animation model allows you to define interpolations for
+spatial positioning, color gradients, image scaling, and others. You can define
+your own subclass of
+[`Animation`](https://docs.flutter.io/flutter/animation/Animation-class.html)
+or use an animation subclass that the framework provides.
+
+To learn more about
+using the other key classes in the
+[`flutter/animation`](https://docs.flutter.io/flutter/animation/animation-library.html)
+package, see the [Animations in Flutter](/animations) guide.
+
+## Render a basic widget animation
+
+To render your widget with an animation, set one or more of the widget’s
+members to the animation object, then use the animation’s value to decide how
+to draw your object.
+
+If you are animating a stateful widget, you can notify your widget of changes
+to the animation’s values by using the
+[`addListener()`](https://docs.flutter.io/flutter/animation/Animation/addListener.html)
+method. The framework invokes the listener callback whenever the value of the
+animation changes. Typically, you’ll want to call the
+[`setState()`](http://docs.flutter.io/flutter/material/State/setState.html)
+method when the animation value change occurs, to trigger the framework to
+rebuild the widget tree.
+
+The following code snippet shows how you might create a simple animation
+effect to make an logo image appear to expand.
+
+```dart
+import 'package:flutter/material.dart';
+import 'package:flutter/animation.dart';
+
+class LogoApp extends StatefulWidget {
+  LogoAppState createState() => new LogoAppState();
+}
+
+class LogoAppState extends State<LogoApp> {
+  Animation<double> animation;
+
+  initState() {
+    AnimationController controller = new AnimationController(
+      duration: const Duration(milliseconds: 5000)
+    );
+    animation = new Tween(begin: 0.0, end: 200.0).animate(controller);
+    animation.addListener(() {
+      setState(() {
+      // the state that has changed here is the animation object’s value
+      });
+    });
+    controller.forward();
+    super.initState();
+  }
+
+  Widget build(BuildContext context) {
+    return new Center(
+      child: new Container(
+        margin: new EdgeInsets.symmetric(vertical: 10.0),
+        height: animation.value,
+        width: animation.value,
+        child: new AssetImage(
+          name: 'assets/flutter_logo.png',
+          fit: ImageFit.contain
+        )
+      )
+    );
+  }
+}
+
+void main() {
+  runApp(
+    new MaterialApp(
+      title: 'My Animation Test',
+      routes: <String, WidgetBuilder> {
+        '/': (BuildContext context) => new LogoApp()
+      }
+    )
+  );
+}
+```
+
+You can simplify the code snippet above further by using the
+[`AnimatedWidget`](http://docs.flutter.io/flutter/material/AnimatedWidget-class.html)
+class. This class hides the state object and listener you previously had to
+implement to run the animation.
+
+```dart
+import 'package:flutter/material.dart';
+import 'package:flutter/animation.dart';
+
+class LogoApp extends AnimatedWidget {
+  LogoApp({Key key, Animation<double> animation})
+      : super(key: key, animation: animation);
+
+  Widget build(BuildContext context) {
+    return new Center(
+      child: new Container(
+        margin: new EdgeInsets.symmetric(vertical: 10.0),
+        height: animation.value,
+        width: animation.value,
+        child: new AssetImage(
+           name: 'assets/flutter_logo.png',
+           fit: ImageFit.contain
+        )
+      )
+    );
+  }
+}
+
+void main() {
+  final AnimationController controller = new AnimationController(
+    duration: const Duration(milliseconds: 5000)
+  );
+  final animation = new Tween(
+    begin: 0.0,
+    end: 200.0)
+  .animate(controller);
+  runApp(
+    new MaterialApp(
+      title: 'My Animation Test',
+      routes: <String, WidgetBuilder> {
+        '/': (BuildContext context) => new LogoApp(
+          animation: animation
+        )
+      }
+    )
+  );
+  controller.forward();
+}
+```
+
+For more complex animations involving additional state, consider using the
+[`AnimatedBuilder`](http://docs.flutter.io/flutter/material/AnimatedBuilder-class.html)
+class.
+
+## Render simultaneous widget animations
+
+The following example shows how you can apply two simultaneous animation
+effects: one that controls the size of the widget and another that controls
+the opacity to fade the widget in and out.
+
+You can use the
+[`addStatusListener()`](http://docs.flutter.io/flutter/animation/Animation/addStatusListener.html)
+method to notify your widget when the current status of an animation changes.
+In this example, we use the status listener to switch the animation’s playback
+between going forward or reverse.
+
+```dart
+import 'package:flutter/material.dart';
+import 'package:flutter/animation.dart';
+
+class LogoApp extends AnimatedWidget {
+
+  // Declare the Tween objects as static because we don't change them.
+  static final _opacityTween = new Tween<double>(begin: 1.0, end: 0.3);
+  static final _sizeTween = new Tween<double>(begin: 0.0, end: 300.0);
+
+  LogoApp({Key key, Animation<double> animation})
+      : super(key: key, animation: animation);
+
+  Widget build(BuildContext context) {
+    return new Center(
+      child: new Opacity(
+        opacity: _opacityTween.evaluate(animation),
+        child: new Container(
+          margin: new EdgeInsets.symmetric(vertical: 10.0),
+          height: _sizeTween.evaluate(animation),
+          width: _sizeTween.evaluate(animation),
+          child: new AssetImage(
+            name: 'assets/flutter_logo.png',
+            fit: ImageFit.contain
+          )
+        )
+      )
+    );
+  }
+}
+
+void main() {
+  final AnimationController controller = new AnimationController(
+    duration: const Duration(milliseconds: 5000)
+  );
+
+  final Animation<double> animation =
+      new CurvedAnimation(parent: controller, curve: Curves.bounceIn);
+
+  animation.addStatusListener((status) {
+    if (status == AnimationStatus.completed)
+      controller.reverse();
+    else if (status == AnimationStatus.dismissed)
+      controller.forward();
+  });
+
+  runApp(
+    new MaterialApp(
+      title: 'My Animation Test',
+      routes: <String, WidgetBuilder> {
+        '/': (BuildContext context) => new LogoApp(
+          animation: animation
+        )
+      }
+    )
+  );
+  controller.forward();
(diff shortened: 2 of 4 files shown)


_data/sidebars/home_sidebar.yml
```diff
@@ -98,6 +98,10 @@ entries:
       url: /text-input/
       output: web
 
+    - title: Animate Widgets
+      url: /animating-widgets/
```
abarth: Probably the title and URL should match.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Introduction
+
+You can use animations to make the user experience on your mobile app feel more
+intuitive, dynamic, and polished. Animations are especially useful when the
+screen changes state, such as when content loads or when new actions become
+available.
+
+The [`flutter/animation`](https://docs.flutter.io/flutter/animation/animation-library.html)
+package in the Flutter SDK provides APIs to create animation effects in user
```
abarth: Technically `animation` is a library rather than a package.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+following:
+
+1. Attach an animation object to a widget or a listener to detect changes to the
+animation object.
+2. Specify how the the animation is *interpolated* (that is, how the animation
+objects’s value and status changes over
+time) as it runs.
+
+Based on the animation object’s properties at a specific frame, Flutter
+dynamically modifies your widget visual appearence, and rebuilds the widget
```
abarth: widget's visual appearance

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+Flutter's flexible animation model allows you to define interpolations for
+spatial positioning, color gradients, image scaling, and others. You can define
+your own subclass of
+[`Animation`](https://docs.flutter.io/flutter/animation/Animation-class.html)
+or use an animation subclass that the framework provides.
+
+To learn more about
+using the other key classes in the
+[`flutter/animation`](https://docs.flutter.io/flutter/animation/animation-library.html)
+package, see the [Animations in Flutter](/animations) guide.
```
abarth: s/package/library/

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Render a basic widget animation
+
+To render your widget with an animation, set one or more of the widget’s
+members to the animation object, then use the animation’s value to decide how
+to draw your object.
```
abarth: draw is kind of an unusual case.  It's much more common to use the animation's value to decide how to build child widgets.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+
+If you are animating a stateful widget, you can notify your widget of changes
+to the animation’s values by using the
+[`addListener()`](https://docs.flutter.io/flutter/animation/Animation/addListener.html)
+method. The framework invokes the listener callback whenever the value of the
+animation changes. Typically, you’ll want to call the
+[`setState()`](http://docs.flutter.io/flutter/material/State/setState.html)
+method when the animation value change occurs, to trigger the framework to
+rebuild the widget tree.
```
abarth: Rather than doing this, we might want to recommend that people use `AnimatedWidget` or `AnimatedBuilder`.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+
+class LogoApp extends StatefulWidget {
+  LogoAppState createState() => new LogoAppState();
+}
+
+class LogoAppState extends State<LogoApp> {
+  Animation<double> animation;
+
+  initState() {
+    AnimationController controller = new AnimationController(
```
abarth: You need the `vsync` parameter now.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+      duration: const Duration(milliseconds: 5000)
+    );
+    animation = new Tween(begin: 0.0, end: 200.0).animate(controller);
+    animation.addListener(() {
+      setState(() {
+      // the state that has changed here is the animation object’s value
+      });
+    });
+    controller.forward();
+    super.initState();
```
abarth: The typical style is to call `super.initState` first.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  Widget build(BuildContext context) {
+    return new Center(
```
abarth: This is a good application of `AnimatedBuilder`.  That will let you skip the `animation` member variable.  You can use a build function like:

```dart
  Animation<double> animation = new Tween(begin: 0.0, end: 200.0).animate(_controller);
  return new Container(
    padding: const EdgeInsets.symmetric(vertical: 10.0),
    child: new AnimatedBuilder(
      animation: animation,
      builder: (BuildContext context) {
        return new Image.asset(
          name: 'assets/flutter_logo.png',
          width: animation.value,
          height: animation.value,
          fit: ImageFit.contain
        );
      }
  )
}
```

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+    AnimationController controller = new AnimationController(
+      duration: const Duration(milliseconds: 5000)
+    );
+    animation = new Tween(begin: 0.0, end: 200.0).animate(controller);
+    animation.addListener(() {
+      setState(() {
+      // the state that has changed here is the animation object’s value
+      });
+    });
+    controller.forward();
```
abarth: You also need to hold onto the controller in a member variable and call it's `dispose` method in the State's `dispose` method.  If you don't do this, the animation will continue to run if your widget is removed from the tree, which is a waste of resources and will trigger an asset.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+}
+
+void main() {
+  runApp(
+    new MaterialApp(
+      title: 'My Animation Test',
+      routes: <String, WidgetBuilder> {
+        '/': (BuildContext context) => new LogoApp()
+      }
+    )
```
abarth: You can simplify this as:

```dart
runApp(
  new MaterialApp(
    title: 'My Animation Test',
    home: new LogoApp()
  )
);
```

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+  LogoApp({Key key, Animation<double> animation})
+      : super(key: key, animation: animation);
+
+  Widget build(BuildContext context) {
+    return new Center(
+      child: new Container(
+        margin: new EdgeInsets.symmetric(vertical: 10.0),
+        height: animation.value,
+        width: animation.value,
+        child: new AssetImage(
```
abarth: This should be `Image.asset`.  Also, the width and height parameter should move to the image (it's more efficient).  Finally, width should be listed before height.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+}
+
+void main() {
+  final AnimationController controller = new AnimationController(
+    duration: const Duration(milliseconds: 5000)
+  );
+  final animation = new Tween(
+    begin: 0.0,
+    end: 200.0)
+  .animate(controller);
```
abarth: I'm not sure this is a good example.  It's extremely unusual to create an AnimationController inside the `main` function.  Also, now with the `vsync` parameter, you'll have trouble figuring out which `vsync` you want to attach your animation to.

I'm not sure it's worth presenting these two examples.  Instead, we should just use `AnimatedBuilder` in the first example.  That's the code you'd normally write and it's simple enough to explain in one go.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+      }
+    )
+  );
+  controller.forward();
+}
+```
+
+For more complex animations involving additional state, consider using the
+[`AnimatedBuilder`](http://docs.flutter.io/flutter/material/AnimatedBuilder-class.html)
+class.
```
abarth: I'd skip this sentence.  `AnimatedBuilder` is good for simple animations too.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+
+```dart
+import 'package:flutter/material.dart';
+import 'package:flutter/animation.dart';
+
+class LogoApp extends AnimatedWidget {
+
+  // Declare the Tween objects as static because we don't change them.
+  static final _opacityTween = new Tween<double>(begin: 1.0, end: 0.3);
+  static final _sizeTween = new Tween<double>(begin: 0.0, end: 300.0);
```
abarth: We would normally declare these at the library scope.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+      : super(key: key, animation: animation);
+
+  Widget build(BuildContext context) {
+    return new Center(
+      child: new Opacity(
+        opacity: _opacityTween.evaluate(animation),
+        child: new Container(
+          margin: new EdgeInsets.symmetric(vertical: 10.0),
+          height: _sizeTween.evaluate(animation),
+          width: _sizeTween.evaluate(animation),
```
abarth: Better to evaluate once and re-use the value than to call `evaluate` twice.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+        )
+      )
+    );
+  }
+}
+
+void main() {
+  final AnimationController controller = new AnimationController(
+    duration: const Duration(milliseconds: 5000)
+  );
```
abarth: Again, this pattern of creating an AnimationController is `main` is extremely unusual and isn't a practice we should recommend.  Instead, we should host the AnimationController in a State object so that it can get the proper `vsync` signal and participate in the rest of the lifecycle.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+    )
+  );
+  controller.forward();
+}
+```
+
+## Dispose your widget animation
+
+It’s good practice to dispose your animation controller and free up your
+resources when they are no longer needed. To do so, call the
```
abarth: It's more than a good practice.  It's required for your code to be correct and you'll now get an assert if you don't do it.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+```dart
+class LogoAppState extends State<LogoApp> {
+
+  // Implementation code omitted for brevity.
+
+  @override
+  void dispose() {
+     controller.dispose();
+     super.dispose();
+  }
```
abarth: We should be doing this for all the examples throughout.  It's not really optional.

animations.md
```diff
@@ -17,27 +17,34 @@ functions directly by reading their current value and listening to their
...
+notifies all the listeners added with
 [`addListener`](http://docs.flutter.io/flutter/animation/Animation/addListener.html).
 Typically, a [`State`](http://docs.flutter.io/flutter/widgets/State-class.html)
 object that listens to an animation will call
 [`setState`](http://docs.flutter.io/flutter/widgets/State/setState.html) on
 itself in its listener callback to notify the widget system that it needs to
 rebuild with the new value of the animation.
 
-This pattern is so common that there are two widgets that help widgets rebuild
+     This pattern is so common that there are two widgets that help widgets rebuild
```
abarth: spurious indent?

animations.md
```diff
@@ -63,20 +69,35 @@ to 1.0 will be `dismissed` when their value is 0.0. An animation might then run
...
+[`AnimationController`](http://docs.flutter.io/flutter/animation/AnimationController-class.html)
+class lets you define important characteristics of the animation, such as its
+duration and playback direction (forward or reverse).
+
+The following snippet shows how you might create an animation that runs for
+500 milliseconds:
+
+```dart
+final AnimationController controller = new AnimationController(
+  duration: const Duration(milliseconds: 500)
```
abarth: We'll need to update this with `vsync: this`

animations.md
```diff
@@ -85,7 +106,90 @@ from 1.0 to 0.0). Similarly, you can create a
...
+and others.
+
+The following snippet shows how you might create a curved animation object that
+wraps a linear animation produced by an animation controller and applies the
+[`easeIn`](http://docs.flutter.io/flutter/animation/Curves/easeIn-constant.html)
+curve type to it.
+
+```dart
+final AnimationController controller =
+    new AnimationController(duration: const Duration(milliseconds: 500));
```
abarth: We usually use two-space indent.

animations.md
```diff
@@ -85,7 +106,90 @@ from 1.0 to 0.0). Similarly, you can create a
...
+
+final curve = new CurvedAnimation(
```
abarth: I'd skip the `final` and supply a type instead.

animations.md
```diff
@@ -85,7 +106,90 @@ from 1.0 to 0.0). Similarly, you can create a
...
+interpolates between the double values of -200.0 to 0.0.
+
+```dart
+final Tween doubleTween = new Tween<double>(begin: -200.0, end: 0.0);
+```
+
+You can also specify a progression between two colors. For example:
+
+```dart
+final Tween colorTween =
```
abarth: I'd drop the `final` from all these example.  Here I'd use the more specific ColorTween type.

animations.md
```diff
@@ -85,7 +106,90 @@ from 1.0 to 0.0). Similarly, you can create a
...
++ [`ThemeDataTween`](https://docs.flutter.io/flutter/material/ThemeDataTween-class.html):
+Interpolation between
+[`ThemeData`](https://docs.flutter.io/flutter/material/ThemeData-class.html) objects.
+
+The following snippet shows how you might create a
+[`Tween`](http://docs.flutter.io/flutter/material/Tween-class.html) object that
+interpolates between the double values of -200.0 to 0.0.
+
+```dart
+final Tween doubleTween = new Tween<double>(begin: -200.0, end: 0.0);
```
abarth: Dropping the type parameter here is dodgy.  Better to keep it in.

animations.md
```diff
@@ -85,7 +106,90 @@ from 1.0 to 0.0). Similarly, you can create a
...
+```
+
+You can also specify a progression between two colors. For example:
+
+```dart
+final Tween colorTween =
+    new ColorTween(begin: Colors.transparent, end: Colors.black54);
+```
+
+To apply a Tween to a given animation object, use the [`animate()`]()
```
abarth: Extra `()`

animations.md
```diff
@@ -85,7 +106,90 @@ from 1.0 to 0.0). Similarly, you can create a
...
+snippet shows how you might create an animation represented by integral values
+from 0 to 255, over a duration of 500 milliseconds, using linear interpolation.
+
+```dart
+final AnimationController animation =
+    new AnimationController(duration: const Duration(milliseconds: 500));
+
+Animation<int> alpha = new IntTween(
+  begin: 0,
+  end: 255)
```
abarth: This whitespace is odd.  I'd just put it all on one line.

animations.md
```diff
@@ -104,26 +208,24 @@ function.
...
+
+Use these methods if you need to combine a tween with an animation to get a
+concrete value:
+
++ [`evaluate()`](http://docs.flutter.io/flutter/animation/Tween/evaluate.html):
+Evaluates the tween at the current value of an animation. This approach is most
+useful for widgets that are already listening to the animation and hence
+rebuilding whenever the animation changes value.
++ [`animate()`](http://docs.flutter.io/flutter/animation/Animatable/animate.html)
+   Animates the tween based on the animation object's properties. Rather than
```
abarth: Inconsistent indent between these two bullets.

animating-widgets.md
jimbeveridge: I strongly disagree that it's "simple enough". I originally wrote my animation tutorial because of the amount of effort it took to understand how AnimatedBuilder worked. Unfortunately the AnimatedBuilder example was removed from this document, as was all of the discussion about how AnimatedBuilder worked.

animating-widgets.md
jimbeveridge: I disagree with this comment for the intended audience of this document. The sentence is to help direct people with their progression of learning.

animating-widgets.md
jimbeveridge: Why were all of the "new MaterialApp" statements added to the examples? These are not Material Apps and so similar statements were removed in early code reviews of these examples.

animating-widgets.md
```diff
@@ -0,0 +1,293 @@
...
+      : super(key: key, animation: animation);
+
+  Widget build(BuildContext context) {
+    return new Center(
+      child: new Container(
+        margin: new EdgeInsets.symmetric(vertical: 10.0),
+        height: animation.value,
+        width: animation.value,
+        child: new AssetImage(
+           name: 'assets/flutter_logo.png',
```
jimbeveridge: Why were these changed from web-based images to local images? With web-based images, readers can copy and paste the examples and they'll work. With local examples, readers have to create a project with the directory structure exactly right, and get the flutter.yaml file right, none of which is shown here. This was identified during usability testing of the original examples.

animating-widgets.md
jimbeveridge: The point of this document is to teach people how the animations work, not to provide boilerplate to copy and paste. I sounds like the document should discourage readers from using the low level calls, which is fine, but (IMO) understanding how it works under the covers was key to building up to how AnimatedBuilder works.

abarth: That makes sense.  However, we should be aware that people are going to copy-and-paste code samples.  If we're going to show people code that we don't want them to copy (e.g., missing dispose calls and starting AnimationControllers in main), then we should make it more obvious that these are code snippets and not complete executable samples.

animating-widgets.md
abarth: Yeah, we can skip that and just run `LogoApp` directly.

animating-widgets.md
jimbeveridge: This code was written last January and things have changed since then, so, instead of making them snippets, I think we should do your proposed changes to add dispose() calls and take the AnimationControllers out of main(). Even if we are showing low level calls, we should still show the proper way to use them.

animating-widgets.md
jimbeveridge: +1

animations.md
jimbeveridge: And we need to add some expository text about vsync.

_data/sidebars/home_sidebar.yml
qchong: The titles in the sidebar (left-nav) are shortened by design, to make them easier to scan. The full title ('Animating Widgets') is shown when users open the link to that page.

animating-widgets.md
qchong: changed to "library".

animating-widgets.md
qchong: fixed.

animating-widgets.md
qchong: fixed.

animations.md
qchong: This is intended markdown formatting to keep the new paragraph part of the same indent block as its antecedent.

animations.md
qchong: fixed.

animations.md
qchong: fixed

animations.md
qchong: fixed.

qchong:
This PR has a dependency on https://github.com/flutter/flutter/issues/1322 for improved animation code samples.

eseidelGoogle:
@sethladd maybe you can find a new owner for this?

sethladd:
Closing as stale. We can reopen if a new owner is found.

PR #287, opened 2016-10-13
Fix the bottom sheet entry in the material design catalog

InMatrix:
This will fix [issue 6290](https://github.com/flutter/flutter/issues/6290).


sethladd:
lgtm

PR #291, opened 2016-10-20
Downplay the design spec links.

Hixie:
In our usability studies, participant P1 got distracted by these links.
They are not immediately useful to people trying to learn Flutter.

I also simplified how we do the catalog so it's more consistent.

This is untested.


qchong:
lgtm

Hixie:
I've checked this in. If you're here because you found that this broke things, please revert it and I'll fix it tomorrow.

PR #292, opened 2016-10-21
Adjust expectations around /layout/

Hixie:
This page is really intended to be the landing page for the constraints
exceptions, so it's confusing to refer to it as just "Layout in
Flutter". Usability study participant P1 was directed to this page when
looking for information on layout algorithms, which was very misleading.

### DIFF
--- a/_data/sidebars/home_sidebar.yml
+++ b/_data/sidebars/home_sidebar.yml
@@ -54,7 +54,7 @@ entries:
       url: /gestures/
       output: web
 
-    - title: Layouts
+    - title: Box Constraints
       url: /layout/
       output: web
 
--- a/layout.md
+++ b/layout.md
@@ -1,9 +1,12 @@
 ---
 layout: page
-title: Layouts in Flutter
+title: Dealing with box constraints in Flutter
 sidebar: home_sidebar
 permalink: /layout/
 ---
+
+You may be directed to this page if the framework detects a problem
+involving box constraints.
 
 * Placeholder for TOC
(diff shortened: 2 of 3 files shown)


_data/sidebars/home_sidebar.yml
```diff
@@ -54,7 +54,7 @@ entries:
       url: /gestures/
       output: web
 
-    - title: Layouts
+    - title: Box Constraints
```
sethladd: We might not need to put it in the sidebar if this is a doc for a limited audience.

layout.md
```diff
@@ -1,10 +1,13 @@
 ---
 layout: page
-title: Layouts in Flutter
+title: Dealing with box constraints in Flutter
 sidebar: home_sidebar
 permalink: /layout/
```
sethladd: we can rename this permalink and issue a redirect, if you'd like.

_data/sidebars/home_sidebar.yml
Hixie: I mean it's for everyone eventually. It's just not generically about layout.

layout.md
Hixie: Yeah that's probably a good idea. Let's do that in a separate set of PRs though, since it'll also involve changes to the framework and isn't required for improving the docs on the short term.

abarth:
The plan here is to eventualyy write a real doc about layout and fold this material into a "troubleshooting" section

sethladd:
lgtm

PR #294, opened 2016-10-21
Fix the sample links in the catalog

Hixie:
TBR @abarth 

fixes breakage from https://github.com/flutter/website/pull/291


abarth:
LGTM

PR #295, opened 2016-10-21
Fix CSS to make sure that h3 amd h4 are actually smaller than h2

mit-mit:
Fixes https://github.com/flutter/flutter/issues/6416

Cross checked with https://github.com/jekyll/minima/blob/master/_sass/minima/_layout.scss

Before:
screen shot 2016-10-21 at 12 10 49

After:
screen shot 2016-10-21 at 12 10 12


sethladd:
lgtm thanks!

PR #297, opened 2016-10-26
Make the material design catalog work again in Chrome

InMatrix:
I did some testing locally. It seems that a missing \</p> caused the problem. The page now works in Chrome on my laptop.


sethladd:
nice, good find. (maybe run an html validator to double check?)

lgtm

Hixie:
wow. that's crazy. that really shouldn't have made any difference.

InMatrix:
I wonder if that surfaced a bug in Chrome. I would expect Chrome to be able
to auto correct simple missing tags.

Hixie:
It's not even a missing tag, </p> is optional in HTML.

PR #300, opened 2016-10-29
Merge the catalog and basic widgets pages

Hixie:
Based on P1 in particular, having two pages is confusing since people
get siloed in one.


Hixie:
The next step is simplifying the top of this page but let's start here.

Hixie:
@abarth

abarth:
LGTM

Hixie:
That broke the tree. Will look closer later.

PR #302, opened 2016-10-31
Merge the catalog and basic widgets pages

Hixie:
Based on P1 in particular, having two pages is confusing since people
get siloed in one.


Hixie:
I'm carrying the LGTM from https://github.com/flutter/website/pull/300 over to this.
The only difference is an extra comma in the JSON file.

PR #305, opened 2016-10-31
Fix some stuff noticed during P3 study.

Hixie:


getting-started.md
```diff
@@ -43,7 +43,9 @@ To run your app from the command-line:
 
 1. Open a terminal and change directories to the root of your app (the same directory that
 contains the `pubspec.yaml` file for your project).
-2. Run the following command.
+2. Run the following command. Once the `flutter` tool is running, if you change the app's
```
sethladd: We might want to split this into a third step. After the app is running, we can ask the user to open any text editor, change a specific thing, then hit 'r'.

widgets.md
```diff
@@ -14,10 +14,22 @@ extra_css:
...
 
-See also:
-
-* [The API reference for the `widgets` library](https://docs.flutter.io/flutter/widgets/widgets-library.html).
-* [The API reference for the `material` library](https://docs.flutter.io/flutter/material/material-library.html).
+Flutter uses a layered approach. At the highest level we provide
+[Material components](#material-design-widgets). These are implemented
+in the [`material`
+library](https://docs.flutter.io/flutter/material/material-library.html).
+That library is implemented in terms of the [`widgets`
```
sethladd: I'm generally in favor of not using 'that', and instead being specific. "The material library is implemented.."

widgets.md
```diff
@@ -14,10 +14,22 @@ extra_css:
...
+library](https://docs.flutter.io/flutter/widgets/widgets-library.html).
+That library in turn is implemented in terms of the [`rendering`
+library](https://docs.flutter.io/flutter/rendering/rendering-library.html),
+and so forth. Even within each library, we have used a layered
```
sethladd: s/have used/used

sethladd: New paragraph at "Even within..." ?

getting-started.md
Hixie: Yeah, I considered that, but it ends up making the stuff about IntelliJ more awkward.

widgets.md
Hixie: Fixed the first one, but I left the second to avoid making this hard to read.

widgets.md
Hixie: Changed to "use".

Hixie: Done

Hixie:
PTAL

getting-started.md
sethladd: Ah, then I'd create a new header before "Next Steps" called "Make a change and reload"

sethladd:
Talked to @Hixie , we'll land this and then tweak in another PR

PR #308, opened 2016-11-01
Move widgets.md to the CSV model, add images

Hixie:


Hixie:
cc @abarth for review

I've no idea if this actually works, but I'll back it out and fix it if it doesn't.

abarth:
LGTM

PR #311, opened 2016-11-01
More breakage fixes for widgets.md

Hixie:
Quote some more commas that weren't quoted in the CSV.
Validator fixes for `img`.
CSS fixes for the layout.
Use markup since apparently markdown in the CSV isn't expanded.


Hixie:
We're getting there! TBR=@abarth, landing untested to fix breakage.

PR #312, opened 2016-11-02
Further improvements to widgets.md.

Hixie:
Rather than marking classes with `<code>`, link them to the dartdocs.
Fix the image URLs.

This is probably the penultimate breakage fix.


Hixie:
TBR=@abarth

PR #314, opened 2016-11-02
attempt to draw the reader to the section that helps them with their intent

sethladd:
cc @hixie

Reasoning: new users frame their intent as "how do I handle an input/tap?". "Managing state" is a solution to the intent/problem.

### DIFF
--- a/widgets-intro.md
+++ b/widgets-intro.md
@@ -344,7 +344,7 @@ widgets have
 [`onPressed`](https://docs.flutter.io/flutter/material/RaisedButton-class.html#onPressed)
 callbacks that are triggered when the user taps the widget.
 
-Managing state and handling input
+Changing widgets in response to input
 ---------------------------------
 
 _Main articles: [`StatefulWidget`](https://docs.flutter.io/flutter/widgets/StatefulWidget-class.html), [`State.setState`](https://docs.flutter.io/flutter/material/State/setState.html)_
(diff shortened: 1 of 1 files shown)


Hixie:
LGTM.

We could go further. P4 was looking for "how to change my widget", so maybe "Handling input by changing widgets" or something?

@abarth may have ideas on how else to phrase this.

widgets-intro.md
```diff
@@ -344,8 +344,8 @@ widgets have
 [`onPressed`](https://docs.flutter.io/flutter/material/RaisedButton-class.html#onPressed)
 callbacks that are triggered when the user taps the widget.
 
-Managing state
---------------
+Managing state and handling input
```
abarth: Yeah, we should remove "managing state" from the title.  That's the solution.  Instead, we should state the problem:

"Changing widgets in response to input" ?

sethladd: can do!

sethladd:
ptal

widgets-intro.md
```diff
@@ -344,8 +344,8 @@ widgets have
 [`onPressed`](https://docs.flutter.io/flutter/material/RaisedButton-class.html#onPressed)
 callbacks that are triggered when the user taps the widget.
 
-Managing state
---------------
+Changing widgets in response to input
+---------------------------------
```
Hixie: missing some hyphens :-)

Hixie:
Still LGTM

PR #318, opened 2016-11-03
We call it debug mode now, more than checked mode.

Hixie:
And there's no --no-checked any more.


qchong:
lgtm, thanks for the updates Ian!

PR #333, opened 2016-12-06
Update the testing documentation

Hixie:


HansMuller:
LGTM.

Is it going to be clear to the reader that you must run flutter test from the root of your project (or whatever we call the top level)?

Hixie:
Updated accordingly.

PR #340, opened 2016-12-29
Enable link checking with html_proofer (non-automated for now)

mit-mit:
Inspired by https://github.com/flutter/flutter/issues/7306, we should check all links

This uses https://github.com/gjtorikian/html-proofer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,12 @@ A tldr version follows:
 * Make your changes.
 * To test your changes, run `bundle exec jekyll serve` (or
   `jekyll serve -w --force_polling`)
+* Prior to submitting, run automatic link validation:
+
+  ```
+  bundle exec jekyll build
+  bundle exec htmlproofer --empty-alt-ignore _site --url-ignore "#"
+  ```
 
 ## Code Snippet Validation
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,5 @@ node_js:
 sudo: false
 before_install: ./tool/before_install.sh
 script: ./tool/travis.sh
+rvm:
+  - 2.3.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 node_js:
   - "5.1"
 sudo: false
-before_install: ./tool/before_install.sh
+before_install:
+  - rvm install 2.2.2
+  - ./tool/before_install.sh
 script: ./tool/travis.sh
-rvm:
-  - 2.3.1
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -25,6 +25,12 @@ A tldr version follows:
 * Make your changes.
 * To test your changes, run `bundle exec jekyll serve` (or
   `jekyll serve -w --force_polling`)
+* Prior to submitting, run automatic link validation:
+
+  ```
+  bundle exec jekyll build
+  bundle exec htmlproofer --empty-alt-ignore _site --url-ignore "#"
```
Hixie: shouldn't this happen in the PR test?

Realistically, I'm not going to be installing jekyll and running this before every checkin. I rely on Travis to do these tests.

mit-mit: This is just the first step, we definitely need something more automated also.

I don't think we should fail a Travis build on this; it's a little too uncontrollable to have the build fail on the availability on some external website. But I plan on enabling daily builds that will check the links using the new Crons feature in Travis. I'm testing that setup for a few days on my fork, and will send a PR for that if it works out.

Hixie:
I think we already do cron-based testing for the web site, FWIW. @sethladd might know more.

abarth:
Yes, we have a Travis cron that runs every day on the web site.

mit-mit:
Yes, but we need adjustments to the script (Travis uses the same .yml) for
the cron case. I'm working on those.

Can I get a LGTM on this PR?

abarth:
I agree with @Hixie that we should just run this check on every PR.  If it's too flaky, we can iterate.

mit-mit:
Ok, that's easy, but we would need to fix the errors I listed in https://github.com/flutter/flutter/issues/7306 first. Can one of you take a look at those? It looks like they are mostly framework classes that no longer exist (maybe renamed?).

abarth:
Yes, that page is pretty out-of-date because we revised the part of the system it is describing substantially.

mit-mit:
@Hixie @abarth I'd like to land this first PR so we can get manual checking in; can I get a LGTM? Once someone with more knowledge of the Flutter framework fixes https://github.com/flutter/flutter/issues/7306, I'll enable automatic checking (I have it running already on my fork).

sethladd:
Is the issue that this PR title is misleading? The title implies that this PR turns on automated link checking. I can't see where we do that in the Travis process?

Perhaps this PR is just about installing/locking new dependencies required by link checking.

This PR by itself LGTM but I'd like to try living with link checking turned on for every PR. If life becomes too painful, we can revisit. But let's try with more checking, not less.

PR #341, opened 2016-12-29
Fix a few broken links

mit-mit:
As discussed in flutter/flutter#7306 and https://github.com/flutter/website/pull/340

### DIFF
--- a/debugging.md
+++ b/debugging.md
@@ -26,12 +26,11 @@ problems.
 
 ## Dart Observatory (statement-level single-stepping debugger and profiler)
 
-If you started your application on an Android device using `flutter
-run`, then, while it is running, you can open the Web page at
-[http://localhost:8181/](http://localhost:8181/) to connect to your
-application directly with a statement-level single-stepping debugger.
-If you're using IntelliJ, you can also debug your application using its
-built-in debugger.
+If you started your application on an Android device using `flutter run`, then,
+while it is running, you can open the Web page at `http://localhost:8181/` to
+connect to your application directly with a statement-level single-stepping
+debugger. If you're using IntelliJ, you can also debug your application using
+its built-in debugger.
 
 Observatory also supports profiling, examining the heap, etc. For more
 information on Observatory, see
--- a/debugging.md
+++ b/debugging.md
@@ -27,7 +27,8 @@ problems.
 ## Dart Observatory (statement-level single-stepping debugger and profiler)
 
 If you started your application on an Android device using `flutter run`, then,
-while it is running, you can open the Web page at `http://localhost:8181/` to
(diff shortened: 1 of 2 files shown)


debugging.md
```diff
@@ -26,12 +26,11 @@ problems.
...
 ## Dart Observatory (statement-level single-stepping debugger and profiler)
 
-If you started your application on an Android device using `flutter
-run`, then, while it is running, you can open the Web page at
-[http://localhost:8181/](http://localhost:8181/) to connect to your
-application directly with a statement-level single-stepping debugger.
-If you're using IntelliJ, you can also debug your application using its
-built-in debugger.
+If you started your application on an Android device using `flutter run`, then,
+while it is running, you can open the Web page at `http://localhost:8181/` to
```
Hixie: We should replace this with "...then, while it is running, you can open the Web page at the address given by the `flutter run` tool to connect..." because the URL changes every time now.

getting-started.md
```diff
@@ -72,13 +70,13 @@ If everything works, you should see your starter app on your device or simulator
 <div id="starter-app-screenshots">
   <div class="box2">
     <figure>
-      <img src="/images/flutter-starter-app-ios.png">
+      <img src="/images/flutter-starter-app-ios.png" alt="iOS screenshot">
```
Hixie: that is not a valid alt value. It is more correct to omit the alt attribute here.

getting-started.md
```diff
@@ -72,13 +70,13 @@ If everything works, you should see your starter app on your device or simulator
...
       <figcaption>Figure 1. iOS (iPhone 6s Plus)</figcaption>
     </figure>
   </div>
   <div class="box2">
     <figure>
-      <img src="/images/flutter-starter-app-android.png">
+      <img src="/images/flutter-starter-app-android.png" alt="Android screenshot">
```
Hixie: ditto

getting-started.md
Hixie: See https://html.spec.whatwg.org/#alt for detailed information on what the alt attribute value should be.

Hixie: https://html.spec.whatwg.org/#a-key-part-of-the-content in particular has examples of screen shot alt texts

mit-mit: Yes, those were not great alt strings.

I don't think omitting the attribute would be correct. From the spec you linked, that is only correct when the image contents are not known.

Rather, in this case the image is *not* a key part of the content, but rather merely a graphical representation of the surrounding text, so they can be left empty per  §4.8.4.4.6.

mit-mit:
Thanks for the feedback, updated

mit-mit:
@Hixie any more comments, or can I get a LGTM?

Hixie:
LGTM.

PR #344, opened 2017-01-05
Fix Firebase publishing step to be on pushes/merges to master only

mit-mit:
The current branching logic deploys to Firebase whenever not running for a pull-request, and when not on master. However, that means it also deploys when running a cron job, or an api job (e.g. a restart from the admin UI).

This changes the branching to only deploy when running a push job (i.e., pushing/merging) into master.

https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

### DIFF
--- a/tool/travis.sh
+++ b/tool/travis.sh
@@ -59,14 +59,11 @@ bundle install
 echo "Building site."
 bundle exec jekyll build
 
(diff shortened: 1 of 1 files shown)


sethladd:
Good idea! lgtm

PR #345, opened 2017-01-05
Update references to Flexible to Expanded

abarth:
We're in the process of renaming Flexible to Expanded to be clearer about what
the widget does.


abarth:
@Hixie

Hixie:
LGTM

PR #351, opened 2017-01-12
Remove section on implementing Mojo services from design principles page

tvolkert:


abarth:
LGTM

Hixie:
We might have to bring something like this back, in time, depending on what kind of APIs we expose.

abarth:
The reference to mojo is just a bit confusing since that term now refers to something in chrome.

Hixie:
oh I've no objection to this specific removal

PR #352, opened 2017-01-13
Update assets-and-images.md

LarkAscending:
1. Updating link from ImageResource to newer ImageStream class (see https://github.com/flutter/flutter/commit/2dfdc840b1b10342cecd2c89292e2f4bf6003948)
2. Updating link from AssetVendor to newer AssetImage.
3. Removing reference to resolution_awareness.dart - there is no replacement example.
4. Adding "assets/" prefix to path in build method in Loading Images.

Partial fix for https://github.com/flutter/flutter/issues/7480

### DIFF
--- a/assets-and-images.md
+++ b/assets-and-images.md
@@ -72,10 +72,10 @@ different locales or regions, reading directions, etc.
 
 Your app can access its assets through an
 [`AssetBundle`](https://docs.flutter.io/flutter/services/AssetBundle-class.html)
-object.
+object. 
 
 The two main methods on an asset bundle allow you to load a string/text asset
-(`loadString`) or an image asset (`loadImage`) out of the bundle,
+(`loadString`) or an image/binary asset (`load`) out of the bundle,
 given a logical key.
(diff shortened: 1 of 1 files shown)


sethladd:
lgtm

thank you!

assets-and-images.md
```diff
@@ -165,12 +165,12 @@ Widget build(BuildContext context) {
 ```
 
 The way this works is through an object called
-[`AssetVendor`](https://docs.flutter.io/flutter/widgets/AssetVendor-class.html)
+[`AssetImage`](https://docs.flutter.io/flutter/widgets/AssetImage-class.html)
 established at the top of the build tree. AssetVendor replaces the default asset
```
Hixie: this whole paragraph is wrong now (AssetVendor was entirely removed).

LarkAscending: Good one!

LarkAscending:
See https://github.com/flutter/flutter/issues/7530 for future work needed on this page.

PR #354, opened 2017-01-14
Update debugging.md

LarkAscending:
1. Changed InkFeatures class name to newer InkFeature class. 
 *  _However, the debugging output continues to reference the old class name, and needs to be updated_
2. Updated widget/element and render tree output for this example. Corrected InkFeatures > InkFeature manually in the output. Left  name of renderObject _RenderInkFeatures as-is, since it's not clear that has changed.


Hixie:
Looks like you left the old tree as well as the new tree? I'm confused by this diff.

What tool did you use to regenerate this? I didn't think `flutter run` showed the pid.

LarkAscending:
Used flutter run. Since all the lines in the replaced examples show the PID, it means they're all new.

Hixie:
I think there's some lines to remove still, unless I'm really misunderstanding the patch.

PR #356, opened 2017-01-14
Update platform-services.md

LarkAscending:
Many changes=big update pass, triggered by updating broken links.
1. HostMessages>PlatformMessages.
1. Subtract Mojo.
1. Structure and naming of files in examples has changed.
1. Names of static methods changed.

Uncertainties:
1. I believe PlatformMessages can now handle binary data.
1. Graphic might be out-of-date.
1. There's no mention of having to import io.flutter.view.FlutterView - is this needed?

### DIFF
--- a/platform-services.md
+++ b/platform-services.md
@@ -113,8 +113,8 @@ class establishes connections with services.
    You can register a callback for receiving
    messages through the [`PlatformMessages`](https://docs.flutter.io/flutter/services/PlatformMessages-class.html)
    class by calling the
-   [`setJSONMessageHandler()`](https://docs.flutter.io/flutter/services/HostMessages/setJSONMessageHandler.html)
-   or [`setStringMessageHandler()`](https://docs.flutter.io/flutter/services/HostMessages/setStringMessageHandler.html)
+   [`setJSONMessageHandler()`](https://docs.flutter.io/flutter/services/PlatformMessages/setJSONMessageHandler.html)
+   or [`setStringMessageHandler()`](https://docs.flutter.io/flutter/services/PlatformMessages/setStringMessageHandler.html)
    static methods.
 
    For example, this snippet shows how you might initialize a service binding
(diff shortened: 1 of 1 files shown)


abarth:
LGTM

platform-services.md
```diff
@@ -51,17 +51,17 @@ non-graphical, data-only information. Examples include:
...
 class to invoke APIs from third-party SDKs that run on iOS and Android devices.
 
 To communicate between the
-[`HostMessages`](https://docs.flutter.io/flutter/services/HostMessages-class.html) class and
+[`PlatformMessages`](https://docs.flutter.io/flutter/services/PlatformMessages-class.html) class and
 platform-specific APIs in the host, you must add the following Flutter components into
 the host's view hierarchy.
 
-+ **On iOS:** Add a [`FlutterViewController`](https://github.com/flutter/engine/blob/master/sky/shell/platform/ios/framework/Headers/FlutterViewController.h)
++ **On iOS:** Add a [`FlutterViewController`](https://github.com/flutter/engine/blob/master/shell/platform/darwin/ios/framework/Headers/FlutterViewController.h)
```
chinmaygarde: We may move files around in the repository. If we ever moved FlutterViewController, it would result in a broken link on the website. A handy trick on GitHub it to press `y`. This will replace the current link (the one containing `blob/master`) with a link that explicitly specifies the SHA (`blob/<SHA>`). We'd still need to update the website but at least users won't see broken links in the meantime. And updating docs would be as simple as doing as looking for all instances of that SHA.

LarkAscending: Good one, let's run it past @sethladd 
Right now the link checker is the way we know something on the site needs to be updated. It's being integrated into Travis CI, so all the links will be checked on every build. If no broken link, we literally won't know to go investigate and update.

chinmaygarde: Is the link checker on all builds of `flutter`, or, the job that rebuilds the website? If its the latter, links may be broken till the next time someone submits a patch to the website.

Disclaimer: I am not sure where/when this job runs.

LarkAscending: Not sure, others will know. Even if it's the latter, I'm saying that link breakage is one important signal that the website needs to change. Often it's not just the link that needs updating...this PR is a good example. It's a cluster of changes, including subtracting Mojo, that were triggered by link breakage. Without the breakage, the flutter and website repos can drift apart silently and smoothly, with no red flags. Normally, I agree that avoiding errors and hurtles is totally excellent. In this case, there would need to be some other reliable indicator or sync mechanism.

PR #360, opened 2017-01-19
Fix broken links

abarth:
Many of these links broke when we removed the duplicate doc pages from the API
docs.

Fixes #7480


abarth:
@LarkAscending

animations.md
```diff
@@ -172,12 +172,6 @@ change state irreversibly when queried.
 There are [various concrete implementations](https://docs.flutter.io/flutter/physics/physics-library.html)
 of the `Simulation` class for different effects.
 
-## Forces
-
-The
-[`Force`](https://docs.flutter.io/flutter/animation/Force-class.html)
-abstract class provides a factory for `Simulation` instances.
```
abarth: This class no longer exists.

LarkAscending:
Yes, we can either remove the section or turn it into something about
SimulationGroup, which is new.

FYI this is missing from the original list of broken links. I see the class
was removed yesterday.

PR #361, opened 2017-01-19
Remove Force class, add SimulationGroup

LarkAscending:
The Force class has been removed, so this commit removes the Forces section.
Also, adding sentence about SimulationGroup and adding link to that class.


abarth:
LGTM

PR #362, opened 2017-01-23
Fix broken link

abarth:


abarth:
@LarkAscending 

(We'll need to update this page after we're done with https://github.com/flutter/flutter/issues/7031 )

PR #368, opened 2017-01-31
Making changes to the website.

Sfshaza:
* Changes how prev/next page navigation is implemented: 
  * The prev/next nav appears when those tags are added to the front matter. You no longer manually include the prev-next-nav.html include file, which I've deleted.
  * The prev-next navigation is now placed above the title, not below it.
  * The prev-next navigation is now added to the bottom of the page.
* Fixes the problem where the spacing between the &lt;li&gt; tags in a &lt;ul&gt; block is too much. (The bulleted lists were taking too much vertical space.)
* Fixes the problem where the &lt;h4&gt; headers are too tiny and italicized.
* Fixes the problem where code font text in an &lt;aside&gt; displays with a white background, instead of the color of the aside's background.

You can see the results of all of these changes staged [here](https://sz-flutter.firebaseapp.com/tutorials/layout/classes.html).

Please review, @sethladd.  Thanks for your assistance, @kwalrath!


sethladd:
lgtm

PR #370, opened 2017-02-02
Use default ios-deploy Homebrew formula

cbracken:
Eliminate instructions to use brew tap flutter/flutter. The standard
homebrew formula for ios-deploy is sufficient. This eliminates the need
for the Flutter team to maintain our own formula for it.

Fixes flutter/flutter#7783


cbracken:
@sethladd @eseidelGoogle @abarth

PR #372, opened 2017-02-07
Adding a _code directory.

Sfshaza:
@sethladd, here's a simple HelloWorld app.


sethladd:
Oh, wow. Are we certain we want the entire app? I thought we just needed the app's .dart file here?

Sfshaza:
Sure, we can do that, though all of them are called main.dart. Maybe retain the directory name and have main.dart under? (Sans lib directory.)

sethladd:
Makes sense!

Sfshaza:
Seth agreed (offline) that we'll use a directory name and it will contain main.dart. I'm closing this PR.

PR #373, opened 2017-02-07
Fixed spelling for NSDocumentDirectory

raju-bitter:
The correct name is NSDocumentDirectory, see https://developer.apple.com/reference/foundation/nssearchpathdirectory/nsdocumentdirectory


mit-mit:
LGTM, thanks @raju-bitter!

PR #374, opened 2017-02-07
recognize comment-style highlight commands in source code.

sethladd:
Had to upgrade our prism.js to respect markup inside of codeblocks.


sethladd:
cc @Sfshaza

Sfshaza:
I haven't tested it personally, but I've seen the output of Seth's test and it looks great!

LGTM

PR #376, opened 2017-02-08
include a range/region of a file, and unindent prettified code

sethladd:


sethladd:
cc @Sfshaza @LarkAscending

Sfshaza:
LGTM! And thanks!

PR #381, opened 2017-02-09
Remove mention of Block, in favour of ListView

Hixie:
@abarth

Also removes MaterialList mentions, in preparation for it being simplified away.


abarth:
LGTM

PR #384, opened 2017-02-12
Fix link to non-existent "Managing State" section

LarkAscending:
Widgets Intro does not have a Managing State section. Instead, pointing users to example of Changing Widgets in Response to Input, which shows how to manage state.
  #383 

When we do have detailed state info, this link should be repointed there.

This broken link was not caught in previous HTMLproofer runs, so we should keep an eye out for other misses...


abarth:
LGTM

PR #386, opened 2017-02-13
Relocate IDE setup steps to separate page

LarkAscending:
The majority of users will probably use a text editor, so streamlining the setup page for them.
Moving the IDE setup steps to a new page...

### DIFF
--- a/_data/sidebars/home_sidebar.yml
+++ b/_data/sidebars/home_sidebar.yml
@@ -26,8 +26,8 @@ entries:
       url: /setup/
       output: web
       
-    - title: IDE set up
-      url: /ide-setup/
+    - title: IntelliJ set up
+      url: /intellij-setup/
       output: web
 
     - title: Get started
--- a/_data/sidebars/home_sidebar.yml
+++ b/_data/sidebars/home_sidebar.yml
@@ -25,6 +25,10 @@ entries:
     - title: Set up
       url: /setup/
       output: web
+      
+    - title: IDE set up
+      url: /ide-setup/
+      output: web
 
     - title: Get started
       url: /getting-started/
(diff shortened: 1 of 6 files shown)


_data/sidebars/home_sidebar.yml
```diff
@@ -25,6 +25,10 @@ entries:
     - title: Set up
       url: /setup/
       output: web
+      
+    - title: IDE set up
```
sethladd: Can we make this specific for IntelliJ ? There may be other IDEs in the future.

LarkAscending: Done!

sethladd:
Thanks for this tweak! Please wait for an lgtm before merging next time. Maybe I missed where the lgtm was given?

Sometimes, very very simple changes (like, fixing a spelling error) can be "TBR" or "to be reviewed" and merged right away. But, if you do, please add a TBR and add someone's name so they are notified about the change.

But generally for changes we like to get a reviewer to sign-off.

mit-mit:
I think we will have to revisit once we add Windows support. There I believe the majority will be IDE users, not the minority.

LarkAscending:
Thanks for looking at this!

@sethladd I haven't seen that written down anywhere, but sure. FYI this seemed like a trivial mod since the substance didn't change. I guess it wouldn't hurt to define that, so we have the same idea of what needs an LGTM. Reverting is simple to do if it's controversial. No worries.

@mit-mit We're seeing users jump through a lot of hoops when setting up an IDE (dependencies, IDE, plugin, etc.). Even if the majority of users  are IDE'rs, it might be a good idea to keep these separate so new users don't get lost in the setup process. And, if we end up supporting multiple IDEs as seems to be the plan, we'll definitely need to separate the setup instructions. This change should be OK in a variety of scenarios. Will keep listening to users and watching the analytics data, of course :-)

Hixie:
@LarkAscending see https://flutter.io/design-principles/#code-review-and-checking-in-code

LarkAscending:
Thank you @Hixie!

Good to know. In this case, Eric asked for the change tho he did not review
the PR.
I'll be cluttering up everyone's Inbox from now on ;-)

PR #388, opened 2017-02-15
Fix IntelliJ HowTo installation link

mit-mit:
The setup instructions no longer have IntelliJ setup, so changing to the IntelliJ page


sethladd:
lgtm

PR #390, opened 2017-02-15
Add xcode-select step for iOS installation

cbracken:
Also tweak hyphenation to differentiate 'command-line' (compound
modifier to 'tools') and 'command line' (noun).


cbracken:
One of two fixes for flutter/flutter#8094.

PR #393, opened 2017-02-17
Add iOS physical device deploy instructions

xster:
#7839


Sfshaza:
Hey! This afternoon the API docs changed and the ScrollSimulation class was removed (and replaced with two classes). It was hanging me up, so I created a PR that Adam approved. I've merged it, so if you get the latest, your submit should work.

eseidelGoogle:
Just going through old PRs.  @xster looks like this is waiting merge?

xster:
Oh, oops, totally forgot about this. Will follow up

PR #397, opened 2017-02-17
Tweaking the code samples and adding one.

Sfshaza:
* Added comments across the examples about using/suppressing visual layout at runtime.
* At the request of Brian, added an example of packed items in a row.
* Moved more of the widget building into variables, for a couple reasons: it's easier to re-use, it's also clearer to read.

Please review, @abarth.


_includes/_code/container/main.dart
```diff
@@ -31,85 +32,75 @@ class MyHomePage extends StatefulWidget {
...
-                        borderRadius:
-                            const BorderRadius.all(const Radius.circular(8.0)),
-                      ),
-                      margin: const EdgeInsets.all(4.0),
-                      child: new Image.asset('images/pic1.jpg'),
-                    ),
+              new Expanded(
+                child: new Container(
+                  decoration: new BoxDecoration(
+                    border: new Border.all(width: 10.0, color: Colors.black38),
```
abarth: const?

abarth: Never mind.  I can't be const.

_includes/_code/packed/pubspec.yaml
```diff
@@ -0,0 +1,40 @@
...
+description: A new flutter project.
+
+dependencies:
+  flutter:
+    sdk: flutter
+
+# For information on the generic Dart part of this file, see the
+# following page: https://www.dartlang.org/tools/pub/pubspec
+
+# The following section is specific to Flutter.
```
abarth: I think in other example you've taken out these comments.

_includes/_code/stack/main.dart
```diff
@@ -31,32 +32,34 @@ class MyHomePage extends StatefulWidget {
...
+        new Container(
+          decoration: new BoxDecoration(
+            backgroundColor: Colors.black45,
+          ),
+          child: new Text(
+            'Mia B',
+            style: new TextStyle(
+                fontSize: 20.0,
+                fontWeight: FontWeight.bold,
+                color: Colors.white),
```
abarth: missing comma after white (and you should reformat after that)

abarth:
LGTM

PR #401, opened 2017-02-22
Move editor setup out from under the Android section

mit-mit:
The editor/IDE paragraphs are currently placed as the last paragraph under the Android section/header. Thus, a developer that chooses to skip Android setup (e.g., an iOS developer) will never reach that paragraph and not be aware of the IDE option.

This simple moves it under a separate header.


LarkAscending:
It looks like we reworded to favor the IDE working environment over text editor. Is that because users may not be familiar w/Dart?

mit-mit:
The goal is to point developers towards IntelliJ *unless* they have (strong) preferences to use something else, which in that case is totally OK. In other words, devs who are indifferent really should be in IntelliJ.

PR #403, opened 2017-02-23
Fix list formatting in code_comp_page layout

kwalrath:
@Sfshaza pointed out that, after she fixed the list line height in https://flutter.io/tutorials/layout/ and other pages, the list items in https://flutter.io/web-analogs/ were too close together vertically. @lwalrath helped me track down what was causing the difference. We confirmed in my local build that this change doesn't have any unexpected effects on web-analogs.

### DIFF
--- a/_layouts/code_comp_page.html
+++ b/_layouts/code_comp_page.html
@@ -11,5 +11,5 @@ layout: default
   </header>
 </div>
     <body>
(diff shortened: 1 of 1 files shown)


Sfshaza:
Thanks for addressing this, Kathy!

Sfshaza:
LGTM!

PR #413, opened 2017-03-03
First pass at refreshed setup and IDE setup pages

LarkAscending:
1. Tested end-to-end on macOS and revised steps to match.
1. Incorporated feedback from 2 UX sessions and 2.5 open issues. 
    1. Added a task checklist, with Required? column. Removed TOC because, weird with both.
    1. Added IntelliJ IDE note and moved Editor setup further up the page.
    1. Added info about how to permanently update your PATH (until we have an installer or a command that does it).
    1. Added syntax examples and verification info.
1. Removed Android SDK step (done by Android Studio now).
1. Added Homebrew troubleshooting steps.
1. Moved Google Analytics from a section w/header that looks like a task to a colored alert box.
1. Updated IntelliJ plugin steps w/new procedure.


setup.md
```diff
@@ -8,72 +8,156 @@ permalink: /setup/
...
+   The above command sets your PATH variable temporarily, for the current terminal window.
+   The steps for modifying this variable permanently for all terminal sessions are machine-
+   specific. Typically you add a line to a file that is executed whenever you open 
+   a new window:
+
+   ```
+   export PATH=~/flutter/bin
+   ```
+and then run the `source <filename>` command to refresh the window. For example on Linux, 
+you might edit and source `~/.bash_profile`. On macOS, it might be `/etc/bashrc`.
```
InMatrix: It's also "~/.bash_profile" at least on my Mac.

mit-mit: Yup, it's `~/.bash_profile`. `/etc/` are system files shared for all users, best not to point folks to them.

setup.md
```diff
@@ -8,72 +8,156 @@ permalink: /setup/
...
-This lets users choose to opt-out of analytics without any data being sent. See Google's
-privacy policy:
-[www.google.com/intl/en/policies/privacy](https://www.google.com/intl/en/policies/privacy/).
+Using our command-line tools, you can use any editor to develop Flutter applications.
+
+We recommend using our IntelliJ plug-ins for a  [rich IDE experience](/intellij-ide/) 
+supporting editing, running, and debugging Flutter apps. See [IntelliJ Setup](/intellij-setup/)
+for detailed steps.
 
 ## iOS setup
```
mit-mit: Should we clarify that this whole iOS section is macOS only?

setup.md
```diff
@@ -8,72 +8,156 @@ permalink: /setup/
 This guide describes how to set up your development environment to run Flutter apps on
 iOS or Android.
 
-* TOC Placeholder
-{:toc}
+New to GitHub? First [install the software](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git){: alert-link}
```
mit-mit: This new GitHub section seems like a potential distraction; there is a lot of text to read through on those links.

On the two current operating systems we support git should be available, and the user really doesn't need to know how git works just for installation purposes, so I think we can omit it.

intellij-setup.md
```diff
@@ -28,23 +28,27 @@ Android Studio (and various other JetBrains editors) is currently not supported.
 
 ### Installing the plugins
 
-We provide two plugins:
+To use IntelliJ with Flutter, you must install two plugins:
```
LarkAscending: Actually, we don't mention that below (we just say the SDK is bundled), and I was not aware. When testing I installed the Dart plugin, then the Flutter plugin, as the instructions said to do - so didn't see the new behavior.

intellij-setup.md
```diff
@@ -28,23 +28,27 @@ Android Studio (and various other JetBrains editors) is currently not supported.
...
 
   * The `Dart` plugin offers code analysis (code validation as you type, code completions, etc.).
   * The `Flutter` plugin powers Flutter developer workflows (running, debugging, hot reload, etc.).
 
 To install the plugins:
```
LarkAscending: OK thanks! (you can always feel free to edit directly, but I'll apply these changes)

setup.md
sethladd: sgtm

LarkAscending: Agreed. This was a request from @sethladd - we discussed and he's OK w/taking it out.

setup.md
```diff
@@ -8,72 +8,156 @@ permalink: /setup/
...
+Using an IDE is optional. However, we recommend using an IntelliJ IDE with Flutter for code completion, 
+inline error checking, and visual debugging features.
+{: .alert-warning}
 
 ## System requirements
 
 To install and run Flutter, your development environment must meet these minimum requirements:
 
 * **Operating Systems**: Mac or Linux (64-bit). Windows support is [in progress](https://github.com/flutter/flutter/issues/138).
+* **Disk Space**: 710 MB on Mac, 558 MB on Linux (does not include disk space for Xcode or Android Studio).
```
LarkAscending: LOL (and even approximate numbers are quite vulnerable to churn...)

setup.md
sethladd: I don't remember this request but then again sometimes I can't find my keys.

setup.md
```diff
@@ -8,72 +8,156 @@ permalink: /setup/
...
-$ git clone https://github.com/flutter/flutter.git
-$ export PATH=`pwd`/flutter/bin:$PATH
-</pre>
+   ```
+   $ git clone https://github.com/flutter/flutter.git
+   $ export PATH=`pwd`/flutter/bin:$PATH
+   ```
 
-Run the following command to see if there are any dependencies you need to install to complete
+   where `pwd` is the current working directory, for example:
```
LarkAscending: Yes, familiar with Unix. On my machine I used `pwd` as-is. The UX'ers read this differently, though, and they were substituting a path w/o knowing what it should be. This change was to reduce confusion for that kind of user (but apparently it caused an issue for this kind of user). Reverting...

setup.md
LarkAscending: There was a reason it was worded this way - on a loaner Mac it was necessary to edit /etc/bashrc to get the PATH to stick. It took a lot of trying and time. Despite that, I'll make this more general...

FYI it's OK (and totally expected) if your mileage varies on this one. The point is YMMV and we can't tell you exactly which file in what directory; it depends how the user's machine is set up. We're not actually pointing to any particular file.

lukechurch: @LarkAscending I'll follow up with a brief email about this

intellij-setup.md
```diff
@@ -5,7 +5,7 @@ sidebar: home_sidebar
 permalink: /intellij-setup/
 ---
 
-This page describes how to set up IntelliJ IDE to develop Flutter apps.
+This page describes how to set up an IntelliJ IDE to develop Flutter apps.
```
lukechurch: s/an/the?

(Having said which this is your area of expertise not mine, so I'll happily stand corrected :))

intellij-setup.md
lukechurch: @mit-mit Is there a reason our users need to think about the Dart plugin at all? Could we just talk about the Flutter plugin and have it do the rest, including pulling in the Dart plugin?

intellij-setup.md
lukechurch: +1

intellij-setup.md
```diff
@@ -28,23 +28,27 @@ Android Studio (and various other JetBrains editors) is currently not supported.
...
   * The `Dart` plugin offers code analysis (code validation as you type, code completions, etc.).
   * The `Flutter` plugin powers Flutter developer workflows (running, debugging, hot reload, etc.).
 
 To install the plugins:
 
-1. Open plugin preferences (**Preferences>Plugins** on macOS, **File>Settings>Plugins** on Linux)
-1. Select **"Browse repositories…"**
-1. Search for `'dart'`, select the Dart plug-in and click `install` (do not restart yet).
+1. Start IntelliJ.
+1. Open plugin preferences (**Preferences>Plugins** on macOS, **File>Settings>Plugins** on Linux).
```
lukechurch: How easy would it be to do an OS detect + 2 versions of the page? That might make it a bit easier to read. We used to do this for Dart I think - and we'll almost certainly need to do it for Windows.

Perhaps something for the future?

intellij-setup.md
```diff
@@ -28,23 +28,27 @@ Android Studio (and various other JetBrains editors) is currently not supported.
...
+navigation panel when you create a new project.
 
-1. Open preferences (**Preferences** on macOS, **File>Settings** on Linux)
-1. Select **Languages & Frameworks>Flutter**
-1. Enter, or browse to, your Flutter SDK directory path in **Flutter SDK path**
+### Configuring the Flutter plugin
+
+1. Open preferences (**Preferences** on macOS, **File>Settings** on Linux).
+1. Select **Languages & Frameworks>Flutter**.
+1. Enter or browse to your Flutter SDK directory in **Flutter SDK path**
```
lukechurch: People need to know how to find the Flutter SDK path (until we can find it automatically). We can either do this by telling them when they git clone it that this is the SDK, or here that they should look for the place where they git cloned it into. I'll leave it up to your discretion as to which.

setup.md
```diff
@@ -8,72 +8,156 @@ permalink: /setup/
...
+    ✗ Xcode end user license agreement not signed; open Xcode or run the command 'sudo xcodebuild -license'.
+
+    ✗ Brew not installed; use this to install tools for iOS device development.
+
+
+      Download brew at http://brew.sh/.
+   ```
 
-The first time you run the `flutter` command, it will download its dependencies and compile
+   The first time you run the `flutter` command, it downloads its dependencies and compiles
```
LarkAscending: The possessive here says that the `flutter` tool itself has dependencies that it needs to run, which it takes care of (so you don't have to). This explains to users why the command takes more time to execute on the first run. If that's not what happens, LMK.

Things like Android Studio are dependencies of the Flutter SDK itself, which users need to take care of.

setup.md
```diff
@@ -92,33 +176,55 @@ may overflow your screen. Set the device scale under the **Window > Scale** menu
...
+   Flutter requires the latest Android SDK and Android SDK Platform-Tools packages.
+In Android Studio, you can verify these are up-to-date, as described in [Update the IDE and SDK Tools](https://developer.android.com/studio/intro/update.html).
+
+1. Add the bundled JDK directory to your PATH variable. For example, on macOS:
+   ```
+   $ export PATH=/Users/obiwan/Applications/Android\ Studio.app/Contents/jre/jdk/Contents/Home/bin:$PATH
+   ```
+   or on Linux:
+   ```
+   $ export PATH=~/Android-Studio/jre/bin:$PATH
```
lukechurch: Do we need to repeat the previous instructions to make this persist, or is being available once sufficient?

setup.md
LarkAscending: Using Linux/Android, and maybe am not understanding what's behind the question... Xcode doesn't run on Linux or Windows. Maybe we're expecting folks to use a VM? Maybe there's some other path?
If yes, we need to say macOS. If no, what's there is equivalent. "To develop Flutter apps for iOS, you need a Mac" says that.

setup.md
xster: +1 on removing this. It's fairly common to omit it on most open source software instructions that uses git.

setup.md
xster: Don't think we need to mention this

setup.md
```diff
@@ -8,72 +8,156 @@ permalink: /setup/
...
+
+   ```
+   export PATH=~/flutter/bin
+   ```
+and then run the `source <filename>` command to refresh the window. For example on Linux, 
+you might edit and source `~/.bash_profile`. On macOS, it might be `/etc/bashrc`.
+
+1. Verify that the `flutter/bin` directory is now in your PATH by running:
+   ```
+   $ echo $PATH
```
xster: Just checking `which flutter` is more direct

setup.md
```diff
@@ -92,33 +176,55 @@ may overflow your screen. Set the device scale under the **Window > Scale** menu
...
-  1. Open the default Xcode workspace in your project under `ios/Runner.xcworkspace`. 
-  2. Select the `Runner` project in the left navigation panel.
-  3. In the Target Settings page, make sure your Development Team is selected under **General > Signing > Team**.
-5. Start your app by running `flutter run`.
+   ```
+   $ brew install ideviceinstaller ios-deploy
+   ```
+1. If you see the following error:
+    ```
+   You must brew link libtasn1 libplist libusb before ideviceinstaller can be installed
```
xster: They seem a bit exceptional for the main setup page. This isn't really flutter related.

setup.md
mit-mit: That is technically 100% correct! But it's a bit of a systems interpretation, and I'm just afraid users might misread it to mean that flutter is able to download everything it needs automatically, which is not the case...

setup.md
mit-mit: Yes, my point was that for practical purposes there is no way to develop for iOS on anything but macOS. But new mobile developers may not know that. 

So having a leading sentence like "Note: iOS development is only available on macOS" seems like a good service to them.

setup.md
LarkAscending: Good suggestion, but I tried `which flutter` and it actually points to flutter/bin/flutter. Since they entered flutter/bin, let's keep it like this to avoid confusion.

setup.md
LarkAscending: Added "own" so it reads "its own". From the context, users should not be confused. If they are confused we'll take another pass.

setup.md
xster: That's intended. It points to the exact location of the executable. If the path wasn't correctly set up, it wouldn't return anything. It more directly indicates whether just using `flutter` will work in the future. 

If they just echo the $PATH, it may still output a .../flutter/bin location even if it doesn't exist or there's nothing in there. PATH is just a string.

setup.md
LarkAscending: This is what it says today. 

It might seem like adding statements like this is a good idea, and will give more clarity. But it has the opposite effect: clutter and confusion. Let's keep an eye on what actual users have trouble with.

setup.md
LarkAscending: Yes, I wondered the same thing. Will add a small comment.

setup.md
LarkAscending: When my PATH doesn't have flutter/bin, it doesn't show in echo $PATH and `flutter` doesn't work.
Since we've just made them put something in flutter/bin (via git clone), we can be relatively sure it's there. I'll look for something in FAQ/Troubleshooting/SO re:which flutter.

FYI there's nothing wrong with this thread in conversation but doc/UX is a different thing. flutter/bin/flutter vs. flutter/bin is called 'cognitive dissonance' and it is The Enemy of Understanding. ;-)

intellij-setup.md
LarkAscending: Uh, that's what the following example is supposed to do, show what that looks like. 
I'll add a sentence.
Here's why I avoided talking about it by name. During setup users might have to think about:
- flutter/ (the Flutter SDK path)
- flutter/bin (the installation directory)
- flutter/bin/flutter (what `which flutter` outputs, the path to the executable)

Too many directories to think about! Where's that installer....?

setup.md
LarkAscending: Disk space is a common hurtle (for everyone but us).

intellij-setup.md
LarkAscending: It is in the plan for future...this design really only works for <2 platforms.

intellij-setup.md
LarkAscending: There are multiple! (see intellij-setup.md)

setup.md
LarkAscending: It's here because the user can get into a loop and not get out, unless they know about `brew doctor`. (Ask me how I know...)
@lukechurch pointed out offline that this might have been corp-related.
Let's see if external users get into the loop...

intellij-setup.md
LarkAscending: Right now they have to click `Yes`. I think you're asking, can we remove that step.
If it's possible, +1 to that idea...

Until then, see the latest commit.

sethladd:
RSLGTM

Thanks for working on this!

PR #414, opened 2017-03-04
Adding Android emulator steps (Mac only for now)

LarkAscending:
Adding steps for Android emulator setup (no KVM setup needed).
* Sourced internally
* Tested locally.
* Mac only for now...Linux and Windows to come...

@cbracken FYI

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -221,6 +221,20 @@ To develop Flutter apps for Android, you can use either a Mac or a Linux (64-bit
 
 1. Start Android Studio.
 
+### Set up the Android emulator
+
+To prepare to run and test your Flutter app on the Android emulator, follow these steps:
+
+1. In **Android Studio>Configure>SDK Manager**, install the system image(s) for the Android
+version you want to emulate.
+1. Launch **Android Studio>Tools>Android>AVD Manager** to define an Android Virtual Device (AVD).
+For detailed steps, see [Managing AVDs](https://developer.android.com/studio/run/managing-avds.html).
+1. In Android Virtual Device Manager, click **Run** in the toolbar.
+   The emulator starts up and displays the default canvas for your selected OS version
+   and device.
+1. Start your app by running `flutter run`. The connected device name is
+   `Android SDK built for <platform>`, where _platform_ is the chip family, such as x86.
+
 ### Set up your Android device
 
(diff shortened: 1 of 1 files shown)


setup.md
```diff
@@ -221,6 +221,20 @@ To develop Flutter apps for Android, you can use either a Mac or a Linux (64-bit
 
 1. Start Android Studio.
 
+### Set up the Android emulator
+
+To prepare to run and test your Flutter app on the Android emulator, follow these steps:
+
+1. In **Android Studio>Configure>SDK Manager**, install the system image(s) for the Android
+version you want to emulate.
+1. Launch **Android Studio>Tools>Android>AVD Manager** to define an Android Virtual Device (AVD).
```
mit-mit: Maybe add a recommendation to use an `x86` image?

tvolkert: Or `x86_64`, and until @chinmaygarde turns on the software backend by default, they need to enable hardware acceleration by checking "Use Host GPU".

eseidelGoogle:
FYI @pq (re: https://github.com/flutter/flutter-intellij/issues/692)

setup.md
LarkAscending: Looked for this setting in AVD Manager, did not find it. Is it associated w/one or more specific Android versions?

tvolkert: Hmm might be - I have sdk 25.

setup.md
```diff
@@ -221,6 +221,20 @@ To develop Flutter apps for Android, you can use either a Mac or a Linux (64-bit
 
 1. Start Android Studio.
 
+### Set up the Android emulator
+
+To prepare to run and test your Flutter app on the Android emulator, follow these steps:
+
+1. In **Android Studio>Configure>SDK Manager**, install the system image(s) for the Android
```
LarkAscending: OK, good to know. Configure was nice because you didn't have to go through loading a project, which takes a while (and you might not have one). I can't access AVD Manager from anywhere but inside Android Studio.

setup.md
LarkAscending: I don't see it on Linux, SDK 25... but a clue - you're using CLI. In the UI it's here:
https://developer.android.com/studio/run/emulator-acceleration.html

PR #415, opened 2017-03-04
Removing outbound icon from layer cake, changing link text color

LarkAscending:
This change was left out of Thinking in Flutter commit, but does not fix https://github.com/flutter/flutter/issues/8545


LarkAscending:
Build failure unrelated to this PR. In process here #424

sethladd:
Thanks!

PR #417, opened 2017-03-06
Fix comment to match method name.

butlermatt:


butlermatt:
Will try to fix my commit as I know that I've signed the CLA before.

PR #421, opened 2017-03-08
optipng all the images in layout tutorial

sethladd:


sethladd:
Saved around 20-30% across the board. Still more to be done.

TBR= @Sfshaza

PR #426, opened 2017-03-09
Add libimobiledevice install step

cbracken:
The most recently published binary version of the libimobiledevice
homebrew formula is relatively out-of-date and doesn't work with recent
Xcode releases. Update instructions to explicitly install
libimobiledevice with --HEAD so as not to pick up the out-of-date
version through an implicit dependency of ideviceinstaller.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -173,6 +173,7 @@ To deploy your Flutter app to a physical iOS device, you’ll need some addition
 iOS devices.
 
    ```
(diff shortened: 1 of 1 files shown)


cbracken:
Fixes #8674

cbracken:
@chinmaygarde @LarkAscending 

/cc @mit-mit

LarkAscending:
LGTM

@cbracken  well done

PR #429, opened 2017-03-09
Change language around Windows

goderbauer:
Setup instructions for Windows will be integrated in a follow-up PR.

Partial fix for https://github.com/flutter/flutter/issues/8654.

/cc @LarkAscending

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -46,7 +46,7 @@ inline error checking, and visual debugging features.
 
 To install and run Flutter, your development environment must meet these minimum requirements:
 
-* **Operating Systems**: Mac or Linux (64-bit). Windows support is [in progress](https://github.com/flutter/flutter/issues/138).
+* **Operating Systems**: Mac or Linux (64-bit). For Windows support see [Flutter on Windows](https://github.com/flutter/flutter/wiki/Flutter-on-Windows).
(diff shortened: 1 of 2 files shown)


setup.md
```diff
@@ -46,8 +46,8 @@ inline error checking, and visual debugging features.
 
 To install and run Flutter, your development environment must meet these minimum requirements:
 
-* **Operating Systems**: Mac or Linux (64-bit). Windows support is [in progress](https://github.com/flutter/flutter/issues/138).
-* **Disk Space**: 700 MB on Mac, 600 MB on Linux (does not include disk space for Xcode or Android Studio).
+* **Operating Systems**: Mac or Linux (64-bit). For Windows instructions (beta), see [Flutter on Windows](https://github.com/flutter/flutter/wiki/Flutter-on-Windows).
```
Hixie: We should eventually merge these instructions in so that this one document covers all platforms, rather than having some second-class citizen platforms on the wiki.

goderbauer: Yes, that's the plan (see PR description). However, until that is ready we want to guide windows users to the right place instead of turning them away :)

Hixie: +1

PR #431, opened 2017-03-10
Add brew update step for macOS install

cbracken:
Ensures that developers are installing up-to-date versions of the
homebrew formulae.


cbracken:
Partial fix to flutter/flutter#8670

PR #432, opened 2017-03-10
Fixing outbound icon issue and a few apwilson changes.

LarkAscending:
1. Only 4 links on the splash page should not have had an outbound icon. Fix involved changed these to relative links.
1. To reduce icon noise, this PR ADDS A CSS RULE to consider docs.flutter.io as part of our own domain. So links to the API docs will not have an outbound icon, with this change. Thus, most of the links are now icon-less.
  This preserves the spirit of the outbound icon and its utility for users, while making it scarcer and less annoying.

1. Includes 9, 10, 17 from @apwilson feedback.

### DIFF
--- a/_sass/_customstyles.scss
+++ b/_sass/_customstyles.scss
@@ -950,7 +950,7 @@ a[href^="http://"]:after, a[href^="https://"]:after {
 }
 
 /* Strip the outbound icon when this class is present */
-a[href^="http://own-domain.com"]:after,
+a[href^="http://own-domain.com"]:after, a[href^="https://docs.flutter.io"]:after, 
 a.no_icon:after {
     content:none !important;
     padding-left: 0;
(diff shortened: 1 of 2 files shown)


FaisalAbid:
Ah crap. so good news is we fixed the firebase issue
Bad news is Github is giving an issue with tokens.

taking a look, shoot

FaisalAbid:
no way we did 5k requests in an hour lol

> For requests using Basic Authentication or OAuth, you can make up to 5,000 requests per hour. 
https://developer.github.com/v3/#rate-limiting

sethladd:
FWIW I think we could experiment with removing the outbound link completely. Up to you.

LarkAscending:
Open to it. Thought we'd try this as an experiment.

The original idea was to give users a heads up that they're going to an
outside site. To draw a virtual boundary, and also to not take credit/blame
for what happens when they click.
Viruses, 404s, low-quality or irrelevant content, redirects, etc.

We curate the majority of the sites marked w/the icon, even now.
Setup has the Apple Developer Site, the DAC site, the Homebrew site, etc.
In plugins, there will be external links.
So that's the scope.

PR #437, opened 2017-03-14
Use color property for defining color of a container.

mdakin:


abarth:
LGTM

mdakin:
@sethladd I think this is good to go now :)

sethladd:
Thanks!

PR #439, opened 2017-03-14
Networking HowTo

mit-mit:
Several folks at First Look asked me how to do networking in Flutter. Here is a first draft of a new page that covers the basics of that, incl. a small example.

### DIFF
--- a/networking.md
+++ b/networking.md
@@ -174,10 +174,10 @@ For full API docs, see:
   * [library `dart:io`][dartio]
 
 [http]:       https://pub.dartlang.org/packages/http
-[client]:     https://www.dartdocs.org/documentation/http/stable/http/Client-class.html
-[get]:        https://www.dartdocs.org/documentation/http/stable/http/get.html
-[read]:       https://www.dartdocs.org/documentation/http/stable/http/read.html
-[readbytes]:  https://www.dartdocs.org/documentation/http/stable/http/readbytes.html
-[post]:       https://www.dartdocs.org/documentation/http/stable/http/post.html
+[client]:     https://www.dartdocs.org/documentation/http/latest/http/Client-class.html
+[get]:        https://www.dartdocs.org/documentation/http/latest/http/get.html
+[read]:       https://www.dartdocs.org/documentation/http/latest/http/read.html
+[readbytes]:  https://www.dartdocs.org/documentation/http/latest/http/readbytes.html
+[post]:       https://www.dartdocs.org/documentation/http/latest/http/post.html
 [convert]:    https://docs.flutter.io/flutter/dart-convert/dart-convert-library.html
 [dartio]:     https://api.dartlang.org/stable/dart-io/dart-io-library.html
--- a/networking.md
+++ b/networking.md
@@ -1,8 +1,11 @@
 ---
 layout: page
-title: Making Networking Requests in Flutter
+title: Making HTTP Networking Requests in Flutter
 permalink: /networking/
 ---
+
+This page discusses how to make HTTP networking requests in Flutter. For
+HTTP servers and sockets, see [dart:io][dartio].
 
 * TOC Placeholder
 {:toc}
@@ -76,10 +79,11 @@ Map, List, or List of Maps containing simple values to the `encode` method:
 String encodedString = JSON.encode([1, 2, { "a": null }]);
 ```
 
-## Example Flutter app decoding JSON returned by a HTTP GET call
+## Example Flutter app decoding JSON returned by a HTTPS GET call
 
-The following example calls the jsontest.com web service API. This responds with
-your local IP address.
+The following example calls the [httpbin.com](http://httpbin.com) web service
+testing API. This responds with your local IP address. Note that it uses secure
+'https' networking.
 
 1. Create a new flutter app with `flutter create`
 
@@ -116,11 +120,11 @@ class _MyHomePageState extends State<MyHomePage> {
   String _ipAddress = 'Unknown';
 
   _getIPAddress() async {
-    String url = "http://ip.jsontest.com/";
+    String url = "https://httpbin.org/ip";
     var httpClient = new http.Client();
     var response = await httpClient.read(url);
     Map data = JSON.decode(response);
-    String ip = data["ip"];
+    String ip = data["origin"];
 
     setState(() {
       _ipAddress = ip;
@@ -153,13 +157,16 @@ class _MyHomePageState extends State<MyHomePage> {
 
 For full API docs, see:
 
+  * [package `http`][http]
   * [`Client` in package `http`][client]
-  * [`dart:convert` library][convert]
+  * [library `dart:convert`][convert]
+  * [library `dart:io`][dartio]
 
 [http]:       https://pub.dartlang.org/packages/http
-[client]:     https://www.dartdocs.org/documentation/http/0.11.3%2B10/http/Client-class.html
+[client]:     https://www.dartdocs.org/documentation/http/stable/http/Client-class.html
 [get]:        https://www.dartdocs.org/documentation/http/stable/http/get.html
 [read]:       https://www.dartdocs.org/documentation/http/stable/http/read.html
 [readbytes]:  https://www.dartdocs.org/documentation/http/stable/http/readbytes.html
 [post]:       https://www.dartdocs.org/documentation/http/stable/http/post.html
 [convert]:    https://docs.flutter.io/flutter/dart-convert/dart-convert-library.html
+[dartio]:     https://api.dartlang.org/stable/dart-io/dart-io-library.html
--- a/networking.md
+++ b/networking.md
@@ -1,0 +1,157 @@
+---
+layout: page
+title: Making Networking Requests in Flutter
+permalink: /networking/
+---
+
+* TOC Placeholder
+{:toc}
+
+## Using the `http` package
+
+Flutter supports the the [http package][http], version `0.11.3+12` or later.
+
+Start by declaring a dependency by adding an http line under `dependencies` in
+`pubspec.yaml`:
+
+```
+dependencies:
+  ...
+  http: '>=0.11.3+12'
+```
+
+## Making requests
+
+You can create an HTTP [Client][client] with the `Client` constructor:
+
+```dart
+import 'package:http/http.dart' as http;
+
+http.Client _httpClient = new http.Client();
+```
+
+The client supports common HTTP operations:
+
+* **HTTP GET:** Use [`get`][get] for a general request, [`read`][read] for a
+ request returning a string, or [`readbytes`][readbytes] for a request
+ returning bytes.
+
+* **HTTP POST:** Use [`post`][post] for a general post.
+
+Note that the API uses [Dart
+Futures](https://www.dartlang.org/tutorials/language/futures) in the return
+values. These are most easily used with the `then()` methods like this:
+
+```dart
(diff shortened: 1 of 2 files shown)


mit-mit:
Note: Due to https://github.com/dart-lang/http/issues/66 link checking will currently fail.

networking.md
```diff
@@ -0,0 +1,162 @@
...
+Start by declaring a dependency by adding an http line under `dependencies` in
+`pubspec.yaml`:
+
+```
+dependencies:
+  ...
+  http: '>=0.11.3+12'
+```
+
+## Making requests
```
sethladd: If this is a sub-header, please use ### , otherwise, we can be more specific in this header: "Making HTTP requests".  (to clarify: which kind of requests?)

networking.md
```diff
@@ -0,0 +1,162 @@
...
+
+You can create an HTTP [Client][client] with the `Client` constructor:
+
+<!-- skip -->
+```dart
+import 'package:http/http.dart' as http;
+
+http.Client _httpClient = new http.Client();
```
sethladd: We discussed snippet style, and it's OK to write something simpler: `var httpClient = new http.Client()`

networking.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+The client supports common HTTP operations:
```
sethladd: operations, such as:

networking.md
```diff
@@ -0,0 +1,162 @@
...
+
+* **HTTP GET:** Use [`get`][get] for a general request, [`read`][read] for a
+ request returning a string, or [`readbytes`][readbytes] for a request
+ returning bytes.
+
+* **HTTP POST:** Use [`post`][post] for a general post.
+
+Note that the API uses [Dart
```
sethladd: Consider putting this after the code snippet, so it's more of a note.

As written, "the API" isn't clear what API you're talking about. But, after the code, you can refer to "the code above".

networking.md
```diff
@@ -0,0 +1,162 @@
...
+Futures](https://www.dartlang.org/tutorials/language/futures) in the return
+values. These are most easily used with the `then()` methods like this:
```
sethladd: I think we want to prefer showing async/await syntax. More modern, and it makes for easier to read code.

networking.md
```diff
@@ -0,0 +1,162 @@
...
+To decode/parse the JSON string response into a Map:
+
+<!-- skip -->
+```dart
+Map data = JSON.decode(response.body);
+// Assume the response body is something like: ["foo", { "bar": 499 }]
+int barValue = data[1]['bar']; // barValue is set to 499
+```
+
+To encode JSON simply pass in a `Map` with the values:
```
sethladd: Maybe say "pass a Map or List of Maps, Lists, and string, boolean, or number literals"

networking.md
```diff
@@ -0,0 +1,162 @@
...
+
+<!-- skip -->
+```dart
+String encodedString = JSON.encode([1, 2, { "a": null }]);
+```
+
+## Example
```
sethladd: Can we be more specific here? perhaps, "Example of HTTP and JSON together". Great for SEO and more clear what the intention of the example, for scanning through the doc.

networking.md
```diff
@@ -0,0 +1,162 @@
...
+
+The following example calls the jsontest.com web service API. This responds with
+your local IP address.
+
+1. Create a new flutter app with `flutter create`
+
+1. Add the http dependency as discussed at the top of this page.
+
+1. Replace the contents of `main.dart` with the following:
```
sethladd: maybe: "lib/main.dart"

networking.md
```diff
@@ -0,0 +1,162 @@
...
+  _MyHomePageState createState() => new _MyHomePageState();
+}
+
+class _MyHomePageState extends State<MyHomePage> {
+  String _ipAddress = 'Unknown';
+
+  void _getIPAddress() {
+    String url = "http://ip.jsontest.com/";
+    http.Client _httpClient = new http.Client();
+    _httpClient.read(url).then((String response) {
```
sethladd: Let's use async/await here.

networking.md
```diff
@@ -0,0 +1,162 @@
...
+  @override
+  _MyHomePageState createState() => new _MyHomePageState();
+}
+
+class _MyHomePageState extends State<MyHomePage> {
+  String _ipAddress = 'Unknown';
+
+  void _getIPAddress() {
+    String url = "http://ip.jsontest.com/";
+    http.Client _httpClient = new http.Client();
```
sethladd: we can use var here.

networking.md
```diff
@@ -0,0 +1,162 @@
...
+
+      setState(() {
+        _ipAddress = ip;
+      });
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    Widget spacer = new SizedBox(height: 32.0);
```
sethladd: Is this best practice, to use a more general class for the type annotation?

Using var here will work in strong mode.

networking.md
mit-mit: Done

networking.md
mit-mit: OK, great, done

networking.md
mit-mit: Done

networking.md
mit-mit: Done

networking.md
mit-mit: Yes, good idea, done

networking.md
mit-mit: Done

networking.md
mit-mit: Done

networking.md
mit-mit: Done

networking.md
mit-mit: Sounds good

networking.md
mit-mit: Yup

networking.md
mit-mit: I don't know. But changed to var.

networking.md
```diff
@@ -0,0 +1,165 @@
+---
+layout: page
+title: Making Networking Requests in Flutter
+permalink: /networking/
+---
+
+* TOC Placeholder
+{:toc}
+
```
sethladd: One more thought:

People might think that all Flutter can do is http for networking. Maybe put a little preamble here, linking to our support for sockets (TCP, UDP), and ensuring the reader knows we can do all sorts of networking but because HTTP is the most common thing, we'll focus on it here, but don't forget about sockets, etc, etc.

Thanks!

mit-mit:
Added preamble mentioning sockets and http server, and updated example code to use https.

mit-mit:
@abarth can you give this a quick review for technical accuracy?

mit-mit:
@LarkAscending can you review this for general writing, please?

networking.md
```diff
@@ -4,17 +4,17 @@ title: Making HTTP Networking Requests in Flutter
 permalink: /networking/
 ---
 
-This page discusses how to make HTTP networking requests in Flutter. For
+This page describes how to make HTTP networking requests in Flutter. For
 HTTP servers and sockets, see [dart:io][dartio].
```
sethladd: I don't think we'd promote that you can build HTTP servers in Flutter :) so it's OK to scope this to just "For more on how to use TCP and UDP sockets from within Flutter, ..."

LarkAscending: Was doing a quick pass, w/o time to research exactly what we were saying here. This gets my vote.

networking.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Making HTTP requests
+
+Next, create an HTTP [Client][client] with the `Client` constructor:
+
+<!-- skip -->
+```dart
+import 'package:http/http.dart' as http;
+
+var httpClient = new http.Client();
```
abarth: We should recommend that folks use https://docs.flutter.io/flutter/services/createHttpClient.html .  It's useful to send all the client creation through that point so you can mock it during testing, etc.

networking.md
```diff
@@ -0,0 +1,174 @@
...
+1. Create a new flutter app with `flutter create`.
+
+1. Add the [http dependency](#using-http-package).
+
+1. Replace the contents of `lib/main.dart` with the following:
+
+```dart
+import 'package:flutter/material.dart';
+import 'package:http/http.dart' as http;
+import 'dart:convert';
```
abarth: Dart style would put this import first with a blank line after it.

networking.md
```diff
@@ -0,0 +1,174 @@
...
+  String _ipAddress = 'Unknown';
+
+  _getIPAddress() async {
+    String url = "https://httpbin.org/ip";
+    var httpClient = new http.Client();
+    var response = await httpClient.read(url);
+    Map data = JSON.decode(response);
+    String ip = data["origin"];
+
+    setState(() {
```
abarth: You need to checkout `mounted` before calling setState to see if your widget is still on the screen.

networking.md
```diff
@@ -0,0 +1,174 @@
...
+      body: new Center(
+        child: new Column(children: <Widget>[
+          spacer,
+          new Text('Your current IP address is:'),
+          new Text('$_ipAddress.'),
+          spacer,
+          new RaisedButton(
+            onPressed: _getIPAddress,
+            child: new Text('Get IP address'),
+          )
```
abarth: Missing trailing comma

networking.md
```diff
@@ -0,0 +1,174 @@
...
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    var spacer = new SizedBox(height: 32.0);
+
+    return new Scaffold(
+      body: new Center(
+        child: new Column(children: <Widget>[
```
abarth: We usually put a newline before `children`

networking.md
```diff
@@ -0,0 +1,174 @@
...
+          spacer,
+          new Text('Your current IP address is:'),
+          new Text('$_ipAddress.'),
+          spacer,
+          new RaisedButton(
+            onPressed: _getIPAddress,
+            child: new Text('Get IP address'),
+          )
+        ]),
```
abarth: Missing comma between `]` and `)`

abarth:
LGTM  ... but not to Travis.

networking.md
mit-mit: Do we always include that check, even for cases like this where the set of widgets is fixed?

abarth: Strictly speaking, it's not needed in this example.  However, people are likely to copy and paste this example code into situations where the check will be needed.  It's better to include the check for the general case to improve their copy/paste success rate.

networking.md
mit-mit: Yup, this made the formatter fix the "newline before `children`" part :-)

mit-mit:
This is done, but requires https://github.com/astashov/dartdocs.org/issues/42 to get the broken links fixed.

sethladd:
Thanks @mit-mit !

networking.md
```diff
@@ -0,0 +1,184 @@
...
+sockets, see [dart:io][dartio].
+
+* TOC Placeholder
+{:toc}
+
+## Using the `http` package
+
+Flutter supports the [http package][http], version `0.11.3+12` or later.
+
+The first step is to declare a dependency by adding an `http...` line under `dependencies` in
```
sethladd: People might be confused what `http...` means. Consider dropping "an `http....`" and just having: "declare a dependency on the `http` package under...."

PR #440, opened 2017-03-14
Renamed list_item.dart to list_tile.dart

HansMuller:
See https://github.com/flutter/flutter/pull/8769


Hixie:
LGTM

Hixie:
might also be listed in a csv file somewhere. gets used to generate the basic widgets list.

PR #443, opened 2017-03-15
Fixed a typo: and and -> and

domrein:
The word "and" was accidentally repeated twice in the Adding Interactivity to your Flutter App tutorial.


domrein:
I signed it!

Sfshaza:
Thanks so much for catching and fixing that, @domrein!

PR #445, opened 2017-03-17
Added brew commands for cocoapod support.

LarkAscending:


setup.md
```diff
@@ -176,16 +176,11 @@ iOS devices.
    $ brew update
    $ brew install --HEAD libimobiledevice
    $ brew install ideviceinstaller ios-deploy
+   $ brew install cocoapods
+   $ brew link cocoapods
```
eseidelGoogle: link is implied by install, only needed when the install fails. :)

eseidelGoogle:
Otherwise lgtm, @cbracken is probably your correct reviewer though. Thanks!

setup.md
eseidelGoogle:
```
 %brew help link                                   [/src/flutter]
brew ln, link [--overwrite] [--dry-run] [--force] formula:
    Symlink all of formula's installed files into the Homebrew prefix. This
    is done automatically when you install formulae but can be useful for DIY
    installations.
```

LarkAscending: It failed for me...possibly b/c of corp machine.
But, good to know.

cbracken: Yep; just `brew install` is sufficient & consistent with our other homebrew instructions.

cbracken: Also feel free to merge this up onto the previous line; so just:
`brew install ideviceinstaller ios-deploy cocoapods`

The reason for splitting out libimobiledevice is because it's currently required to be installed from `HEAD`, unlike the ones on the second line.

PR #446, opened 2017-03-17
Adding first Windows setup info!

LarkAscending:
First take at adding Windows to flutter.io/setup.

Some sort of picker widget (like a dropdown list w/platforms) or zippies for showing/hiding platforms that don't apply, is TBD.

Note that this PR doesn't instruct Windows users to add the JDK to their PATH; we're pretty sure that @johnmccutchan 's PATH detection goodness also works on Windows. It would be good to verify this is the case, i.e. what happens if the user doesn't manually add this to the system PATH variable.

There are several other omissions, like opening a new terminal window. This might be too streamlined - we can add back in.

Ideally, I'd like to test this on a Windows machine before merging.


setup.md
```diff
@@ -76,6 +94,18 @@ To install and run Flutter, your development environment must meet these minimum
...
+#### Windows
+
+1. Add Flutter to your PATH environment variable by opening 
+**Control Panel>System and Security>System>Advanced system settings>Environment Variables...**
+1. Modify the `PATH` system variable to add the full path to the `flutter\bin` directory.
+   For example, `C:\flutter\bin`.
+
+   If you don't have admin access to your machine, open **Control Panel>User Accounts>User Accounts>Change my environment variables**
+   and add a `PATH` user variable with the path to `flutter\bin` as its value. 
+
```
goderbauer: I would add a not here telling people that they need to restart PowerShell or Cmd for this change to PATH to have effect.

mit-mit: Yes, John's change was just for `flutter doctor`, we still need this for `flutter run/build` until https://github.com/flutter/flutter/issues/8515 is fixed.

goderbauer:
Given that people are already trying to run Flutter on Windows in Git Bash (see https://github.com/flutter/flutter/issues/8860) it might be worth pointing out that we only support Cmd and PowerShell on Windows.

setup.md
```diff
@@ -46,20 +46,38 @@ inline error checking, and visual debugging features.
...
 To install and run Flutter, your development environment must meet these minimum requirements:
 
-* **Operating Systems**: Mac or Linux (64-bit). For Windows instructions (beta), see [Flutter on Windows](https://github.com/flutter/flutter/wiki/Flutter-on-Windows).
+* **Operating Systems**: Mac or Linux or Windows. (64-bit)
 * **Disk Space**: 700 MB on Mac, 600 MB on Linux, 400 MB on Windows (does not include disk space for Xcode or Android Studio).
-* **Tools**: Flutter depends on these command-line tools being available on your environment.
-  * `bash`, `mkdir`, `rm`, `git`, `curl`, `unzip`
+* **Tools**: Flutter depends on these command-line tools being available in your environment.
+  * `bash`, `mkdir`, `rm`, `git`, `curl`, `unzip`(Mac and Linux), 
+  * Git with command-line support, PowerShell (Windows)
```
goderbauer: the command line tool `where` is also required on Windows.

mit-mit: Actually, I am not sure I understand the benefit of calling out a bunch of tools that come with the operating system (not just for Windows, same for Mac and Linux)?

goderbauer: @Hixie asked to document this.

Thinking about it now, I'd have to agree with @mit-mit. I think we should only list things that people have to install manually (like Git on Windows). Listing things that are part of your OS is kinda confusing.

Hixie: We should document this _somewhere_. It can be on the CONTRIBUTING.md file if we don't think it's appropriate for the website.

We regularly run into users who don't have critical tools installed. You'd be surprised what isn't considered "part of your OS", especially on Linux-based systems.

setup.md
goderbauer: https://github.com/flutter/flutter/issues/8515 is fixed now.

LarkAscending: OK, removing JDK PATH instructions under "Android Studio", keeping flutter\bin PATH instructions.

setup.md
LarkAscending: Adding `where` for now, in case it helps users troubleshoot w/o confusing them. In an enterprise environment, it's possible to have perm issues that cause weird errors.  If it's confusing we'll remove `where`.

LarkAscending:
@mit-mit Re: Travis failure, is it possible that HTML Proofer is failing again on newly-added files?

There is one valid broken link, in tutorials/interactive/index.html. Will fix that after merging changes from master...

mit-mit:
> @mit-mit Re: Travis failure, is it possible that HTML Proofer is failing again on newly-added files?

No, that is not it. It's the relative links that are wrong. Don't use:
```
<a href="setup-linux/#android-setup">Linux</a>
```

Use:
```
<a href="/setup-linux/#android-setup">Linux</a>
```

_includes/setup/android-setup.md
```diff
@@ -0,0 +1,53 @@
+## Android setup
+
+_This step is optional, and can be performed at any time before
+using Flutter to develop apps for Android._
+
+### Install Android Studio
+
+To develop Flutter apps for Android, you can use either a Mac or a Linux (64-bit) machine:
```
mit-mit: `you can use either a Mac or a Linux (64-bit) machine` is no longer true (it's also windows).

But given all three OS's support Android Studio, just leave out all together

_includes/setup/android-setup-linux.md
```diff
@@ -0,0 +1,89 @@
+## Android setup
```
mit-mit: Why do we need a separate copy of the whole android file for Linux? Can't we just include in the one extra KVM step?

_includes/setup/checklist.md
```diff
@@ -0,0 +1,33 @@
+## Checklist
```
mit-mit: This whole `setup/checklist.md` file doesn't appear to be used anywhere?

_includes/setup/system-linux.md
```diff
@@ -0,0 +1,8 @@
+## System requirements
```
mit-mit: When navigating to either of setup_linux, setup_mac, or setup_win, the whole sidenav collapses:

screen shot 2017-03-24 at 13 06 34

_includes/setup/system-win.md
```diff
@@ -0,0 +1,11 @@
+## System requirements
+
+To install and run Flutter, your development environment must meet these minimum requirements:
+
+* **Operating Systems**: Windows (64-bit)
```
mit-mit: @goderbauer why do we require 64-bit Windows?

_includes/setup/system-win.md
```diff
@@ -0,0 +1,11 @@
...
+* **Disk Space**: 400 MB (does not include disk space for Xcode or Android Studio).
```
mit-mit: Remove Xcode as that does not exist for windows

_includes/setup/system-linux.md
```diff
@@ -0,0 +1,8 @@
...
+
+To install and run Flutter, your development environment must meet these minimum requirements:
+
+* **Operating Systems**: Linux (64-bit)
+* **Disk Space**: 600 MB (does not include disk space for Xcode or Android Studio).
```
mit-mit: Remove Xcode as that does not exist for windows

_includes/setup/next-steps-win-linux.md
```diff
@@ -0,0 +1,5 @@
+## Next steps
+
+Now, follow our [Getting Started guide](/getting-started/)
+to run your first Flutter app on Android. To develop for iOS in addition to Android,
```
mit-mit: I don't understand this next step. How many developers do we think have multiple machines with multiple different operating systems?

And even in that case, I am pretty sure they can find their way to the start of the installation instructions.

I would much rather not have the next-steps.md file be forked per OS. I think we can simply change to a single `next-steps.md` file with:

```
## Next steps

Now, follow our [Getting Started guide](/getting-started/)
to run your first Flutter app.
```

mit-mit:
Great to see Windows support! I added a first round of review comments.

mit-mit:
Also, one general concern: The whole flow starts with this:

screen shot 2017-03-24 at 12 55 32

I have two overall concerns with this:

1. It's confusing to see the checklist here, and then not see it on the per-os pages.

2. Why do we ask for OS? Can't we just detect what OS they are on, and take them to the relevant page?

_includes/setup/android-setup-linux.md
```diff
@@ -0,0 +1,89 @@
...
+1. Enable Kernel-based Virtual Machine (KVM) support on your machine.
+1. Define a virtual device in Android Virtual Device Manager.
+
+#### Enable KVM
+For the Android emulator, you need to configure x86 and x86_64 emulation, which 
+requires the use of hardware virtualization. The steps to enable hardware virtualization 
+are:
+
+1. Reboot into the system BIOS (usually by repeatedly tapping F1 or F10 at the first page during startup).
+1. In **Security > System Security**:
```
goderbauer: Should we note that this is highly platform specific? The setting might be located somewhere else in different BIOSes...

_includes/setup/android-setup-linux.md
```diff
@@ -0,0 +1,89 @@
...
+1. Verify the AVD configuration is correct, and select **Finish**.
+
+   For details on the above steps, see [Managing AVDs](https://developer.android.com/studio/run/managing-avds.html).
+1. In Android Virtual Device Manager, click **Run** in the toolbar.
+   The emulator starts up and displays the default canvas for your selected OS version
+   and device.
+1. Start your app by running `flutter run`. The connected device name is
+   `Android SDK built for <platform>`, where _platform_ is the chip family, such as x86.
+
+### Set up your Android device
```
goderbauer: Personally, I would put this section above the emulator section. It's much simpler to get running on a phone than it is in an emulator and more satisfying as well...

_includes/setup/android-setup.md
```diff
@@ -0,0 +1,53 @@
+## Android setup
```
goderbauer: Is this a dup of the previous file?

_includes/setup/editor-setup.md
```diff
@@ -0,0 +1,7 @@
+## Editor setup
+
+Using our command-line tools, you can use any editor to develop Flutter applications.
```
goderbauer: Should we mention that the command line tool is called `flutter` and you can learn more about it by typing `flutter help` in a terminal?

_includes/setup/get-sdk-win.md
```diff
@@ -0,0 +1,59 @@
...
+Running `flutter doctor` shows any remaining dependencies you may need to install.
+
+### Clone the repo
+
+If this is the first time you're installing Flutter on this machine, clone the
+repository and then add the `flutter` tool to your path:
+
+```
+> git clone https://github.com/flutter/flutter.git
+> setx path "%path%;%cd%\flutter\bin"
```
goderbauer: Nice! I didn't know about `setx`.

_includes/setup/get-sdk-win.md
```diff
@@ -0,0 +1,59 @@
...
+```
```
goderbauer: We should mention that you need to open a new terminal for this to take effect. Otherwise, `flutter doctor` in the next step will just fail.

From https://ss64.com/nt/setx.html:
> Because SETX writes variables to the master environment in the registry, edits will only take effect when a new command window is opened - they do not affect the current CMD or PowerShell session.

_includes/setup/get-sdk-win.md
```diff
@@ -0,0 +1,59 @@
+## Get the Flutter SDK
+
+To get Flutter, use `git` to clone the repository and then add the `flutter` tool to your path.
+Running `flutter doctor` shows any remaining dependencies you may need to install.
+
+### Clone the repo
+
+If this is the first time you're installing Flutter on this machine, clone the
+repository and then add the `flutter` tool to your path:
```
goderbauer: Can we mention that you should do this in Cmd or PowerShell? We currently don't support third-party shells (like Git Bash or Bash on Ubuntu on Windows).

_includes/setup/get-sdk-win.md
```diff
@@ -0,0 +1,59 @@
...
+the setup:
+
+```
+> flutter doctor
+```
+
+This command checks your environment and displays a report to the terminal window.
+The Dart SDK is bundled with Flutter; it is not necessary to install Dart separately.
+Check the output carefully for other software you may need to install or further 
+tasks to perform (shown in **bold** text).
```
goderbauer: nit: The text will only be in bolt on more recent versions of Windows 10. Older versions of Windows 10 and old versions of Windows don't support bolt text in the terminal.

_includes/setup/get-sdk-win.md
```diff
@@ -0,0 +1,59 @@
...
+
+For example:
+```
+✗ Xcode end user license agreement not signed; open Xcode or run the command 'sudo xcodebuild -license'.
```
goderbauer: Xcode seems like a bad example in the Windows instructions :)

_includes/setup/next-steps-win-linux.md
```diff
@@ -0,0 +1,5 @@
...
+follow the [Flutter Setup for Mac](/setup-mac/) instructions.
```
goderbauer: Pointing to Flutter on Mac in the `next-step-win-linux.md` file seems strange. Do we expect Windows/Linux users to also have a mac machine?

_includes/setup/path-win.md
```diff
@@ -0,0 +1,23 @@
+### Update your path
```
goderbauer: Do we need these instructions? `setx` mentioned above should already be permanent (at least that's my understanding after reading https://ss64.com/nt/setx.html). No need to go through the UI.

_includes/setup/system-linux.md
```diff
@@ -0,0 +1,8 @@
...
+* **Tools**: Flutter depends on these command-line tools being available in your environment.
+  * `bash`, `mkdir`, `rm`, `git`, `curl`, `unzip`
```
goderbauer: This is missing the newly added `which` command.

_includes/setup/system-mac.md
```diff
@@ -0,0 +1,8 @@
+## System requirements
+
+To install and run Flutter, your development environment must meet these minimum requirements:
+
+* **Operating Systems**: macOS (64-bit)
+* **Disk Space**: 700 MB (does not include disk space for Xcode or Android Studio).
+* **Tools**: Flutter depends on these command-line tools being available in your environment.
+  * `bash`, `mkdir`, `rm`, `git`, `curl`, `unzip`
```
goderbauer: add `which`

_includes/setup/system-win.md
```diff
@@ -0,0 +1,11 @@
...
+* **Tools**: Flutter depends on these command-line tools being available in your environment.
+  * Git for Windows (with the Use Git from the Windows Command Prompt option), Cmd, PowerShell, `where`
+     
+     If Git for Windows is already installed, make sure you can run `git` commands from
+     Windows PowerShell and the Cmd command prompts.
```
goderbauer: I would say "PowerShell or the Cmd..."

_includes/setup/system-win.md
goderbauer: For all platforms (not just WIndows), Flutter tools currently assumes that you're running on 64bits. If we want to run on 32bit, we probably could, but we would have to add some logic to download the 32bit version of the Dart SDK (and possibly of our other pre-build binaries) if we detect such a platform. Personally, I don't think we will gain much by also supporting 32bit. If you disagree, you should file a feature request and we can discuss it there :)

_includes/setup/system-win.md
LarkAscending: Agreed, this wording is more correct because it's normally one or the other.

_includes/setup/system-mac.md
LarkAscending: Good one!

_includes/setup/path-win.md
LarkAscending: Have not done the negative test...would be great to remove.

LarkAscending:
@mit-mit We have an OS dropdown because it's what was initially asked for. OS detection - tho not perfect - is possible. It takes longer, is not a priority, and right now the setup instructions are already too long to keep all the platforms in one file. Maybe open an issue requesting this? Or just a PR if it's easy enough to do.

Also, the HTML Proofer URL requirements should probably go in a contributing.md file... see #https://github.com/flutter/flutter/issues/8187

_includes/setup/android-setup-linux.md
LarkAscending: Thanks for asking! Considered the nested include file approach. Not easy to maintain, though. For new contributors and team members, it's best to keep the organization simple. That's one reason. Another reason is the main drawback in this organization would be if we expected the Android setup instructions to change frequently. Then, each content change would need to be double-committed. In comparison to the other setup content, this piece seems more stable. Thus, not much down side to this approach.

setup_linux.md
```diff
@@ -0,0 +1,17 @@
+---
+layout: page
+title: Flutter Setup for Linux
+permalink: /setup-linux/
+---
+
```
LarkAscending: Just not at that point yet...good suggestion though!

In these long instructions, the first content a Flutter user interacts adding a TOC makes the page longer! What would help tremendously is some CSS to right-align the TOC and wrap the text around it. Is there an issue for this?

_includes/setup/android-setup-linux.md
```diff
@@ -0,0 +1,89 @@
+## Android setup
+
+_This step is optional, and can be performed at any time before
+using Flutter to develop apps for Android._
+
+### Install Android Studio
+
+To develop Flutter apps for Android, you need to install Android Studio. After installation,
+you also need to start Android Studio once, to
```
LarkAscending: It's in progress, and we're looking for why that's needed.

Any idea why that's needed?

_includes/setup/android-setup.md
LarkAscending: Updated. Left it in, since it's parallel to the statement in the iOS section (underlying UX principle=consistency).

_includes/setup/system-linux.md
LarkAscending: Good one!

_includes/setup/system-win.md
LarkAscending: Good catch!

_includes/setup/system-linux.md
LarkAscending: Done

_includes/setup/get-sdk-win.md
LarkAscending: <sigh> can't run flutter doctor on Windows because of Bit9. Whole machine would need to be reimaged. Will look at the source file for flutter tool output to cook up a windows example.

_includes/setup/next-steps-win-linux.md
LarkAscending: This was in response to PR feedback that we needed to clarify that iOS development requires a Mac. Also, in this PR, there's a comment that apparently many folks are running Windows on a Mac, so yeah - not as strange as it seems. Finally, (@mit-mit who feels strongly) the original wording isn't platform-agnostic. " iOS and Android" doesn't apply to Windows and Linux. 

Changed to "iOS or Android" and re-unified these for now.

_includes/setup/android-setup-linux.md
LarkAscending: Yeah I had the same impulse.

_includes/setup/android-setup.md
LarkAscending: Nope - enable KVM steps in emulator section are required.

_includes/setup/android-setup-linux.md
LarkAscending: Will note that... this particular section is preliminary, since it's high-impact to test, and most Flutter users are not on Linux. So it's for the small subset of users who are on Linux, don't have KVM enabled already, and prefer using an emulator vs. a device.

_includes/setup/get-sdk-win.md
LarkAscending: Good catch!

_includes/setup/system-linux.md
LarkAscending: These aren't in the sidenav. Does our sidenav support 3 levels in the hierarchy?

_includes/setup/get-sdk-win.md
LarkAscending: Done!

_includes/setup/path-win.md
LarkAscending: OK tested this, and read some stuff on the Internet. The way we have the syntax, it just updates the USER variable. Not SYSTEM. Some users will not be able to use setx to update SYSTEM PATH (I wasn't allowed). So, leaving this section as-is.

_includes/setup/get-sdk-win.md
LarkAscending: Done

_includes/setup/system-win.md
mit-mit: Never mind, looks like 32-bit is very common only in the pre-Win 10 world.

_includes/setup/path-win.md
mit-mit: Isn't user variable enough?

_includes/setup/android-setup-linux.md
```diff
@@ -0,0 +1,88 @@
...
+you want Flutter to use a different installation of the Android SDK, you must set the
+`ANDROID_HOME` environment variable to that installation directory.
+
+### Set up the Android emulator
+
+To prepare to run and test your Flutter app on the Android emulator, follow these steps:
+1. Enable Kernel-based Virtual Machine (KVM) support on your machine.
+1. Define a virtual device in Android Virtual Device Manager.
+
+#### Enable KVM
```
mit-mit: I think this is getting too detailed, and IMHO getting outside what we should document vs rely on documentation of our partners. I would just have a small line of text saying hardware accelleration is recommended and then link to Android docs: https://software.intel.com/en-us/android/articles/speeding-up-the-android-emulator-on-intel-architecture

You may also note in these docs:
1. There are steps for all three OS'es we support
2. The steps are different for Intel and AMD, so don't think the current docs we have will work on AMD (that would not have *Intel* VT-d

_includes/setup/path-win.md
goderbauer: I'd think a user variable is enough. (Isn't the setup we are recommending for Linux/Mac also just setting it for the current user?)

_includes/setup/path-win.md
```diff
@@ -0,0 +1,24 @@
...
+
+You can update your PATH variable for the current session only at the command line,
+as shown in [Clone the Flutter repo](./#clone-the-repo). You'll probably want to
```
goderbauer: For windows, we don't describe a way to only set it for the current session anywhere (`setx` is permanent, I think).

As it stands, I don't find the information in this section very useful. We already describe how to set a permanent user variable with `setx` elsewhere, so why do we need to describe how to set it via the UI here? Also, why are we concerned about system variables if a user variable is all we really need?

_includes/setup/system-win.md
goderbauer: Should we be explicit about the Windows version? We don't intent to support any version before Windows 7.

_includes/setup/system-win.md
```diff
@@ -0,0 +1,11 @@
+## System requirements
+
+To install and run Flutter, your development environment must meet these minimum requirements:
+
+* **Operating Systems**: Windows (64-bit)
+* **Disk Space**: 400 MB (does not include disk space for Android Studio).
+* **Tools**: Flutter depends on these command-line tools being available in your environment.
+  * Git for Windows (with the Use Git from the Windows Command Prompt option), Cmd, PowerShell, `where`
```
goderbauer: Maybe add a link to https://git-scm.com/download/win for git? It's the only thing listed here that doesn't ship with Windows by default.

LarkAscending: Good idea, done.

_includes/setup/system-win.md
LarkAscending: We should! Was hoping you'd respond w/this info...

_includes/setup/android-setup-linux.md
LarkAscending: Thanks for the link! That's a really good page.... FYI one of the things that's going on in this PR is I'm testing everything that's possible for me to test. And this procedure 1. does not work on Ubuntu workstation b/c chip doesn't support it, 2. just works on corp Mac, and 3. can't test on Windows. What I notice is, if the steps were already vetted, these types of comments would be minimal. Is someone on the team responsible for testing the setup procedure and its instructions?

If the broader question is, what's the thinking behind having a very long procedure here, the answer is it is a best practice not to bounce users around to various websites during setup. They get lost and/or frustrated. The remedy for long documentation is usually product-related changes upstream (though we're not in control of the KVM steps in this particular example).

For very specific feedback where you feel what's there is actually wrong, it's totally fine to edit the PR directly or submit a separate PR. I'll take a stab at this, but if you have further feedback on it, let's go the direct route, OK?

_includes/setup/path-win.md
LarkAscending: OK, will remove for now. Are we monitoring feedback channels for Windows users, so if there's an issue w/some common configuration, we'll know?

_includes/setup/android-setup-linux.md
LarkAscending: BTW, the Linux version was different because on Linux, VM acceleration is *required*, not recommended. For example on this workstation the Android emulator won't run because AVD manager runs a check for KVM, and that check fails.

_includes/setup/path-win.md
goderbauer: I am monitoring our github issues and gitter channel for Windows issues.

_includes/setup/android-setup-linux.md
mit-mit: Yes, we expect the Android setup instructions to change frequently. We have Q2 work planned to make this much simpler.

_includes/setup/system-linux.md
mit-mit: This issue is still not resolved.

_includes/setup/android-setup-linux.md
mit-mit: I don't understand your response. 

As I mentioned, the steps here are extremely detailed and also incomplete (they only cover Intel; AMD has significant marketshare). I concretely also do not understand the actual steps; where did they come from?

While I agree we should generally not "bounce users around to various websites", this particular case is quite a bit of a corner case. I think it's safe to assume that Linux users are capable of configuring their systems given a link to official Android docs.

With that, I re-iterate my original point: I see move overall value in having a single `android-setup.md` which is not fragmented per OS, and which simply links to the Android docs wrt. configuration of hardware acceleration.

LarkAscending: Re where did they come from, they are internal instructions... feel free to IM me to get the URL if needed. Any other specific questions?

As far as fragmentation/forking, in documentation we put users first vs. file organization on the back end. The justification that works here is, the Android team is maintaining the documentation. If you read the Android instructions carefully, they apply to all supported platforms, not just Linux. That's good news for your concern about forking or fragmentation (with little to no downside), but maybe not such good news for assuming users won't be affected by bouncing them to DAC. They'll all be bounced there.

On this end, just one open question: is someone other than me testing the setup steps on various platforms?

_includes/setup/system-linux.md
mit-mit: Fixed in https://github.com/flutter/website/pull/446/commits/2aa22ecdda004eabe4a2bc42e0901e35b48bb73c

mit-mit:
LGTM

Everything looks great now! Thanks for the long, hard push on getting this done!

PR #447, opened 2017-03-17
New platform plugin documentation

mit-mit:
This is an initial and incomplete draft, so not ready for rigid review yet. But I would much appreciate a structural and conceptual review!

Fixes https://github.com/flutter/flutter/issues/8910
Fixes https://github.com/flutter/flutter/issues/1432
Fixes https://github.com/flutter/flutter/issues/6397


mit-mit:
Page is staged here:
https://flutter-website-staging.firebaseapp.com/platform/platform-plugins/

mit-mit:
CC @LarkAscending @sethladd @abarth

sethladd:
I assume that this replaces https://flutter.io/platform-services/ ? If so, please add a redirect via https://github.com/flutter/website/blob/master/firebase.json

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
+---
+layout: page
+title: Platform-specific APIs and Platform Plugins
+permalink: /platform-plugins/
+---
+
+This guide describes how to connect your Flutter app to platform-specific
+services available on iOS and Android devices. This includes both device APIs
+(like `Clipboard` and `BatteryState`) and third-party platform SDKs (like
```
sethladd: nit: we don't need to use code font for these names, especially Google Analytics.

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+`Google Analytics`).
+
+* TOC
+{:toc}
+
+## Options for calling platform-specific APIs
+
+Mobile devices offer many options to extend your app through platform-specific
+APIs. These are critical to the completeness of most mobile apps, and Flutter
+imposes no limits on your ability to them.
```
sethladd: beware of words like "no limits" :)

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+
+* **Use existing Platform Plugins:** For frequently used APIs and SDKs, use
+ existing Platform Plugins, with no need to write any platform-specific code.
+ [See details below](#use).
+
+* **Create a Platform Plugin:** Write your own platform-specific code using
+ Flutter's Platform Channels. This can be used for private APIs, or for cases
+ where a suitable Platform Plugin is not yet available. [See details below](#create).
+
+* **Publish a Platform Plugin:** Publish your own custom platform-specific as a
```
sethladd: Structural: Is this "use" ? This feels like part of a workflow under "creating".

There are two ways to access platform APIs:

* Use an existing plugin
* Write your own

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+ plugin. This enables others to use it by simply declaring a dependency on the
+ plugin. [See details below](#publish).
+
+## Use existing Platform Plugins {#use}
+
+### Core Flutter Platform Plugin
```
sethladd: Now I'm confused about what a "plugin" is?

It wouldn't hurt to define "plugin" at the top. I would presume it means "an optional piece of code that I need to opt-in to use". But we have many ways to define it, not a problem.

What you list below are core APIs from Flutter. They are indeed "platform integrations" but are they plugins?

(just working on being specific with semantics)

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+The Flutter framework includes a core Platform Plugin for foundational
+platform-specific services, such as:
+
+* [Clipboard](https://docs.flutter.io/flutter/services/Clipboard-class.html):
+Utility methods for interacting with the system's clipboard.
+
+* [SystemSound](https://docs.flutter.io/flutter/services/SystemSound-class.html):
+Provides access to the library of short system specific sounds for common tasks.
+
+* [pathProvider](https://docs.flutter.io/flutter/services/PathProvider-class.html):
```
sethladd: For example, how is this different than dart:io's File and Directory? Are those plugins?

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+   }
+ }
+ ...
+}
+```
+
+### Platform Plugins from the Flutter ecosystem
+
+{% include note.html content="Sharing of Platform Plugins is currently not
+supported, but Flutter is expected to support this in the Future." %}
```
sethladd: lower-case F

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+
+## Create a Platform Plugin {#create}
+
+Flutter uses a flexible system that allows you to call *any* platform-specific
```
sethladd: beware of words like "any". Not sure we can commit to this. (e.g. Maps)

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+API whether available in Java-code on Android, or in ObjectiveC- or Swift-code
```
sethladd: style: IIRC we don't need a hyphen between Java and code

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+on iOS.
+
+Flutter's platform-specific API support does not rely on fragile code
```
sethladd: Can we write this without the word "fragile" ? Some people might love code generation. Let's not make a judgement here, and just state that "does not rely on code generation".

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+generation, but rather on a flexible message passing style:
+
+* The Flutter portion of your app sends messages to it's *host*, the iOS or
```
sethladd: it would help to define what a message here: "(text and binary)"

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+
+Messages are passed between the client (UI) and host (platform) using Platform
+channels as illustrated in this diagram:
+
+![Platform Channels architecture](/images/PlatformChannels.png)
+
+Message and responses are passed asynchronously, to ensure the user interface
+remains responsive.
+
+On the client side, the class
```
sethladd: I think you can drop the words "class" from this paragraph.

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+[`PlatformMessageChannel`](https://docs.flutter.io/flutter/services/PlatformMessageChannel-class.html)
+enables sending general purpose message, and the class
```
sethladd: define "general purpose message" ?

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+[`PlatformMethodChannel`](https://docs.flutter.io/flutter/services/PlatformMethodChannel-class.html)
+enables spending specific messages that correspond to method calls. On the
```
sethladd: not sure what a "specific messages" are?

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+platform side, the class `FlutterMessageChannel` enables receiving messages
+messages from the client, and responding with a result, and the class
+`FlutterMethodChannel`. These classes allow you to developer a platform plugin
```
sethladd: typo: "developer"

And, choose if we capitalize P for platform. Above you say "Platform channels" and here you say "platform plugin"

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+with very little 'boilerplate' code.
+
+### Example: Calling platform-specific iOS and Android code using Platform Channels
+
+The following example demonstrates how to call a new platform-specific API,
+through a concrete example that retrieves and displays the current battery level
+by calling into the Android `BatteryManager` API, and the iOS
+`device.batteryLevel` API, via a single platform message, `getBatteryLevel`.
+
+The standard app template created when you create a new Flutter app can be used
```
sethladd: missing period?

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+
+*Note*: The full source-code for this example is available in
+[`/samples/platform-services/`](https://github.com/flutter/flutter/tree/master/examples/platform_services).
+
+**Step 1: Create the Flutter platform client**
+
+The Flutter platform client code is written inside the app's `State` class.
```
sethladd: This read as if we always do this. Do we want every user of PlatformMethodChannel to be a static const inside a State class?

Also, the app has many State classes.

Maybe: "it's common practice to create a State class that manages a PlatformMethodChannel" or similar?

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+Start by constructing the channel. Since we need just a single platform message,
+`getBatteryLevel`, representing a platform method call, we will use a
+`PlatformMethodChannel`:
+
+```dart
+class _PlatformServicesState extends State<PlatformServices> {
+  static const PlatformMethodChannel platform = const PlatformMethodChannel('battery');
+  ...
```
sethladd: // ...

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+}
+```
+
+The client and host sides of the channel are connected through by the passed
+name argument; here `battery`. Note that channel names must be unique; we
```
sethladd: unique across what?

sethladd: The lead sentence in this paragraph was hard to parse. Is there another way to reword this?

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+recommend using a naming pattern using *TODO*.
+
+Next, we invoke a method on the method channel, specifying the concrete method
+to call via a String `method` argument, and use the returned result to update
+our user interface:
+
+```dart
+String _batteryLevel = 'Unknown battery level.';
```
sethladd: From our previous flutter first look, we learned that code snippets like this, without the context of where to put the snippet, are hard to use. Can you be more clear in the code where exactly to put this?

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+```
+
+The call may fail -- for example if the platform does not support the platform
+API (such as when running in a simulator), so next we add exception handling:
+
+```dart
+  try {
+    final int result = await platform.invokeMethod('getBatteryLevel');
+    batteryLevel = 'Battery level at $result % .';
+  } on PlatformException catch (e) {
```
sethladd: yay!  thanks for adding the error condition!!

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+  });
+}
+```
+
+Next, we add the actual Android Java code that uses the Android battery APIs to
+retrieve the battery level. This code is exactly the same you would have written
+in a native Android app:
+
+```java
+private int getBatteryLevel() {
```
sethladd: As with above, providing a marker for where this code goes will help our users that like to "copy and paste" their way to success. :)

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
...
+```
+
+Note how we use the `response` argument to pass back both a response for both
+the success and error cases, and how we throw an exception if a method call is
+received for which we have no handler.
+
+### Separate Platform Plugin code from UI code
+
+If you expect to use your platform-specific in multiple Flutter apps, it can be
+useful to separate the code into a Platform Plugin located in a directory
```
sethladd: if we clearly define a "Platform Plugin" above, this sentence becomes easier to understand. I'm still not 100% on what a plugin is.

platform-plugins.md
```diff
@@ -0,0 +1,277 @@
+---
+layout: page
+title: Platform-specific APIs and Platform Plugins
```
sethladd: Would we add an example of binary data to this doc? At a minimum, being more clear that it's possible and when I'd use it.

Do we know enough to document patterns of use? Are there common themes/patterns for interop for various different APIs? Some APIs are stateful, some are stateless (like battery). How does that impact my interop design?

platform-plugins.md
mit-mit: I don't expect devs to copy paste this; the full code is available in /examples/platform-services/, but perhaps that wasn't clear?

sethladd: One of the pieces of feedback we got from the previous event was "I lost my place as I was adding the code from the doc into my project". I think a simple comment like "// inside of the foo() method" would be sufficient for snippets like this. A little "sign post".

mit-mit:
> I assume that https://flutter-website-staging.firebaseapp.com/platform/platform-plugins/ replaces https://flutter.io/platform-services/ ? If so, please add a redirect via https://github.com/flutter/website/blob/master/firebase.json

Done

platform-plugins.md
mit-mit: We don't have great answers to these questions yet; we need to write some more plugins first.

platform-plugins.md
mit-mit: Ah, yes misleading heading and grouping, fixed!

platform-plugins.md
mit-mit: So, yes this is confusing mostly as we have stuff in services that really should just be in an optional plugin.

platform-plugins.md
mit-mit: It might be clearer if we change services.dart to operating-system-services.dart or something like that. That is how I think of it. File, Dir, etc. are just core OS services. Battery, Location, etc. are not.

platform-plugins.md
mit-mit: Sure, I'll leave out my personal judgement :-)

platform-plugins.md
mit-mit: This was added earlier.

platform-plugins.md
mit-mit: Got it, done

platform-plugins.md
mit-mit: And with that, I would remove the current `### Core Flutter platform plugin` section, and the whole `## Use existing platform plugins` would change to how you search for a plugin, and add it to your pubspec.yaml file.

mit-mit:
This is now fairly complete. Only major missing section is the iOS one.

@mravn-google @szakarias please review

faq.md
```diff
@@ -141,44 +141,30 @@ we encourage you to email our mailing list:
...
 ## Can I interop with my mobile platform's default programming language?
 
-Yes, though this feature is a work in progress.
-
-We are building a _message pipe_ between Flutter and the host
-operating system. You can use this message pipe to send and receive
-string data (typically JSON data) between your Flutter code
-and Java (on Android) and ObjectiveC (on iOS).
+Yes, Flutter supports calling into the platform, incl. integrating with
+java-code on Android, and ObjectiveC and Swift code on iOS. This is enabled via
```
mravn-google: "Java code" instead of "java-code"

faq.md
```diff
@@ -141,44 +141,30 @@ we encourage you to email our mailing list:
 
 ## Can I access native services and APIs like sensors and local storage?
```
mravn-google: Probably best to avoid the word 'native'. Use 'platform' or 'host' instead. Several places.

faq.md
```diff
@@ -141,44 +141,30 @@ we encourage you to email our mailing list:
...
-
 ## Can I interop with my mobile platform's default programming language?
 
-Yes, though this feature is a work in progress.
-
-We are building a _message pipe_ between Flutter and the host
-operating system. You can use this message pipe to send and receive
-string data (typically JSON data) between your Flutter code
-and Java (on Android) and ObjectiveC (on iOS).
+Yes, Flutter supports calling into the platform, incl. integrating with
```
mravn-google: [nit] Expand incl.

faq.md
```diff
@@ -141,44 +141,30 @@ we encourage you to email our mailing list:
...
-on Android. (An equivalent example project for iOS is in development.)
-
-This is an evolving part of the system, and we're very interested
-to learn how teams want to interop between
-Flutter and the OS and third-party SDKs. We encourage you to email our
-mailing list to let us know what you're building, and if you have
-questions:
-[flutter-dev@googlegroups.com](mailto:flutter-dev@googlegroups.com).
+Here is an [example project](https://github.com/flutter/flutter/tree/master/examples/platform_services)
+that shows how to use the PlatformChannel to access battery state information on
```
mravn-google: There is no `PlatformChannel`, at least not as a Dart type. The `platform_services` example uses a `PlatformMethodChannel`, whereas your link above is to `PlatformMessageChannel`. If we want to use PlatformChannel as a concept in the documentation, we should write somewhere that it covers the `PlatformMessageChannel` and `PlatformMethodChannel` implementation classes (and any later friends). Similarly for the FlutterChannel concept.

platform-plugins.md
```diff
@@ -0,0 +1,305 @@
...
+Google Analytics).
+
+* TOC
+{:toc}
+
+## Options for integrating with platform-specific APIs
+
+Mobile devices offer many options to extend your app through platform-specific
+APIs. These are critical to the completeness of most mobile apps. Flutter offers
+the following ways to integrate with these via 'plugins', packages that contains
```
mravn-google: contain without s

platform-plugins.md
```diff
@@ -0,0 +1,305 @@
...
+implementation for Android, for iOS, or for both.
+
+There are two ways you can use plugins:
+
+* **Use existing platform plugins:** For frequently used APIs and SDKs, use
+ existing platform plugins, with no need to write any platform-specific code.
+ [See details below](#use).
+
+* **Create a platform plugin:** Write your own platform-specific code using
+ Flutter's Platform Channels. This can be used for private APIs, or for cases
```
mravn-google: Here Platform Channel is written in two words, which is probably better, since it is not a Dart type. But we should be consistent.

platform-plugins.md
```diff
@@ -0,0 +1,305 @@
...
+ where a suitable platform plugin is not yet available. [See details below](#create).
+
+And, if desired, you can share plugins with others:
+
+* **Publish a platform plugin:** Publish your own custom platform-specific as a
```
mravn-google: platform-specific what? Noun missing.

platform-plugins.md
```diff
@@ -0,0 +1,305 @@
...
+```
+
+### Platform plugins from the Flutter ecosystem
+
+{% include note.html content="Sharing of platform plugins is currently not
+supported, but Flutter is expected to support this in the future." %}
+
+## Create a platform plugin {#create}
+
+Flutter uses a flexible system that allows you to call platform-specific API
```
mravn-google: APIs?

platform-plugins.md
```diff
@@ -0,0 +1,305 @@
...
+whether available in Java code on Android, or in ObjectiveC or Swift code on
+iOS.
+
+Flutter's platform-specific API support does not rely on code generation, but
+rather on a flexible message passing style:
+
+* The Flutter portion of your app sends messages to it's *host*, the iOS or
+ Android portion of your app, over a platform channel.
```
mravn-google: Here platform channel is spelled in lower-case. Even better. We should be consistent.

platform-plugins.md
```diff
@@ -0,0 +1,305 @@
...
+
+* The *host* listens on the platform channel, and receives the message. It then
+ calls into any number of platform-specific APIs -- using their native
+ programming language -- and sends back a response to the *client*, the Flutter
+ portion.
```
mravn-google: We should probably state somewhere that the communication mechanism is symmetric: the Dart side can also act as host. That goes for both message passing and method calls. Event streams are one-way for now though, with the Dart side acting as receiver.

platform-plugins.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Architectural overview: Platform Channels
+
+Messages are passed between the client (UI) and host (platform) using Platform
+channels as illustrated in this diagram:
+
+![Platform Channels architecture](/images/PlatformChannels.png)
+
+Message and responses are passed asynchronously, to ensure the user interface
```
mravn-google: Messages

platform-plugins.md
```diff
@@ -0,0 +1,305 @@
...
+remains responsive.
+
+On the client side,
+[`PlatformMessageChannel`](https://docs.flutter.io/flutter/services/PlatformMessageChannel-class.html)
+enables sending general purpose messages, and
+[`PlatformMethodChannel`](https://docs.flutter.io/flutter/services/PlatformMethodChannel-class.html)
+enables spending messages that correspond to method calls. On the platform side,
```
mravn-google: sending

platform-plugins.md
```diff
@@ -0,0 +1,305 @@
...
+the class `FlutterMessageChannel` enables receiving messages from the client,
+and responding with a result, and the class `FlutterMethodChannel`. These
```
mravn-google: Unfinished sentence. The class `FlutterMethodChannel` does what?

platform-plugins.md
```diff
@@ -0,0 +1,305 @@
...
+             response.error("UNAVAILABLE", "Battery level not available.", null);
+           }
+
+         } else {
+           throw new IllegalArgumentException("Unknown method " + call.method);
+         }
+       }
+```
+
+Note how we use the `response` argument to pass back both a response for both
```
mravn-google: "both" occurs twice

platform-plugins.md
```diff
@@ -0,0 +1,305 @@
...
+the success and error cases, and how we throw an exception if a method call is
+received for which we have no handler.
```
mravn-google: Maybe better to code that using `response.error` as well. I think that's how one would do it on the iOS side because of the complexities of exception and memory handling.

platform-plugins.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Separate platform plugin code from UI code
+
+If you expect to use your platform-specific in multiple Flutter apps, it can be
```
mravn-google: platform-specific what?

platform-plugins.md
mit-mit: Agree, done everywhere

platform-plugins.md
mit-mit: Added a note

mit-mit:
This is now a full draft! @mravn-google @szakarias please review! 

Content is staged in these two pages:
https://flutter-website-staging.firebaseapp.com/platform/platform-plugins/
https://flutter-website-staging.firebaseapp.com/platform/platform-channels/

platform-plugins.md
mit-mit: Updated plan: We will review content in [`/lib/src/services/`](https://github.com/flutter/flutter/tree/master/packages/flutter/lib/src/services), and see if we can "harvest" a few plugins from that. From our initial analysis it looks like PathProvider and URLLauncher are good candidates. We will then take those, turn them into plugins, and publish them! 

And then I will change the current clipboard example and the weird "The Flutter framework includes a core platform plugin" phrase to simply talk about how to use a published plugin (I will likely change the example from clipboard to URLLauncher).

mit-mit:
@jakobr-google please review

platform-plugins.md
```diff
@@ -0,0 +1,108 @@
...
+
+* **Write custom platform code:** Write your own platform-specific code using
+ Flutter's platform channels. This can be used for private APIs, or for cases
+ where a suitable platform plugin is not yet available. See the [platform channels](/platform-channels/)
+ page for details.
+
+And, if desired, you can create and share plugins:
+
+* **Create a platform plugin** If you wish to use your platform-specific code in
+ multiple Flutter apps, or share it with others (see 'publish'), than you can
```
szakarias: Remove (see 'publish') since it confuses.

szakarias: than -> then

platform-plugins.md
```diff
@@ -0,0 +1,108 @@
...
+The Flutter framework includes a core platform plugin for foundational
+platform-specific services, such as:
+
+* [Clipboard](https://docs.flutter.io/flutter/services/Clipboard-class.html):
+Utility methods for interacting with the system's clipboard.
+
+* [SystemSound](https://docs.flutter.io/flutter/services/SystemSound-class.html):
+Provides access to the library of short system specific sounds for common tasks.
+
+* [pathProvider](https://docs.flutter.io/flutter/services/PathProvider-class.html):
```
szakarias: pathProvider -> PathProvider

platform-plugins.md
```diff
@@ -0,0 +1,108 @@
...
+ getClipboardContents() async {
+   var clipboardData = await Clipboard.getData(Clipboard.kTextPlain);
+   if (clipboardData != null) {
+     setState(() {
+       clipboardContents = clipboardData.text;
+     });
+   }
+ }
+
+ // clipboardContents can now be accessed the build() method.
```
szakarias: accessed the -> accessed in the

platform-channels.md
```diff
@@ -0,0 +1,249 @@
...
+class _PlatformServicesState extends State<PlatformServices> {
+  // Construct platform channel inside State class.
+  static const platform = const PlatformMethodChannel('battery');
+
+  // TODO: Get battery level.
+}
+```
+
+Next, we invoke a method on the method channel, specifying the concrete method
+to call via a String `method` argument. We then use the returned result to update
```
mravn-google: Please reconsider the word `argument`, as it may be confusing: there is also the question of passing arguments to the method called on the channel. That is not shown in your example, but should probably be mentioned here.

platform-channels.md
```diff
@@ -0,0 +1,249 @@
...
+
+Messages are passed between the client (UI) and host (platform) using platform
+channels as illustrated in this diagram:
+
+![Platform channels architecture](/images/PlatformChannels.png)
+
+Messages and responses are passed asynchronously, to ensure the user interface
+remains responsive.
+
+On the client side, `PlatformMessageChannel` ([API][PlatformMessageChannel])
```
szakarias: nit: below you say "the class Flutter*Channel", here you just say Platform*Channel.

platform-channels.md
mit-mit: Changed to 'via a String identifier'

platform-channels.md
```diff
@@ -0,0 +1,249 @@
...
+
+For more specialized use cases, you may also use the
+[`BinaryCodec`](https://docs.flutter.io/flutter/services/BinaryCodec-class.html),
+[`StringCodec`](https://docs.flutter.io/flutter/services/StringCodec-class.html),
+or
+[`JSONMessageCodec`](https://docs.flutter.io/flutter/services/JSONMessageCodec-class.html).
+
+### Example: Calling platform-specific iOS and Android code using platform channels
+
+The following example demonstrates how to call a new platform-specific API,
```
szakarias: Remove "new" ?

szakarias: Remove the first "example", that is "The following demonstrates ...."

szakarias: nit: In general this sentence is a bit too long for my taste.

platform-channels.md
```diff
@@ -0,0 +1,249 @@
...
+
+*Note*: The full source-code for this example is available in
+[`/samples/platform-services/`](https://github.com/flutter/flutter/tree/master/examples/platform_services).
+
+**Step 1: Create the Flutter platform client**
+
+We start by adding support for updating our app's `State` class with the current
+battery state.
+
+First, we construct the channel. Since we need just a single platform message,
```
szakarias: nit: Reverse the sentence, that is start by saying "We use `PlatformMethodChannel` ...."

platform-channels.md
```diff
@@ -0,0 +1,249 @@
...
+
+The client and host sides of a channel are connected through a channel name
+passed in the channel constructor. In this example we are using `battery`. Note
+that all channel names used in a single app must be unique; we recommend using a
+naming pattern *TODO*.
+
+<!-- skip -->
+```dart
+class _PlatformServicesState extends State<PlatformServices> {
+  // Construct platform channel inside State class.
```
szakarias: Maybe remove this comment?

platform-channels.md
```diff
@@ -0,0 +1,249 @@
...
+  static const platform = const PlatformMethodChannel('battery');
+
+  // TODO: Get battery level.
+}
+```
+
+Next, we invoke a method on the method channel, specifying the concrete method
```
szakarias: Maybe just : "specifying the name of the concrete method to call. We then ... " ?

platform-channels.md
```diff
@@ -0,0 +1,249 @@
...
+to call via a String `method` argument. We then use the returned result to update
+our user interface:
+
+<!-- skip -->
+```dart
+  // Get battery level.
```
szakarias: Maybe remove this comment too?

platform-channels.md
```diff
@@ -0,0 +1,249 @@
...
+        @Override
+        public void onMethodCall(MethodCall call, Response response) {
+          // TODO
+        }
+  });
+}
+```
+
+Next, we add the actual Android Java code that uses the Android battery APIs to
+retrieve the battery level. This code is exactly the same you would have written
```
szakarias: same as ?

platform-channels.md
```diff
@@ -0,0 +1,249 @@
...
+in a native Android app.
+
+Add the following as a new method in the activity class, below the `onCreate`
+method:
```
szakarias: Technically, it does not need to be below `onCreate`.

platform-channels.md
mit-mit: Fixed

platform-channels.md
mit-mit: Fixed

mit-mit:
@mravn-google any more feedback, or can I get a LGTM?

xster:
This is super awesome!

PR #452, opened 2017-03-21
Fixes to PR Deployer

FaisalAbid:
This fixes

1. Properly OAuth Generated GitHub Token (added to Travis already)
2. Bug where two branches would be deployed on the same instance.

### DIFF
--- a/tool/prdeployer.dart
+++ b/tool/prdeployer.dart
@@ -32,9 +32,11 @@ main(List<String> arguments) async {
       // deploy again
       projectToDeploy = value["name"];
     } else {
-      bool oldBranch = await isBranchOld(value["branch"]);
-      if (oldBranch) {
-        await fbClient.patch("${firebaseRoot}/$key.json", {"branch": null});
+      if (value["branch"] != null) {
+        bool oldBranch = await isBranchOld(value["branch"]);
+        if (oldBranch) {
+          await fbClient.patch("${firebaseRoot}/$key.json", {"branch": null});
+        }
       }
     }
   });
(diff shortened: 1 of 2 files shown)


tool/prdeployer.dart
```diff
@@ -41,14 +41,14 @@ main(List<String> arguments) async {
 
   if (projectToDeploy == null) {
     await projectsData.forEach((String key, Map<String, String> value) async {
-      if (projectToDeploy != null) {
-        // project already set
-        // todo: remove this foreach with a better iterator pattern
+      if(projectToDeploy != null){
```
sethladd: space after if

tool/prdeployer.dart
```diff
@@ -41,14 +41,14 @@ main(List<String> arguments) async {
...
         return;
       }
-      projectToDeploy = value["name"];
-      await fbClient.patch("${firebaseRoot}/$key.json", {"branch": branchName});
-      await postLinkToGithub(projectToDeploy, currentPullRequest);
+      if(value["branch"] == null){
```
sethladd: space after if

Or, just run the dartfmt tool :)

FaisalAbid: Ah i wonder why webstorm didnt run the tool when I saved. Good catch! Fixing

sethladd:
Yeah, WebStorm doesn't run dartfmt automatically. Maybe there's a way to trigger that?

FaisalAbid:
Oh I know that, I have it set to cmd shift -f and I do it right before save. Its become a habit now to just hit it and save Lol, I guess it didnt work that time!

sethladd:
lgtm

sethladd:
Please merge in master, it has a fix so this PR will pass the linkchecker. Thanks!

FaisalAbid:
Sorry! missed your comment in my Github noise!

sethladd:
OK, here we go! lgtm :)

PR #457, opened 2017-03-21
Signal that it's OK to create this file if it doesn't exist

sethladd:
Ideally, we would automate these steps somehow. Until we have a more automated process, we can reassure our new users that it's OK to create this file if it doesn't exist.

This is a temp fix, an installer script/app would make for a better long-term fix.


sethladd:
Closes https://github.com/flutter/flutter/issues/8770

sethladd:
cc @mit-mit

mit-mit:
LGTM

sethladd:
One build succeeded, the other timed out linkchecking.

PR #458, opened 2017-03-21
Remove broken links

abarth:
We're updating this page, but in the meantime we need to remove the broken
links to unblock the deploy script.


sethladd:
lgtm

sethladd:
unnamed

PR #460, opened 2017-03-21
Fix linkchecker hang

mit-mit:
Move link checker config to a Rakefile to make it more maintainable.

Extend that config to skip httpbin.com, as that site appears to make the checker hang.


sethladd:
lgtm

(curious why we need to run this through rake?)

mit-mit:
> lgtm

Let's see if travis passes this time first :-)

> (curious why we need to run this through rake?)

It was getting pretty unwieldy to pass all the config parameters through the command line, and with rake they are in a nicely formatted Rakefile (see the new file in the PR)

mit-mit:
Woot, it's passing!

sethladd:
Nice!

PR #462, opened 2017-03-22
Redirect with and without a trailing slash

sethladd:


sethladd:
Fixes https://github.com/flutter/flutter/issues/8958

sethladd:
cc @mit-mit

PR #467, opened 2017-03-23
Remove "add Java from Android Studio to your path" step

goderbauer:
This is no longer necessary, see https://github.com/flutter/flutter/pull/8965.


mit-mit:
LGTM

PR #470, opened 2017-03-23
Style tweak for defining a new term

sethladd:
We use italics when we introduce a new term that needs a definition.

cc @mit-mit


abarth:
LGTM

mit-mit:
Thanks, good catch!

PR #475, opened 2017-03-27
Add pod setup step to iOS install instructions.

jakobr-google:
CocoaPods needs to download the master spec repo, which can take quite a
while. Adding a `pod setup` step to the installation instructions, so
users take the hit during install, instead of on the first build.

Fixes flutter/flutter#8983.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -38,7 +38,7 @@ to test and iterate using a simulator or physical device.
  </tr>
 </table>
 
-Using an IDE is optional. However, we recommend using an IntelliJ IDE with Flutter for code completion, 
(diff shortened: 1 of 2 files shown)


setup.md
```diff
@@ -177,6 +177,7 @@ iOS devices.
    $ brew install --HEAD libimobiledevice
    $ brew install ideviceinstaller ios-deploy
    $ brew install cocoapods
+   $ pod setup
```
mit-mit: This line (and 179) should be moved up to a separate Pod section (the present section is only for those that deploy to physical devices)

jakobr-google: Done. PTAL.

PR #477, opened 2017-03-27
Fix links broken in travis

mit-mit:


mit-mit:
TBR @HansMuller for the `AboutDrawerItem` part

Sfshaza:
Thanks, @mit-mit!

PR #478, opened 2017-03-28
Move platform docs to guides (they were never howtos)

mit-mit:


_data/sidebars/home_sidebar.yml
```diff
@@ -74,6 +74,14 @@ entries:
       url: /widgets/
       output: web
 
+    - title: Plugins & platform-specific code
```
sethladd: Did we come up with something to replace "platform-specific"? That's too easy to misinterpret as "cupertino vs material design".

Strawman: "Plugins for SDKs and APIs"

_data/sidebars/home_sidebar.yml
```diff
@@ -74,6 +74,14 @@ entries:
...
+      url: /platform-plugins/
+      output: web
+
+    - title: Custom platform-specific code with platform channels
```
sethladd: What's our convention? "PlatformChannel" or "platform channel" ?

_data/sidebars/home_sidebar.yml
mit-mit: No we didn't. "Platform-specific code" is what we started using when we deemed "Native code" too confusing. I'm worried about doing another "indirection".

_data/sidebars/home_sidebar.yml
mit-mit: Platform channel is the overarching concept, so we use that generally. There is actually no type called `PlatformChannel` so we don't use that (there are several specific ones, e.g. `PlatformMethodChannel`)

sethladd:
lgtm

PR #479, opened 2017-03-28
Minor fixes to font style, indentation on iOS releasing page

cbracken:


cbracken:
A couple minor fixes to get the style consistent with the rest of the document.

ios-release-build.md
```diff
@@ -81,29 +81,29 @@ Navigate to your target's settings in Xcode:
 
 Next, you'll verify the most important settings:
 
-In the `Identity` section:
+In the **Identity** section:
```
LarkAscending: The guideline here is to use bold font for things users have to click or select. It's visually noisy, and if it's used too much it can be distracting. So I intentionally left these un-bolded. They're not inconsistent....

PR #481, opened 2017-03-28
remove wacky JS fixing sidenav scrolling

sethladd:
Now the nav stays visible, even on a laptop.

I tested on desktop and laptop, wide and narrow screens.

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -1,6 +1,6 @@
 {% include custom/sidebarconfigs.html %}
 
-<ul id="mysidebar" class="nav">
+<ul id="mysidebar" class="nav affix">
     <li class="sidebarTitle">{{sidebar[0].product}} {{sidebar[0].version}}</li>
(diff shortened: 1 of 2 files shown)


sethladd:
@mit-mit ptal

mit-mit:
I can't tell the difference, but I'm also not entirely sure what issue I should be looking for. Can you elaborate?

I thought it was this flickering issue when clicking items, but that doesn't seem to be fixed:
flicker

sethladd:
If you're on a laptop, a macbook 13 inch, for example, the side nav would have scrolled out of view. It would not have stayed in place.

With this patch, the left nav stays in place even on a laptop (unless you're on mobile, it will scroll then).

sethladd:
re: flickering, we're going to fix that by simply removing all the javascript from the left nav :) But that's a separate patch

sethladd:
TBR= @mit-mit

PR #485, opened 2017-03-31
Homepage graphics

sethladd:
Staged at https://sethladd-flutter-io-staging.firebaseapp.com/

Addresses https://github.com/flutter/flutter/issues/9066


sethladd:
@bkobash @LarkAscending what do you guys think? Staged the SVGs.

LarkAscending:
With the white bezels around the devices, the home page is squinting-bright on my monitor.

At first I thought adding the devices around the screenshots would be the natural next step and was skeptical of the UX feedback. Now looking at them, there's really hardly any visual difference. Is it actually helpful? 
@lukechurch should look at this...

LGTM. Iteration is good.

LarkAscending:
Forgot the most important part - Thanks Brian - the images look great!

sethladd:
I think the bezels really drive home the point: these are _mobile_ screenshots :)

sethladd:
Hm, on Windows, the fonts in the SVG don't look like Roboto:

capture

InMatrix:
It's not Roboto on iOS either. It looks similar to the screenshot on
Windows.

Regarding the bezels, have we considered a darker color? White bezels are
really close to the background.

bkobash:
Ah. I will have to dig into embedding fonts within the SVG file. Support seems a bit inconsistent (even with today's browsers), so we may need to have a few options available. I can also darken the bezel a bit.

bkobash:
Here's an [updated ZIP](https://github.com/flutter/website/files/886452/flutter-graphics.zip) that hopefully addresses the font issue. 

Essentially, I added a reference to the Google Fonts API within the SVG's defs. I also cleaned up the textblocks so it would be a little easier to edit later.

However, in order for it to work within a web browser, it looks like you'll have to use the OBJECT tag, not the IMG tag. (The OBJECT tag also gets the image references to work in hero-shrine.svg.)  I included a sample HTML file in the ZIP to check against Chrome and Firefox.

LarkAscending:
Trends in mobile device chassis tend to last ~ a year, maybe 18 months. The existing iPhone images in Getting Started already look dated. Do we have a plan in mind for revisiting and refreshing the bezel images?

sethladd:
Thanks @bkobash ! Here's Safari, Chrome, and Firefox (on a mac), img tag on top, object tag on bottom:

screen shot 2017-03-31 at 1 31 35 pm

Notice how the browser is scaling the img tag version to fit the screen (neat!)

I see a font difference, though. But the top version LGTM. I'll check on Windows next.

@LarkAscending good point, phones change designs. One strategy could be to use a stylized frame, as we're doing here, instead of a very-accurate frame like we're using on Getting Started.

bkobash:
In order for Roboto to get rendered properly, we might have to use the object tag. If I use the img tag and don't have Roboto installed on my local machine, the type elements default back to Times. To achieve similar responsiveness as the img version, we may need to write a few more CSS rules. 

See example with IMG on top, OBJECT on bottom

eseidelGoogle:
Correct.  SVGs used in  tags don't get to make network requests or run javascript, etc.

<object> tag should be a fine way to go?  If you have any troubles the flutter team has lots of (dusty) expertise around the quirks of SVG and html if needed.

sethladd:
Cool, thanks. Using object tag on bottom, across these screen shots. Note: the width is responsive here, it's shrinking according to tab width.

screen shot 2017-04-03 at 10 01 17 am

sethladd:
And here's the screenshots:

screen shot 2017-04-03 at 10 09 38 am

sethladd:
@bkobash @LarkAscending ptal: https://sethladd-flutter-io-staging.firebaseapp.com

sethladd:
@bkobash sent over this diagram of the layer cake:

[diagram-layercake.svg.zip](https://github.com/flutter/website/files/894236/diagram-layercake.svg.zip)

sethladd:
OK, y'all: ptal: https://sethladd-flutter-io-staging.firebaseapp.com/

This is inclusive of:

* darker phone frames
* SVG all the things

sethladd:
The layer diagram is now updated. Staged at https://sethladd-flutter-io-staging.firebaseapp.com/

Thanks @abarth for noticing!

sethladd:
@LarkAscending @abarth ptal thanks!

LarkAscending:
LGTM! Resizing behavior is awesome... Colors too.

Two q's:
* Did we intentionally skip some of the libraries (like physics) in the layer cake diagram? 
* When we update the Shrine app for iOS/Android differences, will we be able to easily update the side-by-side images?

sethladd:
For the layer cake, I asked @bkobash to simply re-create the same layer diagram we link to from the homepage: https://docs.google.com/presentation/d/1cw7A4HbvM_Abv320rVgPVGiUP2msVs7tfGbkgdrTy0I/edit which was created by @abarth so I trust it shows the good stuff :)

The new screenshots + phone frames are now easier to update than before, which is pretty cool. The frames are implemented as SVG so we can swap the screenshots without recreating the whole thing (assumes our dimensions remain the same). In other words, the frames are just _frames_. :)

PR #487, opened 2017-03-31
Make sample code easier to follow. Add Swift code.

mit-mit:


platform-channels.md
```diff
@@ -107,49 +109,73 @@ prefix', e.g. `samples.flutter.io/battery`.
...
+  // Get battery level.
 }
 ```
 
 Next, we invoke a method on the method channel, specifying the concrete method
-to call via the String identifier `getBatteryLevel`. We then use the returned
-result to update our user interface inside `setState`:
+to call via the String identifier `getBatteryLevel`. The call may fail -- for
+example if the platform does not support the platform API (such as when running
+in a simulator), so we wrap the `invokeMethod` call in a try-catch statement:
```
mravn-google: Replace : by .

platform-channels.md
```diff
@@ -107,49 +109,73 @@ prefix', e.g. `samples.flutter.io/battery`.
...
+    return new Material(
+      child: new Center(
+        child: new Column(
+          mainAxisAlignment: MainAxisAlignment.spaceEvenly,
+          children: <Widget>[
+            new RaisedButton(
+              child: new Text('Get Battery Level'),
+              onPressed: _getBatteryLevel,
+            ),
+            new Text(_batteryLevel, key: new Key('Battery level label')),
```
mravn-google: What's the Key used for?

platform-channels.md
```diff
@@ -313,6 +374,121 @@ is called, we report that instead.
 }];
 ```
 
+You should now be able to run the app on IOS. If you are using the iOS
```
mravn-google: iOS

platform-channels.md
```diff
@@ -313,6 +374,121 @@ is called, we report that instead.
...
+  (call: FlutterMethodCall, result: FlutterResultReceiver) -> Void in
+  if ("getBatteryLevel" == call.method) {
+    receiveBatteryLevel(result: result);
+  } else {
+    result(FlutterMethodNotImplemented);
+  }
+});
+```
+
+You should now be able to run the app on IOS. If you are using the iOS
```
mravn-google: iOS

platform-channels.md
```diff
@@ -313,6 +374,121 @@ is called, we report that instead.
...
+1. If present, delete the files `AppDelegate.h`, `AppDelegate.m`, and
+`Supporting files/main.h`. Select 'Move to trash' when asked.
+
+1. Add a new `AppDelegate.swift` main file:
+  * Right-click the Runner>Runner folder, and select **New File...**. 
+  * Select `Swift File`, and click **Next**
+  * Name the file `AppDelegate.swift`
+  * Select **Create Bridging Header** when asked.
+
+1. Open `Runner-Bridging-Header.h`, and a Flutter import:
```
mravn-google: add a

platform-channels.md
```diff
@@ -313,6 +374,121 @@ is called, we report that instead.
...
+}
+```
+
+1. Validate that the project builds (shortcut `cmd-b`).
+
+With the project converted to use Swift for the main application class we can
+now begin.
+
+First we override `didFinishLaunchingWithOptions` to contain a
+FlutterMethodChannel tied to the channel name `samples.flutter.io/battery`:
```
mravn-google: FlutterMethodChannel should be code font?

platform-channels.md
```diff
@@ -313,6 +374,121 @@ is called, we report that instead.
...
+    receiveBatteryLevel(result: result);
+  } else {
+    result(FlutterMethodNotImplemented);
+  }
+});
+```
+
+You should now be able to run the app on IOS. If you are using the iOS
+Simulator, note that it does not support battery APIs, and the app will thus
+display 'battery info unavailable'.
```
mravn-google: Battery

platform-channels.md
```diff
@@ -80,8 +80,10 @@ display the current battery level. It uses the Android `BatteryManager` API, and
 the iOS `device.batteryLevel` API, via a single platform message,
 `getBatteryLevel`.
 
-*Note*: The full source-code for this example is available in
-[`/examples/platform_channel/`](https://github.com/flutter/flutter/tree/master/examples/platform_channel).
+*Note*: The full, run-able source-code for this example is available in
```
mit-mit: Apparently it is indeed a work (with two n's, sigh)

platform-channels.md
```diff
@@ -80,8 +80,10 @@ display the current battery level. It uses the Android `BatteryManager` API, and
...
+[`/examples/platform_channel/`](https://github.com/flutter/flutter/tree/master/examples/platform_channel)
+for Android with Java and iOS with Objective-C. For iOS with Swift, see
+[`/examples/platform_channel_swift/`](https://github.com/flutter/flutter/tree/master/examples/platform_channl_swift).
```
mit-mit: fixed

platform-channels.md
mit-mit: yup

platform-channels.md
mit-mit: It was left over from an old experiment; thanks for catching that!

platform-channels.md
mit-mit: Yup

platform-channels.md
mit-mit: Fixed

platform-channels.md
mit-mit: Fixed

platform-channels.md
mit-mit: Fixed

platform-channels.md
mit-mit: Fixed

PR #489, opened 2017-04-03
Redirect with and without trailing slash

sethladd:


sethladd:
Closes https://github.com/flutter/flutter/issues/9133

sethladd:
cc @abarth ptal thanks!

abarth:
LGTM

PR #496, opened 2017-04-07
Skip failing test case

mit-mit:
Skip failing test as per https://github.com/flutter/flutter/issues/9248


mit-mit:
Merging, TBR @sethladd @abarth

abarth:
Can you file a bug about reverting this patch when we can?

PR #500, opened 2017-04-10
PR Bot - now with CloudFunctions

FaisalAbid:
CloudFunctions is just used to post a comment to GitHub, this will allow me to also debug when things go wrong.

I have also added a enable/disable flag for the bot, if things break down, a simple travis flag disables it.

### DIFF
--- a/tool/prdeployer.dart
+++ b/tool/prdeployer.dart
@@ -2,17 +2,18 @@ import 'package:firebase/firebase_io.dart';
 import 'dart:async';
 import 'package:github/server.dart';
 import 'dart:io';
+import 'package:http/http.dart';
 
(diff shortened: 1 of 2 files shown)


FaisalAbid:
I was wondering if I should get rid of GitHub all together, what if they start rate limiting the public calls too (though they shouldn't as it doesn't say anywhere they will - but also does not say they wont!)

Perhaps we give it a try for now and go for it, we can easily enable/disable PR bot now through a flag in travis if the public non authenticated calls get rate-limited.

FaisalAbid:
Lets merge this and 🤞  -

sethladd:
merged!

PR #501, opened 2017-04-15
Update Android source, output paths

cbracken:
Sources are now under <app_dir>/android/app/src. Also corrects the
location of the built APK.


abarth:
LGTM

PR #504, opened 2017-04-15
Update docs.flutter.io and api.dartlang.org links

cbracken:
Adjusts links for recent punctuation changes to dart:ui, dart:io,
dart:core links.


abarth:
LGTM

PR #519, opened 2017-04-24
Remove old text about GlobalKeys and Focus

abarth:
Developers no long need to worry about this topic because the framework does it
for them. We should probably have an article about focus management for
advanced developers, but that's probably best done in the dartdocs for
FocusNode or FocusScope.

Fixes #7514


abarth:
@LarkAscending @sethladd

abarth:
Fixes https://github.com/flutter/flutter/issues/7514

PR #530, opened 2017-04-28
Fix breaking change

mit-mit:
Update backgroundColor -> color to match https://github.com/flutter/flutter/pull/9648

### DIFF
--- a/_includes/_code/container/main.dart
+++ b/_includes/_code/container/main.dart
@@ -38,7 +38,7 @@ class _MyHomePageState extends State<MyHomePage> {
   Widget build(BuildContext context) {
     var container = new Container(
       decoration: new BoxDecoration(
-        backgroundColor: Colors.black26,
+        color: Colors.black26,
       ),
       child: new Column(
         children: [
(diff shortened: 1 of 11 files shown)


mit-mit:
Fixes travis failure:

https://travis-ci.org/flutter/website/builds/226718731

```
Analyzing /home/travis/build/flutter/website/_includes/_code/container/main.dart...
  warning • The named parameter 'backgroundColor' isn't defined at _includes/_code/container/main.dart:41:9 • undefined_named_parameter
1 issue found.
(Ran in 5.6s)
```

sethladd:
Looks like we need to update backgroundImage, too.

sethladd:
lgtm

_includes/_code/container/main.dart
```diff
@@ -38,7 +38,7 @@ class _MyHomePageState extends State<MyHomePage> {
   Widget build(BuildContext context) {
     var container = new Container(
       decoration: new BoxDecoration(
-        backgroundColor: Colors.black26,
+        color: Colors.black26,
```
Hixie: For cases like this where there's a BoxDecoration with nothing but a `color`, in a Container, you can replace the `decoration` property with a `color` property that just takes the Color and avoid the BoxDecoration entirely.

Hixie: This applies to most cases here.

PR #533, opened 2017-05-03
Update git clone instructions to point to alpha (aka "release") branch

mit-mit:


sethladd:
Thanks! Do we have any wording around the output of "flutter doctor" that we need to tweak and point to alpha?

sethladd:
lgtm for this change

PR #535, opened 2017-05-03
Clarify what debug builds are

mit-mit:


mraleph:
I think it makes sense to clarify it like this - to avoid confusing people with checked mode.

lgtm

PR #536, opened 2017-05-03
Assets: Make it clearer that the directory name is unimportant.

Hixie:
Fixes https://github.com/flutter/flutter/issues/7891


abarth:
LGTM

PR #548, opened 2017-05-09
Update files page to reflect that PathProvider is now a package

mit-mit:
Update website to match https://github.com/flutter/flutter/pull/9916

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -9,3 +9,5 @@ dev_dependencies:
     sdk: flutter
   path: any
   test: ^0.12.13
+  url_launcher:
+  path_provider:
--- a/reading-writing-files.md
+++ b/reading-writing-files.md
@@ -5,13 +5,15 @@ sidebar: home_sidebar
 permalink: /reading-writing-files/
 ---
 
-This guide describes how to read and write files in Flutter by using the [`PathProvider`](https://docs.flutter.io/flutter/services/PathProvider-class.html) class.
+This guide describes how to read and write files in Flutter by using the
+[`PathProvider`](https://pub.dartlang.org/packages/path_provider) plugin and
+Dart's [`IO`](https://api.dartlang.org/stable/dart-io/dart-io-library.html) library.
 
 * TOC
 {:toc}
 
 ## Introduction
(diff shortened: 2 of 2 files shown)


mit-mit:
Failing in code validation:
https://travis-ci.org/flutter/website/builds/230380653
```
/home/travis/build/flutter/website/example/widgets_intro_md_8.dart...
    error • Target of URI doesn't exist: 'package:path_provider/path_provider.dart' at example/reading_writing_files_md_1.dart:5:8 • uri_does_not_exist
  warning • The method 'getApplicationDocumentsDirectory' isn't defined for the class '_FlutterDemoState' at example/reading_writing_files_md_1.dart:39:25 • undefined_method
2 issues found.
(Ran in 6.7s)
```

@sethladd does the code validator work with imports of packages from Pub?

sethladd:
@devoncarew might remember what we need to do to get more imports into the checker.

sethladd:
Ah, I remember. https://github.com/flutter/website/blob/master/tool/travis.sh#L33 is where the tool is run from. The extract.dart is looking for a pubspec.yaml. Maybe we need to add a new dependency to the pubspec.yaml that extract.dart expect?

@devoncarew which pubspec.yaml do we need to edit, if we want to add a dependency for the extract-and-analyze script for the website?

devoncarew:
Will take a look -

devoncarew:
@sethladd, got context back for this repo; @mit-mit, in your PR, add `path_provider: any` to the top-level pubspec.yaml for the repo - the build should go green :)

pubspec.yaml
```diff
@@ -9,3 +9,5 @@ dev_dependencies:
     sdk: flutter
   path: any
   test: ^0.12.13
+  url_launcher:
```
devoncarew: I think you need a version here - `any` should do.

mit-mit: Hmm, it's working great in several samples like this, I think this is the same as `any`?

mit-mit:
Passing, @sethladd can I get a LGTM?

mit-mit:
Merging to get the website unblocked. TBR @sethladd @szakarias

pubspec.yaml
devoncarew: (OK, I was unfamiliar w/ that format)

PR #549, opened 2017-05-09
Minor cleanup around widget catalog

Hixie:
cc @sethladd


sethladd:
@FaisalAbid any clues as to why the auto-stager didn't work here?

_includes/catalogpage.html
```diff
@@ -22,9 +22,7 @@
         {% if comp.categories contains include.category %}
         <li class="cards__item">
             <div class="catalog-entry">
-                <div class="catalog-image-holder">
-                    {% if comp.image contains '<svg' %}{{comp.image}}{% else %}<img alt="" src="{{comp.image}}" />{% endif %}
-                </div>
+                <div class="catalog-image-holder">{{comp.image}}</div>
```
sethladd: we don't need the `` tag here?

FaisalAbid: @Hixie Images can be SVG or png's etc, so this would render just the url in many cases

FaisalAbid:
@sethladd Not sure, the code didnt even execute looking at travis, will take a look

FaisalAbid:
@sethladd it looks like a change in travis

```Travis CI makes encrypted variables and data available only to pull requests coming from the same repository. These are considered trustworthy, as only members with write access to the repository can send them.

Pull requests sent from forked repositories do not have access to encrypted variables or data.```

Since @Hixie sent it from a forked repo, all the encrypted variables in Travis are removed, including the ones PR bot uses.

sethladd:
Ah! Good to know. Thanks!

_includes/catalogpage.html
Hixie: I changed the JSON file too.

Hixie:
Isn't sending PRs from a forked repo the way one is supposed to do things?

_includes/catalogpage.html
sethladd: Ah, thanks! That diff was collapsed.

sethladd:
lgtm

PR #552, opened 2017-05-10
Fix the analysis step

Hixie:
I'm a bad boy and directly committed my last change and it broke the build. My bad.


cbracken:
lgtm

PR #568, opened 2017-05-18
Document how to edit plugin using Swift

mravn-google:


mravn-google:
cc @mit-mit Would we want to do something like `flutter create -swift --plugin hello` instead?

mit-mit:
Yes, we would love a swift command line option; tracking in https://github.com/flutter/flutter/issues/7312. 

But I think having docs for this in the meantime is a great idea. Let me review this now!

platform-plugins.md
```diff
@@ -210,6 +210,59 @@ The iOS platform code of your plugin is located in `Pods/Development Pods/hello/
...
+Before editing the iOS platform code in Xcode, first make sure that
+the code has been built at least once by Flutter by executing
+`cd hello/example; flutter build ios`.
+
+Next,
+
+1. Launch Xcode
+1. Select 'File > Open', and select the `hello/example/ios/Runner.xcworkspace` file.
+   The iOS platform code of your plugin is located in `Pods/Development Pods/hello/Classes/` in the Project Navigator.
+1. Select the `Classes` folder, then select `File > New > File...` and add a new Cocoa Touch class.
```
mit-mit: Do we need the 'Cocoa Touch class' template? Looks like we are replacing all the contents anyway, so wouldn't the plain 'Swift class' work?

platform-plugins.md
```diff
@@ -210,6 +210,59 @@ The iOS platform code of your plugin is located in `Pods/Development Pods/hello/
...
+    #import "HelloPlugin.h"
+    #import <hello/hello-Swift.h>
+    
+    @implementation HelloPlugin
+    + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar>*)registrar {
+        [RealHelloPlugin registerWithRegistrar:registrar];
+    }
+    @end
+    ```
+1. Build the project again using Flutter by executing `flutter build ios` in
```
mit-mit: Why is this extra build step required? Won't it get build as part of the Run in the next step?

platform-plugins.md
```diff
@@ -210,6 +210,59 @@ The iOS platform code of your plugin is located in `Pods/Development Pods/hello/
...
+
+Next,
+
+1. Launch Xcode
+1. Select 'File > Open', and select the `hello/example/ios/Runner.xcworkspace` file.
+   The iOS platform code of your plugin is located in `Pods/Development Pods/hello/Classes/` in the Project Navigator.
+1. Select the `Classes` folder, then select `File > New > File...` and add a new Cocoa Touch class.
+   Name the class `RealHelloPlugin`, make it a sub-class of `NSObject`, and select `Swift` as language.
+   Press `Next` and accept the location for the new class.
+1. Replace the code for the new class with
```
mit-mit: Missing a colon after 'with'

platform-plugins.md
mravn-google: Yes, but you cannot launch from within Xcode without having done `pod install` again. Which happens when you do `flutter build ios`. I'll look at a way to rephrase to something simpler

platform-plugins.md
mravn-google: Probably. I just took the first option in the Xcode UI ☺️

mravn-google:
@mit-mit PTAL

PR #570, opened 2017-05-18
Add objc syntax highlighting

mravn-google:


platform-plugins.md
```diff
@@ -246,7 +246,7 @@ Next,
     }
     ```
 1. Replace the implementation in `HelloPlugin.m` with a forwarding call:
-    ```objective-c
+    ```objc
```
mit-mit: I think it's `objectivec`? See https://github.com/flutter/website/blob/master/platform-channels.md

mravn-google: Yes. I found the `objc` someplace else, but it doesn't work with out site. Was just about to change it.

PR #572, opened 2017-05-22
icon is now in widget

sethladd:
Fixes https://github.com/flutter/flutter/issues/9932


Hixie:
LGTM

Hixie:
I love that we have scripts that catch this.

PR #573, opened 2017-05-23
Fix a number of padding issues

mit-mit:
Fixes https://github.com/flutter/flutter/issues/10104

Optimizes layout on mobile in general

### DIFF
--- a/_sass/_homepage.scss
+++ b/_sass/_homepage.scss
@@ -140,6 +140,9 @@
                 margin-bottom: 16px;
             }
         }
+        section {
+            padding: 12px;
+        }
     }
 
     .homepage__key_point_title {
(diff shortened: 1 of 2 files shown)


_sass/_homepage.scss
```diff
@@ -165,4 +155,23 @@
...
+        .key-points {
+            display: block;
+
+            p {
+                margin-bottom: 16px;
+            }
+        }
+ 
+        section {
+            padding: 12px;
```
sethladd: we strive for multiples of eight. When creating a new style, please use a multiple of eight.

mit-mit: Got it, fixed

sethladd:
Thanks so much!

lgtm

PR #574, opened 2017-05-24
Document language choice in plugin template

mravn-google:


platform-plugins.md
```diff
@@ -158,6 +158,14 @@ This creates a plugin project in the `hello/` folder with the following content:
 * `example/`:
    - A Flutter app that depends on the plugin, and illustrates how to use it.
 
+By default, the plugin project will be set up to use Objective-C for iOS code and
```
mit-mit: maybe simplify 'will be set up to use' to 'will use' ?

PR #575, opened 2017-05-24
slight tweak to homepage for native features

sethladd:
We can add a link to all of our plugins later, once we have a few more.

cc @mit-mit 

Tries to make https://github.com/flutter/flutter/issues/10281 better for now


mit-mit:
One small comment, otherwise LGTM!

PR #576, opened 2017-05-25
Add a small doc section on the Timeline function

xster:
I keep forgetting how it works. Might as well write it down.


Hixie:
Yes! Documenting every question we ask is a great way to make our docs useful.

LGTM modulo the complaints Travis has.

PR #579, opened 2017-05-26
encourage people to thumbs up issues

sethladd:
Suggest another way to get support.


sethladd:
cc @mit-mit

mit-mit:
Love it! LGTM

PR #582, opened 2017-05-27
Small manual fix to underlined Get Started link on homepage.

LarkAscending:
Root cause is 
.card {   a {
    text-decoration: underline;
  }
}

in _catalog.scss is taking precedence over 
.get-started-button {
    text-decoration: none;
...
}
.get-started-button:hover {
    text-decoration: none;
...
}

in _homepage.scss

This humble fix doesn't address root cause but is the minimal solution for removing the underline on only the homepage. (The Widget Catalog really does need the underline for its links.) The other option is  a new layout for the homepage, and a good amount of CSS troubleshooting.

### DIFF
--- a/_sass/_homepage.scss
+++ b/_sass/_homepage.scss
@@ -1,4 +1,5 @@
 .get-started-button {
+    text-decoration: none;
     background-color: #0091EA;
     color: #fff;
     padding: 16px;
@@ -11,6 +12,7 @@
 }
 
 .get-started-button:hover {
+    text-decoration: none;
     color: #fff;
 }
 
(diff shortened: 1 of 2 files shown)


sethladd:
Thanks! I'd love to see if we can fix this in the CSS file.

What do you think about adding this to `_homepage.scss` ?

```
.homepage {

    a {
        text-decoration: none;
    }
```

At line 70?

LarkAscending:
Will spend a few minutes testing that... in case it's problematic here's the rationale for a quick fix: 
1. There's already a dedicated class for the get started button. 
2. There are only 2 of these buttons on the homepage, and if we're planning to add more buttons it makes sense to do something scalable. If 2 is the upper limit then a quick local fix matches the scale of the problem.

sethladd:
lgtm thanks!

Makes the buttons look much nicer now :)

Fixes https://github.com/flutter/flutter/issues/10127

PR #583, opened 2017-05-27
removing sudo from install instructions

sethladd:
it should be up to the developer, also doesn't work in windows


sethladd:
cc @Hixie @mit-mit

mit-mit:
LGTM

PR #588, opened 2017-06-01
Fixing overflowing margin issue for select pages on iOS

LarkAscending:
This PR addresses https://github.com/flutter/flutter/issues/10120 and https://github.com/flutter/flutter/issues/10153. 

#10153 is fixed by giving a single instance of inline CLI syntax its own, dedicated line in the MD file.

#10120 is diagnosed, but not yet fixed. This file has multiple instances of special characters throwing kramdown for a loop. See issue conversation for the chunk of platform-plugins.md that is causing problems. I'm making a partial commit now, with formatting fixes, but the way forward is likely to address forward slash characters (/) in inline syntax. We should look at adding a kramdown option to _config.yml, to deal with these automatically (rather than individually).

### DIFF
--- a/_includes/setup/get-sdk-win.md
+++ b/_includes/setup/get-sdk-win.md
@@ -19,7 +19,7 @@ To update an existing version of Flutter, see [Upgrading Flutter](/upgrading/)
 To run the `flutter` command in any terminal session, you need to add it to your PATH environment variable:
 
 * Go to "Control Panel > User Accounts > User Accounts > Change my environment variables"
-* Under "User variables" check if there is an entry called "Path":
+* Under "User variables" check if there is an entry named `Path`:
     * If the entry does exist, append the full path to `flutter\bin` using `;` as a separator from existing values.
     * If the entry does not exist, create a new user variable named `Path` with the full path to `flutter\bin` as its value.
 
@@ -28,7 +28,7 @@ Close and re-open any terminal window (e.g. command prompt or PowerShell) to app
 ### Run flutter doctor
 
 Open a new Command Prompt or PowerShell window and run the following command to
-see if there  are any dependencies you need to install to complete the setup:
+see if there are any dependencies you need to install to complete the setup:
 
 ```
 > flutter doctor
@@ -64,5 +64,5 @@ and basic crash reports. This data is used to help improve Flutter tools over ti
 Analytics is not sent on the very first run or for any runs involving `flutter config`,
 so you can opt out of analytics before any data is sent. To disable reporting, 
 type `flutter config --no-analytics` and to display the current setting, type 
-`flutter config`. See Google's privacy policy:[www.google.com/intl/en/policies/privacy](https://www.google.com/intl/en/policies/privacy/).
+`flutter config`. See [Google's privacy policy](https://www.google.com/intl/en/policies/privacy/).
 {: .alert-warning}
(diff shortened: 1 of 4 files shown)


mit-mit:
I still see the https://github.com/flutter/flutter/issues/10153 issue in Chrome Dev Tools with the iPhone 5 profile when I run `LarkAscending:ios_margins` with `bundle exec jekyll serve`:

127 0 0 1-4000-setup-macos- iphone 5

Isn't this some kind or margin / min width is too high issue, or something like that?

LarkAscending:
I'm running with the Simulator/iPhone 7 profile and I don't see it.
Will create a new profile for iPhone 5 and debug further..,

To the margin question, no it can't be that type of issue. You can see that
by removing the problem chunk from platform-plugins.md and regenerating.
If you still see it, try the iPhone 7 profile in DevTools. Since we can
make the issue go away by removing content, it's not a layout issue.
Also these MD files are governed by the same site-wide layouts as all the
other pages on flutter.io.

LarkAscending:
So, on iPhone 5 it looks like all the setup_* pages have this issue (along
with platform-plugins, of course).

The overlapping includes are get-sdk and editor-setup and android-setup.
I'll take a look at those...

LarkAscending:
The files get-sdk.md and get-sdk-win.md had an ambiguous colon (:) character in the last para. Restructured the link to remove the offending character, and made a small change to ios-setup.md to display correctly on iPhone 5 (Simulator).

platform-plugins.md still has issues - PTAL.

mit-mit:
Confirmed that setup pages look great now!

mit-mit:
I think I found the underlying issue! `platform-plugins` has several lines like this:

```
* `android/src/main/java/com/yourcompany/hello/HelloPlugin.java`:
```

Kramdown translates `bla bla` in a back-tick block to `<code>bla bla</code>`, and per our stylesheet `<code>` means `white-space: pre-wrap;` which means never line-break. And because the filenames on this page are so very long, the line basically doesn't fit on a narrow screen.

I'm not sure how to fix it though...

LarkAscending:
Really great troubleshooting and diagnosis! 1st_place_medal: 
* Matches what I've seen here. That specific line with the long path is a problem.
* Also confirming that is the Kramdown behavior wrt single graves, or backticks.

Re: a solution, we can add CSS to allow line breaks in <code></code>. Question is, do we want to do that across the whole site? Are there any cases where we wouldn't want that?

If yes and we need to limit the impact, another option might be to insert line breaks in the MD where we have a long path and add an inner declaration to tell the parser what to do. Something like this...
{::options hard_wrap="true" /} just before the line with the long path. [Parser options here](https://kramdown.gettalong.org/rdoc/Kramdown/Options.html). gfm_quirks is another interesting one...

Further ideas welcome...

LarkAscending:
@mit-mit Just to clarify, am blocked on a response re: line breaks. If we go that route, tending from you on platform-plugins. I'll also help out as needed on this end but I'm focused on setup (which is fixed).

mit-mit:
Sorry, missed the question.

We already line break in code snippets with Dart code, etc

The "problem child" are long paths. If we could make these line break when they don't fit, that would be good. It would be ideal if the line break was at a slash, e.g.

android/src/main/java/com/yourcompany/
hello/HelloPlugin.java

mit-mit:
@LarkAscending any news on this?

Hixie:
@LarkAscending I'm going to close this PR to take it off our review queue, but please don't hesitate to reopen it (or a new one) when you're ready for review again.

LarkAscending:
I'm not the assignee ;-)  so didn't pick this up after being away for a couple of weeks.

@mit-mit was expecting a commit or separate PR from you to address your suggestion. Only the plugins page is affected at this point. It's been a while but I believe the suggested fix means reverting all the commits in this PR. Over to you...

PR #589, opened 2017-06-01
Mention why debug/run might be disabled

mit-mit:
Fixes https://github.com/flutter/flutter/issues/9810


mit-mit:
@InMatrix incorporated your review feedback from flutter/flutter#9810, PTAL

PR #592, opened 2017-06-05
Adding Edit Source and File Issue buttons to every page.

LarkAscending:
HIGH IMPACT CHANGE - see  https://github.com/flutter/flutter/issues/10416 for discussion. This is an incremental step toward inviting Flutter users to contribute to flutter.io.

image

This PR adds two buttons just below the title of every page on flutter.io. 
* The Edit Source button links to the current file on the master branch in the flutter/website repo. For example, clicking this button on the Assets and Images page at https://flutter.io/assets-and-images/ will take users to this URL: `https://github.com/flutter/website/blob/master/assets-and-images.md`.
* The File Issue button creates a new issue that is prepopulated with the current page title and URL. In the above example, the URL would be `https://github.com/flutter/flutter/issues/new?title=Issue%20from%20website%20page%20Adding%20Assets%20and%20Images%20in%20Flutter&body=From%20URL:%20https://flutter.io/assets-and-images.md`

### Notes
The pencil icon is not ideal - it's horizontal (instead of diagonal like the paperclip) and smaller. Investigations and experiments with CSS, getbootstrap, other icons, and HTML entities led nowhere. Would like this icon to be larger and diagonal to match the paperclip.

Tested on Chrome and Safari on Linux and macOS. Also tested on Safari on the iOS Simulator w/iPhone5 profile. Not yet tested on IE/Windows, or Android. Suggestions on testing requirements welcome...

@cc InMatrix

### DIFF
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -6,15 +6,17 @@ layout: default
   
   {% unless page.hide_title %}
   <header class="post-header">
-    <h1 class="post-title">{{ page.title }} </h1>
-      <div class="btn-group" role="group">
-         <a href="https://github.com/flutter/website/blob/master/{{page.path}}" class="btn btn-default">
+      <div class="btn-group contribute" role="group">
+         <a href="https://github.com/flutter/website/blob/master/{{page.path}}" class="btn btn-sm">
             <i class="glyphicon glyphicon-pencil"></i> Edit Source
          </a>
-         <a href="https://github.com/flutter/flutter/issues/new?title=Issue from website page {{page.title}}&body=From URL: {{site.url}}/{{page.path}}" class="btn btn-default">
+         <a href="https://github.com/flutter/flutter/issues/new?title=Issue from website page {{page.title}}&body=From URL: {{site.url}}/{{page.path}}" class="btn btn-sm">
             <i class="glyphicon glyphicon-paperclip"></i> File Issue
         </a>
      </div>
+   <div>
+    <h1 class="post-title">{{ page.title }} </h1>
+   </div>
 
   </header>
   {% endunless %}
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -6,7 +6,16 @@ layout: default
   
   {% unless page.hide_title %}
   <header class="post-header">
-    <h1 class="post-title">{{ page.title }}</h1>
+    <h1 class="post-title">{{ page.title }} </h1>
+      <div class="btn-group" role="group">
+         <a href="https://github.com/flutter/website/blob/master/{{page.path}}" class="btn btn-default">
+            <i class="glyphicon glyphicon-pencil"></i> Edit Source
+         </a>
+         <a href="https://github.com/flutter/flutter/issues/new?title=Issue from website page {{page.title}}&body=From URL: {{site.url}}/{{page.path}}" class="btn btn-default">
+            <i class="glyphicon glyphicon-paperclip"></i> File Issue
+        </a>
+     </div>
+
   </header>
   {% endunless %}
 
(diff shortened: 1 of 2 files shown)


Hixie:
I think it would look a bit more professional if we had these on the right rather than right under the title.
Alternatively, we could put them in the tool bar, next to "API - GitHub - Support".

I don't want our site to look like it's a work in progress, we should look confident that the site is solid, while still inviting participation.

LarkAscending:
Sorry if you feel it looks unprofessional... not sure how to respond to that or how 'professional' or 'work in progress' can be objectively defined. But we can certainly place the buttons somewhere else. In the banner is a bit far away and we want folks to notice. OK if we try the right margin?

For the future, is there another site design that you are using as a mental model? That you'd like flutter.io to emulate?

Hixie:
"Professional" is probably the wrong word... Not really sure how to express what I mean more precisely.

What I'm trying to get as is the difference between a Google doc in "publish" mode and a Google doc in "edit" mode. IMHO our Web site should feel like we have published it, as opposed to feeling like we're still editing it. I understand that it's hard to do that while simultaneously encouraging people to feel like they can contribute and participate. :-)

InMatrix:
It seems that this is a broadening of scope from the issue we discussed in the study review. Just a reminder, what I thought would be helpful was to ask the community to contribute code samples and images to _APIs_  that we know are under-documented. We can wait until we have good examples and guidelines against which community contributors can self-check their PRs.

I'm not sure if it's useful to put "Edit Source" and "File Issue" on every webpage on flutter.io without clarifying which part of the page is work in progress and which part is solid. I'd suggest we roll this out carefully and slowly. Perhaps, we can start with new tutorials/howtos, label them as "beta" content, and use these buttons to more explicitly solicit feedback and edits from the community.

LarkAscending:
I think I understand now what @Hixie is saying; there are some projects on GitHub that are really unfinished (works-in-progress) and 100% dependent on community contributions for doc. And the quality of their doc usually reflects this. It can feel a bit incomplete and ad hoc, whereas Flutter has put a lot of effort into its docs. 

I believe we're looking for a balance between being clear and obvious that we accept contributions, while also conveying somehow that Flutter doc *has* actually been vetted and is being actively tended. 

It might be possible to mark certain pages/sections we are actively seeking contributions for. With a colored header, for example, similar to the marking for note or warning text. These buttons are still useful and needed but we might want that too.

@InMatrix Point taken on the API doc contributions... dartdocs might be able to reuse what we do for flutter.io. But if you prefer we can spin off a different issue for this work.

Hixie:
@LarkAscending Yeah that's a much better way to express what I meant, thanks.

LarkAscending:
New appearance:
image

_layouts/page.html
```diff
@@ -6,7 +6,18 @@
   
   {% unless page.hide_title %}
   <header class="post-header">
-    <h1 class="post-title">{{ page.title }}</h1>
+      <div class="btn-group contribute" role="group">
+         <a href="https://github.com/flutter/website/blob/master/{{page.path}}" class="btn btn-sm">
+            <i class="glyphicon glyphicon-pencil"></i> Edit Source
+         </a>
+         <a href="https://github.com/flutter/flutter/issues/new?title=Issue from website page {{page.title}}&body=From URL: {{site.url}}/{{page.path}}" class="btn btn-sm">
```
sethladd: Nice!

sethladd:
I like the two new links! How does it look on mobile? (aka squeezed all the way narrow)

For the "file an issue", can we use the GitHub icon? I see a GitHub icon in fontawesome, which I think we still use, as an option.

LarkAscending:
Larger buttons display above the title text on Small and Medium mobile devices:
image
On Large mobile devices, the buttons share the header row and make the title text wrap around.

Smaller buttons make the title text wrap on more devices (both Medium and Large).
image

sethladd:
lgtm!

PR #594, opened 2017-06-07
Update getting-started.md

mit-mit:
Now that we don't link directly to this page from the top-bar, remove some of the duplicated preamble


sethladd:
lgtm

PR #599, opened 2017-06-09
Kicking off Sample Catalog template work...

LarkAscending:
This is a collaborative PR with an initial, seed commit for the Sample Catalog.
cc @InMatrix 

* An overview page, `sample_catalog.md`, lists all the sample apps via template logic. That's the main goodness here. Ultimately, when we can transition to vanilla bootstrap, the small images will be on the left and text on the right (thanks to `media-object` classes). For right now it's a simple vertical list.

* Links go to detail pages, like `samples/basic_app_bar.md`. The side-by-side images need a little wrangling to use the custom class `catalog`, which just aligns their top edges and makes it all look even. The {.<classname>} syntax has worked in the past to mark paragraphs in Markdown, but not here. An alternative that is ugly but legal is to just use HTML  tags with the "class=" syntax. HTML can be embedded directly in MD files, without ceremony :-)


Hixie:
Rather than having a separate catalog for samples, would it be better to just link to the samples from the widget catalog?

_data/samples/widgets.yml
```diff
@@ -0,0 +1,12 @@
+- name: "AppBar"
+  link: "https://docs.flutter.io/flutter/material/AppBar-class.html"
+  image: "<img class='media-object' src='/images/samples/ios_animated_list_very_small.png' alt=''>"
+  description: "Lorem ipsum"
+  detail-page: "/catalog/basic_app_bar/"
```
sethladd: can we qualify this path? we may have other catalogs in the future. qualifying to samples might help.

_sass/_customstyles.scss
```diff
@@ -334,6 +334,10 @@ img.screenshotSmall {
     max-width: 300px;
 }
 
+img.catalog {
```
sethladd: Let's create a new _samples_catalog.scss and put all the samples styles there. I've found it makes it easier to manage when we separate out styles by function.

sample_catalog.md
```diff
@@ -0,0 +1,25 @@
+---
+layout: page
+title: Sample Catalog
+permalink: /sample_catalog/
+---
+
+
+This page lists all the sample apps in the Sample Catalog.
```
sethladd: is _this_ the sample catalog?

samples/basic_app_bar.md
```diff
@@ -0,0 +1,130 @@
+---
+layout: page
+catalog: basic_app_bar
+title: "AppBar Basics"
```
sethladd: maybe add "sample" somehow to the title, for SEO?

samples/basic_app_bar.md
```diff
@@ -0,0 +1,130 @@
...
+permalink: /catalog/basic_app_bar/
+---
+
+
+![iOS AppBar](/images/samples/ios_animated_list_small.png){.catalog}
+![Android AppBar](/images/samples/animated_list_small.png){.catalog}
+
+
+### Description
+An AppBar with a title, actions, and an overflow dropdown menu.
```
sethladd: will developers knows what an AppBar is, or do we need to link it?

samples/basic_app_bar.md
```diff
@@ -0,0 +1,130 @@
...
+The two most common choices are available as action buttons and the remaining
+choices are included in the overflow dropdow menu.
+
+### Related
+See also:
+- The "Layout-Structure" section of the material design specification:
+    <https://material.io/guidelines/layout/structure.html#structure-app-bar>
+
+### Sample App
+```dart
```
sethladd: When Flutter changes, will we be notified that this code needs to be changed?

samples/basic_app_bar.md
```diff
@@ -0,0 +1,130 @@
...
+        body: new Padding(
+          padding: const EdgeInsets.all(16.0),
+          child: new ChoiceCard(choice: _selectedChoice),
+        ),
+      ),
+    );
+  }
+}
+
+class Choice {
```
sethladd: How can we make it ultra clear which code is "optional" (specific to this sample), and which code is required (will always need) ?

Do I need to make a Choice class when making an AppBar?

LarkAscending:
Thanks for the interest! It's early days yet, and this PR isn't ready for review. For this amount of feedback on the general direction, maybe a meeting would help...

_sass/_customstyles.scss
LarkAscending: I wouldn't recommend taking this step yet - it's possible no new styles will be needed.

In general, if we have a set of practices and criteria for reviewing CSS, our approach on /website should be surfaced and described in Contributing.

sample_catalog.md
LarkAscending: This is the overview page.

LarkAscending:
@Hixie The current plan is to add an auto-generated link to the Sample Catalog to each widget in the Widget Catalog. The samples will be filtered by the widgets they describe...

IMHO users will need the samples easily accessible, rather than embedded in the Widget Catalog. But maybe a discussion and/or meeting on the general approach would be helpful.

Hixie:
> The current plan is to add an auto-generated link to the Sample Catalog to each widget in the Widget Catalog. The samples will be filtered by the widgets they describe...
>
> IMHO users will need the samples easily accessible, rather than embedded in the Widget Catalog.

That sounds reasonable to me.

Hixie:
What's the status of this PR?

LarkAscending:
@sethladd might know... I believe he/Hans created a separate parallel PR. So it's probably an orphan; I haven't heard anything, though.

Hixie:
Ok since nobody seems to want to own this PR, I'm going to close it. :-)

PR #603, opened 2017-06-15
Clarify where to find keytool

mit-mit:
Fixes https://github.com/flutter/flutter/issues/10158


wmleler:
Much clearer. Thanks!

PR #605, opened 2017-06-16
Update package documentation

mit-mit:
Fixes https://github.com/flutter/flutter/issues/10070


developing-packages.md
```diff
@@ -0,0 +1,174 @@
...
+title: Developing Packages & Plugins
+permalink: /developing-packages/
+---
+
+* TOC
+{:toc}
+
+## Package introduction
+
+Dart packages enable the creation of modular code that can easily be shared. A
```
sethladd: Should we just say "_Packages_ ..." instead of "Dart packages" ?

developing-packages.md
```diff
@@ -0,0 +1,174 @@
...
+### Package types {#types}
+
+Flutter supports several kinds of packages:
+
+* *Dart packages*: General packages written in Dart without any dependencies on
+  any specific frameworks. They can be used across many frameworks such as
+  [AngularDart](https://webdev.dartlang.org/angular) and Flutter. A concrete
+  example is the [`path`](https://pub.dartlang.org/packages/path) package.
+
+* *Flutter packages*: General packages which contain Flutter specific
```
sethladd: I'm not sure we want to make this distinction. The _kind_ of package is the same: a pub package.

I think the user cares about one thing:

* where do I go to see all the packages I can use as a Flutter developer

The differences you talk about here are implementation details that probably don't matter. Only: "will it work with Flutter?"

developing-packages.md
```diff
@@ -0,0 +1,174 @@
...
+  functionality and thus have a dependency on the Flutter framework, restricting
+  their use to Flutter only. A concrete example is the
+  [`fluro`](https://pub.dartlang.org/packages/fluro) package.
+
+* *Flutter plugins*: Flutter packages which in additional to the Dart code also
```
sethladd: I think the package/plugin distinction probably isn't too helpful. A plugin isn't a sub-type of package.

maybe some packages offer plugins, but it's still a package.

developing-packages.md
```diff
@@ -0,0 +1,174 @@
...
+  version, author, etc.
+
+* A `lib` directory containing the public code in the package, minimally a
+  single `<package-name>.dart` file.
+
+For additional details, see the [Dart library package](https://www.dartlang.org/guides/libraries/create-library-packages) documentation.
+
+### Package types {#types}
+
+Flutter supports several kinds of packages:
```
sethladd: Maybe the framing is better as : "there are two kinds of code you can put into a Flutter package: dart code, and native code. If you are writing Dart code that has a dependency on Flutter, do this: ...."

(instead of "different kinds of packages", let's use "packages can contain different kinds of code")

developing-packages.md
```diff
@@ -0,0 +1,174 @@
...
+  [`fluro`](https://pub.dartlang.org/packages/fluro) package.
+
+* *Flutter plugins*: Flutter packages which in additional to the Dart code also
+  contain a platform-specific implementation for Android, for iOS, or for both.
+  A concrete example is the
+  [`battery`](https://pub.dartlang.org/packages/battery) plugin.
+
+## Developing packages
+
+### Create a package project
```
sethladd: You use "-ing" in many titles, but you use "Create" here. Maybe "Creating..." is more consistent

sethladd: also, do we mean "project" ? I think that's superfluous. "Creating a package" works.

developing-packages.md
```diff
@@ -0,0 +1,174 @@
...
+1. Add a main package file `lib/example.dart` with the following contents:
+    ``` dart
+    library example;
+
+    /// Add public package code (Classes, etc.) here:
+    ```  
+
+## Developing plugins: platform-specific packages {#plugin}
+
+Packages enable sharing Dart code. If you want to develop packages that call
```
sethladd: Packages enable sharing code. I wouldn't qualify it to "Dart code" here. That makes it easier later to explain a plugin.

developing-packages.md
```diff
@@ -0,0 +1,174 @@
...
+into platform-specific APIs available on Android, on iOS, or on both, you need
+to develop a platform-specific package, a so-call plugin.
```
sethladd: rm "so-called". Either we canonically define a plugin, or we shouldn't use the term :)

developing-packages.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Create a plugin project
```
sethladd: Creating a plugin

developing-packages.md
```diff
@@ -0,0 +1,174 @@
...
+
+You can run the example app by pressing the &#9654; button.
+
+   
+## Publishing packages {#publish}
+
+Once you have implemented a package, you can publish it on
+[Pub](https://pub.dartlang.org/), so that other developers can easily use it.
+
+Prior to publishing, make sure to complete the `pubspec.yaml`, `README.md`, and
```
sethladd: what does "complete" mean? Is there something they _must_ do before publishing? If so, let's tell them.

platform-plugins.md
```diff
@@ -4,267 +4,6 @@ title: Plugins&#58; Integrating with platform-specific code
 permalink: /platform-plugins/
 ---
 
-This guide describes how Flutter apps can integrate with platform-specific code
-available on iOS and Android devices. This includes both device APIs (like
-[`url_launcher`](https://pub.dartlang.org/packages/url_launcher) and
-[`battery`](https://pub.dartlang.org/packages/battery)) and third-party platform SDKs 
-(like [Firebase](https://github.com/flutter/plugins/blob/master/FlutterFire.md)).
+<meta http-equiv="Refresh" content="0; url=/using-packages/" />
```
sethladd: Let's use the Firebase rules for redirects, so we don't need to keep this page around.

Check out https://github.com/flutter/website/blob/master/firebase.json

There's a few redirects in there now.

using-packages.md
```diff
@@ -0,0 +1,262 @@
...
+title: Using Packages
+permalink: /using-packages/
+---
+
+Flutter supports using shared packages contributed by other developers to the
+Flutter and Dart ecosystems. This allows you to quickly build your app without
+having to develop everything from scratch.
+
+Existing packages enable many use cases, for example, file path manipulation
+([`path`](https://pub.dartlang.org/packages/path)), custom navigation/route
```
sethladd: nit: is path really relevant to Flutter developers? What about a JSON library? or, better yet: http package

using-packages.md
```diff
@@ -0,0 +1,262 @@
...
+* TOC
+{:toc}
+
+## Using packages
+
+### Searching for packages
+
+Packages are published to the *[Pub](https://pub.dartlang.org)* package repository. 
+
+The list [Flutter packages](https://pub.dartlang.org/flutter/packages) displays
```
sethladd: did you mean "The list of Flutter packages" ?

using-packages.md
```diff
@@ -0,0 +1,262 @@
...
+packages that have been developed specifically for Flutter (i.e., that declare a
+dependency on the Flutter framework). Flutter-specific packages are presented
+with a Flutter logo adornment.
+
+The standard Pub search field supports searching among all published packages.
+Note however that many packages depend on other frameworks, such as the
+[AngularDart](https://webdev.dartlang.org/angular/) web framework, and those
+will not work with Flutter. We hope to offer a better search experience with
+support for filtering to only packages that are compatible with Flutter in the
+future.
```
sethladd: Is there a bug we can link to?

cc @kevmoo

using-packages.md
```diff
@@ -0,0 +1,262 @@
...
+
+{% include note.html content="Flutter is still a young platform, and only a
```
sethladd: FWIW I don't think you need this disclaimer.

using-packages.md
```diff
@@ -0,0 +1,262 @@
...
+fairly small set of Flutter packages are currently available on pub. We
+encourage all Flutter developers to [develop](/developing-packages/) new
+packages." %}
+
+### Adding a package dependency to an app
+
+To add a plugin 'plugin1' to an app:
+
+1. Open the `pubspec.yaml` file located inside your app folder, and add
+`plugin1:` under `dependencies`.
```
sethladd: Bonus points to show a small example here.

using-packages.md
```diff
@@ -0,0 +1,262 @@
...
+
+1. Get the plugin:
+   * From the terminal: Run `flutter packages get`
+   * From IntelliJ: Click 'Packages Get' in the action ribbon at the top of `pubspec.yaml`
+
+1. Build or run your app. As part of this, Flutter will add the package code to
```
sethladd: This is a little confusing. The code was already added to my add when I installed ("get") the package.

I'm not sure we need this last bullet. Only two steps! :)

maybe, Step Three is "import the package's library into your app"

using-packages.md
```diff
@@ -0,0 +1,262 @@
...
+[lockfile](https://www.dartlang.org/tools/pub/glossary#lockfile). This ensures
+that you get the same version again if you, or another developer on your team,
+run `flutter packages get`.
+
+If you want to upgrade to a new version of the package, for example to use new
+features in that package, run `flutter packages upgrade` ('Upgrade dependencies'
+in IntelliJ). This will retrive the highest available version of the package,
+which is allowed by the version constraint you have specified in `pubspec.yaml`.
+
+### Dependencies on unpublished packages
```
sethladd: This seems generic and duplicates existing docs. Should we just link to somewhere?

using-packages.md
```diff
@@ -0,0 +1,262 @@
...
+    plugin1:
+      git:
+        url: git://github.com/flutter/plugin1.git
+```
+
+For more details, see the [Pub Dependencies article](https://www.dartlang.org/tools/pub/dependencies).
+
+## Examples
+
+### Example: Using the CSS Colors package {#css-example}
```
sethladd: You say "CSS Colors package" and then you start talking about URL Launcher. Is that intentional?

using-packages.md
```diff
@@ -0,0 +1,262 @@
...
+    ```dart
+    dependencies:
+      flutter:
+        sdk: flutter
+      css_colors: ^1.0.0
+    ```
+
+1. Run `flutter packages get` in the terminal, or click 'Packages get' in IntelliJ
+
+1. Open `lib/main.dart` and replace its full contents with:
```
sethladd: BTW great example!!!

using-packages.md
```diff
@@ -0,0 +1,262 @@
...
+phone's default browser open, and the Flutter homepage appear.
+
+
+
+### Example: Using the URL Launcher plugin to launch the browser {#url-example}
+
+The [URL Launcher](https://pub.dartlang.org/packages/url_launcher) plugin
+enables you to open the default browser on the mobile platform to display a
+given URL. It demonstrates how packages may also contain platform-specific code
+(we call these packages 'plugins'). It is supported on both Android and iOS.
```
sethladd: It doesn't seem like there's any difference between installing and using a package, or a package+plugin.

If there is no difference, do we need this example?

sethladd:
Thank you very much for working on this!!

developing-packages.md
mit-mit: I totally agree for package *consumers*. But you re commenting on 'developing-packages.md'; I don't see how we can avoid knowledge of the differences for producers.

Look at 'using-packages.md', I just talk about packages there.

mit-mit: Whoops, sorry this was meant as a reply to your "overall taxanomy" comment up top

sethladd: as a package developer, it's still one kind of package. It can contain a few different things: dart library, native code, examples, LICENSE, etc. But it's one package format.

The details you talk about are important, I'm commenting on the phrasing. s/types of packages/kinds of things inside a package

Hixie: Please don't call it "native" code. There's nothing more native about Java than Dart on Android, they both compile down to machine code, and neither are what the kernel is written in.

developing-packages.md
wmleler: I would keep "Dart packages" here.

mit-mit: Yes, I will keep it. There are also java, swift, etc. packages that a plugin developer will be familiar with.

using-packages.md
kevmoo: https://github.com/dart-lang/pub-dartlang-dart/issues/39

developing-packages.md
mit-mit: Landed on a compromise: There are *Dart packages* and *Plugin packages*. We do need this distinction for *authors* as the steps for developing one are very different. We do _not_ need the distinction for *consumers* as the API of both types (and all packages in general) is Dart.

developing-packages.md
mit-mit: See above, agree consumers don't care. Producers need to care (and this is in `developing-packages.md`)

developing-packages.md
mit-mit: See above

developing-packages.md
mit-mit: Good point, removed

developing-packages.md
mit-mit: Removed

developing-packages.md
mit-mit: Rewritten

developing-packages.md
mit-mit: donr

developing-packages.md
mit-mit: Rewritten (don't need to tell them which specific tags to add, as the required fields are added by our template)

platform-plugins.md
mit-mit: Nice, thanks

using-packages.md
mit-mit: I think path is relevant, but agree that http is even more relevant!

using-packages.md
mit-mit: yup

using-packages.md
mit-mit: thanks

using-packages.md
mit-mit: Yup, I think we have enough now that we can remove this

using-packages.md
mit-mit: Oh, my bad, left over from an older version of this

using-packages.md
mit-mit: :-)

using-packages.md
mit-mit: Changed

using-packages.md
mit-mit: I couldn't find a short set of existing docs for this, so given the important of understanding decided a bit of duplication was OK. I do link to the Pub versioning guide for complete details.

using-packages.md
mit-mit: I think this help delivered that very message; if you are a consumer there is no difference.

mit-mit:
Thanks for all the great feedback! Uploaded a v2, please take a look.

developing-packages.md
```diff
@@ -0,0 +1,175 @@
...
+  version, author, etc.
+
+* A `lib` directory containing the public code in the package, minimally a
+  single `<package-name>.dart` file.
+
+For additional details, see the [Dart library package](https://www.dartlang.org/guides/libraries/create-library-packages) documentation.
+
+### Package types {#types}
+
+Flutter supports several kinds of packages:
```
sethladd: I think we're being confusing here. There is one type of package: a pub package (or, call it a Dart package). It can contain many different kinds of things: Dart code, plugin code, or a combination.

I agree that package developers might want to know the different kinds of things they can put in it.

If you're willing to change this phrasing to something like:

* Flutter uses pub for its package format and distribution.
* A pub (Dart) package can contain many things: Dart code, plugin code, example code, assets, and more.

Maybe it's as simple as "Dart packages can contain several kinds of code: Dart, plugin"

So, it's not about supporting different kinds of packages, it's that there are different kinds of contents.

developing-packages.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Step 2: Implement the package
+
+For pure Dart packages, simply add the functionality inside the main
+`lib/<package name>.dart` file, or in several files in the `lib` directory.
+
+To test the package, add [unit tests](https://flutter.io/testing/#unit-testing)
+in a `test` directory.
+
+## Developing plugin packages {#plugin}
```
sethladd: Then this would become "Developing packages with plugins"

developing-packages.md
mit-mit: I don't see the confusion; actually, I find what you are suggesting more confusing, and over-abstracted.

To use an analogy, the term "xml file" is commonly used rather than the more generic "text file containing text in xml format". It's just convenient to use more concrete terms.

mit-mit:
Yes, a plugin package is still a Dart package; in fact, it's a specialization of a Dart package. Made that clearer.

mit-mit:
TravisCI failures:

```
- ./_site/developing-packages/index.html
  *  External link https://github.com/flutter/website/blob/master/developing-packages.md failed: 404 No error
- ./_site/using-packages/index.html
  *  External link https://github.com/flutter/website/blob/master/using-packages.md failed: 404 No error
```

These are the auto-inserted links for the 'Edit Source' action -- those fail, as these new pages are not on GitHub yet, as the PR hasn't merged yet. Will have to update the Rakefile to skip these.

PR #617, opened 2017-06-28
Update .travis.yml to fix build failures

mit-mit:
Update nodejs version in an attempts to fix build failure


mit-mit:
Didn't work; trying a roll of Firebase tools (are are i minor versions behind)

mit-mit:
Not needed; issue was fixed in https://github.com/flutter/website/pull/622

PR #620, opened 2017-06-29
Other plugins support

NMateu:
Add bridging to GeneratedPluginRegistrant and register it in AppDelegate


NMateu:
I signed it!

mravn-google:
These changes are an improvement, but we should replace the whole section with use of `flutter create -i swift`

mit-mit:
Thanks for contributing @NMateu !

mit-mit:
@mravn-google tracking that issue in https://github.com/flutter/flutter/issues/11032

PR #631, opened 2017-07-12
Remove glyphicons from the CSS

filiph:
The `.glyphicons` class is never used — and if it was, it wouldn't work. The file (`fonts/glyphicons-halflings-regular.eot`) is missing, which leads to wasted CSS at best and a wasted HTTP request per pageview at worst. 

The site uses Font Awesome for icons.


sethladd:
lgtm

thank you!

PR #632, opened 2017-07-12
we're more reactive than functional-reactive

sethladd:


sethladd:
cc @filiph for review

PR #635, opened 2017-07-18
Fix caps and item length in nav items

mit-mit:
- Make the 'Building beautiful UIs - Codelab' item shorter so it doesn't wrap
- Make caps consistent


sethladd:
lgtm thanks!

PR #637, opened 2017-07-19
Update jquery to 3.x

mit-mit:
Fixes http://b/63830355


mit-mit:
I did some spot testing, and could not find any issues. Who knows exactly what we use jquery for so that we can do complete testing?

sethladd:
lgtm though I didn't exhaustively test

PR #638, opened 2017-07-21
Fix background color of operators

mit-mit:
Fixes https://github.com/flutter/flutter/issues/11335


mit-mit:
Issues appears fixed on staging server.

PR #641, opened 2017-07-26
Add a few zero-width spaces to allow line breaks

mit-mit:
Fixes https://github.com/flutter/flutter/issues/10153

The current line is so long it's causing weird scroll port issues.

screen shot 2017-07-26 at 13 16 45


sethladd:
lgtm

PR #646, opened 2017-08-02
Removing "functional" from "functional-reactive" term.

Sfshaza:
Staged: https://sz-flutter.firebaseapp.com/faq/

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -112,7 +112,7 @@ built with Flutter to users. We'd love to hear what you're building!
 
 * Heavily optimized, mobile-first 2D rendering engine with excellent support for
   text
-* Modern reactive-style framework
+* Modern reactive framework
 * Rich set of widgets for Android and iOS
 * APIs for unit and integration tests
 * Interop and plugin APIs to connect to the system and 3rd-party SDKs
--- a/faq.md
+++ b/faq.md
@@ -110,8 +110,9 @@ built with Flutter to users. We'd love to hear what you're building!
 
 ### What is inside the Flutter SDK?
 
-* Heavily optimized, mobile-first 2D rendering engine with excellent support for text
-* Reactive framework
+* Heavily optimized, mobile-first 2D rendering engine with excellent support for
+  text
+* Modern reactive-style framework
 * Rich set of widgets for Android and iOS
 * APIs for unit and integration tests
 * Interop and plugin APIs to connect to the system and 3rd-party SDKs
@@ -129,7 +130,7 @@ and one of the many editors that support [editing Dart](https://www.dartlang.org
 
 ### Does Flutter come with a framework?
 
-Yes! Flutter ships with a reactive style framework, inspired by React.
+Yes! Flutter ships with a modern framework, inspired by React.
 Flutter's framework is designed to be layered and customizable (and optional).
 Developers can choose to use only parts of the framework, or a different
 framework.
@@ -232,11 +233,9 @@ The primary criteria we used to pick a programming language were the following:
   performance and delivers predictable performance, without periodic
   pauses that would cause dropped frames.
 
-* _Fast allocation_. The Flutter framework uses a reactive style
-  programming model, whose performance depends heavily on the underlying memory
-  allocator efficiently handling small, short-lived allocations. The
-  reactive style was developed in languages with this property and
-  does not work efficiently in languages that lack this facility.
+* _Fast allocation_. The Flutter framework uses a functional-style flow
+  that depends heavily on the underlying memory
+  allocator efficiently handling small, short-lived allocations.
 
 Dart scores highly on all four dimensions. In addition, we have the opportunity
 to work closely with the Dart community, which is actively investing resources
--- a/faq.md
+++ b/faq.md
@@ -111,7 +111,7 @@ built with Flutter to users. We'd love to hear what you're building!
 ### What is inside the Flutter SDK?
 
 * Heavily optimized, mobile-first 2D rendering engine with excellent support for text
-* Functional-reactive framework
+* Reactive framework
 * Rich set of widgets for Android and iOS
 * APIs for unit and integration tests
 * Interop and plugin APIs to connect to the system and 3rd-party SDKs
@@ -129,9 +129,8 @@ and one of the many editors that support [editing Dart](https://www.dartlang.org
 
 ### Does Flutter come with a framework?
 
-Yes! Flutter ships with a functional-reactive style framework, inspired by
-React. Flutter's framework is designed to be layered and customizable
-(and optional).
+Yes! Flutter ships with a reactive style framework, inspired by React.
+Flutter's framework is designed to be layered and customizable (and optional).
 Developers can choose to use only parts of the framework, or a different
 framework.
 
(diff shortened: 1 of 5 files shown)


Hixie:
I'm assuming that is the wrong bug link, and that you meant to link to https://github.com/flutter/flutter/issues/11474.

faq.md
```diff
@@ -111,7 +111,7 @@ built with Flutter to users. We'd love to hear what you're building!
 ### What is inside the Flutter SDK?
 
 * Heavily optimized, mobile-first 2D rendering engine with excellent support for text
-* Functional-reactive framework
+* Reactive framework
```
Hixie: We're no more reactive than functional, so I disagree that this is a correct change.

wmleler: Modern, reactive-style framework

faq.md
```diff
@@ -129,9 +129,8 @@ and one of the many editors that support [editing Dart](https://www.dartlang.org
 
 ### Does Flutter come with a framework?
 
-Yes! Flutter ships with a functional-reactive style framework, inspired by
-React. Flutter's framework is designed to be layered and customizable
-(and optional).
+Yes! Flutter ships with a reactive style framework, inspired by React.
```
wmleler: modern reactive-style

faq.md
Hixie: I think it'd be fine to remove the label "reactive" as well and just say something like "Modern UI framework"

faq.md
Hixie: How about "Flutter ships with a modern framework, inspired by React"?

technical-overview.md
```diff
@@ -35,19 +35,20 @@ What are some advantages of Flutter? It helps you:
...
-of OEM widget sets
+    *   Benefit from a rich set of Material Design and Cupertino (iOS-flavor)
+        widgets built using Flutter's own framework
+    *   Realize custom, beautiful, brand-driven designs, without the
+        limitations of OEM widget sets
 
 ## Core principles
 
-Flutter includes a functional-reactive framework, a 2D rendering engine, ready-made
+Flutter includes a reactive framework, a 2D rendering engine, ready-made
```
wmleler: modern reactive-style

widgets-intro.md
```diff
@@ -9,7 +9,7 @@ permalink: /widgets-intro/
 
 ## Introduction
 
-Flutter widgets are built using a functional-reactive framework, which takes
+Flutter widgets are built using a reactive framework, which takes
```
wmleler: modern reactive-style

faq.md
```diff
@@ -233,10 +232,10 @@ The primary criteria we used to pick a programming language were the following:
   performance and delivers predictable performance, without periodic
   pauses that would cause dropped frames.
 
-* _Fast allocation_. The Flutter framework uses a functional-reactive style
+* _Fast allocation_. The Flutter framework uses a reactive style
```
Hixie: It's actually the functional part that's more relevant here, not the reactive part. The reactive programming model doesn't necessarily imply much memory allocation. What's relevant here is that we use a functional-style flow where many immutable objects get allocated to describe the scene, then the previous scene's objects get released.

faq.md
```diff
@@ -233,10 +232,10 @@ The primary criteria we used to pick a programming language were the following:
...
   programming model, whose performance depends heavily on the underlying memory
   allocator efficiently handling small, short-lived allocations. The
-  functional-reactive style was developed in languages with this property and
+  reactive style was developed in languages with this property and
```
Hixie: Is this statement still accurate?

The "functional-reactive" style was developed in different languages than the "reactive" style. I would be wary of changing this statement without verifying the accuracy of the new claim.

Hixie:
Our framework is not "reactive-style". We don't use streams at all in the framework itself, for instance. The framework's pipeline is not asynchronous. The primary data flow doesn't use events.

faq.md
```diff
@@ -233,11 +233,9 @@ The primary criteria we used to pick a programming language were the following:
...
   pauses that would cause dropped frames.
 
-* _Fast allocation_. The Flutter framework uses a functional-reactive style
-  programming model, whose performance depends heavily on the underlying memory
-  allocator efficiently handling small, short-lived allocations. The
-  functional-reactive style was developed in languages with this property and
-  does not work efficiently in languages that lack this facility.
+* _Fast allocation_. The Flutter framework uses a functional-style flow
+  that depends heavily on the underlying memory
+  allocator efficiently handling small, short-lived allocations.
```
Hixie: I'd like to keep the second sentence if we can. It makes it clearer why this is something Dart offers but that, say, Java does not.

Hixie:
The update is better, but I think we should really avoid the term "reactive-style". It's even more inaccurate than "functional reactive", IMHO.

What we really need is a term for the specific paradigm that we actually most prominently use, the whole build function thing, which I think is what most people are getting at with "functional-reactive" when they use it loosely. I really can't find a term that is used to describe this. None of the React marketing materials seem to talk about the style of programming used in the framework.

faq.md
Sfshaza: @Hixie , who could validate that second sentence? Kevmoo? Someone else on the Dart team?

Hixie: Well if we just leave it the way it was then I don't think we need to particularly change anything. :-)

@abarth is the one who originally wrote this sentence so he might have thoughts on how to rephrase it if we want to remove mention of "functional-reactive" but keep it accurate.

abarth: You could just say "This style was developed ..."

faq.md
```diff
@@ -110,8 +110,9 @@ built with Flutter to users. We'd love to hear what you're building!
 
 ### What is inside the Flutter SDK?
 
-* Heavily optimized, mobile-first 2D rendering engine with excellent support for text
-* Functional-reactive framework
+* Heavily optimized, mobile-first 2D rendering engine with excellent support for
+  text
+* Modern reactive framework
```
Hixie: How about "Modern react-style framework"?

Though personally I still think we should say "functional-reactive". I don't really understand what's wrong with the term.

faq.md
Hixie: I think the original text is fine, but if we have to change it, I would recommend something like:

"_Fast allocation_. The Flutter framework uses a react-style programming model, whose performance depends heavily on the underlying memory allocator efficiently handling small, short-lived allocations. This style was developed in languages with this property and does not work efficiently in languages that lack this facility."

technical-overview.md
```diff
@@ -35,21 +35,22 @@ What are some advantages of Flutter? It helps you:
...
+        widgets built using Flutter's own framework
+    *   Realize custom, beautiful, brand-driven designs, without the
+        limitations of OEM widget sets
 
 ## Core principles
 
-Flutter includes a functional-reactive framework, a 2D rendering engine, ready-made
-widgets, and development tools. These components work together to help you
-design, build, test, and debug apps. Everything is organized around a few core
+Flutter includes a modern modern reactive framework, a 2D rendering engine,
```
Hixie: replace "modern modern reactive" with "react-style" or "functional-reactive". (I'd prefer the second.)

widgets-intro.md
```diff
@@ -9,7 +9,7 @@ permalink: /widgets-intro/
 
 ## Introduction
 
-Flutter widgets are built using a functional-reactive framework, which takes
+Flutter widgets are built using a modern reactive framework, which takes
```
Hixie: Similarly, "modern react-style framework" or "functional-reactive framework".

Hixie:
I'm less and less convinced that this change is an improvement.

faq.md
Sfshaza: OK, I will change it to this. Though I ran the original text past @a-siva and he suggested removing it.

Hixie:
I would prefer not making the change at all, but I can live with this version.
I'll let someone who actually wants the change review it.

Sfshaza:
Wm just lgtm'd it again (after recent changes). Thanks for all feedback.

PR #648, opened 2017-08-10
Add website instructions for changing icons and launch screens

xster:
https://github.com/flutter/flutter/issues/8147


assets-and-images.md
```diff
@@ -206,3 +206,48 @@ level classes, like
 or
 [`ImageCache`](https://docs.flutter.io/flutter/services/ImageCache-class.html),
 you'll also notice parameters related to scale.)
+
+## Platform assets
+
+There will also be occasions to work with assets in the platform projects directly. Below are 2 common cases where assets are used before the Flutter framework is loaded and running.
```
Hixie: s/2/two/

assets-and-images.md
```diff
@@ -206,3 +206,48 @@ level classes, like
...
+
+![Launch icon](/images/assets-and-images/icon.png)
+
+#### Android
+
+In your Flutter project's root directory, navigate to `.../android/app/src/main/res`. The various bitmap resource folders such as `mipmap-hdpi` already contain placeholder images named `ic_launcher.png`. Simply replace them with your desired assets respecting the recommended icon size per screen density as indicated by the [Android Developer Guide](https://developer.android.com/guide/practices/ui_guidelines/icon_design_launcher.html#size).
+
+![Android icon location](/images/assets-and-images/android-icon-path.png)
+
+Note: if you rename the .png files, you must also match the new name in your `AndroidManifest.xml`'s `<application>` tag's `android:icon` attribute.
```
Hixie: Remove "Note:", it doesn't add anything (see style guide).

assets-and-images.md
```diff
@@ -206,3 +206,48 @@ level classes, like
...
+
+## Updating the launch screen
+
+<p align="center">
+  <img src="images/assets-and-images/launch-screen.png" alt="Launch screen" />
+</p>
+
+Flutter also uses native platform mechanisms to draw transitional launch screens to your Flutter app while the Flutter framework loads. This launch screen will persist until Flutter renders the first frame of your application.
+
+This also implies if you don't call [runApp](https://docs.flutter.io/flutter/widgets/runApp.html) in the `void main()` function of your app, the launch screen will persist forever.
```
Hixie: Maybe after the reference to runApp put in parentheses `(or more specifically, if you don't call [window.render] in response to [window.onDrawFrame] being invoked)`, since you don't actually have to call runApp per se.

Hixie: Also, @HansMuller, this maybe what we should do for i18n. Just never actually call render for the scenes we don't want to put up.

Hixie:
This is fantastic. 
LGTM

assets-and-images.md
xster: done

assets-and-images.md
xster: I actually wanted to achieve the effects found in some of the other pages, i.e. users don't have to read this if they're just following along. Added that <aside> box.

assets-and-images.md
xster: Is it too advanced for readers? Maybe in another <aside> box?

xster:
Will submit after https://github.com/flutter/flutter/pull/11505 which is pending https://github.com/flutter/engine/pull/3956

assets-and-images.md
Hixie: I trust our readers to understand the complicated stuff. It's the best way for them to get comfortable with it. We don't want people thinking the framework is magical.

xster: Added

assets-and-images.md
```diff
@@ -206,3 +206,52 @@ level classes, like
...
+
+![Android icon location](/images/assets-and-images/android-icon-path.png)
+
+<aside class="alert alert-info" markdown="1">
+**Note:** if you rename the .png files, you must also match the new name in your `AndroidManifest.xml`'s `<application>` tag's `android:icon` attribute.
+</aside>
+
+#### iOS
+
+In your Flutter project's root directory, navigate to `.../ios/Runner/Assets.xcassets/AppIcon.appiconset`. The directory already contains placeholder images. Simply replace them with the appropriately sized images as indicated by their filename as dictated by the Apple [Human Interface Guidelines](https://developer.apple.com/ios/human-interface-guidelines/graphics/app-icon). Keep the original file names.
```
mit-mit: Due to https://github.com/flutter/flutter/issues/10153, long backtick strings like this will cause the viewport to no longer fit on mobile devices:

screen shot 2017-08-14 at 14 29 14


I've found the easiest workaround for now to be to rewrite sentences like:

    navigate to `.../ios/Runner/Assets.xcassets/AppIcon.appiconset`.

 to:

    navigate to `AppIcon.appiconset` located in .../ios/Runner/Assets.xcassets/.

xster: Thanks for the catch @mit-mit! It's an unfortunate issue. Wonder if we should just generally make monospace fonts smaller for small viewports

PR #651, opened 2017-08-14
What programming paradigm does Flutter's framework use?

Hixie:
This keeps coming up as a topic of conversation, so it seems a FAQ entry may be worthy.

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -499,6 +499,81 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
 profile or release mode using the menu entries **Run > Flutter Run in Profile Mode**
 or **Release Mode**.
 
+### What programming paradigm does Flutter's framework use?
+
+Flutter is a multiparadigm programming environment. Many programming techniques
+developed over the past few decades are used where they are appropriate, to solve
+problems in a clean way. In no particular order:
+
+* Composition: The primary paradigm used by Flutter is that of using small objects with narrow
+scopes of behavior, composed together to obtain more complicated effects. Most widgets
+in the Flutter widget library are built in this way.
+
+* Functional programming: Entire applications can be built with only
+[StatelessWidget](https://docs.flutter.io/flutter/widgets/StatelessWidget-class.html)s,
+which are essentially functions that describe how arguments map to other functions,
+bottoming out in primitives that compute layouts or paint graphics.
+Additionally, heavy use is made of immutable data structures, including the entire
+[Widget](https://docs.flutter.io/flutter/widgets/Widget-class.html)
+class hierarchy as well as numerous supporting classes such as
+[Rect](https://docs.flutter.io/flutter/dart-ui/Rect-class.html) and
+[TextStyle](https://docs.flutter.io/flutter/painting/TextStyle-class.html).
+On a smaller scale, Dart's
+[Iterable](https://docs.flutter.io/flutter/dart-core/Iterable-class.html) API,
+which makes heavy use of the functional style (map, reduce, where, etc), is
+frequently used to process lists of values in the framework.
+
+* Event-driven programming: User interactions are represented by event objects that are
+dispatched to callbacks registered with event handlers. Screen updates are triggered
+by a similar callback mechanism. The [Listenable](https://docs.flutter.io/flutter/foundation/Listenable-class.html)
+class, which is used as the basis of the animation system, formalises a
+subscription model for events with multiple listeners.
+
+* Class-based object-orientated programming: Most of the APIs of the framework are built using classes
+with inheritance. We use an approach whereby we define very high-level APIs in our base classes, then
+specialise them iteratively in subclasses. For example, our render objects have a base class
+([RenderObject](https://docs.flutter.io/flutter/rendering/RenderObject-class.html))
+that is agnostic regarding the coordinate system, and then we have a subclass
+([RenderBox](https://docs.flutter.io/flutter/rendering/RenderBox-class.html))
+that introduces the opinion that the geometry should be based on the Cartesian coordinate system (x/width and y/height).
+
+* Prototype-based object-orientated programming: The
+[ScrollPhysics](https://docs.flutter.io/flutter/widgets/ScrollPhysics-class.html) class
+chains instances to compose the physics that apply to scrolling dynamically at runtime. This
+lets the system compose, for example, paging physics with platform-specific physics, without
+the platform having to be selected at compile time.
+
+* Imperative programming: Straight forward imperative programming, usually paired with state encapsulated
+within an object, is used where it provides the most intuitive solution. For example, tests are
+written in an imperative style, first describing the situation under test, then listing the invariants
+that the test must match, then advancing the clock or inserting events as necessary for the test.
+
+* Reactive programming: The widget and element trees are sometimes described as reactive, since new inputs 
+provided in a widget's constructor are immediately propagated as changes to lower-level widgets by
+the widget's build method, and changes made in the lower widgets (e.g. in response to user input) propagate
+back up the tree via event handlers. Aspects of both functional-reactive and imperative-reactive are
+present in the framework.
+
+* Declarative programming: The build functions of widgets are typically a single expression with multiple levels
+of nested constructors, very much a declarative approach to building UI. This can also be combined with
+the imperative style to build UIs that would be harder to describe in a pure-declarative approach.
+
+* Generic programming: Types can be used to help catch programming errors early. The Flutter framework
+uses generic programming to help in this regard. For example, the
+[State](https://docs.flutter.io/flutter/widgets/State-class.html) class is parameterized in terms
+of the type of its associated widget, so that it 
+
+* Concurrent programming: Flutter makes heavy use of
+[Future](https://docs.flutter.io/flutter/dart-async/Future-class.html)s and other asynchronous APIs.
+For example, the animation system reports when an animation is finished by completing a future. The
+image loading system similarly uses futures to report when a load is complete.
+
+* Constraint programming: The layout system in Flutter uses a weak form of constraint programming to
+determine the geometry of a scene. Constraints (e.g. for cartesian boxes, a minimum and maximum width
+and a minimum and maximum height) are passed from parent to child, and the child selects a resulting
+geometry (e.g. for cartesian boxes, a size, specifically a width and a height) that fulfills those
+constraints. By using this technique, Flutter can usually lay out a whole scene with a single pass.
+
 ## Project
 
 ### Where can I get support?
--- a/faq.md
+++ b/faq.md
@@ -478,7 +478,7 @@ We have also found that our "code-first" better allows for features like
 hot reload and dynamic environment adaptations.
 
 It is possible to create a custom language that is then
-converted to widgets on the fly. Since build methods are "just code", they
+converted to widgets on the fly. Because build methods are "just code", they
 can do anything, including interpreting markup and turning it into widgets.
 
 ### My app has a Slow Mode banner/ribbon in the upper right. Why am I seeing that?
@@ -501,9 +501,9 @@ or **Release Mode**.
 
 ### What programming paradigm does Flutter's framework use?
 
-Flutter is a multiparadigm programming environment. Many programming techniques
-developed over the past few decades are used where they are appropriate, to solve
-problems in a clean way. In no particular order:
+Flutter is a multi-paradigm programming environment. Many programming techniques
+developed over the past few decades are used in Flutter, each to solve
+a particular set of problems in a clean way. In no particular order:
 
 * Composition: The primary paradigm used by Flutter is that of using small objects with narrow
 scopes of behavior, composed together to obtain more complicated effects. Most widgets
@@ -548,20 +548,29 @@ within an object, is used where it provides the most intuitive solution. For exa
 written in an imperative style, first describing the situation under test, then listing the invariants
 that the test must match, then advancing the clock or inserting events as necessary for the test.
 
-* Reactive programming: The widget and element trees are sometimes described as reactive, since new inputs 
+* Reactive programming: The widget and element trees are sometimes described as reactive, because new inputs 
 provided in a widget's constructor are immediately propagated as changes to lower-level widgets by
 the widget's build method, and changes made in the lower widgets (e.g. in response to user input) propagate
 back up the tree via event handlers. Aspects of both functional-reactive and imperative-reactive are
 present in the framework.
 
 * Declarative programming: The build functions of widgets are typically a single expression with multiple levels
-of nested constructors, very much a declarative approach to building UI. This can also be combined with
+of nested constructors, very much a declarative approach to building UI (such nested expressions could be mechanically
+transformed to or from any suitably expressive markup language). This can also be combined with
 the imperative style to build UIs that would be harder to describe in a pure-declarative approach.
 
 * Generic programming: Types can be used to help catch programming errors early. The Flutter framework
 uses generic programming to help in this regard. For example, the
 [State](https://docs.flutter.io/flutter/widgets/State-class.html) class is parameterized in terms
-of the type of its associated widget, so that it 
+of the type of its associated widget, so that the Dart analyzer can catch mismatches of states and widgets.
+Similarly, the [GlobalKey](https://docs.flutter.io/flutter/widgets/GlobalKey-class.html) class takes a type
+parameter so that it can access a remote widget's state in a type-safe manner (with runtime checking),
+the [Route](https://docs.flutter.io/flutter/widgets/Route-class.html) interface is parameterized with the
+type that it is expected to use when [popped](https://docs.flutter.io/flutter/widgets/Navigator/pop.html),
+and collections such as [List](https://docs.flutter.io/flutter/dart-core/List-class.html)s,
+[Map](https://docs.flutter.io/flutter/dart-core/Map-class.html)s, and
+[Set](https://docs.flutter.io/flutter/dart-core/Set-class.html)s are all parameterized so that mismatched
+elements can be caught early either by the analyzer or at runtime during debugging.
 
 * Concurrent programming: Flutter makes heavy use of
 [Future](https://docs.flutter.io/flutter/dart-async/Future-class.html)s and other asynchronous APIs.
--- a/faq.md
+++ b/faq.md
@@ -552,10 +552,10 @@ that the test must match, then advancing the clock or inserting events as necess
 provided in a widget's constructor are immediately propagated as changes to lower-level widgets by
 the widget's build method, and changes made in the lower widgets (e.g. in response to user input) propagate
 back up the tree via event handlers. Aspects of both functional-reactive and imperative-reactive are
-present in the framework.
+present in the framework. 
 
 * Declarative programming: The build functions of widgets are typically a single expression with multiple levels
-of nested constructors, very much a declarative approach to building UI (such nested expressions could be mechanically
+of nested constructors, written using a strictly declarative subset of Dart (such nested expressions could be mechanically
 transformed to or from any suitably expressive markup language). This can also be combined with
 the imperative style to build UIs that would be harder to describe in a pure-declarative approach.
 


Hixie:
cc @wmleler for opinions and review.

faq.md
```diff
@@ -499,6 +499,81 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
 profile or release mode using the menu entries **Run > Flutter Run in Profile Mode**
 or **Release Mode**.
 
+### What programming paradigm does Flutter's framework use?
+
+Flutter is a multiparadigm programming environment. Many programming techniques
```
sethladd: multi-paradigm ?

faq.md
```diff
@@ -499,6 +499,81 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+developed over the past few decades are used where they are appropriate, to solve
```
sethladd: s/where they are appropriate/in Flutter

sethladd:
I definitely want to see the talk: "The 10 programming paradigms used in Flutter"

faq.md
```diff
@@ -499,6 +499,81 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+[Rect](https://docs.flutter.io/flutter/dart-ui/Rect-class.html) and
+[TextStyle](https://docs.flutter.io/flutter/painting/TextStyle-class.html).
+On a smaller scale, Dart's
+[Iterable](https://docs.flutter.io/flutter/dart-core/Iterable-class.html) API,
+which makes heavy use of the functional style (map, reduce, where, etc), is
+frequently used to process lists of values in the framework.
+
+* Event-driven programming: User interactions are represented by event objects that are
+dispatched to callbacks registered with event handlers. Screen updates are triggered
+by a similar callback mechanism. The [Listenable](https://docs.flutter.io/flutter/foundation/Listenable-class.html)
```
wmleler: I would be much happier if it used some kind of Observable interface, so the callbacks did not have to be called explicitly. This introduces a dependency that can get in the way of many optimizations.

faq.md
```diff
@@ -499,6 +499,81 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+lets the system compose, for example, paging physics with platform-specific physics, without
+the platform having to be selected at compile time.
+
+* Imperative programming: Straight forward imperative programming, usually paired with state encapsulated
+within an object, is used where it provides the most intuitive solution. For example, tests are
+written in an imperative style, first describing the situation under test, then listing the invariants
+that the test must match, then advancing the clock or inserting events as necessary for the test.
+
+* Reactive programming: The widget and element trees are sometimes described as reactive, since new inputs 
+provided in a widget's constructor are immediately propagated as changes to lower-level widgets by
```
wmleler: s/since/because

faq.md
```diff
@@ -499,6 +499,81 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+the widget's build method, and changes made in the lower widgets (e.g. in response to user input) propagate
+back up the tree via event handlers. Aspects of both functional-reactive and imperative-reactive are
+present in the framework.
+
+* Declarative programming: The build functions of widgets are typically a single expression with multiple levels
```
wmleler: This is the only point that I might take issue with. I don't really see how "build functions of a single expression with multiple levels of nested constructors" has anything to do with declarative programming (much less "very much").

faq.md
```diff
@@ -499,6 +499,81 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+of nested constructors, very much a declarative approach to building UI. This can also be combined with
+the imperative style to build UIs that would be harder to describe in a pure-declarative approach.
+
+* Generic programming: Types can be used to help catch programming errors early. The Flutter framework
+uses generic programming to help in this regard. For example, the
+[State](https://docs.flutter.io/flutter/widgets/State-class.html) class is parameterized in terms
+of the type of its associated widget, so that it
```
branflake2267: ending?

faq.md
```diff
@@ -499,6 +499,81 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+chains instances to compose the physics that apply to scrolling dynamically at runtime. This
+lets the system compose, for example, paging physics with platform-specific physics, without
+the platform having to be selected at compile time.
+
+* Imperative programming: Straight forward imperative programming, usually paired with state encapsulated
+within an object, is used where it provides the most intuitive solution. For example, tests are
+written in an imperative style, first describing the situation under test, then listing the invariants
+that the test must match, then advancing the clock or inserting events as necessary for the test.
+
+* Reactive programming: The widget and element trees are sometimes described as reactive, since new inputs
```
branflake2267: Would a reference to StatefulWidget fit in this section? Do you think it would be nice to counter the Stateless Widget claims above?

faq.md
Hixie: Can you elaborate on the difference between Listenable and Observable?

faq.md
Hixie: done

faq.md
Hixie: tweaked the wording to: "Many programming techniques developed over the past few decades are used in Flutter, each to solve a particular set of problems in a clean way."

faq.md
Hixie: How would you say that StatefulWidget fit into the reactive programming model?

faq.md
Hixie: done

faq.md
Hixie: Isn't that exactly what declarative programming is? Nested expressions?

Hixie: I've added "(such nested expressions could be mechanically transformed to or from any suitably expressive markup language)", not sure how else to clarify this.

faq.md
Hixie: oops! Added: ...so that the Dart analyzer can catch mismatches of states and widgets. Similarly, the [GlobalKey](https://docs.flutter.io/flutter/widgets/GlobalKey-class.html) class takes a type parameter so that it can access a remote widget's state in a type-safe manner (with runtime checking), the [Route](https://docs.flutter.io/flutter/widgets/Route-class.html) interface is parameterized with the type that it is expected to use when [popped](https://docs.flutter.io/flutter/widgets/Navigator/pop.html), and collections such as [List](https://docs.flutter.io/flutter/dart-core/List-class.html)s, [Map](https://docs.flutter.io/flutter/dart-core/Map-class.html)s, and [Set](https://docs.flutter.io/flutter/dart-core/Set-class.html)s are all parameterized so that mismatched elements can be caught early either by the analyzer or at runtime during debugging.

Hixie:
Updated per comments, PTAL.

wmleler:
Nope! That's more like functional programming (assuming you don't have any
side effects).
For example, Prolog (if you removed the "cut" operator) was all rules.
The language I created was completely declarative, and it was all rules as
well.

There are other declarative languages that are not rule based.
I've mentioned Lucid to you in the past that was both declarative and
functional,
and could be used to build programs that had "state" (although you keep
denying that that is possible).

Anyway, the next time I'm in MTV I would love to have a big discussion with
you about declarative languages.
It is something I spent a lot of time on, and something that still informs
what I do (for example, CSS is a
declarative language, although it isn't much of a language -- not even
close to Turing complete).

wmleler:
Would depend on whether those expressions had side effects. As long as you
specify that they have to have referential transparency, you are probably
ok.

faq.md
wmleler: Yes, but I don't have time right now.

faq.md
wmleler: Your "nested expressions" contain constructors, which by definition have side effects, so you definitely do not have referential transparency. You would have to jump through some hoops in order to make this declarative. For example, memoizing the results so even multiple calls return the exact same constructed object. Maybe other things too. I would have to spend some time looking at it.

faq.md
Hixie: Ok. When you have a moment, file a bug or send me an e-mail or something. In the meantime I'm disregarding this comment this it's not actionable.

faq.md
Hixie: Do you agree that XAML is declarative?

wmleler:
If it has referential transparency, then yes. And even if it is, that is
not the same thing as what you are claiming about Flutter.

The problem is that there is no guarantee that a constructor does not have
a side effect in Flutter. Even something as innocent as reference counting
could become exposed and could destroy declarative semantics.

Hixie:
Constructors for widgets can't have side-effects, because the entire widget hierarchy is declared as `@immutable`.

faq.md
branflake2267: Hm, maybe it doesn't fit here. But I was thinking that the mutable StatefulWidget reacts to the state and/or inputs. Or where the StatefulWidget reacts to the change in the variables over time, seemingly reactive to me.

wmleler:
I'm not sure that @immutable on constructors will be good enough. I would think that @immutable would only keep the object from being mutated after it is constructed. That would not preserve referential transparency -- which would require that calling the constructor with the same arguments would return exactly the same object.  Of course, I'm not totally up to speed on Dart.

Hixie:
Sorry, my bad. I meant `const`, not `@immutable`.

Hixie:
@wmleler I went to try to clarify the "declarative" section some more, but I really don't see anything in it that's wrong. What we do is exactly like what people do with HTML (which I hope we agree is declarative). Creating HTML elements can have side-effects (e.g. mutating the DOM tree, kicking off network loads), but that doesn't make it imperative. How does what Flutter do in its widget build functions differ from that?

wmleler:
HTML is not declarative, and I don't know anyone who claims it is.

branflake2267:
@wmleler why do you think HTML is imperative?

wmleler:
I didn't say it was imperative. I said it is not declarative. They are not
antonyms.

Without script tags or embedded code, HTML isn't even a programming
language, so arguing whether it is declarative or not is pretty much a red
herring. With script tags or embedded code, it is definitely not
declarative.

Hixie:
[Wikipedia page on Declarative programming](https://en.wikipedia.org/wiki/Declarative_programming): "Many markup languages such as HTML, MXML, XAML, XSLT or other user-interface markup languages are often declarative. HTML, for example, only describes what should appear on a webpage - it specifies neither control flow rendering a page nor its possible interactions with a user."

[Quora question "Is HTML declarative or imperative or neither? Why?"](https://www.quora.com/Is-HTML-declarative-or-imperative-or-neither-Why): "HTML is a declarative language" [...] "It's a declarative domain-specific markup language."

Me (HTML spec editor for 10 years): HTML is declarative.

HTML is a markup language for creating DOM elements (what in Flutter we would call elements). In exactly the same way that a Widget tree is inflated into Flutter elements, HTML tags are inflated into DOM elements. I really think there's a very close parallel here.

I would agree that if you took the subset of Dart that lets you build widget trees as a single expression, that wouldn't count as a programming language either.

branflake2267:
@wmleler I was asking to see where that would go. Because I thought HTML was declarative in nature. I was looking to see if there was something I haven't considered. Keeping my mind open...

If I were to dig into the meaning of programming and languages, I look at programming as an action, a verb which uses languages to define a computational process. While HTML could be considered document markup, it appears to be considered yet another language you can speak in terms of writing a process for a computer to do some function(s). Although it's not imperative in nature it specifies some output. So I would think theoretically speaking, programming in general includes all languages spoken to a computer, no matter the language type. Maybe this is a bit pedantic here. 

In other words, from my perspective then, it all looks like programming to me, using a variety of languages to get the job done.

Hixie:
This is in the context of a FAQ entry that talks about a variety of paradigms that we use in narrowly scoped parts of the framework. I would agree that Flutter isn't declarative in general. This section is just talking about when you describe widgets in single-expression build methods, which is essentially a declarative programming approach. You specify widgets just like a markup language would, you specify constraints in a declarative fashion (arguments), you specify the relationships (e.g. `.of` methods). This seems very much in line with at least Wikipedia's definition of "declarative programming", and so seems similar to the other entries in the proposed text.

wmleler:
Whatever you want. My knowledge of what is declarative programming only
includes, you know, actual programming languages. I guess I was fooled by
you titling this "What programming paradigm does Flutter's framework use?"

Hixie:
@wmleler I definitely want to make sure you're on board (our [design principles](https://flutter.io/design-principles/#conflict-resolution) talk about avoiding agreement by attrition or agreement by compromise).

Do you have any pointers for what you would consider a declarative programming language so I can educate myself on the topic and more carefully consider how this applies in the context of build methods? (All I know on the topic so far is what Wikipedia says, but Wikipedia isn't always a good guide, and maybe it's wrong on this topic.)

branflake2267:
@wmleler How do you define programming? I'm guessing you're matching that with only can write logical instructions. And then that would have to match with, programing is imperative only. And then declarative markup doesn't fit that model. Do you think the meaning programming could be too finite in your case? 

I like the title for what it's worth. I'm butting in from the outside of your org to say the least. :)

wmleler:
To me, a programming language is Turing complete. Pretty simple definition,
and it is the commonly accepted definition.

If you allow HTML to be called a programming language, then it is a quick
trip down a very slippery slope to saying that any text input to a computer
is a programming language. If HTML is a programming language, then is JSON?
After all, I can convert any JSON into HTML and vice-versa.

My dissertation was on a class of declarative languages called "constraint
languages". My contribution was the first constraint language that was
Turing complete (and it was still purely declarative).

The universe of Turing-complete languages is pretty large without allowing
arbitrary data formats, including HTML. I'd be OK with calling HTML a
"language", just not a programming language. You can't really write
programs with it, you can only describe a data structure.

On Thu, Aug 17, 2017 at 2:19 PM, Brandon Donnelson <notifications@github.com

Hixie:
@wmleler That makes sense. What would you call the programming style where you explicitly avoid Turing-complete expressiveness (and indeed any control flow at all), and instead limit yourself to a subset of an otherwise Turing-complete language that is sufficient for describing a tree structure that can then be processed to generate an object tree? The goal of which being to make it possible to reason about the structure (something which is famously not possible with a Turing-complete language), for example for IDEs, structured search-and-replace, pattern analysis, etc.

wmleler:
If you indeed avoid any control flow or side effects, I would call it
"pretty cool" :-)  Actually, I think the closest term might be a
declarative DSL (domain specific language). Of course, you actually could
embed a real declarative DSL, like React Native does with JSX.

Oh, there are plenty of Turing-complete languages where you can reason
about the structure -- Prolog, pure Scheme, among others. Maybe not the way
you want, though.

Hixie:
I don't mean a DSL, per se, just a subset of an existing language. For example, in Dart one could say "the subset that can describe const expressions". That's definitely not a DSL, but it's also not a Turing complete programming language; it has no control flow.

Avoiding control flow is easy (it's what we do for build methods that just return a nested expression). Avoiding side-effects is either impossible or easy depending on how you define side-effects. (If you include allocation, it's pretty much impossible even for non-programming markup languages or indeed even plain text files; if you don't include allocation, then again it's what we do for build methods that just return a nested expression.)

Not sure how you'd be able to reason about something like pure Scheme any more than pure Dart. If it's Turing-complete, you're more or less SOL in terms of doing anything beyond static analysis, because of the halting problem. (If you just mean static analysis, then it's not clear why you'd need something like Scheme to do it; you can do it with Dart too.) I'm specifically talking about a subset of Dart (or other languages) that doesn't have control flow and isn't Turing complete, so that you can treat it as equivalent to a markup language (literally mechanically interchangeable).

Anyway, I thought the term for this was "declarative programming" (based on [the Wikipedia definition of the term](https://en.wikipedia.org/wiki/Declarative_programming)). I've also heard the terms "markup language" and "data description language" used for this but I'm not sure that's as accurate (again based on the Wikipedia definitions). What term should we use in the FAQ to refer to this programming paradigm? We use it heavily in Flutter, and I feel it's a useful one to mention, as it's our alternative to things like HTML, Android XML, XAML, JSX, or Nib files.

Hixie:
@wmleler ping

wmleler:
That's a good question. If you want I can ask some of my language friends
what they would call it.

It is a DSL -- a language for describing certain kinds of tree structures,
but there should be a better and more descriptive name.

In many ways, it is equivalent to a data-interchange format, like JSON or
JSX, but with the bonus that it is a subset of a real programming language
(like JSON). Although it can call constructors, which can complicate things.

Hixie:
If it's the "DS" part of "DSL", what would the domain be? Is XML a DSL? It's also "a language for describing certain kinds of tree structures". I think it's pretty established that XML itself isn't a DSL, would you not agree? (It's a metalanguage that can be used to describe DSL vocabularies.)

I would have called this "declarative". It's literally "a style of building the structure and elements of computer programs—that expresses the logic of a computation without describing its control flow", which is Wikipedia's definition of "declarative programming". Wikipedia also gives the alternative definition of "any style of programming that is not imperative", which seems to pretty firmly describe expressions.

I would be curious to hear about what other people would call this style, yes.

wmleler:
What "logic of a computation" does it describe?  That's the main point (and
defining characteristic) of a declarative language. It defines a
computation (without any prescription for its control flow). Your subset of
Dart definitely prescribes control flow if it is calling constructors. And
what it is describing is a data structure, not any "logic of a computation".

Hixie:
It describes the app UI: the logic of computation is figuring out the layout, the painting, the handling of events, how to update stateful widgets cheaply, how to update widgets that depend on ambient state efficiently, and so on. Basically, all the logic in widgets/framework.dart, in rendering/*, etc.

branflake2267:
I think Google's definition for the word 'declarative' in computing pins what Flutter's framework does quite well. The definition, "denoting high-level programming languages that can be used to solve problems without requiring the programmer to specify an exact procedure to be followed." 

So as a engineer, when I implement Flutter's framework and I declare to the widget here's your attributes when I construct. I feel like I'm declaring my intent when it's constructed. Do you think it's declarative in nature from that point of view?

Hixie:
@wmleler any updates from your language friends? I'd really like to land this sooner rather than later.

wmleler:
Sorry, I've been swamped preparing for GDE. Leaving town Friday!

I sent off the question to two friends. I'll see if they have a good answer.

For now, you might just call it a "declarative subset of Dart".

Hixie:
@wmleler I'm not sure how "declarative subset of Dart" would fit in the text. Can you elaborate?

The text we're talking about is from a list of programming paradigms, and says:

> * Declarative programming: The build functions of widgets are typically a single expression with multiple levels of nested constructors, very much a declarative approach to building UI (such nested expressions could be mechanically transformed to or from any suitably expressive markup language). This can also be combined with the imperative style to build UIs that would be harder to describe in a pure-declarative approach.

How would you rephrase this to address your concern?

wmleler:
Declarative programming: The build functions of widgets are typically a
single expression with multiple levels of nested constructors, written
using a strictly declarative subset of Dart (such nested expressions could
be mechanically transformed to or from any suitably expressive markup
language). This can also be combined with the imperative style to build UIs
that would be harder to describe in a pure-declarative approach.

Hixie:
Sounds good! PTAL.

Hixie:
@wmleler PTAL

mit-mit:
@wmleler can you take a final look at this, please?

faq.md
wmleler: I'm not familiar with the Listenable interface, so let me give the underlying issue I'm concerned about. Separation of concerns wants to separate the instigator of an action from the handler of that action. A procedure does this for the handler (i.e., a function foo() does not know or care who calls it). While an event does it for the instigator (an event doesn't know or care who handles it). In reactive programming, you care about the latter. This can be done with "event-driven programming" but this is nothing new. It also requires an event to be created for every possible thing that might need to be handled. I was using the term "Observable interface" in the general sense (not Dart specific) to mean an interface that can be easily added to any class to allow it to be observed (e.g., one or more handlers subscribed to it). Does Listenable do this?

I was confused because your language seems to be describing Listenable as some kind of callback mechanism, which is not an Observable interface. In fact, I would not want for us to brag that using callbacks is "event-driven programming". In a callback, there is little separation of concerns.

This is especially confusing because in Flutter, a stateful widget has to explicitly call setState. For example, in the default Flutter app, inside class _MyHomePageState there is a comment that says: "This call to setState tells the Flutter framework that something has changed in this State". That is not event-driven programming (and is certainly not Reactive) because the state (the observable object) has to explicitly tell the framework that it changed). It is also not functional. Would the use of a Listenable interface fix this? I don't see one being used here.

faq.md
```diff
@@ -499,6 +499,90 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+([RenderObject](https://docs.flutter.io/flutter/rendering/RenderObject-class.html))
+that is agnostic regarding the coordinate system, and then we have a subclass
+([RenderBox](https://docs.flutter.io/flutter/rendering/RenderBox-class.html))
+that introduces the opinion that the geometry should be based on the Cartesian coordinate system (x/width and y/height).
+
+* Prototype-based object-orientated programming: The
+[ScrollPhysics](https://docs.flutter.io/flutter/widgets/ScrollPhysics-class.html) class
+chains instances to compose the physics that apply to scrolling dynamically at runtime. This
+lets the system compose, for example, paging physics with platform-specific physics, without
+the platform having to be selected at compile time.
```
wmleler: This is good. It defines a benefit (something that can be done at compile time). But I think this could be worded a bit better to make it clearer what problem it is solving.

faq.md
```diff
@@ -499,6 +499,90 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+* Imperative programming: Straight forward imperative programming, usually paired with state encapsulated
+within an object, is used where it provides the most intuitive solution. For example, tests are
+written in an imperative style, first describing the situation under test, then listing the invariants
+that the test must match, then advancing the clock or inserting events as necessary for the test.
+
+* Reactive programming: The widget and element trees are sometimes described as reactive, because new inputs 
+provided in a widget's constructor are immediately propagated as changes to lower-level widgets by
+the widget's build method, and changes made in the lower widgets (e.g. in response to user input) propagate
+back up the tree via event handlers. Aspects of both functional-reactive and imperative-reactive are
+present in the framework.
```
wmleler: This would make far more sense if it were not for the fact that having to call setState explicitly to trigger building of the tree is basically the antithesis of Reactive programming.

faq.md
```diff
@@ -499,6 +499,90 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+* Reactive programming: The widget and element trees are sometimes described as reactive, because new inputs 
+provided in a widget's constructor are immediately propagated as changes to lower-level widgets by
+the widget's build method, and changes made in the lower widgets (e.g. in response to user input) propagate
+back up the tree via event handlers. Aspects of both functional-reactive and imperative-reactive are
+present in the framework. 
+
+* Declarative programming: The build functions of widgets are typically a single expression with multiple levels
+of nested constructors, written using a strictly declarative subset of Dart (such nested expressions could be mechanically
+transformed to or from any suitably expressive markup language). This can also be combined with
+the imperative style to build UIs that would be harder to describe in a pure-declarative approach.
```
wmleler: Definitely need an example, and a statement of the benefit or problem solved. Does this allow Flutter to reorder the building of the widget tree, or do transformations on the code prior to building the tree? If not, then this is imperative since it also proscribes how the tree must be created, not just what it looks like. Basically, why did you build the tree this way? What's the advantage?

faq.md
```diff
@@ -499,6 +499,90 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+
+* Generic programming: Types can be used to help catch programming errors early. The Flutter framework
+uses generic programming to help in this regard. For example, the
+[State](https://docs.flutter.io/flutter/widgets/State-class.html) class is parameterized in terms
+of the type of its associated widget, so that the Dart analyzer can catch mismatches of states and widgets.
```
wmleler: good. Has an example and a benefit!

faq.md
```diff
@@ -499,6 +499,90 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+[Map](https://docs.flutter.io/flutter/dart-core/Map-class.html)s, and
+[Set](https://docs.flutter.io/flutter/dart-core/Set-class.html)s are all parameterized so that mismatched
+elements can be caught early either by the analyzer or at runtime during debugging.
+
+* Concurrent programming: Flutter makes heavy use of
+[Future](https://docs.flutter.io/flutter/dart-async/Future-class.html)s and other asynchronous APIs.
+For example, the animation system reports when an animation is finished by completing a future. The
+image loading system similarly uses futures to report when a load is complete.
+
+* Constraint programming: The layout system in Flutter uses a weak form of constraint programming to
```
wmleler: Not sure why you say "weak".

faq.md
```diff
@@ -499,6 +499,90 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+determine the geometry of a scene. Constraints (e.g. for cartesian boxes, a minimum and maximum width
+and a minimum and maximum height) are passed from parent to child, and the child selects a resulting
+geometry (e.g. for cartesian boxes, a size, specifically a width and a height) that fulfills those
+constraints. By using this technique, Flutter can usually lay out a whole scene with a single pass.
```
wmleler: Good benefit. I might even add "in Order N time where N is the number of widgets".

faq.md
```diff
@@ -499,6 +499,90 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+* Prototype-based object-orientated programming: The
+[ScrollPhysics](https://docs.flutter.io/flutter/widgets/ScrollPhysics-class.html) class
+chains instances to compose the physics that apply to scrolling dynamically at runtime. This
+lets the system compose, for example, paging physics with platform-specific physics, without
+the platform having to be selected at compile time.
+
+* Imperative programming: Straight forward imperative programming, usually paired with state encapsulated
+within an object, is used where it provides the most intuitive solution. For example, tests are
+written in an imperative style, first describing the situation under test, then listing the invariants
+that the test must match, then advancing the clock or inserting events as necessary for the test.
```
wmleler: Benefit? Problem solved? Why was it done this way?

faq.md
```diff
@@ -499,6 +499,90 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+class, which is used as the basis of the animation system, formalises a
+subscription model for events with multiple listeners.
+
+* Class-based object-orientated programming: Most of the APIs of the framework are built using classes
+with inheritance. We use an approach whereby we define very high-level APIs in our base classes, then
+specialise them iteratively in subclasses. For example, our render objects have a base class
+([RenderObject](https://docs.flutter.io/flutter/rendering/RenderObject-class.html))
+that is agnostic regarding the coordinate system, and then we have a subclass
+([RenderBox](https://docs.flutter.io/flutter/rendering/RenderBox-class.html))
+that introduces the opinion that the geometry should be based on the Cartesian coordinate system (x/width and y/height).
```
wmleler: This is a good benefit, but it could be explained a little better.

faq.md
```diff
@@ -499,6 +499,90 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+developed over the past few decades are used in Flutter, each to solve
+a particular set of problems in a clean way. In no particular order:
+
+* Composition: The primary paradigm used by Flutter is that of using small objects with narrow
+scopes of behavior, composed together to obtain more complicated effects. Most widgets
+in the Flutter widget library are built in this way.
+
+* Functional programming: Entire applications can be built with only
+[StatelessWidget](https://docs.flutter.io/flutter/widgets/StatelessWidget-class.html)s,
+which are essentially functions that describe how arguments map to other functions,
```
wmleler: What do you mean "how arguments map to other functions"? Are you mapping arguments to functions, or are you just calling other functions using arguments that have been mapped?

faq.md
```diff
@@ -499,6 +499,90 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+bottoming out in primitives that compute layouts or paint graphics.
```
wmleler: If the tail of your chain of mapping functions ends in primitives that compute layouts, the state of which then affects what gets drawn, then this doesn't seem to be functional at all.

faq.md
```diff
@@ -499,6 +499,90 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
+Additionally, heavy use is made of immutable data structures, including the entire
+[Widget](https://docs.flutter.io/flutter/widgets/Widget-class.html)
+class hierarchy as well as numerous supporting classes such as
+[Rect](https://docs.flutter.io/flutter/dart-ui/Rect-class.html) and
+[TextStyle](https://docs.flutter.io/flutter/painting/TextStyle-class.html).
+On a smaller scale, Dart's
+[Iterable](https://docs.flutter.io/flutter/dart-core/Iterable-class.html) API,
+which makes heavy use of the functional style (map, reduce, where, etc), is
+frequently used to process lists of values in the framework.
```
wmleler: What is the problem being solved? What is the benefit? Why do you use immutable data structures? The use of immutable data structures is not the same as functional programming (use of the former does not imply the latter, although the latter does require the former). Are you conflating the two?

faq.md
```diff
@@ -499,6 +499,90 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
...
 
+### What programming paradigm does Flutter's framework use?
+
+Flutter is a multi-paradigm programming environment. Many programming techniques
+developed over the past few decades are used in Flutter, each to solve
+a particular set of problems in a clean way. In no particular order:
+
+* Composition: The primary paradigm used by Flutter is that of using small objects with narrow
+scopes of behavior, composed together to obtain more complicated effects. Most widgets
+in the Flutter widget library are built in this way.
```
wmleler: Example? Benefit? Problem solved?

faq.md
```diff
@@ -499,6 +499,90 @@ If you are using the Flutter plugin for IntelliJ, you can launch the app in
 profile or release mode using the menu entries **Run > Flutter Run in Profile Mode**
 or **Release Mode**.
 
+### What programming paradigm does Flutter's framework use?
+
+Flutter is a multi-paradigm programming environment. Many programming techniques
+developed over the past few decades are used in Flutter, each to solve
+a particular set of problems in a clean way. In no particular order:
```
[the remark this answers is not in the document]
Hixie: Yeah, you're right that saying that each one solves a particular problem sets the framing poorly. I've rephrased the first paragraph to just be:

```
Flutter is a multi-paradigm programming environment. Many programming
techniques developed over the past few decades are used in Flutter. We
use each one where we believe the strengths of the technique make it
particularly well-suited. In no particular order:
```

faq.md
Hixie: The paragraph gives an example: "Most widgets". The benefits are subjective, I'm not sure how to objectively describe them here without setting us up for subjective arguments (which I don't think benefit the project).

faq.md
Hixie: a StatelessWidget is literally equivalent to a function with the signature `(constructor arguments) -> Widget`. If you build your app exclusively with StatelessWidgets, then you are mapping the constructor arguments to more StatelessWidgets, which as previously established are equivalent to functions, so they are "essentially functions that describe how arguments map to other functions".

faq.md
Hixie: That's like saying that Scheme isn't functional because it bottoms out to a C program that does allocation and I/O. :-)

The point here is to discuss paradigms that are used in limited parts of the system. When you deal with StatelessWidget, you are using a functional programming style. Beyond that, we use other paradigms.

faq.md
Hixie: As above, I think we should avoid going into details of "why" here since it would lead to subjective arguments which isn't the point of the FAQ entry. The entry is just trying to answer the question "what paradigms does Flutter use", the answer to "why" could result in many book deals down the road. :-)

faq.md
Hixie: > This can be done with "event-driven programming" but this is nothing new.

That's rather the thesis of this entire FAQ entry. As the introduction says, "Many programming techniques developed over the past few decades are used in Flutter". It's not trying to say that we invented any of them.

> I was using the term "Observable interface" in the general sense (not Dart specific) to mean an interface that can be easily added to any class to allow it to be observed (e.g., one or more handlers subscribed to it). Does Listenable do this?

Yes. For example, Animation (and its subclasses such as AnimationController) implements Listenable, and when you want to be driven by an animation you subscribe to it. The Animation doesn't know anything about who is subscribed to it.

> a stateful widget has to explicitly call setState. [...] That is not event-driven programming

Agreed, setState is unrelated to this.

faq.md
Hixie: Can you propose alternative text? I'm not sure what is unclear here.

faq.md
Hixie: Can you suggest some text for making it clearer? I'm not sure how else to phrase this.

faq.md
Hixie: As above, I think we should avoid going into details of "why" here since it would lead to subjective arguments which isn't the point of the FAQ entry.

faq.md
Hixie: Yup, this isn't describing how `State` works. We don't use the reactive style for `State`. This is talking about where we do use the reactive style, which is in the `Element` tree and build methods.

faq.md
Hixie: > Definitely need an example, and a statement of the benefit or problem solved.

As above, I think we should avoid getting too specific because that would take away from the point of the FAQ entry. We give a pointer for people who want to learn more ("build methods").

> Does this allow Flutter to reorder the building of the widget tree, or do transformations on the code prior to building the tree? If not, then this is imperative since it also proscribes how the tree must be created, not just what it looks like.

Flutter does internally control the order in which widgets are built, but I don't know that one could really say that the declarative style "allows" it. Just being declarative doesn't necessarily mean that things can be arbitrarily reordered; if anything, it is usually used to _force_ an order (by requiring that the system determine the required order of operations).

faq.md
Hixie: That's probably too strong a claim to make in the FAQ (it depends on much more, e.g. on whether you use the intrinsic dimension APIs, or whether you have off-stage widgets, etc).

faq.md
Hixie: We used to have a stronger form, cassowary, but we removed it since nobody made use of it.

Hixie: (Basically I don't want to oversell our use of constraints programming. We're certainly influenced by the concept and use it to some extent, but we don't make as heavy use of it as we do of functional, reactive, declarative, or generic programming.)

Hixie:
The patch in this PR was made on the Web site, since we keep iterating on it I've brought it locally to my machine and in submitting that had to make a new PR: https://github.com/flutter/website/pull/668

I'm therefore closing this PR in favour of the new one.

PR #657, opened 2017-08-24
Remove mention of Roboto fonts in the custom font docs

jason-simmons:
These fonts are no longer included by the Flutter tools.


cbracken:
lgtm

PR #658, opened 2017-08-30
Adding/rearranging animations pages

Sfshaza:
This PR:
- Adds a new landing page
- Moves the old landing page to the new /animations subdirectory and calls it a tech overview
- Add's Jim Beveridge's tweening doc (with his permission)

Staged: https://flutter-io-deploy-two.firebaseapp.com/animations/

Please review @InMatrix @Hixie


animations/tween-animations.md
```diff
@@ -0,0 +1,727 @@
...
+functions directly by reading their current value and listening to their
+state changes or they can use the animations as the basis of more elaborate
+animations that they pass along to other widgets.
+
+## Animation<double>
+
+In Flutter, an Animation object knows nothing about what is onscreen.
+An Animation is an abstract class that understands its current value
+and its state (completed or dismissed). One of the more commonly used
+animation types is Animation<double>.
```
Hixie: here and elsewhere (e.g. the header), you need to put backticks around class names otherwise the `<` characters end up interpreted as HTML.

animations/tween-animations.md
```diff
@@ -0,0 +1,727 @@
...
+
+A few facts about Flutter animations:
+
+* An Animation object in Flutter is a class that sequentially generates
```
Hixie: this is specifically talking about `Animation<double>`

animations/tween-animations.md
```diff
@@ -0,0 +1,727 @@
...
+    duration: const Duration(milliseconds: 2000), vsync: this);
+{% endprettify %}
+
+AnimationController derives from Animation<double>, so it can be used
+wherever an Animation object is needed. However, the AnimationController
+has additional methods to control the animation. For example, you start
+an animation with the `.forward()` method. The generation of numbers is
+tied to the screen refresh, so typically 60 numbers are generated per
+second. After each number is generated, each Animation object calls the
+attached Listener objects.
```
Hixie: i would just say "attached listeners", since they're usually closures or methods, not objects (except in the pedantic sense that closures, methods, and functions are objects).

animations/tween-animations.md
```diff
@@ -0,0 +1,727 @@
...
+
+When creating an AnimationController, you pass it a `vsync` argument.
+This option keeps the current widget tree in display memory until
+Flutter's rendering engine completes its refresh cycle. To use your
```
Hixie: The sentence "This option keeps the current widget tree in display memory until Flutter's rendering engine completes its refresh cycle." is not at all correct. :-)

What vsync does is it ties the AnimationController to a particular place in the widget tree, so that when that part of the widget tree isn't on the screen, the AnimationController stops calling its callback (though without pausing the clock, so time still passes). Specifically, the vsync argument specifies a TickerProvider.

animations/tween-animations.md
```diff
@@ -0,0 +1,727 @@
...
+final Tween doubleTween = new Tween<double>(begin: -200.0, end: 0.0);
+{% endprettify %}
+
+A Tween is a stateless object that takes only `begin` and `end`.
+The sole job of a Tween is to define a mapping from an input range
+to an output range. The input range is commonly 0.0 to 1.0,
+but that’s not a requirement.
+
+A Tween inherits from Animatable<T>, not from Animation<T>.
+An Animatable does not have to output double.
```
Hixie: Nor does an `Animation<T>` (it produces a T). I'd probably move this sentence elsewhere lest people think that it is the explanation of the difference between Animation and Animatable.

Hixie: if you do want to explain the difference here, it's what you say in the paragraph below -- Animations have state and a current value and can be listened to, while Animatables describe how to map an `Animation<double>`'s value to another value.

animations/tween-animations.md
```diff
@@ -0,0 +1,727 @@
...
+each configured with a single Tween.
+
+## Animation notifications
+
+An Animation object can have Listeners and StatusListeners,
+defined with `addListener()` and `addStatusListener()`.
+A Listener is called whenever the value of the animation changes.
+The most common behavior of a Listener is to call `setState()`
+to cause a rebuild. A StatusListener is called when an animation begins,
+ends, moves forward, or moves reverse, as defined by AnimationStatus.
```
Hixie: might be simpler to describe this as "starts, stops, or changes direction"

animations/tween-animations.md
```diff
@@ -0,0 +1,727 @@
...
+The next section has an example of the `addListener()` method,
+and [Monitoring the progress of the animation](#monitoring) shows an
+example of `addStatusListener()`.
+
+## Rendering animations
+
+The examples so far show how to generate a sequence of numbers over time.
+Nothing has been rendered to the screen.
+To render with an Animation<> object, store the Animation object as a
```
Hixie: `<>`?

animations/tween-animations.md
```diff
@@ -0,0 +1,727 @@
...
+  [[highlight]]}[[/highlight]]
+
+  @override
+  Widget build(BuildContext context) {
+    return new Center(
+      child: new Container(
+        margin: new EdgeInsets.symmetric(vertical: 10.0),
+        height: [[highlight]]animation.value,[[/highlight]]
+        width: [[highlight]]animation.value,[[/highlight]]
+        child: new FlutterLogo,
```
Hixie: missing parens on the constructor here

animations/tween-animations.md
```diff
@@ -0,0 +1,727 @@
...
+            setState(() {
+              // the animation object’s value is the changed state
+            });
+          });
+{% endprettify %}
+
+The difference is that, in the first case,
+the animation variable hasn't been set when the closure is created.
+Therefore, if you need to use `animation.value` in `setState()`,
+you must use the second form.
```
Hixie: it's been set by the time the closure runs, which should be sufficient.
maybe you mean that it hasn't been declared (and is thus out of scope)? If so, the example should use "var", "final", or a type on the assignment line to make it clear that that's where the variable is defined. (In this example it's not defined there, it's a member variable).

animations/index.md
```diff
@@ -0,0 +1,115 @@
...
+"scroll driven layout demo".
+{% endcomment %}
+
+### Physics-based animation
+
+In physics-based animation, motion is modeled to resemble real-world
+behavior. When you toss a ball, for example, where and when
+it lands depends on how fast it was tossed, how heavy it is, and how
+far off the ground. Similarly, dropping a ball attached to a spring falls
+(and bounces) differently than dropping a ball attached to a string.
```
InMatrix: Should we point advanced users to [AnimationController.animateWith](https://docs.flutter.io/flutter/animation/AnimationController/animateWith.html) and [SpringSimulation](https://docs.flutter.io/flutter/physics/SpringSimulation-class.html)? We can acknowledge that we're working on docs for physics-based animations.

animations/index.md
```diff
@@ -0,0 +1,115 @@
...
+## Other resources
+
+Learn more about Flutter animations at the following links:
+
+* [Animations: Technical Overview](/animations/overview.html)<br>
+A look at some of the major classes in the animations library,
+and Flutter's animation architecture.
+
+* The [animation
+library](https://docs.flutter.io/flutter/animation/animation-library.html)
```
InMatrix: The landing page of animation library is not very useful as it is today. For one thing, many classes useful for building animations are outside of this library, e.g., [SizeTransition](https://docs.flutter.io/flutter/widgets/SizeTransition-class.html), etc. I'll file a bug to add more info to the animation library landing page. For the time being, I'd recommend link to https://flutter.io/widgets/animation/

animations/tween-animations.md
```diff
@@ -0,0 +1,727 @@
...
+
+    animation.addStatusListener((status) {
+      if (status == AnimationStatus.completed) {
+        controller.reverse();
+      } else if (status == AnimationStatus.dismissed) {
+        controller.forward();
+      }
+    });
+    controller.forward();
+  }
```
Hixie: you can also get this same effect with an appropriately configured sawtooth tween and "repeat" on the animation controller.

animations/index.md
```diff
@@ -0,0 +1,115 @@
...
+in the [Flutter API docs](https://docs.flutter.io/)<br>
+The animation API for the Flutter framework.
+If you are writing a [hero
+animation](https://docs.flutter.io/flutter/widgets/Hero-class.html),
+you may also want to reference the
+[Navigator](https://docs.flutter.io/flutter/widgets/Navigator-class.html) and
+[PageRoute](https://docs.flutter.io/flutter/widgets/PageRoute-class.html)
+widgets.
```
InMatrix: Mention these classes where we talked about hero animation?

animations/tween-animations.md
```diff
@@ -0,0 +1,727 @@
...
+
+* TOC Placeholder
+{:toc}
+
+The animation system in Flutter is based on typed
+[`Animation`](https://docs.flutter.io/flutter/animation/Animation-class.html)
+objects. Widgets can either incorporate these animations in their build
+functions directly by reading their current value and listening to their
+state changes or they can use the animations as the basis of more elaborate
+animations that they pass along to other widgets.
```
InMatrix: When we say "animation" here, it'd be helpful to distinguish between the general concept of animation and Dart objects with the Animation type.

animations/tween-animations.md
InMatrix: What is that "particular place" in the widget tree? Is it the State object from which the TickerProvider is obtained?

animations/tween-animations.md
```diff
@@ -0,0 +1,726 @@
...
+[Dart Language Tour.](https://www.dartlang.org/guides/language/language-tour)
+</aside>
+
+##  Simplifying with AnimatedWidget
+
+The call to `setState()` is an obvious candidate for refactoring,
+which you’ll find in a helper class named AnimatedWidget.
+The simplified code is shown below. LogoApp now derives from
+AnimatedWidget instead of from StatefulWidget, and your
+AnimatedWidget does not need to maintain a State object to
```
InMatrix: But the parent of the AnimatedWidget needs to maintain a State object to hold the animation. I'm not sure if we should explain that.

animations/tween-animations.md
```diff
@@ -0,0 +1,726 @@
...
+  runApp(new LogoApp());
+}
+{% endprettify %}
+
+LogoApp passes the Animation object to the base class and uses
+`animation.value` to set the height and width of the container, so
+it works exactly the same as before.
+
+You can find the source for this example,
+[animate2,](https://raw.githubusercontent.com/InMatrix/animation_tutorial/master/lib/animate2.dart)
```
InMatrix: Should I move the code to a place in the Flutter repo?

animations/tween-animations.md
```diff
@@ -0,0 +1,726 @@
...
+
+  initState() {
+    super.initState();
+    controller = new AnimationController(
+        duration: const Duration(milliseconds: 2000), vsync: this);
+    animation = new Tween(begin: 0.0, end: 300.0).animate(controller)
+      [[highlight]]..addStatusListener((state) => print("$state"));[[/highlight]]
+    controller.forward();
+  }
+  //...
```
InMatrix: Can we make `//...` stand out a bit more? It's easy to copy and paste it and believe it's a self-contained code snippet.

animations/tween-animations.md
```diff
@@ -0,0 +1,726 @@
...
+  }
+}
+{% endprettify %}
+
+The middle three blocks in the diagram are all created in the
+`build()` method in GrowTransition. The GrowTransition widget itself
+is stateless and holds the set of final variables necessary to
+define the transition animation. The build() function creates and
+returns the AnimatedBuilder, which takes the (Anonymous builder)
+method and the LogoWidget object as parameters. The work of
```
InMatrix: LogoWidget object is indirectly passed into the (Anonymous builder) method through the child property of GrowTransition.

animations/tween-animations.md
```diff
@@ -0,0 +1,726 @@
...
+  build(BuildContext context) {
+    return new Container(
+      margin: new EdgeInsets.symmetric(vertical: 10.0),
+      child: new FlutterLogo(),
+    );
+  }
+}
+{% endprettify %}
+
+The middle three blocks in the diagram are all created in the
```
InMatrix: I'm a little confused by the "ContainerTransition" in the diagram. It's not in the code. Is it the same thing as GrowTransition?

animations/tween-animations.md
```diff
@@ -0,0 +1,726 @@
...
+
+You can find the source for this example,
+[animate5,](https://raw.githubusercontent.com/InMatrix/animation_tutorial/master/lib/animate5.dart)
+on GitHub.
+
+## Next steps
+
+There are many more classes to explore, including specialized Tween
+classes, animations specific to Material Design, ReverseAnimation,
+physics simulations and `fling()` methods. This doc should have
```
InMatrix: Maybe we can link to relevant API docs here. Some developers might be able to figure things out, and if not, hopefully they will ask about them and file issues.

animations/tween-animations.md
```diff
@@ -0,0 +1,726 @@
...
+given you the foundation to understand these other parts of Flutter
+animation.
+
+{% comment %}
+The Flutter team is looking for use cases for more complex animations,
+such as cascading events that delay the start of Animation B until three
+seconds after Animation A starts. Please contact abarth@ if you have
+use cases.
+{% endcomment %}
```
InMatrix: Remove this comment?

animations/tween-animations.md
```diff
@@ -0,0 +1,726 @@
...
+
+## CurvedAnimation
+
+A CurvedAnimation modifies an Animation object's output to make the
+progression be a non-linear curve:
+
+<!-- skip -->
+{% prettify dart %}
+final CurvedAnimation curve =
+    new CurvedAnimation(parent: controller, curve: Curves.easeIn);
```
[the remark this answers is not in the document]
Hixie: Yes, it's not uncommon to chain multiple Animations.

It's the parent because that's the one that will drive this one.

animations/tween-animations.md
Hixie: It ties it to the TickerProvider; typically that's the State that mixes in the TickerProviderStateMixin. TickerProviderStateMixin further defers to the subtree that's rooted at the nearest TickerMode widget.

Sfshaza:
I've incorporated @InMatrix's feedback on the landing page. I still have to incorporate @Hixie's feedback and update the converted tutorial. This is still in process.

animations/index.md
```diff
@@ -0,0 +1,158 @@
...
+makes it easy to implement a variety of animation types.
+Many widgets, especially
+[Material Design widgets](https://flutter.io/widgets/material/),
+come with the standard motion effects defined in their design spec,
+but it's also possible to customize these effects.
+
+The following resources are a good place to start learning the Flutter
+animation framework. Each of these docs shows, step by step, how
+to write animation code.
+Note that more docs are in the works on how to implement common design patterns,
```
Hixie: nit: i would omit the words "Note that" from this sentence.

animations/index.md
```diff
@@ -0,0 +1,158 @@
...
+such as shared element transitions, and physics-based animations.
+If you have a specific request, please
+[file an issue](https://github.com/flutter/flutter/issues).
+
+* [Tutorial: Animations in Flutter](/tutorials/animation/) <img src="/images/ic_new_releases_black_24px.svg" alt="this doc is new!"> NEW<br>
```
Hixie: The alt text on this image is redundant with the text that follows it. We should have one or the other, not both.

animations/index.md
```diff
@@ -0,0 +1,158 @@
...
+Codelab demonstrating how to build a simple chat app. [Step 7 (Animate
+your app)](https://codelabs.developers.google.com/codelabs/flutter/index.html#6)
+shows how to animate the new message&mdash;sliding it from the input area up
+to the message list.
+
+## Animation types
+
+Animations fall into one of two categories: tween- or physics-based.
+The following sections explain what these terms mean, and points you to
+resources where you can learn more. Note that, in some cases,
```
Hixie: nit: omit "Note that"

animations/index.md
```diff
@@ -0,0 +1,158 @@
...
+the best documentation we currently have is example code in the
+Flutter gallery. More docs and examples will be forthcoming.
```
Hixie: nit: here and elsewhere, maybe replace "docs" with "documents"?

animations/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Physics-based animation
+
+In physics-based animation, motion is modeled to resemble real-world
+behavior. When you toss a ball, for example, where and when
+it lands depends on how fast it was tossed, how heavy it is, and how
+far off the ground. Similarly, dropping a ball attached to a spring falls
+(and bounces) differently than dropping a ball attached to a string.
+
+Note that more docs and examples on physics-based animations will be
```
Hixie: nit: i would remove "note that".

We may want to remove the promises for more documentation in general, actually. If we add more docs, great. If we never do, then these promises aren't helpful.

animations/index.md
```diff
@@ -0,0 +1,158 @@
...
+animate adding an element to a list, or removing a selected element.
+The internal Dart list is synced as the user modifies the list using
+the plus (+) and minus (-) buttons.
+
+### Shared element transition
+
+In this pattern, the user selects an element&mdash;often an
+image&mdash;from the page, and the UI animates the selected element
+to a new page with more detail. In Flutter, you can easily implement
+shared element transitions between routes (pages) using the Hero widget.
```
Hixie: would it make sense, here and elsewhere, to link the class names to the API docs for those classes? Or is it better to only have the "also see" links below?

tutorials/animation/index.md
```diff
@@ -0,0 +1,874 @@
...
+an animation with the `.forward()` method. The generation of numbers is
+tied to the screen refresh, so typically 60 numbers are generated per
+second. After each number is generated, each Animation object calls the
+attached Listener objects. To create a custom display list for each
+child, see
+[RepaintBoundary](https://docs.flutter.io/flutter/widgets/RepaintBoundary-class.html).
+
+When creating an AnimationController, you pass it a `vsync` argument.
+This option keeps the current widget tree in display memory until
+Flutter's rendering engine completes its refresh cycle. To use your
```
Hixie: The sentence "This option keeps the current widget tree in display memory until Flutter's rendering engine completes its refresh cycle." is wrong.

What 'vsync' does is tie the ticking of the animation controller to the visibility of the widget, so that when the animating widget goes off-screen, the ticking stops, and when the widget is restored, it starts again (without stopping the clock, so it's as if it had been ticking the whole time, but without using CPU).

tutorials/animation/index.md
```diff
@@ -0,0 +1,874 @@
...
+custom State object as the `vsync`, include the `TickerProviderStateMixin`
+when defining the custom State class.
+
+<aside class="alert alert-success" markdown="1">
+**Note**: In some cases, a position might exceed the AnimationController's
+0.0-1.0 range. For example, the `fling()` function allows you to provide
+velocity, force, and position (via the Force object).
+The position can be anything and so can be outside of the 0.0 to 1.0 range.
+
+A CurvedAnimation can also exceed the 0.0 to 1.0 range,
```
Hixie: at this point you haven't introduced CurvedAnimation yet.

tutorials/animation/index.md
```diff
@@ -0,0 +1,874 @@
...
+  double transform(double t) {
+    return math.sin(t * math.PI * 2);
+  }
+}
+{% endprettify %}
+</aside>
+
+CurvedAnimation and AnimationController are both of type
+Animation&lt;double&gt;, so you can pass them interchangeably.
+Note that the CurvedAnimation wraps the object it’s modifying.
```
Hixie: nit: here and elsewhere, remove the words "note that" whenever they occur. They don't add anything.

tutorials/animation/index.md
```diff
@@ -0,0 +1,874 @@
...
+final Tween doubleTween = new Tween<double>(begin: -200.0, end: 0.0);
+{% endprettify %}
+
+A Tween is a stateless object that takes only `begin` and `end`.
+The sole job of a Tween is to define a mapping from an input range
+to an output range. The input range is commonly 0.0 to 1.0,
+but that’s not a requirement.
+
+A Tween inherits from Animatable&lt;T&gt;, not from Animation&ltT&gt;.
+An Animatable does not have to output double.
```
Hixie: > An Animatable does not have to output double.

Nor does an Animation. (It sounds like you're saying this as a contrast to Animation.)

tutorials/animation/index.md
```diff
@@ -0,0 +1,874 @@
...
+For example, ColorTween specifies a progression between two colors.
+
+<!--- skip -->
+{% prettify dart %}
+final Tween colorTween =
+    new ColorTween(begin: Colors.transparent, end: Colors.black54);
+{% endprettify %}
+
+A Tween object does not store any state. Instead, it exposes two methods
+`evaluate(Animation<double> animation)`, which applies the mapping function
```
Hixie: you say two methods but only list one

tutorials/animation/index.md
```diff
@@ -0,0 +1,874 @@
...
+to the current value of the animation. The current value of the `Animation`
+object can be found in the `.value` method. The evaluate function also
+performs some housekeeping, such as ensuring that begin and end are
+returned when the animation values are 0.0 and 1.0, respectively.
+
+#### Tween.animate
+
+To use a Tween object, you configure an Animation to use the Tween
+(Animatable) with the `animate()` function on the Animatable object.
```
Hixie: this sentence is convoluted.

tutorials/animation/index.md
```diff
@@ -0,0 +1,874 @@
...
+
+void main() {
+  runApp(new LogoApp());
+}
+{% endprettify %}
+
+The following shows the same code modified to animate the
+logo to grow from nothing to full size. Note that when
+defining an AnimationController, you must pass in a `vsync` object.
+The presence of `vsync` prevents offscreen animations from consuming
```
Hixie: this definition of vsync is accurate. use it in the other place too if you need it in both places. :-)

animations/index.md
Sfshaza: Travis fails unless there's ALT text, or I would have left it off. The visible text is important to clarify the meaning of the icon.

tutorials/animation/index.md
```diff
@@ -0,0 +1,874 @@
...
+        [[highlight]]duration: const Duration(milliseconds: 2000), vsync: this);[[/highlight]]
+    [[highlight]]animation = new Tween(begin: 0.0, end: 300.0).animate(controller)[[/highlight]]
+      [[highlight]]..addListener(() {[[/highlight]]
+        [[highlight]]setState(() {[[/highlight]]
+          [[highlight]]// the state that has changed here is the animation object’s value[[/highlight]]
+        [[highlight]]});[[/highlight]]
+      [[highlight]]});[[/highlight]]
+    [[highlight]]controller.forward();[[/highlight]]
+  [[highlight]]}[[/highlight]]
+
```
Hixie: this demo leaks the animation controller. You need a dispose method where you call animation.dispose().

tutorials/animation/index.md
```diff
@@ -0,0 +1,874 @@
...
+[[highlight]]animation = tween.animate(controller);[[/highlight]]
+[[highlight]]animation.addListener(()[[/highlight]] {
+            setState(() {
+              // the animation object’s value is the changed state
+            });
+          });
+{% endprettify %}
+
+The difference is that, in the first case,
+the animation variable hasn't been set when the closure is created.
```
Hixie: that doesn't matter. What matters is whether the variable is in scope, and it is, so both techniques would work fine.

tutorials/animation/index.md
```diff
@@ -0,0 +1,874 @@
...
+Therefore, if you need to use `animation.value` in `setState()`,
+you must use the second form.
```
Hixie: that is, this last sentence here is wrong.

tutorials/animation/index.md
```diff
@@ -0,0 +1,874 @@
...
+render tree. Like AnimatedWidget, AnimatedBuilder automatically
+listens to notifications from the Animation object, and marks
+the widget tree dirty as necessary, so you don't need to call
+`addListener()`.
+
+The widget tree for the
+[animate5](https://raw.githubusercontent.com/InMatrix/animation_tutorial/master/lib/animate5.dart)
+example looks like this:
+
+<img src="images/AnimatedBuilder-WidgetTree.png" alt="a widget tree with Container pointing to ContainerTransition, pointing to AnimatedBuilder, pointing to (AnonymousBuilder), pointing to LogoWidget">
```
Hixie: Capitalize the first word of the alt text, finish it with a period.

tutorials/animation/index.md
```diff
@@ -0,0 +1,874 @@
...
+  GrowTransition({this.child, this.animation});
+
+  final Widget child;
+  final Animation<double> animation;
+
+  @override
+  Widget build(BuildContext context) {
+    return new AnimatedBuilder(animation: animation,
+        builder: (BuildContext context, Widget child) {
+          return new Center(
```
Hixie: the center could be placed outside the AnimatedBuilder to be more efficient.

tutorials/animation/index.md
```diff
@@ -0,0 +1,874 @@
...
+{% prettify dart %}
+class LogoApp extends StatefulWidget {
+  LogoAppState createState() => new LogoAppState();
+}
+
+class LogoAppState extends State<LogoApp> with TickerProviderStateMixin {
+  Animation animation;
+  AnimationController controller;
+
+  initState() {
```
Hixie: `@override`

tutorials/animation/index.md
```diff
@@ -0,0 +1,874 @@
...
+    return new GrowTransition(
+        child: new LogoWidget(),
+        animation: animation
+    );
+  }
+
+  @override
+  dispose() {
+    super.dispose();
+    controller.dispose();
```
Hixie: here and elsewhere: convention is to call `super.dispose()` last, since in principle once you've called it your object is dead and you shouldn't do anything anymore.

animations/index.md
Hixie: We should fix Travis. It's invalid HTML to have redundant alt="" text. See https://html.spec.whatwg.org/multipage/images.html#alt

tutorials/animation/index.md
Sfshaza: Thanks! The source code is under Tao's repo. I've asked him to fix.

tutorials/animation/index.md
Sfshaza: I deleted those sentences. thx

Sfshaza:
Thanks, Ian, for your thorough review! I believe I've fixed everything, except where noted. I removed all @overrides from the examples, since the analyzer is happy without. I've sent Tao email asking him to update the code on GitHub, with a list of things to update. (I have tested all the updated code, but it's on my hard drive, not GitHub.)

Sfshaza:
The dupe _code is because I want to move the layout examples under a layout directory, but Travis will fail if I don't do this as a 2-step process, because it checks links. Once the code is in place, I can submit another PR to change the links and delete the old cold.

Sfshaza:
With Devon's fix, Travis is happy again. @Hixie, can this PR be pushed?

Hixie:
LGTM.

PR #662, opened 2017-09-05
Document how to handle dependency conflicts

mravn-google:


developing-packages.md
```diff
@@ -220,4 +220,48 @@ Pod::Spec.new do |s|
...
+configurations.all {
+    resolutionStrategy { 
+        force 'com.google.guava:guava:23.0-android'
+    }
+}
+```
+
+Cocoapods does not currently offer override functionality for iOS-specific dependencies. But conflict
+resolution by override should be used only as a last resort anyway. Developers generally rely on package
+authors making good use of version _ranges_ rather than specific versions, e.g.:
```
Hixie: The text about last resort and using ranges seems like it applies to the other two as well and should be put earlier and made more generic, leaving only "Cocoapods does not currently offer override functionality for iOS-specific dependencies.", maybe saying "Please contact your package developer and ask them to use ranges instead" or something.

developing-packages.md
```diff
@@ -220,4 +220,48 @@ Pod::Spec.new do |s|
...
+```ruby
+s.dependency 'AFNetworking', '~> 3'   # Any 3.x.y will do
+```
+instead of 
+```ruby
+s.dependency 'AFNetworking', '3.1.0'  # Only 3.1.0 will do
+```
+If `some_package` declares a `~> 3` dependency on `AFNetworking` in its `Podfile`, then Cocoapods can
+automatically reconcile that dependency with the declaration of, say, a `3.1.0` or `~> 3.1` dependency
+in `other_package`. Similar remarks apply to dependency resolution in pub and Gradle.
```
Hixie: ah in fact you even say that it applies to the others. I think we should move it up because people won't see this if they're not looking at cocoapod-specific stuff.

PR #663, opened 2017-09-06
Document use of asset images in package dependencies.

szakarias:


assets-and-images.md
```diff
@@ -207,6 +207,46 @@ or
...
+
+ To include, say the first image, the `pubspec.yaml` of the application should specify it in the `assets` section:
+
+```yaml
+flutter:
+  assets:
+    - packages/fancy_backgrounds/backgrounds/background1.png
+```
+
+Note that the `lib/` is implied, so it should not be included in the asset path.
```
Hixie: Avoid empty prose like "Note that", you can just start with "The".

Hixie:
LGTM

Hixie:
@LarkAscending may wish to review also.

szakarias:
Fixes https://github.com/flutter/flutter/issues/9249

assets-and-images.md
```diff
@@ -207,6 +207,46 @@ or
 [`ImageCache`](https://docs.flutter.io/flutter/services/ImageCache-class.html),
 you'll also notice parameters related to scale.)
 
+### Asset images in package dependencies
+
+To load an image from a package dependency, the `package` argument must be provided to [`AssetImage`](https://docs.flutter.io/flutter/services/AssetImage-class.html).
```
mit-mit: Maybe link `package` to https://flutter.io/using-packages/ which has a short intro to what packages are?

assets-and-images.md
```diff
@@ -207,6 +207,46 @@ or
...
+ new AssetImage('icons/heart.png', package: 'my_icons')
+```
+
+Assets used by the package itself should also be fetched using the `package` argument as above.
+
+#### Bundling of package assets
+
+If the desired asset is specified in the `pubspec.yaml` file of the package, it is bundled automatically with the application. In particular, assets used by the package itself must be specified in its `pubspec.yaml`.
+
+A package can also choose to have assets in its `lib/` folder that are not specified in its `pubspec.yaml` file. In this case, for those images to be bundled, the application has to specify which ones to include. For instance, a package named `fancy_backgrounds` could have the following files:
```
mit-mit: Maybe add 'in its `pubspec.yaml`' after 'which ones to include' ?

assets-and-images.md
```diff
@@ -207,6 +207,46 @@ or
...
+
+* .../lib/backgrounds/background1.png
+* .../lib/backgrounds/background2.png
+* .../lib/backgrounds/background3.png
+
+ To include, say the first image, the `pubspec.yaml` of the application should specify it in the `assets` section:
```
mit-mit: Missing comma after 'say'

PR #664, opened 2017-09-07
Update path-mac-linux.md

Hixie:
Based on https://github.com/flutter/flutter/issues/11974


Hixie:
cc @LarkAscending

Sfshaza:
LGTM!

LarkAscending:
Good addition! 
LGTM.

A nit - the list numbering scales better when we leverage Markdown's automatic numbering feature. It's no problem and I don't think the CONTRIBUTING guide has anything to say about it. So this is FYI for future reference only.

Hixie:
Yeah, I wondered about that. I figured this would make it easier to read while editing, and since it's so short... but yeah.

PR #666, opened 2017-09-11
Update .analysis_options

devoncarew:
- add `enableSuperMixins: true` to the analysis options, for consistency w/ how we analyze flutter projects (fix https://github.com/flutter/flutter/issues/12019)

@Sfshaza, @Hixie

### DIFF
--- a/.analysis_options
+++ b/.analysis_options
@@ -4,3 +4,4 @@ analyzer:
     - '_layouts/**'
   language:
     enableAssertInitializer: true
+    enableSuperMixins: true


devoncarew:
also, @pq

Hixie:
We should just use https://github.com/flutter/flutter/blob/master/packages/flutter/lib/analysis_options_user.yaml IMHO.

devoncarew:
> We should just use https://github.com/flutter/flutter/blob/master/packages/flutter/lib/analysis_options_user.yaml 

Generally, I agree. The file for this repo does also specifies some excludes, specific to this repo.

Hixie:
It looks like this file doesn't include any of the lints, though. Are they inherited somehow? And if so why aren't the language flags inherited too?

devoncarew:
> It looks like this file doesn't include any of the lints, though.

No, none of the lints are included. I'd prefer not to inline a copy of them here though.

> Are they inherited somehow? 

No, there's no inheritance with analysis options files.

Hixie:
It seems like we need an inheritance mechanism, or to duplicate the lint list (we already duplicate the repo one three times, why not this one twice...), or to have some code that autogenerates this one on the fly.

devoncarew:
Makes sense. I added https://github.com/flutter/flutter/issues/12041 to track that. I think the third options is pretty feasible.

Sfshaza:
Can we merge this, @Hixie @devoncarew ?

devoncarew:
I'll take that as a lgtm :) (and flutter/flutter#12041 to track longer term unification work).

From talking with @danrubel and @bwilkerson, there is actually a mechanism to import one analysis options file into another (`include: ...`). It is not yet documented however, and hasn't (as far as I'm aware) been tested with excludes.

PR #670, opened 2017-09-14
Moving the code around (part 1)

Sfshaza:
I have to do this in stages or Travis will fail. Once the animation code is in place, I have links to update.


sethladd:
RSLGTM

PR #671, opened 2017-09-14
Updates to FAQ from recent conversations

sethladd:


sethladd:
cc @kasperl @mit-mit

faq.md
```diff
@@ -46,6 +46,17 @@ Fundamentally, Flutter is for users that want beautiful apps,
...
+### How experienced of a programmer/developer do I have to be to use Flutter?
+
+Flutter is approachable to programmers familiar with object-oriented
+concepts (classes, methods, variables, etc) and imperative programming
+concepts (loops, conditionals, etc).
+
+No prior mobile experience is required in order to learn and use Flutter.
+
+We have seen people with very little programming experience learn
+and use Flutter for prototyping and app development.
```
Hixie: maybe link to the I/O talk about designers using flutter?

faq.md
```diff
@@ -58,6 +69,11 @@ stock platform apps can also be built with Flutter.
 You can build full-featured apps with Flutter, including
 camera, geolocation, network, storage, 3rd-party SDKs, and more.
 
+### Who makes Flutter?
+
+Flutter is an open source project, with contributions from Google
+and the community.
```
Hixie: we are part of the community, this makes it sound like we're separate.
Maybe "Flutter is an open source project, primarily led by Google engineers, with many contributions from other members of the wider community."?

Hixie:
LGTM

faq.md
mit-mit: yes, that would be good!

faq.md
```diff
@@ -202,6 +219,11 @@ more motion-rich UIs and brand-first designs.
 In order to achieve that level of customized, beautiful design,
 Flutter is architectured to drive pixels instead of the OEM widgets.
 
+By providing our own widgets, we also aim to lower the overall cost
+of maintenance for your apps. You can create and maintain a
+single codebase for your UI,
```
mit-mit: 'single codebase for your UI' can be misread to think we are closer to Xamarin.Forms, i.e. a single UI that is identical on iOS and Android. Maybe expand to something like:

By providing our own widgets, we aim to lower app development cost, and enhance your ability to deliver consistent user experiences across platforms. You can create and maintain a single codebase for your UI, even with platform-specific features if desired. And you can build and publish for both iOS and Android concurrently, without having to do careful and costly planning to align two separate codebases.

mit-mit: Or maybe this fits better in the 'share layout code' question in https://github.com/flutter/website/pull/673/files ?

faq.md
sethladd: works for me, thanks!

faq.md
sethladd: done

faq.md
sethladd: Thanks. 'single codebase for your UI' has been our way to answer "do you write a single UI?" , "you can, but you don't have to. With Flutter, you do work in a single codebase (same language, same widget toolkit, same framework), but you can make your UI as common or as different as you want."

sethladd: updated.

sethladd: Thank you for all the comments! re: biggest attraction and cost savings, when we ask startups why they look at non-OEM toolkits, it's because they often are looking to lower costs during development and more importantly the maintenance costs. Lowering costs isn't sexy, which is why it's important to continue pointing out (as you do) that with Flutter you get to lower costs AND deliver something with higher quality and a brand-first design.

The "consistent experience" angle isn't always what startups want (after numerous interviews with startup designers in 2016). The word "consistent" too often triggers "lowest common denominator" or "a third thing that doesn't feel natural", so while I agree that Flutter's strength is in brand-first (implied: consistent brand) UIs, we don't deliver 100% consistent experiences because we embrace platform "muscle memory".

mit-mit: I think you misunderstood my comment. 

I agree that they don't want a consistent experience, as in a consistent/identical UI.

BUT, I have heard numerous small startups describe the very real challenge of having two separate code bases leading to two separate teams leading to two different *feature sets* on each platform leading to frustrated customers. The ability to deliver the same feature set from a single code base -- while at the same time being able to specialize the UI -- is a major selling point here!

sethladd: Yes, fixing the "two separate teams" issue is real. I'll update "who is flutter for" because it's for eng managers seeking to solve that problem.

PR #672, opened 2017-09-14
update faq based on question from conference

sethladd:
cc @mit-mit

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -202,6 +202,21 @@ more motion-rich UIs and brand-first designs.
 In order to achieve that level of customized, beautiful design,
 Flutter is architectured to drive pixels instead of the OEM widgets.
 
+### What happens when my mobile OS updates and introduces new widgets?
+
+The Flutter team watches the adoption and demand for new mobile
+widgets from iOS and Android, and aims to work with the community
+to build support for new widgets. This may come in the form
+of lower-level framework features, new composable widgets, or new
+widget implementations. Flutter's
+layered architecture is designed to support numerous widget libraries,
+and we encourage and support the community in building and maintaining
+widget libraries.
+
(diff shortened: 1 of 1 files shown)


Hixie:
LGTM

faq.md
```diff
@@ -202,6 +202,21 @@ more motion-rich UIs and brand-first designs.
...
+The Flutter team watches the adoption and demand for new mobile
+widgets from iOS and Android, and aims to work with the community
+to build support for new widgets. This may come in the form
+of lower-level framework features, new composable widgets, or new
+widget implementations. Flutter's
+layered architecture is designed to support numerous widget libraries,
+and we encourage and support the community in building and maintaining
+widget libraries.
+
+Relatedly, Flutter's interop and plugin system is designed to allow
```
mit-mit: I don't think 'relatedly' is an actual work? Why not just use 'related'?

faq.md
```diff
@@ -202,6 +202,21 @@ more motion-rich UIs and brand-first designs.
 In order to achieve that level of customized, beautiful design,
 Flutter is architectured to drive pixels instead of the OEM widgets.
 
+### What happens when my mobile OS updates and introduces new widgets?
+
+The Flutter team watches the adoption and demand for new mobile
+widgets from iOS and Android, and aims to work with the community
+to build support for new widgets. This may come in the form
+of lower-level framework features, new composable widgets, or new
+widget implementations. Flutter's
```
mit-mit: Should we add a line saying something like 'And because the widgets are compiled into your app, they are then available on any supported OS version' after 'widget implementations'?

sethladd: Sure, but does that help answer the question? Or, maybe that's another answer to "why not oem widgets" ?

PR #673, opened 2017-09-14
updating faq based on recent conversations

sethladd:
cc @mit-mit

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -388,7 +388,7 @@ the "bottom nav bar" pattern can now be naturally found across iOS and Android.
 The brand identity and customization of the app's aesthetic design is now
 becoming more important than strictly adhering to traditional platform aesthetics.
 
-Given the importantance on brand-driven design, plus the fact that
+Given the importance on brand-driven design, plus the fact that
 modern apps often have many customized design elements (including branding,
 widgets, animations, and information architectures) sharing
 layout and UI code across platforms is a common practice.
--- a/faq.md
+++ b/faq.md
@@ -382,20 +382,24 @@ you like to make your own custom widgets that have full range of expression.
 
 ### Why would I want to share layout code across iOS and Android?
 
-More and more, we see mobile app layouts and designs converging
-to be more brand-driven and unified across platforms. For example,
-the "bottom nav bar" pattern can now be naturally found across iOS and Android.
-The brand identity and customization of the app's aesthetic design is now
-becoming more important than strictly adhering to traditional platform aesthetics.
+More and more, we see mobile app layouts and designs evolving
+to be more brand-driven and unified across platforms. This implies
+a strong motivation to share layout and UI code across iOS and
+Android.
 
-Given the importance of brand-driven design, plus the fact that
-modern apps often have many customized design elements (including branding,
-widgets, animations, and information architectures) sharing
-layout and UI code across platforms is a common practice.
+The brand identity and customization of the app's aesthetic design is now
+becoming more important than strictly adhering to traditional platform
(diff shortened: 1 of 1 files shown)


faq.md
```diff
@@ -380,6 +380,24 @@ framework, we've decomposed complex widgets to pieces that separately implement
...
+
+Given the importantance on brand-driven design, plus the fact that
+modern apps often have many customized design elements (including branding,
+widgets, animations, and information architectures) sharing
+layout and UI code across platforms is a common practice.
+
+When designers choose to implement different app layouts for iOS and Android,
+Flutter makes that possible. Developers are free to check the mobile OS
+and render different layouts, though we find this to be not too common in
+practice.
```
Hixie: Maybe reverse the order here, starting with "you don't have to", and then moving on to "but you may want to because"?

sethladd: do you mean reverse the order of these last two sentences ? where is the "here" you are referring to? :)

Hixie: I meant the whole three paragraphs.

The question "why would i" implies "i don't want to". So the quickest way to diffuse the question is "you don't have to". But our value comes in part from the way that you can, so then following up with the rest of the answer here would provide the context to actually answer the question.

Right now it reads "Why would I do this dumb thing?" "I contradict you, it's not dumb, because you should do things different! bla bla bla bla bla. Oh also, you don't actually have to."

I'm suggesting having it read "Why would I do this dumb thing?" "You don't have to. You may want to because bla bla bla bla bla."

faq.md
```diff
@@ -380,6 +380,24 @@ framework, we've decomposed complex widgets to pieces that separately implement
...
 
+### Why would I want to share layout code across iOS and Android?
+
+More and more, we see mobile app layouts and designs converging
+to be more brand-driven and unified across platforms. For example,
+the "bottom nav bar" pattern can now be naturally found across iOS and Android.
+The brand identity and customization of the app's aesthetic design is now
+becoming more important than strictly adhering to traditional platform aesthetics.
+
+Given the importantance on brand-driven design, plus the fact that
```
a14n: typo: importantance

sethladd: thanks :)

faq.md
```diff
@@ -380,6 +380,24 @@ framework, we've decomposed complex widgets to pieces that separately implement
 the visual, interaction, and motion design. You can remix these widgets however
 you like to make your own custom widgets that have full range of expression.
 
+### Why would I want to share layout code across iOS and Android?
+
+More and more, we see mobile app layouts and designs converging
+to be more brand-driven and unified across platforms. For example,
+the "bottom nav bar" pattern can now be naturally found across iOS and Android.
```
mit-mit: I suggest we swap the order of 'brand-driven' and 'unified across platforms' so that they are in the same order as the examples in the next two sentences.

faq.md
```diff
@@ -380,6 +380,24 @@ framework, we've decomposed complex widgets to pieces that separately implement
...
+The brand identity and customization of the app's aesthetic design is now
+becoming more important than strictly adhering to traditional platform aesthetics.
+
+Given the importance on brand-driven design, plus the fact that
```
mit-mit: s/on/of ?

sethladd: good catch

faq.md
sethladd: I'll reorder to match

sethladd:
@mit-mit @Hixie I've updated based on your comments. PTAL

sethladd:
travis failures are because I didn't merge with master in this PR.

PR #678, opened 2017-09-18
Update hello world example to match current behavior

mattrenquist:
The current hello world example does not run. Adding 'textDirection' to match the code in examples/hello_world/lib/main.dart fixes the bug and allows the code to run. The website documentation should be updated to match the current code.

### DIFF
--- a/widgets-intro.md
+++ b/widgets-intro.md
@@ -35,7 +35,14 @@ with a widget:
 import 'package:flutter/material.dart';
 
 void main() {
-  runApp(new Center(child: new Text('Hello, world!', textDirection: TextDirection.ltr)));
+  runApp(
+    new Center(
+      child: new Text(
+        'Hello, world!',
+        textDirection: TextDirection.ltr,
+      ),
+    ),
+  );
 }
 ```
 
--- a/widgets-intro.md
+++ b/widgets-intro.md
@@ -35,7 +35,7 @@ with a widget:
 import 'package:flutter/material.dart';
 
 void main() {
-  runApp(new Center(child: new Text('Hello, world!',textDirection: TextDirection.ltr)));
+  runApp(new Center(child: new Text('Hello, world!', textDirection: TextDirection.ltr)));
 }
 ```
 


mattrenquist:
Signed the CLA on 10:30 PST 9/17/2017.

Hixie:
Can you add a space after the comma?

Hixie:
The Travis failure is because we're requiring the use of dartfmt, but here dartfmt is making the code uglier by wrapping in a poor way. We should probably exclude this file from dartfmt verification.

cc @sethladd

mattrenquist:
Added the space. Thanks.

Hixie:
or maybe we can tell dartfmt to accept longer lines? iirc @mit-mit solved our most recent issue with Travis in this repo, maybe he has a quick way to address this one too. :-)

mit-mit:
`dartfmt` can be passed a `--line-length` argument to specify a longer wrapping length than it's default of 80. But we invoke `dartfmt` in a [single location](https://github.com/flutter/website/blob/master/tool/travis.sh#L41) for all formating checks, so would need to make that change globally.

mit-mit:
@Hixie @sethladd did you figure out how we want to handle this formatting issue?

Hixie:
I have no particular opinion on how to solve this. I don't think dartfmt can handle this code in a good way. I'd be fine with excluding this file, changing the line width globally (assuming it doesn't make anything else worse), not using dartfmt at all, or any other similar solution.

Sfshaza:
A longer-than-80-char line length means that code can wrap and look bad when displayed inside a web page.

sethladd:
Or, add trailing commas to the snippet, reformat it with dartfmt

Hixie:
@sethladd or @Sfshaza, can one of you update this to be wrapped as the Web site expects and then land this? Thanks!

mit-mit:
I suggest we just add some trailing commas and auto-format, which would give us:

```
import 'package:flutter/material.dart';

void main() {
  runApp(
    new Center(
      child: new Text(
        'Hello, world!',
        textDirection: TextDirection.ltr,
      ),
    ),
  );
}
```

sethladd:
formatting lgtm

mit-mit:
Pushed that change to this PR

mit-mit:
Travis is happy now, merging.

@mattrenquist thanks for contributing, and sorry for being so slow with this PR.

mattrenquist:
No worries. Glad I could help out even a little. Thanks.

PR #682, opened 2017-09-18
International Flutter Apps tutorial

HansMuller:
Note: a few of the link targets don't exist yet because flutter.io isn't up to date with the master branch.


tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
+---
+layout: tutorial
+title: "International Flutter Apps"
```
sethladd: Suggested: Internationalizing Flutter Apps (more action oriented)

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+
+permalink: /tutorials/international/
+---
+
+<div class="whats-the-point" markdown="1">
+
+<b> <a id="whats-the-point" class="anchor" href="#whats-the-point" aria-hidden="true"><span class="octicon octicon-link"></span></a>What you'll learn:</b>
+
+* How to track the device's locale (the user's preferred language).
```
sethladd: Track? Or "determine" ?

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+* How to manage locale-specific app values.
```
sethladd: Is there another word for "values" ? Not sure... but seems really vague...

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+* How to extend the locales that Flutter supports.
```
sethladd: "extend and add" ?

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+
+</div>
+
+<a name="How To Write International Flutter Applications"></a>
```
sethladd: I don't think you need this line. Markdown will add an ID (anchor) based on the section text you have below.

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+## How To Write International Flutter Applications
```
sethladd: We usually use a pattern of "____-ing something" for section titles. Such as "Internationalizing Flutter Applications"

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+
+If your app might be deployed to users who speak another language then
```
sethladd: another compared to what?

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+supports. Flutter provides widgets and classes that help with
+internationalization and the Flutter libraries themselves are
+internationalized.
+
+### The `Locale` class and the `Localizations` Widget
+
+The [`Locale`](https://docs.flutter.io/flutter/dart-ui/Locale-class.html)
+class is used to identify the user's language. Mobile devices support
+setting the locale for all applications, usually via a system settings
+menu. Internationalized apps respond by redisplaying values that are
```
sethladd: redisplaying? that sounds like "display a value, then display that value again". Maybe "displaying" or "display"...

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+locale-specific. For example if the user switches the device's locale
+from English to French then a Text widget that displayed "Hello World"
+would be rebuilt with "Bonjour le monde".
```
sethladd: So, there's a listener on Locale like "onLocalChange" ? If so, should we link to that?

tutorials/international/index.md
HansMuller: I was using "resources" but that's vague and pretentious, so I switched to just "values". 

Usually the values are just strings, but they don't have to be.

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+
+The
+[`Localizations`](https://docs.flutter.io/flutter/widgets/Localizations-class.html)
+widget defines the locale for its child and the localized resources
+that the child depends on. The
+[`WidgetsApp`](https://docs.flutter.io/flutter/widgets/WidgetsApp-class.html)
+widget creates a Localizations widget and rebuilds it if the system's locale changes.
+
+You can always lookup an app's current locale with `Localizations.localeOf()`:
```
sethladd: We might consider a section title like "Determining the user's locale" because I imagine that's a common journey for the developer. Making it an explicit section title makes it easy to find.

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+delegate must define an async
+[`load()`](https://docs.flutter.io/flutter/widgets/LocalizationsDelegate/load.html)
+method that produces an object which encapsulates a collection of
+localized values. Typically these objects define one method per localized value.
+
+In a large app, different modules or packages might be bundled with
+their own localizations. That's why the Localizations widget manages a
+table of objects, one per LocalizationsDelegate. To retrieve the
+object produced by one of the LocalizationsDelegate's `load` methods,
+you specify a BuildContext and the object's type.
```
sethladd: link BuildContext? (we seem to link most everything else :)

tutorials/international/index.md
HansMuller: You don't need an onLocaleChange listener or any listener.

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+class. Instances of this class are created by a LocalizationDelegate
+provided by the
+[`MaterialApp`](https://docs.flutter.io/flutter/material/MaterialApp-class.html)
+ class. They can be retrieved with `Localizations.of`:
+
+```dart
+Localizations.of<MaterialLocalizations>(context, MaterialLocalizations);
+```
+
+This particular `Localizations.of()` expression is used frequently, so the MaterialLocalizations class provides a convenient shortand:
```
sethladd: nit: break this line at 80 chars

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+  return Localizations.of<MaterialLocalizations>(context, MaterialLocalizations);
+}
+
+/// References to the localized values defined by MaterialLocalizations
+/// are typically written like this:
+
+tooltip: MaterialLocalizations.of(context).backButtonTooltip,
+```
+
+### A Class for the App's Localized Resources
```
sethladd: What is the ____-ing equivalent title for this?

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+to look them up.
+
+```dart
+class MyLocalizations {
+  MyLocalizations(this.locale);
+
+  final Locale locale;
+
+  static Future<MyLocalizations> load(Locale locale) {
+    return initializeMessages(locale.toString())
```
sethladd: could we use async/await here?

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+  static MyLocalizations of(BuildContext context) {
+    return Localizations.of<MyLocalizations>(context, MyLocalizations);
+  }
+
+  String get helloWorld => Intl.message('helloWorld', name: 'helloWorld', locale: locale.toString());
+  // ... more Intl.message() methods like get helloWorld
+}
+```
+
+A class based on the `intl` package imports a generated message catalog that provides the `initializeMessages()` function and the per-locale backing store for `Intl.message()`. The message catalog is produced by an `intl` tool that analyzes the source code for classes that contain `Intl.message()` calls. In this case that would just be the `MyLocalizations` class.
```
sethladd: nit: please break line at 80 chars

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+
+One could choose another approach for loading localized resources and looking them up while still conforming to the structure of MyLocalizations. The An Alternative Class for the App's Localized Resources below provides such an example.
```
sethladd: nit: long line

tutorials/international/index.md
HansMuller: I didn't because it's unlikely to be helpful.

https://docs.flutter.io/flutter/widgets/BuildContext-class.html

I guarantee no one reading about i18n will get though the first paragraph :-)

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+
+As you can see, the app's `AppBar` title comes from an instance of
+`MyLocalizations` produced by `MyLocalizations.load()` via
+`MyLocalizationsDelegate`. If the device's locale changes, then the
+app will be rebuilt with a new `MyLocalizations` object, one
+constructed for the new locale. Any reference to
+`MyLocalizations.of()` anywhere in the app will causes its context to
+be rebuilt if the locale changes.
+
+### The App's `supportedLocales` Parameter
```
sethladd: What is a good ____-ing title here? What am I accomplishing in this section?

tutorials/international/index.md
HansMuller: Maybe we don't need an -ing equivalent.

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+
+Although the Flutter's Material library includes support for about 16
+languages, only English language translations are available by
```
sethladd: as a reader, I'm left asking: "how does Flutter support more languages, if only english is available" ?

tutorials/international/index.md
```diff
@@ -0,0 +1,281 @@
...
+       }
+       // ...
+    );
+  }
+}
+
+```
+
+
+### An Alternative Class for the App's Localized Resources
```
sethladd: Curious, will this be confusing to developers? What's our opinion? If our opinion is to use intl, we might not need this section...

tutorials/international/index.md
HansMuller: I'd rather not make this method async

tutorials/international/index.md
HansMuller: The feedback I got from an earlier version of this doc was that a non-intl package example would be helpful. Our opinion is that you can use intl if you want.

tutorials/international/index.md
HansMuller: I did start with that. IMHO it's too many syllables :-)

sethladd: The title is for SEO, so "Internationalizing" or "Internationalization" is likely what people will search for.

tutorials/international/index.md
HansMuller: Track implies determination as well staying in sync.

tutorials/international/index.md
HansMuller: I changed the title of this section to: Tracking the Locale: The `Locale` class and the `Localizations` Widget

HansMuller:
Updated per review feedback.

tutorials/international/index.md
```diff
@@ -0,0 +1,287 @@
+---
+layout: tutorial
+title: "Internationalizing Flutter Apps"
+
+permalink: /tutorials/international/
```
sethladd: could we also update this permalink to match? internationalizing

tutorials/international/index.md
sethladd: fair enough.

tutorials/international/index.md
sethladd: no problem, as long as we had a reason not to use async. When all things are equal, async/await sells Dart more.

PR #683, opened 2017-09-19
Document use of fonts in package dependencies

szakarias:


custom-fonts.md
```diff
@@ -59,6 +59,38 @@ var buttonText = const Text(
...
+declaration above is in the `pubspec.yaml` of a package named `my_package` which the app depends on. Then creating the
+TextStyle is done as follows:
+
+```dart
+const TextStyle(fontFamily: 'Raleway', package: 'my_package')
+```
+
+This is also how the package itself should create the style.
+
+A package can also provide font files in its `lib/` folder which will not automatically be included in the app. Instead the
```
mit-mit: long line

custom-fonts.md
```diff
@@ -59,6 +59,38 @@ var buttonText = const Text(
...
+app can use these selectively when declaring a font. Suppose a package named `my_package` has:
+
+```
+lib/fonts/Raleway-Medium.ttf
+```
+
+Then the app can declare a font like in the example below:
```
mit-mit: Maybe use 'include' rather than 'declare'?

custom-fonts.md
```diff
@@ -59,6 +58,42 @@ var buttonText = const Text(
...
+TextStyle is done as follows:
+
+```dart
+const textStyle = const TextStyle(
+  fontFamily: 'Raleway',
+  package: 'my_package',
+);
+```
+
+This is also how the package itself should create the style.
```
xster: I was reading this doc and had some questions I wasn't clear on. When referring to the 'package itself', is it saying that if a package was to internally consume the font it bundled, it should still say TextStyle(fontFamily: 'name', package: 'self_package_name')? or did you mean something else?

custom-fonts.md
```diff
@@ -59,6 +58,42 @@ var buttonText = const Text(
...
+
+A package can also provide font files in its `lib/` folder which will not automatically be
```
xster: When you say also here, what is the other alternative the also is referring to? Is it putting fonts outside the 'lib' folder and some different behaviour arises? Or not needing to refer to the package's fonts specifically in the app's yaml (like the text above seems to suggest since you didn't say pubspec changes in the app were needed before consuming via const TextStyle(...

custom-fonts.md
szakarias: Yes, that is what I mean.

custom-fonts.md
szakarias: It's not needing to refer to it in the app's yaml. 

If the package declares a font family in its yaml the app can consume as shown in the example since the relevant font files will be bundled with the app. Another use case is a package that has several font files but doesn't necessarily declare a font family in its yaml. Then it's up to the app to declare a font family and choose the relevant files.

custom-fonts.md
Hixie: Please make sure the text (on the Web site and in the API docs since we seem to have the same text in both places) is updated to clarify the issues @xster raised.

PR #687, opened 2017-09-26
Add note about pinning a package ref using the ref argument

mit-mit:


using-packages.md
```diff
@@ -147,6 +147,7 @@ dependencies:
         path: packages/package1        
 ```
 
+Finally, you can use `ref` argument to pin the dependency to a specific git commit, branch, or tag.
```
szakarias: insert 'the' between 'use' and `ref`.

PR #690, opened 2017-10-02
Help SEO

sethladd:
My search fu didn't find this page. Adding moar Flutter.

also, it's a single plugin? I changes "plugins" to "plugin"... because I thought there is only one plugin for Flutter?


sethladd:
cc @mit-mit

mit-mit:
lgtm

(technically it's two plugins; Flutter & Dart, but I think most like you think of it as just one)

PR #694, opened 2017-10-11
Make the everything-is-a-widget principle more about views.

amirh:
I feel like current explanation hints that no logic should live outside of the widget tree, which IIUC is not what we want to say.

### DIFF
--- a/technical-overview.md
+++ b/technical-overview.md
@@ -55,9 +55,9 @@ principles.
 
 ### Everything's a Widget
 
-Widgets are the basic building blocks of a Flutter app's UI. Each widget is an
+Widgets are the basic building blocks of a Flutter app's user interface. Each widget is an
 immutable declaration of part of the user interface.  Unlike other frameworks that
-separate views, view-controllers, layouts, and other properties, Flutter has a
+separate views, view controllers, layouts, and other properties, Flutter has a
 consistent, unified object model: the widget.
 
 A widget can define:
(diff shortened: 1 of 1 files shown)


sethladd:
lgtm

cc @Hixie

technical-overview.md
```diff
@@ -55,17 +55,16 @@ principles.
 
 ### Everything's a Widget
 
-Widgets are the basic building blocks of every Flutter app. Each widget is an
+Widgets are the basic building blocks of a Flutter app's UI. Each widget is an
```
Hixie: expand abbreviations

technical-overview.md
```diff
@@ -55,17 +55,16 @@ principles.
...
 immutable declaration of part of the user interface.  Unlike other frameworks that
-separate views, controllers, layouts, and other properties, Flutter has a
+separate views, view-controllers, layouts, and other properties, Flutter has a
```
Hixie: "view controllers" (no hyphen)

technical-overview.md
```diff
@@ -55,17 +55,16 @@ principles.
...
 consistent, unified object model: the widget.
```
Hixie: we have controllers (separate from widgets) too. And our widgets have properties.

technical-overview.md
```diff
@@ -55,17 +55,16 @@ principles.
...
 
 A widget can define:
 
 *   a structural element (like a button or menu)
 *   a stylistic element (like a font or color scheme)
 *   an aspect of layout (like padding)
-*   some business logic
```
Hixie: i mean, it can...

Hixie:
LGTM modulo minor edits

PR #696, opened 2017-10-12
specify that intellij community is free

sethladd:


sethladd:
For https://github.com/flutter/flutter/issues/12093

cc @filiph or @mit-mit for review

LarkAscending:
Note that the product names have changed... IntelliJ IDEA Community and IntelliJ IDEA Ultimate. No "edition".

sethladd:
Ah, thanks!

alexander-doroshko:
The product names have not changed. But you may use both longer and shorter variant. Even the About dialog contains both.
image

PR #697, opened 2017-10-12
improved analysis options files and fix strong mode issues

devoncarew:
- fix some strong mode issues in the dart code
- update the analysis options files to use strong mode

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,7 +1,11 @@
 analyzer:
-  exclude:
-    - '_includes/**'
-    - '_layouts/**'
+  strong-mode: true
   language:
     enableAssertInitializer: true
     enableSuperMixins: true
+  exclude:
+    - '_includes/**'
+    - '_layouts/**'
+linter:
+  rules:
+    - directives_ordering
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -6,6 +6,3 @@ analyzer:
   exclude:
     - '_includes/**'
     - '_layouts/**'
-linter:
-  rules:
-    - directives_ordering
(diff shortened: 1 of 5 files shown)


analysis_options.yaml
```diff
@@ -1,7 +1,8 @@
 analyzer:
```
mit-mit: Why generally is this file different from https://github.com/flutter/flutter/blob/master/analysis_options.yaml#L24 ?

tool/extract.dart
```diff
@@ -112,5 +111,5 @@ void clean() {
   Iterable<FileSystemEntity> files = exampleDir
       .listSync()
       .where((FileSystemEntity entity) => entity is File && entity.path.endsWith('.dart'));
-  files.forEach((File file) => file.deleteSync());
+  files.forEach((FileSystemEntity file) => file.deleteSync());
```
mit-mit: Why is this needed given the where clause above filters to files only?

devoncarew: This was to fix a strong mode issue, but I updated it to make the definition of `files` more specific (`Iterable<File>`).

analysis_options.yaml
devoncarew: I think just historically, and to avoid copying and pasting copies of the flutter analysis options. I updated this to try using an analysis options `include:` directive; we'll see if travis is happy with that.

Hixie: wow, i didn't know include: was a thing. If it is we may be able to simplify a number of other files.

devoncarew: Yup, developed I think for use by Flutter, but not yet rolled out in use. We'll give it a go here and see how it works. /cc @danrubel

PR #701, opened 2017-10-16
Make i18n tutorials heading match the doc it links to

HansMuller:
International => Internationalizing


Hixie:
LGTM but googlebot is unhappy

PR #702, opened 2017-10-17
Clarify use of fonts from packages

szakarias:


custom-fonts.md
```diff
@@ -71,11 +71,13 @@ const textStyle = const TextStyle(
...
 ```
 
-This is also how the package itself should create the style.
+If the package internally uses the font it defines, it should still specify
+the `package` argument when creating the text style as in the example above.
 
-A package can also provide font files in its `lib/` folder which will not automatically be 
-included in the app. Instead the app can use these selectively when declaring a font. 
-Suppose a package named `my_package` has:
+A package can also provide font files without declaring a font. These files
```
xster: might be vague whether 'declaring' is referring to the pubspec or the TextStyle instantiation.

custom-fonts.md
```diff
@@ -95,6 +97,15 @@ Then the app can declare a font like in the example below:
 
 The `lib/` is implied, so it should not be included in the asset path.
 
+In this case, since it is a locally defined font, the TextStyle is created
```
xster: is this TextStyle instantiation in the package or in the app? I assume both?

szakarias:
Thanks!

custom-fonts.md
szakarias: The app. Changed the text.

PR #708, opened 2017-10-23
Use responsive YouTube embed

filiph:
The default YouTube embed breaks the page on low-width screens (left and right of text beyond viewport, and no way to scroll horizontally to view it).

Reported here: https://twitter.com/SergiAndReplace/status/922425801248854017

### DIFF
--- a/_sass/_responsive_embeds.scss
+++ b/_sass/_responsive_embeds.scss
@@ -1,0 +1,29 @@
+// Use this instead of the default YouTube embed code so that the embed
+// is reposnisve.
+//
+// Example default YouTube embed:
+//     <iframe width="560" height="315"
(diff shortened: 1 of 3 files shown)


filiph:
Here's the problem as a screenshot. It's not possible to scroll left or right on mobile (probably because of viewport meta).

screenshot_20171023-104724

Sfshaza:
Fascinating! Who knew that Youtube would be unresponsive, by default.  ;) 

LGTM!

filiph:
Merged before formal LGTM because the issue breaks the site.

Hixie:
Sfshaza gave a formal LGTM, no?

Sfshaza:
I think he wanted a more technical review from @sethladd. I'm not a go-to web infrastructure sort of person. But it's merged and working, so I think we're good.

PR #711, opened 2017-10-24
Adding hot reload page...

LarkAscending:
Adding new page on hot reload + sidebar entry.

/cc @a-siva @rmacnak-google @mit-mit

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -45,7 +45,7 @@
   <li class="sidebar-title">Test, deploy, workflows</li>
 
     <ul class="sidebar-items">
-      <li><a href="/hot-reload/">Hot reload your app</a></li>
+      <li><a href="/hot-reload/">Hot reload app changes</a></li>
       <li><a href="/testing/">Test your app</a></li>
       <li><a href="/debugging/">Debug your app</a></li>
       <li><a href="/formatting/">Format your source code</a></li>
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -45,6 +45,7 @@
   <li class="sidebar-title">Test, deploy, workflows</li>
 
     <ul class="sidebar-items">
+      <li><a href="/hot-reload/">Hot reload your app</a></li>
       <li><a href="/testing/">Test your app</a></li>
       <li><a href="/debugging/">Debug your app</a></li>
       <li><a href="/formatting/">Format your source code</a></li>
--- a/hot-reload.md
+++ b/hot-reload.md
@@ -1,0 +1,279 @@
+---
+layout: page
+title: Hot Reloading Flutter Apps
+permalink: /hot-reload/
+---
+
+* TOC
+{:toc}
+
+## Using hot reload
+
+Flutter's hot reload feature helps you quickly and easily experiment, build
+UIs, add features and fix bugs. Hot reload works by injecting updated source
+code files into the running Dart Virtual Machine (VM). After the VM updates
+classes with the new versions of fields and functions, the Flutter framework
+automatically rebuilds the widget tree, allowing you to quickly view the effects
+of your changes.
+
+To hot reload a Flutter app:
+
+1.  Run the app from a supported IntelliJ IDE or a terminal window. Either a
+physical or virtual device can be the target.
+1.  Modify one of the Dart files in your project. Most types of code changes can
+be hot reloaded; for a list of changes that require a full restart, see
+[Understanding limitations](#understanding-limitations).
+1.  If you're working in an IntelliJ IDE that supports Flutter's IDE tools,
+click the Hot Reload button on the toolbar:
+
+![alt_text](/images/intellij/hot-reload.gif "image_tooltip")
+
+If you're running the app at the command line using `flutter run`, type `r` in
+the terminal window. 
+
+After a successful hot reload operation, you'll see a message in the console
+similar to:
+
+```
+Performing hot reload...
+Reloaded 1 of 448 libraries in 2,777ms.
+```
+The app's UI is updated to reflect your change, and the current state of the app
+— the value of the counter variable in the above example — is preserved. Your
+app continues to execute from where it was prior to running the hot reload
+command. The code is updated and execution continues.
+
+## Troubleshooting hot reload
+
+When you start working with hot reload and using it frequently, you'll likely
+encounter the types of code changes that might need a full app restart:
+
+
+
+*   Initializing global variables
+*   Initializing static fields
+*   Changing the `main()` method of the app
+
+You can adapt your code to work around the above cases so that hot reload is
+still an option, for example by evaluating an initialization expression after
+making a change. To do this in IntelliJ, you could select the `reinitialize`
+option from the context menu in a field declaration. Without adaptations,
+however, hot reload will generate an exception in these cases.
+
+This section gives detailed examples of edge cases for hot reload where there
+is no active call stack in the Dart VM. They apply to most Flutter app
+development cycles. Other subtle interactions that can occur when there is an
+active call stack (such as when an animation is in progress) are not covered
+here.
+
+### Recent UI change is excluded
+
+Even when a hot reload operation appears successful and generates no exceptions,
+some code changes might not be visible in the refreshed UI. This behavior is
+common after changes to the app's `main()` method.
+
+As a general rule, if the modified code is downstream of the root widget's
+build method, then hot reload behaves as expected. However, if the modified code
+won't be re-executed as a result of rebuilding the widget tree, then you won't
+see its effects after hot reload.
+
+For example, consider the following code:
+```
+import 'package:flutter/material.dart';
+
+void main() {
+  runApp(new MyApp());
+}
+
+class MyApp extends StatelessWidget {
+  Widget build(BuildContext context) {
+    return new GestureDetector(
+      onTap: () => print('tapped'));
+  }
+}
+```
+
+After running this app, you might change the code as follows:
+
+```
+import 'package:flutter/widgets.dart';
+void main() {
+  runApp(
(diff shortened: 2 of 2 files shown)


_includes/sidebar.html
```diff
@@ -45,6 +45,7 @@
   <li class="sidebar-title">Test, deploy, workflows</li>
 
     <ul class="sidebar-items">
+      <li><a href="/hot-reload/">Hot reload your app</a></li>
```
mit-mit: Maybe 'Hot reload app changes'?

hot-reload.md
```diff
@@ -0,0 +1,279 @@
...
+permalink: /hot-reload/
+---
+
+* TOC
+{:toc}
+
+## Using hot reload
+
+Flutter's hot reload feature helps you quickly and easily experiment, build
+UIs, add features and fix bugs. Hot reload works by injecting updated source
```
mit-mit: comma before and?

hot-reload.md
```diff
@@ -0,0 +1,279 @@
...
+
+To hot reload a Flutter app:
+
+1.  Run the app from a supported IntelliJ IDE or a terminal window. Either a
+physical or virtual device can be the target.
+1.  Modify one of the Dart files in your project. Most types of code changes can
+be hot reloaded; for a list of changes that require a full restart, see
+[Understanding limitations](#understanding-limitations).
+1.  If you're working in an IntelliJ IDE that supports Flutter's IDE tools,
+click the Hot Reload button on the toolbar:
```
mit-mit: Hot reloading on save is now the default; see https://flutter.io/intellij-ide/#fast-edit-and-refresh-development-cycle

hot-reload.md
```diff
@@ -0,0 +1,279 @@
...
+the terminal window. 
+
+After a successful hot reload operation, you'll see a message in the console
+similar to:
+
+```
+Performing hot reload...
+Reloaded 1 of 448 libraries in 2,777ms.
+```
+The app's UI is updated to reflect your change, and the current state of the app
```
mit-mit: We don't just update the UI of the app, we update the behavoir of the app aswell.

hot-reload.md
```diff
@@ -0,0 +1,279 @@
...
+```
+myWidget is not a subtype of StatelessWidget
+```
+
+## Understanding limitations
+
+In addition to edge cases, you might also encounter the rare scenarios where hot
+reload is not supported at all.  These include:
+
+*   Changed source files have compilation errors
```
mit-mit: This really isn't a "limitation", it's a feature

hot-reload.md
```diff
@@ -0,0 +1,279 @@
...
+*   Enumerated types are renamed into regular classes or regular classes are
```
mit-mit: 'renamed' sounds odd here, I would use 'changed'

hot-reload.md
```diff
@@ -0,0 +1,279 @@
...
+command. The code is updated and execution continues.
+
+## Troubleshooting hot reload
+
+When you start working with hot reload and using it frequently, you'll likely
+encounter the types of code changes that might need a full app restart:
+
+
+
+*   Initializing global variables
```
mit-mit: It would be good to synchronize this list with the examples given below. For example, `main()` is third on the list here but first in the detail section below.

hot-reload.md
```diff
@@ -0,0 +1,279 @@
...
+```
+Performing hot reload...
+Reloaded 1 of 448 libraries in 2,777ms.
+```
+The app's UI is updated to reflect your change, and the current state of the app
+— the value of the counter variable in the above example — is preserved. Your
+app continues to execute from where it was prior to running the hot reload
+command. The code is updated and execution continues.
+
+## Troubleshooting hot reload
```
mit-mit: I think it would be helpful to explain here in an overview fashion how hot reload actually works so as to give the developers some intuition for why some reloads don't have an effect.

For example, something like this:

Hot reload works by injecting the updated source code into the running Dart VM. As a result, a source code chance only has a visible effect if the changed source code is actually re-run after the change. The following section discussed cases where you may encounter this.

hot-reload.md
LarkAscending: Good point, done!

hot-reload.md
LarkAscending: Removed "UI" qualifier.

_includes/sidebar.html
LarkAscending: Done.

hot-reload.md
LarkAscending: This is optional, but done.

hot-reload.md
LarkAscending: Added "either... or select *File > Save All**.

hot-reload.md
LarkAscending: Changed!

hot-reload.md
LarkAscending: Added a few sentences here...

hot-reload.md
LarkAscending: Users will likely see it as a limitation, and not in a negative way...it can be very good to have something feature-like first on the list.

If you can think of a sentence to add that gives some reassuring context here, I'll add it.

hot-reload.md
```diff
@@ -0,0 +1,281 @@
...
+
+If code changes affect the state of your app (or its dependencies), the Flutter
+framework gives you two options for reconciling old data with new code: throw
+away state or modify the existing state. Throwing away state is what happens
+during a full restart. Depending on the situation, you might prefer to apply
+changes in a more selective way, by marking the widget tree as dirty and
+recomputing the derived layout trees, while leaving the UI state alone. 
+
+Modifying the existing state is what happens during hot reload; however, in this
+scenario the data your app has to work with is not fully consistent with the
```
rmacnak-google: is not -> might not be

hot-reload.md
```diff
@@ -0,0 +1,281 @@
...
+get bar => foo;
+```
+
+Read more about the [differences between the `const` and `final` keywords](https://news.dartlang.org/2012/06/const-static-final-oh-my.html) in Dart. 
+
+
+### Previous state is combined with new code 
+
+If code changes affect the state of your app (or its dependencies), the Flutter
+framework gives you two options for reconciling old data with new code: throw
```
rmacnak-google: Maybe drop "framework"? The implementation is split across the command line tools, the engine and the framework.

hot-reload.md
```diff
@@ -0,0 +1,281 @@
...
+class Color {
+  Color(this.i, this.j);
+  final Int i;
+  final Int j;
+    	}
+    ```
+
+*   Generic type declarations are modified. For example, if you change:
+    ```
+Class A<T> {
```
rmacnak-google: class

hot-reload.md
```diff
@@ -0,0 +1,281 @@
...
+  T i;
+};
```
rmacnak-google: No semicolon

hot-reload.md
```diff
@@ -0,0 +1,281 @@
...
+    ```
+	to:
+
+    ```
+Class A<T, V> {
+  T i;
+  V v;
+};
```
rmacnak-google: No semicolon

hot-reload.md
```diff
@@ -0,0 +1,281 @@
...
+*   Generic type declarations are modified. For example, if you change:
+    ```
+Class A<T> {
+  T i;
+};
+    ```
+	to:
+
+    ```
+Class A<T, V> {
```
rmacnak-google: class

hot-reload.md
```diff
@@ -0,0 +1,279 @@
...
+];
+```
+and then hot reload, the change is not reflected. 
+
+Conversely, in the following example:
+```
+const foo = 1;
+final bar = foo;
+print(foo);
+print(bar);
```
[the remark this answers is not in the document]
LarkAscending: Yes, that works. Good one!

hot-reload.md
```diff
@@ -0,0 +1,279 @@
...
+
+
+running the app for the first time prints `1` and `1`. Then if you make the
+following change: 
+
+```
+const foo = 2;    //modified
+final bar = foo;
+print(foo);
+print(bar);
```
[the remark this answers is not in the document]
LarkAscending: Done.

hot-reload.md
LarkAscending: Good point, done.

hot-reload.md
LarkAscending: Done!

hot-reload.md
LarkAscending: Yep, done.

hot-reload.md
LarkAscending: Correct, good catch!

hot-reload.md
LarkAscending: Correct, good catch!

hot-reload.md
LarkAscending: Yep, done.

Hixie:
LGTM

hot-reload.md
```diff
@@ -0,0 +1,279 @@
...
+
+
+
+*   Initializing global variables
+*   Initializing static fields
+*   Changing the `main()` method of the app
+
+You can adapt your code to work around the above cases so that hot reload is
+still an option, for example by evaluating an initialization expression after
+making a change. To do this in IntelliJ, you could select the `reinitialize`
```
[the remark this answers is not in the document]
mit-mit: I don't think we have this, created a PR to remove it: https://github.com/flutter/website/pull/714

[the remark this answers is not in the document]
LarkAscending: Am working this separately...please see review comments on #714 .

PR #714, opened 2017-10-25
Align hot reload and IntelliJ docs

mit-mit:
1. Refer to hot reload from IntelliJ
1. Change UI instruction to preferred interaction
1. Remove details about 're-initialize' (the product doesn't have that support, as mentioned in the code review)

### DIFF
--- a/hot-reload.md
+++ b/hot-reload.md
@@ -46,7 +46,7 @@ A code change has a visible effect only if the modified Dart code is run again
 after the change. The following section describes common code patterns where the
 modified code will _not_ run again after hot reload.
 
-## Understanding when a full app restart is needed
+## Troubleshooting hot reload
 
 When you start working with hot reload and using it frequently, you'll likely
 encounter the types of code changes that might need a full app restart:
--- a/hot-reload.md
+++ b/hot-reload.md
@@ -7,7 +7,7 @@ permalink: /hot-reload/
 * TOC
 {:toc}
 
-## Using hot reload
+## Using hot reload to quickly view changes
 
 Flutter's hot reload feature helps you quickly and easily experiment, build
 UIs, add features, and fix bugs. Hot reload works by injecting updated source
@@ -23,13 +23,10 @@ physical or virtual device can be the target.
 1.  Modify one of the Dart files in your project. Most types of code changes can
 be hot reloaded; for a list of changes that require a full restart, see
 [Understanding limitations](#understanding-limitations).
-1.  If you're working in an IntelliJ IDE that supports Flutter's IDE tools,
-either click the Hot Reload button on the toolbar:
-
-![alt_text](/images/intellij/hot-reload.gif "image_tooltip")
-
-or select **File > Save All**. If you're running the app at the command line 
-using `flutter run`, type `r` in the terminal window. 
+1.  If you're working in an IntelliJ IDE that supports Flutter's IDE tools, select Save All (`cmd-s`/`ctrl-s`), or
+click the Hot Reload button on the toolbar:
+   ![alt_text](/images/intellij/hot-reload.gif "image_tooltip")
+   If you're running the app at the command line using `flutter run`, type `r` in the terminal window. 
 
 After a successful hot reload operation, you'll see a message in the console
 similar to:
@@ -47,7 +44,7 @@ A code change has a visible effect only if the modified Dart code is run again
 after the change. The following section describes common code patterns where the
 modified code will _not_ run again after hot reload.
 
-## Troubleshooting hot reload
+## Understanding when a full app restart is needed
 
 When you start working with hot reload and using it frequently, you'll likely
 encounter the types of code changes that might need a full app restart:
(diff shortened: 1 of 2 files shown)


hot-reload.md
```diff
@@ -7,7 +7,7 @@ permalink: /hot-reload/
 * TOC
 {:toc}
 
-## Using hot reload
+## Using hot reload to quickly view changes
```
LarkAscending: Prefer the previous version. Headings must be simple and short. The splash page on flutter.io says this, so users will already know. Also, we elaborate below.

hot-reload.md
```diff
@@ -23,13 +23,12 @@ physical or virtual device can be the target.
 1.  Modify one of the Dart files in your project. Most types of code changes can
 be hot reloaded; for a list of changes that require a full restart, see
 [Understanding limitations](#understanding-limitations).
-1.  If you're working in an IntelliJ IDE that supports Flutter's IDE tools,
-either click the Hot Reload button on the toolbar:
+1.  If you're working in an IntelliJ IDE that supports Flutter's IDE tools, select Save All (`cmd-s`/`ctrl-s`), or
```
LarkAscending: Which of these is the more common scenario? It should go first in the sequence.

hot-reload.md
```diff
@@ -23,13 +23,12 @@ physical or virtual device can be the target.
...
+click the Hot Reload button on the toolbar:
 
-![alt_text](/images/intellij/hot-reload.gif "image_tooltip")
+   ![alt_text](/images/intellij/hot-reload.gif "image_tooltip")
```
LarkAscending: Nice! The image should also probably be resized, eventually.

hot-reload.md
```diff
@@ -23,13 +23,12 @@ physical or virtual device can be the target.
...
 
-or select **File > Save All**. If you're running the app at the command line 
-using `flutter run`, type `r` in the terminal window. 
+   If you're running the app at the command line using `flutter run`, type `r` in the terminal window.
```
LarkAscending: Missing bold formatting for File > Save All. https://developers.google.com/style/ui-elements

hot-reload.md
```diff
@@ -47,7 +46,7 @@ A code change has a visible effect only if the modified Dart code is run again
 after the change. The following section describes common code patterns where the
 modified code will _not_ run again after hot reload.
 
-## Troubleshooting hot reload
+## Understanding when a full app restart is needed
```
LarkAscending: @a-siva @rmacnak-google  This set of changes also changes the meaning. We discussed how to work around the edge cases, right? Is this now the same as the last section "Understanding limitations"?

What point do we want to make? I'm perfectly willing to research whether it's possible to reinitialize in IntelliJ... if that's the direction you want to go in.

intellij-ide.md
```diff
@@ -101,16 +101,8 @@ You need to connect a device, or start a simulator, to proceed.
...
-![Hot reload in IntelliJ](/images/intellij/hot-reload.gif)
-
-To edit your code and hot reload changes:
-
-1. Start your app with 'Run' (without breakpoints) or 'Debug' (with breakpoints).
-1. Make desired changes in your source code.
-1. Do not press the 'Stop' button; let your app continue to run.
-1. To see your changes invoke **Save All** (`cmd-s` / `ctrl-s`),
-   or click the **Hot Reload button** (the button with the lightning bolt icon).
+your changes near instantly with it's 'hot reload' support. See the [hot reload page](/hot-reload/)
```
LarkAscending: Recommend rewording to "almost instantly with the hot reload feature.  See Hot Reloading Flutter Apps (link) for details."

If we keep this version "it's" > "its" .
https://developers.google.com/style/possessives

hot-reload.md
mit-mit: Change reverted

hot-reload.md
mit-mit: Keyboard shortcut is the most common, that is why I reordered them

hot-reload.md
mit-mit: Done

hot-reload.md
mit-mit: Yes, I intentionally changed the meaning.

'Troubleshooting' is something you do when something is broken and doesn't have it's intended function. That is not the issue here. Hot reload behaves as designed, and for good reasons.

LarkAscending: We could change the header to "Refactoring to enable hot reload"... WDYT?

Re: the meaning of the word "troubleshooting", this is from the user's perspective which is different from ours. The content is about hot reload not behaving the way a user was expecting it to, and they're going to make small code changes to remedy that. The question of whether or not they're in this situation due to team design decisions and/or technical constraints, that mixes our perspective in where it's not relevant. Their goal is to enable hot reload in as many situations as possible...

The following section, "Understanding limitations", is about full app restart. That's when no code changes can help.

mit-mit: In the spirit of landing this PR, I have reverted this change.

I'll be in MTV next week, then we can spend some together on this topic.

PR #717, opened 2017-10-26
update pubspecs and docs to use material font package

xster:
Submit after https://github.com/flutter/flutter/pull/12730


xster:
Holding off for a while

PR #721, opened 2017-10-27
Add RecyclerView reference

mehmetf:
Android developers typically do not develop with ListView. Almost everyone knows RecyclerView is a more efficient way of building lists just like ListView.Builder is in Flutter. We should mention it in the doc.


sethladd:
cc @FaisalAbid to double check and review

PR #723, opened 2017-11-01
Link to RenderObject doc from the Framework Tour

InMatrix:
Make information about RenderObject easier to discover.


InMatrix:
Should we link to RenderBox instead of RenderObject? RenderBox is the one I really want to expose more here, though the original doc says RenderObject. However, in this the [Box Constraints](https://flutter.io/layout/) doc, it says "In Flutter, widgets are rendered by render boxes."

Hixie:
LGTM

PR #724, opened 2017-11-01
Add links to API docs

InMatrix:


InMatrix:
The main class I want to expose is RenderBox, but since I was at it, I added links for all the classes mentioned in this doc.

Hixie:
LGTM

PR #729, opened 2017-11-07
Update using-packages.md

mit-mit:
Addressing https://github.com/dart-lang/site-www/issues/421#issuecomment-342307259

### DIFF
--- a/using-packages.md
+++ b/using-packages.md
@@ -18,6 +18,9 @@ platform SDKs (like
 
 If you are looking to develop a new package, please see [developing packages](/developing-packages/).
 
+If you are looking to add assets, images, or fonts to your application, please
+see [Assets & Images](https://flutter.io/assets-and-images/).
+
 * TOC
 {:toc}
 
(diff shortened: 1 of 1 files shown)


using-packages.md
```diff
@@ -18,6 +18,9 @@ platform SDKs (like
 
 If you are looking to develop a new package, please see [developing packages](/developing-packages/).
 
+If you are looking to add assets, images, or fonts to your application, please
```
sethladd: maybe hint as to why this is related to Using Packages ?

e.g. "Some package provide assets such as fonts and images. Learn more...."

sethladd:
lgtm but I think we can clarify the relationship between packages and assets, especially for the "I'm using a package that has an asset, and I want to load and use that asset" use case.

using-packages.md
mit-mit: Would just "If you are looking to add assets, images, or fonts, whether stored in files or packages, please..." work?

sethladd: sure!

mit-mit:
thanks for the quick review!

PR #732, opened 2017-11-10
Fix for broken Text widget example

mjohnsullivan:
Fix for broken Text widget example that now requires directionality to be defined; I added some text outlining why this is now needed.

### DIFF
--- a/widgets-intro.md
+++ b/widgets-intro.md
@@ -35,7 +35,8 @@ with a widget:
 import 'package:flutter/material.dart';
 
 void main() {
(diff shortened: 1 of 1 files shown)


Hixie:
LGTM

marceloHashzen:
I've just started my first tutorial to learn Flutter. Just thought it is nonsense have to directly say that the text should render from left to right, it should be default like every other language or framework...

Hixie:
@marceloHashzen why should the default be left to right rather than right to left?

see also the discussion at https://docs.flutter.io/flutter/dart-ui/TextDirection-class.html

PR #734, opened 2017-11-16
Added missing opening [[highlight]] tags

Skylled:
I'm not 100% sure if the new tag placement is correct, so please double-check.
I spotted this problem because it made `</span>` tags appear in the example code like so:
```
    final Animation<double> animation = listenable;
    return new Center(
      child: new Opacity(</span>
        opacity: _opacityTween.evaluate(animation),</span>
        child: new Container(
          margin: new EdgeInsets.symmetric(vertical: 10.0),
          height: _sizeTween.evaluate(animation),
          width: _sizeTween.evaluate(animation),
          child: new FlutterLogo(),
        ),
      ),
    );
  }
```

### DIFF
--- a/tutorials/animation/index.md
+++ b/tutorials/animation/index.md
@@ -801,8 +801,8 @@ class AnimatedLogo extends AnimatedWidget {
   Widget build(BuildContext context) {
     final Animation<double> animation = listenable;
     return new Center(
-      child: new Opacity([[/highlight]]
-        opacity: _opacityTween.evaluate(animation),[[/highlight]]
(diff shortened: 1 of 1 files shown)


sethladd:
Thanks!

@Sfshaza can you take a quick look? is this PR addressing the original intention?

mit-mit:
cc @Sfshaza can you take a quick look? It looks good to me.

Sfshaza:
Sorry I missed this before and thanks for catching, @Skylled!  (I was on vacay for awhile.)

Your fix looks good, but I found and fixed this before going on vacation.

PR #735, opened 2017-11-16
Adding examples, removing "chain" from animation tutorial.

Sfshaza:
Before I leave on vacay, I'd like to tweak the animation tutorial a bit. Also adding staggered animation examples.

@HansMuller


_includes/code/animation/basic_staggered_animation/main.dart
```diff
@@ -0,0 +1,204 @@
...
+
+import 'dart:async';
+
+import 'package:flutter/material.dart';
+import 'package:flutter/scheduler.dart' show timeDilation;
+
+class StaggerAnimation extends StatelessWidget {
+  StaggerAnimation({ Key key, this.controller }) :
+
+    // Each animation defined here transforms its value between the subset of
```
HansMuller: This comment, which appears to be just as I wrote it, makes no sense.

I suppose this is better:
```
    // Each animation defined here transforms its value during the subset
    // of the controller's duration defined by the animation's interval.
    // For example the opacity animation transforms its value during
    // the first 10% of the controller's duration.
    //
    // Most of the Intervals cause the animations' transforms to be
    // "staggered", i.e. to start one after another.
```

_includes/code/animation/basic_staggered_animation/main.dart
```diff
@@ -0,0 +1,204 @@
...
+
+  final Animation<double> controller;
+  final Animation<double> opacity;
+  final Animation<double> width;
+  final Animation<double> height;
+  final Animation<EdgeInsets> padding;
+  final Animation<BorderRadius> borderRadius;
+  final Animation<Color> color;
+
+  Widget _buildAnimation(BuildContext context, Widget child) {
```
HansMuller: It might be worth putting a comment here.

```
  // This function is called each the controller "ticks" a new frame.
  // When it runs, all of the animation's values will have been
  // updated to reflect the controller's current value.
```

HansMuller:
LGTM

I think a few changes in https://flutter.io/animations/overview.html are needed too. The section called "Chaining animatables" could just be called "Composing animatables" (the method is still called .chain(), there's not changing that for now).

Similarly the "Chainable animations" could just be renamed "Composing animations".

Sfshaza:
Thanks! I'd like to remove overview eventually, but I updated it for now.

PR #736, opened 2017-11-18
Reordering sections in hot reload

LarkAscending:
Sections are now ordered according to the user's ability to adapt their code and enable hot reload:
1. Preconditions 
1. State
1. Dart VM variants of state
1. Limitations

Also:
* Moved all interesting scenarios up to the same level (rather than nested under Troubleshooting, Understanding Limitations).
* Added error message that happens when the syntax is wrong.
* Made small text tweaks to make the new structure work.

This PR resolves open comments on https://github.com/flutter/website/pull/714
/cc @a-siva @mjohnsullivan

### DIFF
--- a/hot-reload.md
+++ b/hot-reload.md
@@ -45,7 +45,7 @@ command. The code is updated and execution continues.
 
 A code change has a visible effect only if the modified Dart code is run again
 after the change. The next sections describe common situations where the
-modified code will _not_ run again after hot reload. In some cases,  
+modified code will _not_ run again after hot reload. In some cases, 
 small changes to the Dart code will enable you to continue using hot reload
 for your app.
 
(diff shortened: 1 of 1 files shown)


hot-reload.md
```diff
@@ -43,74 +44,85 @@ app continues to execute from where it was prior to running the hot reload
...
-With a full restart, the program starts from the beginning, executes the new
-version of `main()`, and builds a widget tree that displays the text `Hello`.
-
-However, if you hot reload the app after this change, `main()` is not 
-re-executed, and the widget tree is rebuilt with the unchanged instance of 
-`MyApp` as the root widget. The result is no visible change after hot reload.
-
+```
+myWidget is not a subtype of StatelessWidget
+```
```
mit-mit: I think we need to add a comment under this failure message along the lines of "In these cases, issue a full restart to see the updated application."

mit-mit:
I really like this organization. I think it flows very well now!

I made two smaller comments.

mit-mit:
lgtm!

PR #738, opened 2017-11-20
Update android-setup.md

mit-mit:
Fixes https://github.com/flutter/flutter/issues/13071


xster:
I think we'd want to keep both instructions. Oreo is just 0.3% of the market share :S

LarkAscending:
Yes, the DAC site uses this wording (and an extra step);

> 2. (Only on Android 8.0 or higher) Select System.

Any objections if we just link to the [Android page](https://developer.android.com/studio/debug/dev-options.html) for dev mode/debugging instead?

mit-mit:
Updated to refer to Android docs, PTAL

mit-mit:
@LarkAscending @xster can you take another look, please?

xster:
LG

PR #742, opened 2017-11-28
Fix 'Getting Started' action buttons on front page

mit-mit:
Currently the "big blue" getting started buttons link to the second subitem ('Create and run your first app') in leftnav, which then in a big blue call-out tells the user to go to the first subitem ('Flutter installation & setup').

This PR streamlines that so it actually goes directly to the first item.

### DIFF
--- a/index.md
+++ b/index.md
@@ -21,7 +21,7 @@ hide_title: true
         </div>
 
         <div class="homepage__button_row">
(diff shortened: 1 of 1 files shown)


sethladd:
lgtm

LarkAscending:
LGTM (and thanks for taking care to change both the buttons, not just the obvious top one)!

PR #746, opened 2017-12-02
Update link text

LarkAscending:
Change "See IntelliJ Setup" > "See IDE Setup" to do two things:
* Better match the left nav entry "Setting up the IDE". These are still different, which is not optimal, but more similar now.
* Accommodate AS as an IDE option.


mit-mit:
lgtm

PR #747, opened 2017-12-05
Add more widget test debug tips

xster:


Hixie:
You may also want to link to the debugging doc.

Also a hint we should list here is that you can `flutter run` a test to see it on the device, and tap things on the screen to get finders for those things. (That's broken right now but I'm fixing it so don't worry about it being broken.)

Hixie:
LGTM

xster:
@Hixie, are you referring to https://flutter.io/debugging? Seems like it's a bit too live debug focused for this use case

xster:
Ummm seems like I can't rebase branches on the flutter user :S

The travis issue is fixed in https://github.com/flutter/website/pull/749. Going to submit and monitor.

PR #754, opened 2017-12-08
Renaming example name to use underbars, not dashes

Sfshaza:
To comply with Dart naming conventions.

In Hans's absence, can you review, @Hixie?


Hixie:
LGTM

PR #755, opened 2017-12-08
Make it easy to find packages

sethladd:


sethladd:
cc @mit-mit

mit-mit:
lgtm

great idea, thanks!

PR #756, opened 2017-12-12
Add a link to China instructions

InMatrix:


mit-mit:
This seems to visually too distinct and upfront for a note that applies to less than the majority of our users.

InMatrix:
@mit-mit This message is likely to apply to the majority of our new users this week, so I'd like to make it easy to notice. I'm happy to downplay it next week.

Hixie:
I don't think it's _that_ visually distinct. Certainly LGTM for this week, but I it may even be reasonable in general going forward, IMHO.

InMatrix:
I'm going to merge this. Sorry about having to do it without consensus. We can certainly revisit the placement of the message next week.

mit-mit:
No worries, I'm fine with this for now

PR #757, opened 2017-12-12
Update change policy to mention updating changelog

Hixie:


sethladd:
lgtm!

Hixie:
The "push" build fails presumably because this was submitted when the tree was broken, but the "pr" build passes presumably because if I land it'll be green? I'm going to try landing and hope that everything works out...

PR #760, opened 2017-12-13
Update `build()` to return a MaterialApp.

pq:
As written, `build()` produces a runtime exception:

```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
No Directionality widget found.
RichText widgets require a Directionality widget ancestor.
The specific widget that could not find a Directionality ancestor was:
The ownership chain for the affected widget is:
RichText ← Text ← Semantics ← Center ← Padding ← Container ← Listener ← _GestureSemantics ←
RawGestureDetector ← GestureDetector ← ⋯
Typically, the Directionality widget is introduced by the MaterialApp or WidgetsApp widget at the
top of your application widget tree. It determines the ambient reading direction and is used, for
example, to determine how to lay out text, how to interpret "start" and "end" values, and to resolve
EdgeInsetsDirectional, AlignmentDirectional, and other *Directional objects.
```

Wrapping the returned `Material` in a `MaterialApp` fixes this.

simulator screen shot - iphone 8 - 2017-12-13 at 05 58 24

(Unused `_counter` reference is also removed.)

/cc @mit-mit @mravn-google

### DIFF
--- a/platform-channels.md
+++ b/platform-channels.md
@@ -128,7 +128,6 @@ import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 ...
 class _MyHomePageState extends State<MyHomePage> {
-  int _counter = 0;
   static const platform = const MethodChannel('samples.flutter.io/battery');
 
   // Get battery level.
@@ -171,17 +170,19 @@ refreshing the valuer.
 ```dart
 @override
 Widget build(BuildContext context) {
-  return new Material(
-    child: new Center(
-      child: new Column(
-        mainAxisAlignment: MainAxisAlignment.spaceEvenly,
-        children: [
-          new RaisedButton(
-            child: new Text('Get Battery Level'),
-            onPressed: _getBatteryLevel,
-          ),
-          new Text(_batteryLevel),
-        ],
+  return new MaterialApp(
+    home: new Material(
+      child: new Center(
+        child: new Column(
+          mainAxisAlignment: MainAxisAlignment.spaceEvenly,
+          children: [
+            new RaisedButton(
+              child: new Text('Get Battery Level'),
+              onPressed: _getBatteryLevel,
+            ),
+            new Text(_batteryLevel),
+          ],
+        ),
       ),
     ),
   );


platform-channels.md
```diff
@@ -171,17 +170,19 @@ refreshing the valuer.
...
-        mainAxisAlignment: MainAxisAlignment.spaceEvenly,
-        children: [
-          new RaisedButton(
-            child: new Text('Get Battery Level'),
-            onPressed: _getBatteryLevel,
-          ),
-          new Text(_batteryLevel),
-        ],
+  return new MaterialApp(
+    home: new Material(
```
mravn-google: We probably don't need the `Material` then.

pq: If you mean something like this:

```dart
  @override
    return new MaterialApp(
        home: new Center(
          child: new Column(
            mainAxisAlignment: MainAxisAlignment.spaceEvenly,
            children: [
              new RaisedButton(
                child: new Text('Get Battery Level'),
                onPressed: _getBatteryLevel,
              ),
              new Text(_batteryLevel),
            ],
          ),
      ),
    );
  }
```

I actually tried it however it looks like `MaterialButton` widgets require a `Material` widget ancestor.

```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
The following assertion was thrown building MaterialButton(dirty, state:
No Material widget found.
MaterialButton widgets require a Material widget ancestor.
In material design, most widgets are conceptually "printed" on a sheet of material. In Flutter's
material library, that material is represented by the Material widget. It is the Material widget
that renders ink splashes, for instance. Because of this, many material library widgets require that
there be a Material widget in the tree above them.
To introduce a Material widget, you can either directly include one, or use a widget that contains
Material itself, such as a Card, Dialog, Drawer, or Scaffold.
The specific widget that could not find a Material ancestor was:
MaterialButton
The ownership chain for the affected widget is:
MaterialButton ← RaisedButton ← Column ← Center ← _ModalScopeStatus ←
PageStorage-[GlobalKey#f01fd] ← RepaintBoundary ← Stack ← _CupertinoBackGestureDetector ←
DecoratedBox ← ⋯
```

mravn-google: Ah, right. `MaterialApp` is usually used with a `Scaffold` which is what supplies the `Material` I guess. Nevermind then.

mravn-google: For the record, the corresponding code in https://github.com/flutter/flutter/blob/master/examples/platform_channel/lib/main.dart does wrap the `Material` widget in a `MaterialApp`. It just does so in `main` rather than in the `build` method from which the code for the web page was copied.

Hixie:
Isn't this supposed to replace the MyHomePage from the `flutter create` app? In which case, the MaterialApp is already present, since it's created by the MyApp widget.

pq:
@Hixie : you know, I think I overstepped here.  The example is maybe a little confusing but I don't like that my remedy means the sample code diverges from the full example [source](https://github.com/flutter/flutter/blob/master/examples/platform_channel/lib/main.dart).  I'll fix that up.

Thanks for the catch!

mit-mit:
Current changes look good, thanks

PR #766, opened 2017-12-18
Make fragment identifier links visible below the header.

Hixie:


Hixie:
Fixes https://github.com/flutter/flutter/issues/10525

_sass/_layout.scss
```diff
@@ -191,3 +191,9 @@
         }
     }
 }
+
+/* Make fragment identifier links visible below the header. */
+:target:before {
+    content: "";
+    display: block;
+    height: 112px; }
```
yjbanov: nit: the rest of the file places closing `}` on the next line.

yjbanov:
In fact, RSLGTM!

_sass/_layout.scss
Hixie: Well that's messed up. That's how I had it, but then I noticed the rest of the file it was in was different, so I changed it, then found it was the wrong file, then moved it to this file, and the style is different. -_-

sethladd:
thank you!

PR #768, opened 2017-12-21
fix layout.md

adwd:
Hi! Thank you for great framework and documentation.
I removed unnecessary parenthesis and line break from `layout.md`.

previous

fixed

And I also removed unnecessary slash from `developing-packages.md`, which fails `rake checklinks`.


sethladd:
lgtm thank you!

PR #770, opened 2017-12-28
Changed compound tween example to fade in as size increases

appsforartists:
It makes more sense for something to become easier to see as it gets bigger than it does for it to vanish as it does.


sethladd:
Thank you very much!

cc @HansMuller @Sfshaza for their quick review

Sfshaza:
Thanks, @appsforartists!

PR #774, opened 2018-01-07
 get-sdk.md: Update installation docs

adi23arora:
This clarifies the way dependencies are
downloaded when any flutter command is
executed for the first time.

### DIFF
--- a/_includes/setup/get-sdk.md
+++ b/_includes/setup/get-sdk.md
@@ -42,7 +42,7 @@ For example:
       visit https://flutter.io/setup/#android-setup for detailed instructions.
 ```
 
-The first time you run the `flutter` command, it downloads its own dependencies and compiles
+The first time you run a flutter command (such as `flutter doctor`), it downloads its own dependencies and compiles
 itself. Subsequent runs should be much faster.
 
 The following sections describe how to perform these tasks and finish the setup process.


adi23arora:
I signed it!

mit-mit:
You can actually run the `flutter` command with any sub-command (`doctor`, `run`, etc.) and it will download dependencies if needed.

adi23arora:
@mit-mit It looks very confusing for new users because first running ``flutter doctor`` already downloads that dependencies and when running ``flutter`` command it shows the list of available options to be used with ``flutter`` and seems that there's no use of executing ``flutter`` command only (just provides available uses of command as shown with ``--help`` or ``-h`` option or argument)

mit-mit:
Got it. Maybe we can do something like this:

The first time you run a flutter command (such as `flutter doctor`), it ...

adi23arora:
Yes, this looks better and more helpful than the current one.

mit-mit:
Great, and you push an update to that in this PR then?

adi23arora:
Yes, sure.

adi23arora:
@mit-mit Updated PR, please review.

mit-mit:
Looks good, thanks for contributing!

PR #777, opened 2018-01-16
Possible code fix

kentcb:
I'm brand new to Dart+flutter, so I may have this wrong. But it seems to me that `curve` should be using `controller` as the `parent`?


kentcb:
CLA signed.

Hixie:
LGTM
Thanks!

We should probably change this to use a CurveTween instead of a CurvedAnimation though.

Hixie:
Thank you for your contribution!

PR #778, opened 2018-01-17
Fix networking.md to use proper http paths

ericmand:
See https://api.dartlang.org/stable/1.24.3/dart-core/Uri/Uri.http.html


ericmand:
I signed it!

sethladd:
Thanks! lgtm

PR #782, opened 2018-01-22
Improve the Flutter arch layer cake

QuirijnGB:
To my understanding Material widgets and Cupertino widgets are siblings, Material isn't built on top of Cupertino. I updated the SVG to reflect that.


Hixie:
Material is build on top of Cupertino. Right now it's pretty limited, e.g. MaterialPageRoute depends on CupertinoPageRoute, and the material scroll bar and text editing toolbar widgets defer to cupertino, but eventually we'll be doing this for all the widgets.

QuirijnGB:
As per @Hixie's comment

PR #783, opened 2018-01-22
Upgrade Nokogiri

mit-mit:


firebase.json
```diff
@@ -57,6 +57,11 @@
         "source" : "/intellij-ide/",
         "destination" : "/using-ide/",
         "type" : 301
+      },
+      {
+        "source": "/app/appcastwin.json",
+        "destination": "http://flutterapp.com",
+        "type": 301
```
sethladd: curious, why do we do this?

can we leave a comment in this file?

PR #787, opened 2018-01-30
highlight the existing assists a bit better

devoncarew:
@mit-mit @Sfshaza @stevemessick


mit-mit:
formatting changes lgtm

PR #796, opened 2018-02-12
Update json_serializable and build_runner versions & remove obsolete library directive from sample

roughike:
Fixes flutter/flutter#14645.

Confirmed working with Flutter 0.1.0.


sethladd:
rslgtm

PR #799, opened 2018-02-14
Initial version of new Getting Started Flow

mit-mit:
Just for conceptual review

Only Android Studio and Terminal flows are filled out.


mit-mit:
Fixed uploader in  PR 800

PR #803, opened 2018-02-16
Move imports above code depending on imports

mit-mit:
Fixes https://github.com/flutter/flutter/issues/14738


InMatrix:
Thanks for the fix. Could you move imports above code in the Kotlin version of the example as well?

mit-mit:
Done, PTAL

mit-mit:
cc @InMatrix please take another look

InMatrix:
image

PR #804, opened 2018-02-16
Adding a "jank no more" page.

Sfshaza:
This is not ready to merge. I'm throwing it up so that it can be seen.

Staged: https://sz-flutter-2.firebaseapp.com/ui-performance/


Sfshaza:
Sharing this with folk from today's meeting.  

@filiph @wmleler

ui-performance.md
```diff
@@ -0,0 +1,470 @@
...
+* TOC Placeholder
+{:toc}
+
+## A workflow for diagnosing jank
+
+<div class="whats-the-point" markdown="1">
+
+<b> <a id="whats-the-point" class="anchor" href="#whats-the-point" aria-hidden="true"><span class="octicon octicon-link"></span></a>What's the point?</b>
+
+* Profile your app on a physical device.
```
xster: Seems like these bullet points are addressing 'steps to diagnose' rather than 'what's the point' which is the header of this section.

ui-performance.md
```diff
@@ -0,0 +1,470 @@
...
+* Check performance on the slowest device your users might use.
+* Start by enabling the performance overlay.
+
+</div>
+
+### Connect to a physical device
+
+Almost all performance debugging for Flutter applications should be
+conducted on a physical Android or iOS device, with your Flutter application
+running in profile mode. Using debug mode, or running apps on simulators
```
xster: I would put profile and debug modes in backquotes since they refer to a very specific thing.

ui-performance.md
```diff
@@ -0,0 +1,470 @@
...
+or emulators, is generally not indicative<sup>*</sup> of the final
+behavior of release mode builts. _You should consider checking performance
+on the slowest device that your users might reasonably use._
+
+<aside class="alert alert-info" markdown="1">
+<sup>*</sup> Why you should run on a real device:
```
xster: The header is referring to the real vs simulator dimension but the bullet points below are mixing the real vs simulator dimension and the debug vs release dimension

ui-performance.md
```diff
@@ -0,0 +1,470 @@
+---
```
Hixie: note the screen shots will need updating; we changed the style of the performance graphs recently for some reason.

ui-performance.md
```diff
@@ -0,0 +1,470 @@
...
+The
+[PerformanceOverlay](https://docs.flutter.io/flutter/widgets/PerformanceOverlay-class.html)
+widget, from the
+[widget library](https://docs.flutter.io/flutter/widgets/widgets-library.html),
+displays performance statistics using two graphs.
+The following screenshot shows the performance overlay running on the Flutter
+Gallery example:
+
+<center><img src="images/performance-overlay-green.png" alt="screenshot of performance overlay showing zero jank"></center>
+<center>The graphs produced by the performance overlay.<br>THe vertical green bar represents the current frame.</center>
```
Hixie: typo `THe` -> `The`

ui-performance.md
```diff
@@ -0,0 +1,470 @@
...
+Flutter’s profile mode compiles and launches your application almost
+identical to release mode, but with just enough additional functionality
+to allow debugging performance problems.
+For example, profile mode provides tracing information to
+[Observatory](https://dart-lang.github.io/observatory/) and other tools.
+
+Launch the app in profile mode as follows:
+
+<ul markdown="1">
+<li markdown="1">In Android Studio, use the **Run > Profile…** menu item.
```
devoncarew: The Android Studio steps are the same as the IntelliJ ones.

ui-performance.md
```diff
@@ -0,0 +1,470 @@
...
+
+### Displaying the performance overlay
+
+You can toggle the performance overlay in several ways:
+
+* Using the Flutter Inspector
+* From the command line
+* Programmatically
+
+Flutter Inspector is available in the VSCode, Android Studio, and IntelliJ plugins,
```
devoncarew: The Flutter Inspector...

devoncarew: You can strike VSCode here - the long term plan is to have it available, but not near term.

ui-performance.md
```diff
@@ -0,0 +1,470 @@
...
+
+The easiest way to enable the PerformanceOverlay widget is by enabling it in the Flutter
+Inspector, which is available through the Flutter plugin for your IDE. The Inspector
+view opens by default when running an application. If the inspector is not available,
+you can display it as follows.
+
+In Android Studio and IntelliJ IDEA:
+
+1. Select **View > Tool Windows > Flutter Inspector**.
+1. In the gear's menu, select **Show Performance Overlay**.
```
devoncarew: This is now accessed from the Inspector toolbar (not the gear menu).

screen shot 2018-03-02 at 7 41 29 pm

The `Toggle Performance Overlay` item.

Sfshaza: Oh, thanks for letting me know, @devoncarew!

filiph:
FYI: You'll have to merge with master before landing this, otherwise it won't build (there's a breakage in the latest `flutter` tool which makes Travis builds fail, and which I "fixed" by pinning the Flutter tool on travis to the `beta` channel).

wmleler:
Let me know when this is landed so I can put a link to it in the "Why
Flutter uses Dart" article.

ui-performance.md
```diff
@@ -0,0 +1,470 @@
...
+The following screenshot shows the performance overlay running on the Flutter
+Gallery example:
+
+<center><img src="images/performance-overlay-green.png" alt="screenshot of performance overlay showing zero jank"></center>
+<center>The graphs produced by the performance overlay.<br>THe vertical green bar represents the current frame.</center>
+
+Each graph represents the last 300 frames for the following two threads used in every Flutter
+app:
+
+* Engine<br>
```
xster: These names were confusing (since the `Engine` label really mainly pointed to code running in the framework). We renamed them UI and GPU instead to match the threads.

ui-performance.md
```diff
@@ -0,0 +1,470 @@
...
+    it takes you directly to the timeline view.
+    If you are looking at Observatory’s front page,
+    click the “view <ul>timeline</ul>” link.
+
+    Initially, timeline page contains no timeline data. In the **Recorded Streams Profile**,
+    select the **Flutter Developer profile** radio button. This simultaneously selects the **Dart**,
+    **Embedder**, and **GC** streams.
+</li>
+
+<li markdown="1"> Run the app and get it to the state that you want to examine.
```
xster: I'd use a different word than 'run' since the app is already running (that's how we got the observatory URL).

ui-performance.md
```diff
@@ -0,0 +1,470 @@
...
+to make the framework slow. You can learn more by adding custom slices to the timeline
+using the [Timeline](https://docs.flutter.io/flutter/dart-developer/Timeline-class.html)
+class in the
+[dart:developer](https://docs.flutter.io/flutter/dart-developer/dart-developer-library.html)
+package.
+
+[PENDING: would be nice to have example of code with time slices, but not required]
+
+Iterate until you can determine what in your code is causing the performance issue.
+
```
xster: @chinmaygarde can explain it better but my understanding is that there's a limit to how much timeline data we're storing in memory and when refreshing, only the last few seconds' data will be shown in Chrome tracing. We should describe it so people won't be wondering why their Timeline calls aren't appearing.

ui-performance.md
wmleler: Also, the ribbon in the upper right now says "DEBUG" instead of "SLOW MODE".

Sfshaza:
This PR has been replaced by [this one](https://github.com/flutter/website/pull/870).

Closing.

PR #805, opened 2018-02-18
[Material] Updating copy mentioning Material

willlarche:
Corrects capitalization and renames common name of material library to Material Components.

*Help me make sure the places where I changed code for the site are still correct. I believe they are.

### DIFF
--- a/_data/catalog/index.json
+++ b/_data/catalog/index.json
@@ -32,7 +32,7 @@
     "id":"material"
   },
   {
-    "name": "Cupertino (Apple-style widgets)",
+    "name": "Cupertino (iOS-style widgets)",
     "description": "Beautiful and high-fidelity widgets for current iOS design language.",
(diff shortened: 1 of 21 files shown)


sethladd:
rslgtm

@HansMuller want to take a look?

HansMuller:
...looking

HansMuller:
I don't think it's a good idea to replace references to "iOS" with "Apple-flavor" just yet. There are hundreds of references to "iOS" in the Flutter repo, and we should be consistent.

willlarche:
Do you want that done first? I don't mind doing that repo first.

HansMuller:
It would be simpler to leave the half-dozen iOS-Apple changes out of this PR.

willlarche:
PTAL

HansMuller:
LGTM

PR #816, opened 2018-02-24
Add a way to check links and redirects

filiph:
Includes sitemap as of today, 23 Feb 2018.

Doesn’t include an automated, CI solution.


Sfshaza:
LGTM  Thanks for spelling this out and moving us towards a more complete linkcheck!

PR #817, opened 2018-02-24
Add first version of flutter.io/docs

filiph:
This will be the “index” of our docs (as the homepage will lack sidenav pretty soon).


Sfshaza:
This seems OK, but I'd like to see it staged.

InMatrix:
One downside is that it will now require 2 clicks instead of 1 to get to API docs from any page on flutter.io, since it's no long part of the top navigation. We should track how this change affects API doc usage.

docs.md
```diff
@@ -0,0 +1,94 @@
...
+* [Add Interactivity to Your Flutter App](/tutorials/interactive/)  
+  Walkthrough showing how to add a stateful widget to your app.
+
+* [A Tour of the Flutter Widget Framework](/widgets-intro/)  
+  A short, conceptual tour through the Flutter framework.
+
+&nbsp;
+
+
+## Want to skill up?
```
sethladd: will this be easy to understand for non-native speakers?

docs.md
```diff
@@ -0,0 +1,94 @@
...
+* [Text input](/text-input/)  
+  How to set up basic text input.
+
+* [Debugging Flutter Apps](/debugging/)  
+  Tools and tips for debugging your app.
+  
+* [Flutter Widget Inspector](/inspector/)  
+  The Flutter Inspector, a tool available in the Flutter plugin for your IDE.
+
+&nbsp;
```
sethladd: is this intention?

docs.md
filiph: I _think_ so. 

@Sfshaza, what are our options to rephrase?

docs.md
filiph: Yes. It's ugly but the special-case CSS would be even uglier. When I have time, I'll see if changing the CSS everywhere to give more margin between `ul` and `h2` won't break some page somewhere.

filiph:
Alright, I fixed the example link, and confirmed we're not breaking any other links. I think I'll merge.

PR #821, opened 2018-02-26
Update get-sdk-win's flutter doctor section

xqwzts:
Updates the `flutter doctor` instructions on the windows get-sdk page to match those of the Linux/Mac page.

### DIFF
--- a/_includes/setup/get-sdk-win.md
+++ b/_includes/setup/get-sdk-win.md
@@ -41,15 +41,18 @@ not support third-party shells like Git Bash.
 This command checks your environment and displays a report to the terminal window.
 The Dart SDK is bundled with Flutter; it is not necessary to install Dart separately.
 Check the output carefully for other software you may need to install or further 
-tasks to perform .
+tasks to perform (shown in **bold** text).
 
 For example:
-
 ```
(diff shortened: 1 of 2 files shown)


_includes/setup/get-sdk-win.md
```diff
@@ -41,15 +41,18 @@ not support third-party shells like Git Bash.
 This command checks your environment and displays a report to the terminal window.
 The Dart SDK is bundled with Flutter; it is not necessary to install Dart separately.
 Check the output carefully for other software you may need to install or further 
-tasks to perform .
+tasks to perform (shown in **bold** text).
```
gspencergoog: Perhaps just say "(marked with an ✗)" instead, since markdown doesn't show the bold in code blocks (so people looking for the bold text in the example below won't be confused).

xqwzts: Thanks for catching that. We could use html tags in the markdown to format the block using `<pre>` and make the error line bold with a `<strong>` tag.

xqwzts: @gspencergoog I made the quick switch to html tags, I could still change the text as well/instead if that would be preferred.

gspencergoog: No, that works for me.

PR #825, opened 2018-02-27
Tweaking the docs landing page. 

Sfshaza:
Please review, @filiph! (If you like, feel free to merge.)

Staged: https://sz-flutter.firebaseapp.com/docs/

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -24,7 +24,7 @@ permalink: /docs/
 ## New to Flutter?
 
 Once you've gone through [Get Started](/get-started/install/),
-including [Write Your First Flutter App](/get-started/codelab/)
+including [Write Your First Flutter App,](/get-started/codelab/)
 here are some next steps.
 
 * [Flutter for Android Developers](/flutter-for-android/)<br>
--- a/docs.md
+++ b/docs.md
@@ -23,10 +23,9 @@ permalink: /docs/
 
 ## New to Flutter?
 
-Start here.
-
-* [Write Your First Flutter App](/get-started/codelab/)<br>
-  Create a Flutter app that generates names for startup companies.
+Once you've gone through [Get Started](/get-started/install/),
+including [Write Your First Flutter App](/get-started/codelab/)
+here are some next steps.
 
 * [Flutter for Android Developers](/flutter-for-android/)<br>
   Review these tips if you have Android experience.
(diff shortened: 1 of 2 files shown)


docs.md
```diff
@@ -23,72 +23,82 @@ permalink: /docs/
 
 ## New to Flutter?
 
-* [Write Your First Flutter App](/get-started/codelab/)  
-  A walkthrough on creating a Flutter app that generates names for startup companies.
+Start here.
```
filiph: This is confusing when we have "Get started" above. Maybe we want to say something like "once you've gone through Get Started [LINK], here are the immediate next steps" or something like that. And this means we drop "Write Your First Flutter App" from here since that's included in the Get Started flow (and should be almost a requirement).

docs.md
```diff
@@ -23,72 +23,82 @@ permalink: /docs/
...
 
-* [Flutter for Android Developers](/flutter-for-android/)  
-  Tips for Android developers.
+* [Write Your First Flutter App](/get-started/codelab/)<br>
```
filiph: FWIW, in Markdown, the `<br>` at the end can be replaced by just adding two spaces at the end. I like the explicitness of `<br>`, though, so feel free to keep this.

docs.md
```diff
@@ -23,72 +23,82 @@ permalink: /docs/
...
+  Learn how to add a stateful widget to your app.
+
+* [A Tour of the Flutter Widget Framework](/widgets-intro/)<br>
+  Learn more about Flutter's react-style framework.
 
 &nbsp;
 
 
-## Want to skill up?
+## Want to improve your skills?
```
filiph: I'm leaning towards the original "skill up" but not strongly.

docs.md
Sfshaza: Done and Done.

docs.md
Sfshaza: I did not know that!!!!

docs.md
Sfshaza: I think Seth's point is that non-native English speakers might find that difficult to parse.

filiph: Yeah, but:

1. I'm non-native and I understood
2. It's not that hard to figure it out
3. It's not like this line is critical to grok

PR #827, opened 2018-02-27
Make the final app gif look more like an app

filiph:
Without the border, it looks like an annoying ad, apparently.

This also makes the gif smaller and more smooth.


filiph:
cc @Sfshaza, I'll be merging this now.

Preview:

screen shot 2018-02-26 at 19 01 29

PR #851, opened 2018-03-02
Update flutter-for-react-native.md

wmleler:
I made a complete pass. I think we should push this PR, and then you can do a pass (rather than you making a bunch of changes in this PR). But do review it to make sure I didn't introduce any bugs.

### DIFF
--- a/flutter-for-react-native.md
+++ b/flutter-for-react-native.md
@@ -1056,8 +1056,7 @@ There are three main ways to manage state:
 * The parent manages the widget’s state.
 * A mix-and-match approach
   
-##### How do you decide which approach to use? The following principles should help you decide:
-
+How do you decide which approach to use? The following principles should help you decide:
 * If the state in question is user data, for example the checked or unchecked mode of a checkbox, or the position of a slider, then the state is best managed by the parent widget.
 * If the state in question is aesthetic, for example an animation, then the widget itself best manages the state.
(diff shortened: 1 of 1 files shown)


Sfshaza:
LGTM

Sfshaza:
Yay!

flutter-for-react-native.md
```diff
@@ -1873,7 +1863,7 @@ class _LogoFadeState extends State<LogoFade> with TickerProviderStateMixin {
 
 ### How do I add swipe animation to Cards?
 
-In React Native, you either use PanResponder or third-party libraries for swipe animation. In Flutter, we can achieve this animation by using the [`Dismissible`](https://docs.flutter.io/flutter/widgets/Dismissible-class.html) widget and nest the child widgets.
+In React Native, you either use PanResponder or third-party libraries for swipe animation. In Flutter, we can achieve this animation simply by using the [`Dismissible`](https://docs.flutter.io/flutter/widgets/Dismissible-class.html) widget and nest the child widgets.
```
Hixie: we should be wary of adding self-congratulating adjectives like this. Let our audience determine if it's simple or not.

For example, using Dismissible isn't actually all that simple; you have to use a stateful widget, track callbacks, update lists, regenerate children, etc. So people might read this, then get upset that we misled them about how simple it was.

PR #852, opened 2018-03-02
Fixes the JSX code snippet highlighting for inline styles. 

sanketsahusoft:
@wmleler Please take a look at this.


wmleler:
I went ahead and merged this so we could do some more work on this.

PR #854, opened 2018-03-02
Fixed:  Image url not working

AkshayChordiya:
The image url wasn't working when the app is run. Changed the url to the one which is working properly now.


AkshayChordiya:
I signed it.

Sfshaza:
LGTM. Thanks, @AkshayChordiya!

PR #863, opened 2018-03-05
Update MaterialApp ctor argument mentioned in debugging.md

Phrohdoh:
Previously `debugging.md` mentioned `debugShowGrid` as an argument for `MaterialApp`'s constructor.
I believe this came from the `stocks` example as `MaterialApp` has `debugShowMateralGrid` not `debugShowGrid`.


Sfshaza:
Thanks for catching that, @Phrohdoh, and fixing it!

PR #865, opened 2018-03-05
Update flutter-for-react-native.md

joefazz:
This sentence read: 'In order to make Row' and then proceeded to show how to arrange things in a column.


joefazz:
I signed it!

Sfshaza:
Thanks for catching that error, @joefazz, and then fixing it!

PR #866, opened 2018-03-05
Fixes incorrect reference to IntelliJ

kozw:
Changes IntelliJ to Xcode in iOS platform code section

### DIFF
--- a/developing-packages.md
+++ b/developing-packages.md
@@ -145,7 +145,7 @@ You can run the example app from Android Studio by pressing the &#9654; button.
 We recommend you edit the iOS code usign Xcode.
 
 Before editing the iOS platform code in Xcode, first make sure that
-the code has been built at least once (i.e., run the example app from IntelliJ,
+the code has been built at least once (i.e., run the example app from Xcode,
 or in a terminal execute `cd hello/example; flutter build ios --no-codesign`).
 
 Next,


kozw:
I signed it

Sfshaza:
Thanks for submitting your PR, @kozw! @filiph, can you confirm?

filiph:
Hi @kozw, thanks for the PR! I actually think the original is correct, although confusing:

> Before editing the iOS platform code in Xcode, first make sure that
> the code has been built at least once (i.e., run the example app from IntelliJ,
> or in a terminal execute `cd hello/example; flutter build ios --no-codesign`).

The first run of `flutter build ios --no-codesign` makes things set up for development on that platform. Running the app in Xcode will not actually run the `flutter` tool, since Xcode doesn't know about it.

cc @mit-mit to confirm

Assuming I'm not wrong, let's figure out how to explain this better. It does seem weird to bring up IntelliJ when teaching how to write Swift in Xcode for Flutter apps.

mit-mit:
@mravn-google in the new setup, should you just run `flutter packages get`?

mravn-google:
> @mravn-google in the new setup, should you just run flutter packages get?

Not quite. `flutter packages get` now does everything needed to make platform-specific tools work.

For Android that simply means the `android/` project is correctly set up for your favorite Gradle-based IDE or Gradle command line tools.

For iOS that means the `ios/` project is correctly set up for CocoaPods and Xcode, in that order. Cocoapods are needed only when plugins are involved. If they are, you need to execute `pod install` in the `ios/` folder after which you can open the resulting `Runner.xcworkspace` in Xcode. As I understand it, that is standard procedure for working with iOS pods.

mravn-google:
In case you're wondering: We do not want to do `pod install` as part of `flutter packages get` because that may require you to download the CocoaPods repo (GB of data). Which would be a total waste, if all you wanted was to work on the Android side.

Hixie:
So should we close this PR?

mit-mit:
Yes. The current page says IntelliJ intentionally: You need to build there first before opening in xCode

PR #873, opened 2018-03-09
Add hyphens to links in post content

raxityo:
This avoids extending width of post content if links have long words (most likely to happen for words separated by underscores).


Hixie:
LGTM

Hixie:
Thanks!

PR #882, opened 2018-03-11
Delete a box design of sentence

qt-luigi:
A sentence of "... at the following link ..." of "Problems?" of Step 6, it is only surrounded a box design (by div, pre, and code tags). Other steps, it is a normal design sentence.


Sfshaza:
Thanks for catching, @qt-luigi!

qt-luigi:
Thanks for merging, @Sfshaza !

PR #884, opened 2018-03-12
Add a few more recipes and two migrations

brianegan:
Hey all, this PR starts the migration of some content into the Cookbook. I'll continue to follow this pattern if the direction looks good!

  * Add Hero Recipe
  * Add Snackbar Recipe
  * Migrate Tabs example
  * Migrate Reading and Writing files example
  * Update Fetch example with simplified demonstration of how to use the Data in a Flutter app

I've tried to include the appropriate redirects. Please let me know if I've made a mistake!


brianegan:
Hey, @mjohnsullivan or @filiph, not sure if either of you can help me with a review! Thanks :)

mjohnsullivan:
LGTM, but we already have a section on hero animations here: https://flutter.io/animations/hero-animations/. We should chat about how to keep all our animations docs in the same place/linked.

brianegan:
Ack, dang it! Didn't see that one :( Happy to discuss further, either via email or a quick hangout.

Thanks for  the review!

PR #887, opened 2018-03-14
Remove space breaking a markdown link.

jyaif:
The extra space character prevents the link from being clickable.


Sfshaza:
Thanks, @jyaif!

PR #889, opened 2018-03-15
Added a version of Flutter Cookbook in PDF

jdamacena:
I merged all the markdown files of the cookbook, exported a version of it in PDF and I also added a link to it in the index page of the markdown version.

### DIFF
--- a/cookbook/index.md
+++ b/cookbook/index.md
@@ -6,7 +6,7 @@ permalink: /cookbook/
 
 This cookbook contains recipes that demonstrate how to solve common problems 
 while writing Flutter apps. Each recipe is self-contained and can be used as 
-reference to help you build up an application.
+reference to help you build up an application. Also available in [PDF](/downloads/Flutter%20Cookbook.pdf).
 
 ## Design basics
 


jdamacena:
I signed it.

brianegan:
Woah, this is way cool! Thanks so much for the contribution :)

I'm very open to adding this, but I would love to see if we could automate this since the Cookbook is undergoing a lot of changes / additions. Did you do all this manually, or did you combine everything with some kind of Dart library or script?

@mjohnsullivan or @filiph ideas on whether you like this idea and if it's something we might be able to put into the build pipeline?

sethladd:
Could we simply ensure the print styles for the Cookbook enable someone to easily "print to PDF", which Chrome does really well?

I'm very hesitant to actually host a PDF via our site (keeping it up to date, blowing up our git repo size with many new versions of the PDF). BUT I definitely want our Cookbook to be "print to PDF" compatible... and that's probably just ensuring we have a print styles CSS?

brianegan:
Definitely agree we don't want to check-in PDF files that we have to maintain manually, but was thinking of generating one as part of the build process. 

However, I just checked and the print styles. They look good and that's an interesting idea! Perhaps we could use Jekyll to generate a version of the cookbook with all of the content in a single page so folks could download / print the whole thing in one shot.

Don't know how common that use-case is, however.

jdamacena:
I totally agree with @sethladd,  a way to print the cookboook was actually what I was looking for when I compiled that PDF. And besides, with the Print to PDF approach, we could have the option to print every tutorial on the site, and not only the cookbook, if it makes sense. Also I did that manually, I had to fix the internal links, and the images, and get rid of some headers on the beggining of the files. We could also create a cover to the printable version, that would be cool.

filiph:
I agree that it's probably a bad idea to try and manually update a PDF. The cookbook is very much a living document.

That said, I love @jdamacena's idea of having something you can print out or take with you on a flight. I agree it would make sense to have one long page with all the recipes.

The only way to do this, afaik, is to move all the recipes to `_includes/cookbook` or similar, and then a) include each in the respective `cookbook/something.md` and also include all of them in sequence in `cookbook/everything.md`. You'd have to remember to manually put it there, but that shouldn't be a problem (and we can test for this).

(There's another way, to use `_data`, but that would require us to write the cookbooks _inside_ yaml files, which would be a horrible experience.)

@brianegan, how do you feel about doing the include stuff? I personally don't see it as high priority.

@jdamacena, can I close this PR? If you put the PDF somewhere, I'm happy to link to it with a post on our Twitter account. As a snapshot it's perfect. But without the automation in place, I don't think it belongs on flutter.io.

jdamacena:
@filiph Yes, I think you can close this PR, I created a repository and put the PDF there, https://github.com/jdamacena/FlutterCookbookSnapshots. Should we create an issue for this include thing?

filiph:
Thanks! I'll close this one.

@brianegan, do you want to go with the include method I described above. I am not convinced it's worth the trouble, and you would be the one implementing the changes, so I'll leave it to you.

brianegan:
@filiph I'd say let's leave it for now and see if we get more requests for the feature. I think it's a cool idea, but maybe not as important as getting more recipes up and running.

PR #894, opened 2018-03-18
A few minor typos

wmleler:
Will also need to recreate the results image. Which will have the added benefit of updating the "Slow Mode" ribbon to "Debug".


filiph:
LGTM

filiph:
I don't think I have a screenshot like that. And I just realized the update of the image isn't a prerequisite to land this PR, so I merged. We can update the screenshot later.

wmleler:
Yeah, I'm sure there are going to be a bunch to update.

PR #895, opened 2018-03-19
Facebook Login Package

vivekkiran:
Since the https://pub.dartlang.org/packages/flutter_facebook_connect is no longer maintained, updated it with https://pub.dartlang.org/packages/flutter_facebook_login


mit-mit:
lgtm, thanks @vivekkiran !

PR #897, opened 2018-03-20
Update index.md

buamod:
Dart supports nested multiline comments. I feel it should be like this


buamod:
I signed it!

filiph:
Thanks for the PR! Dart does support `/* */` comments but the style guide [recommends](https://www.dartlang.org/guides/language/effective-dart/documentation#dont-use-block-comments-for-documentation) `//` even for multiline comments. Most IDEs make it extremely easy to write and toggle these comments nowadays, it's impossible to "leak" them accidentally, and they generally save 2 almost blank lines (`/*` and `*/`).

I'm going to close this issue.

You're raising a good point here: some people might be thinking that Dart doesn't support multiline "bracket" comments if they look at most Dart code.

PR #899, opened 2018-03-21
De dupe ide steps

mit-mit:
Currently the install step (https://flutter.io/get-started/install/) has duplication of the IDE setup step (which was moved to https://flutter.io/get-started/editor/)

### DIFF
--- a/_includes/setup/editor-setup.md
+++ b/_includes/setup/editor-setup.md
@@ -1,8 +1,0 @@
-## Editor setup
-
-Using the `flutter` command-line tools, you can use any editor to develop Flutter applications.
-Type `flutter help` at a prompt to view the available tools.
-
-We recommend using our plug-ins for a [rich IDE experience](/using-ide/) 
-supporting editing, running, and debugging Flutter apps. See [Editor Setup](/get-started/editor/)
-for detailed steps.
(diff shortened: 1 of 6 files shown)


filiph:
I'm not completely sold on this. I think there's value of telling developers that they can use Flutter in many IDEs. Maybe we should reword and not "link ahead". Can we get another opinion on this? Shams is OOO.

mit-mit:
Several developers were getting confused and thought they could skip installing Android Studio if they wanted to use VS Code. That is not the case; we currently rely on Android Studio being installed to vendor all Android dependencies.

filiph:
I understand, but let's not throw the baby out with the water. Could we just not link to the IDE step from here, and add the caveat with Android Studio? Something like this:

```
## Editor setup

Using the `flutter` command-line tools, you can use any editor to develop Flutter applications.
Type `flutter help` at a prompt to view the available tools.

We recommend using our plug-ins for a rich IDE experience
supporting editing, running, and debugging Flutter apps. We will get to the step of configuring your IDE shortly — but first, we need to install some dependencies.
```

mit-mit:
> Could we just not link to the IDE step from here, and add the caveat with Android Studio? 

@filiph yes, I agree! But if you scroll down just a few lines on that page it already does exactly that! See right after the 'Editor setup' section here:
https://flutter.io/setup-windows/#editor-setup

mit-mit:
@filiph, see the green box here. This is indeed duplicated, so I'd like to land  this de-dupe.

screen shot 2018-04-01 at 19 00 42

filiph:
My bad, thanks for pointing it out. Merging!

PR #901, opened 2018-03-21
Add new SDK archive page

mit-mit:
Add a new page displaying SDK bundles

### DIFF
--- a/js/archive.js
+++ b/js/archive.js
@@ -66,9 +66,10 @@ function updateTable(releases, os) {
     releasesForChannel.forEach(function (release, index) {
       // If this is the first row after the cut-off, insert the "Show more..." link.
       if (index === releasesToShow) {
-        var showAll = $("<a />").text("Show all...").attr("href", "#").click(function () {
+        var showAll = $("<a />").text("Show all...").attr("href", "#").click(function (event) {
           $(this).closest("table").removeClass("collapsed");
           $(this).closest("tr").remove();
+          event.preventDefault();
         });
         $("<tr>").append($("<td colspan=\"3\"></td></tr>").append(showAll)).appendTo(table);
       }
--- a/js/archive.js
+++ b/js/archive.js
@@ -1,0 +1,94 @@
+// Number of releases to show by default (rest will be behind a "show all" link).
+var releasesToShow = 5;
+
+// Fetches Flutter release JSON for the given OS and calls the callback once the data is available.
+var fetchFlutterReleases = function (os, callback) {
+  // OS: windows, macos, linux
+  var url = "https://storage.googleapis.com/flutter_infra/releases/releases_" + os + ".json";
+  $.ajax({
+    type: "GET",
+    url: url,
+    dataType: "json",
+    success: function (data) {
+      callback(data, os);
+    },
+    error: function (xhr, textStatus, errorThrown) {
+      console.log(textStatus);
+    }
+  })
+}
+
+function convertReleaseFormat(releases, os) {
+  // Convert releases into a format that's easier to work with:
+  // Old
+  // "1f3eb5034ff2372956620b2a9eb88683eee9495e": {
+  // 	"dev": {
+  // 		"release_date": "2018-02-27T00:26:37.058993Z",
+  // 		"version": "v0.1.6",
+  // 		"windows_archive": "dev/windows/flutter_windows_v0.1.6-dev.zip"
+  // 	}
+  // },
+  // New:
+  // {
+  //   hash: 1f3eb5034ff2372956620b2a9eb88683eee9495e,
+  //   channel: dev,
+  // 	 release_date: Date("2018-02-27T00:26:37.058993Z"),
+  // 	 version: "v0.1.6",
+  //   archive: "dev/windows/flutter_windows_v0.1.6-dev.zip"
+  // }
+  return Object.keys(releases.releases).map(function (hash) {
+    var release = releases.releases[hash];
+    var channel = Object.keys(release)[0];
+    return {
+      hash: hash,
+      channel: channel,
+      release_date: new Date(Date.parse(release[channel].release_date)),
+      version: release[channel].version,
+      archive: release[channel][os + "_archive"]
+    }
+  })
+}
+
+function updateTable(releases, os) {
+  // Convert the data for easier parsing and sort.
+  var releaseData = convertReleaseFormat(releases, os).sort(function (r1, r2) {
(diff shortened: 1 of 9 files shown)


DanTup:
@mit-mit I've pushed changes - they won't work until the JSON files on googlestorage have a header like `access-control-allow-origin: *` to allow them to be pulled in via ajax.

Still need to

- [ ] Test in all browsers (what browsers/versions do we intended to support?)

mit-mit:
Thanks much @DanTup. Can you please confirm that you are "ok with your commits being contributed to this project" per the cla bot? If so, just leave a comment here stating that.

DanTup:
I confirm that I am ok with my commits being contributed to the project (and I think my own PRs here show that I've completed the CLA) 👍

DanTup:
@mit-mit It'll be easier for me to test this from multiple browsers once the allow-origin thing is sorted and we can stage this somewhere, so let me know once that's sorted.

mit-mit:
CORS was configured, and this was pushed to staging: https://flutter-website-staging.firebaseapp.com/sdk-archive/ !

js/archive.js
```diff
@@ -0,0 +1,99 @@
...
+      callback(data, os);
+    },
+    error: function (xhr, textStatus, errorThrown) {
+      errorCallback(os);
+    }
+  })
+}
+
+function convertReleaseFormat(releases, os) {
+  // Convert releases into a format that's easier to work with:
```
gspencergoog: If you want a different format in the source data, I can do that.  The website is the only consumer of this, so it's not going to break anyone.

Would you rather have the 'releases' member just be an array of dictionaries laid out like "New:" below?  I didn't do that because then I have to do duplicate checking when I add a new one to the list (to be sure that there isn't already one with the same hash/channel combination), but I can do that pretty easily.

DanTup: Your call, this code already works, but simplifying it is always good. The main reason I converted it was that it wasn't sorted (object keys don't have an order, so they basically come through sorted by the has in the json), but I then flattened it to simplify some other code.

If you want to change it, having an array pre-sorted by date would be good. Moving complexity to the server seems like a good idea IMO because we control the environment (the less code running in all the different browsers, the better!).

DanTup:
@mit-mit I need to do some more testing of this in other browsers before it's merged (I haven't tested IE/Edge).

BTW - did something break with the access-origin change? It was fine earlier when first noted it was fixed, but now I'm seeing it broken again.

mit-mit:
> BTW - did something break with the access-origin change? It was fine earlier when first noted it was fixed, but now I'm seeing it broken again.

It still seems to work for me!?

DanTup:
Weird, seems ok again now. It was broken on forced refreshes (same error) just a few minutes ago. Ah well, I'll keep an eye on it. I'll test on Windows tomorrow and make any required changes.

js/archive.js
```diff
@@ -0,0 +1,100 @@
...
+    });
+  }
+}
+
+function updateTableFailed(os) {
+  var tab = $("#tab-os-" + os);
+  tab.find(".loading").text("Failed to load releases");
+}
+
+// Send requests to render the tables.
```
filiph: Is this running on every page of flutter.io? It seems it does.

Consider only running the fetch when on a particular page or when a particular element is found.

DanTup: Oh, whoops! When I mocked it up the JS was only included in my test page, but you're right this is happening everywhere. I'll fix in the morning.

js/archive.js
DanTup: By the way - we're currently pulling three files that are mostly the same, just with different archive links - would it be difficult/worthwhile to just have one file that has three archive links within each release object?

gspencergoog: No, we should leave them separate, because they are updated independently from three different build servers, and if they were one file, then it would be vulnerable to race conditions when updating the data (the windows build server could read the file, the linux one updates it, and then the windows one replaces it, reverting the linux update, for instance).  We could implement a locking scheme, but that would complicate the code, and I'd rather not do that until we have to.

sdk-archive.md
```diff
@@ -0,0 +1,57 @@
...
+---
+
+The Beta channel contains the most stable Flutter builds. See [Flutter’s
+channels](https://github.com/flutter/flutter/wiki/Flutter-build-release-channels)
+for details.
+
+<div id="tab-set-os" class="sdk-archives">
+	<ul class="tabs__top-bar">
+		<li class="tab-link current" data-tab="tab-os-windows">Windows</li>
+		<li class="tab-link" data-tab="tab-os-macos">Mac OS</li>
```
gspencergoog: Not that Apple spells it "macOS" on all their materials.

js/archive.js
DanTup: Ah, makes sense. I forgot this was more than just the git clone. LMK what you want to do about the format - if we're going to change it, doing it before we merge this is best!

DanTup:
@mit-mit just saw this again... Here's the HTTP readers - note the Date in the header is now, but the Last-Modified is two days ago (and the accept header was missing). Is there something serving up a cached version of this from before the header was added?

screen shot 2018-03-22 at 07 35 08

js/archive.js
DanTup: Fixed!

sdk-archive.md
DanTup: Fixed!

DanTup:
Seems to work fine in Edge and IE11 too.

js/archive.js
gspencergoog: I am working on a PR that will generate what you asked for: I figure it'll be easier to do now before we're live.  I'll send it to you for review.

gspencergoog: My PR has been merged, and I've updated the existing metadata to the new format.

DanTup:
@mit-mit I've pushed a change to handle the data in the new format that @gspencergoog pushed; you'll need to update the staging site for it to start working again.

Let me know if anything else needs tweaking.

mit-mit:
Staging updated: https://flutter-website-staging.firebaseapp.com/sdk-archive/

mit-mit:
@gspencergoog if you think we are ready to merge, please approve/lgtm.

For now this would just be an new page not linked to. Then I'll do another PR next week to update the main installation page with a link to most recent beta, and a link to this archive.

gspencergoog:
32384589-a60f0e74-c078-11e7-9bc1-e5b5287aea9d

gspencergoog:
Ship it!

PR #904, opened 2018-03-22
Replace tappable heart with tappable tile in getting-started

wreppun:
The getting-started (Write Your First Flutter App) walkthrough references making the heart tappable (and saving suggestions).  The code actually makes the ListTile tappable, which leads to some minor inconsistencies in the text.

This PR changes the language to refer to tappable tiles instead of tappable hearts.

There is an additional PR which, instead of changing the language, changes the code to use an `IconButton`.


wreppun:
Alternate approach at #905

PR #905, opened 2018-03-22
Use IconButton in get-started tutorial

wreppun:
The get-started (Write Your First Flutter App) walkthrough references making the heart tappable (and saving suggestions).  The code actually makes the ListTile tappable, which leads to some minor inconsistencies in the text.

This PR uses IconButton to wrap the Icon and make _only_ the heart tappable.  The strikethrough/highlighting is a little awkward -- even though the Icon widget is kept, the nature of the diff after adding the wrapping IconButton led me to use strikethrough on the existing trailing parameter.

There is an alternate PR which, instead of changing the code, changes the language to reference tappable tiles instead of tappable hearts.

### DIFF
--- a/get-started/codelab/index.md
+++ b/get-started/codelab/index.md
@@ -682,7 +682,8 @@ Add the highlighted lines below:
 </li>
 
 <li markdown="1"> Make the hearts tappable in the `_buildRow`
-    function. If a word entry has already been added to favorites,
+    function by wrapping the Icon in an IconButton. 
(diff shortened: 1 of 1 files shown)


wreppun:
Alternate approach at #904

filiph:
Thanks for the _two_ PRs! I landed the other one (https://github.com/flutter/website/pull/904) since that was how we envisioned the app in the first place (i.e. use the whole tile to favorite).

PR #916, opened 2018-03-28
Fix Build

goderbauer:
Fixes https://github.com/flutter/flutter/issues/15293.

Due to a bug in the analyzer, the "exclude" section of our analysis_options.yaml was previously ignored. The bug was fixed upstream in https://github.com/dart-lang/sdk/commit/748cfdff7edc987ed0c28d01e847aeb910f04ebb causing our build to fail.

/cc @DanTup


DanTup:
Aha, that's what it was :-)

Changes lgtm, and the Travis build for the PR is green 👍

PR #915, opened 2018-03-28
Make Windows requirements more specific

goderbauer:
In a follow-up PR we will add a check to flutter_tools to fail if PowerShell 5.0 is not installed.

Preparation for https://github.com/flutter/flutter/issues/11698

/cc @pq

PR #917, opened 2018-03-29
Make Windows requirements more specific

goderbauer:
In a follow-up PR we will add a check to flutter_tools to fail if PowerShell 5.0 is not installed.

Preparation for https://github.com/flutter/flutter/issues/11698

/cc @pq


goderbauer:
Submitting with approval from https://github.com/flutter/website/pull/915.

PR #921, opened 2018-03-31
Fix networking example

xqwzts:
The networking example has this code:

```dart
final json = json.decode(response.body);
```

Which fails with the following error from the analyzer:
> Local variable 'json' can't be referenced before it is declared.

This probably came along with `json` being renamed from the now-deprecated `JSON`.

Renaming the variable to `_json` fixes it.

### DIFF
--- a/cookbook/networking/fetch-data.md
+++ b/cookbook/networking/fetch-data.md
@@ -83,9 +83,9 @@ we'll need to:
 ```dart
 Future<Post> fetchPost() async {
   final response = await http.get('https://jsonplaceholder.typicode.com/posts/1');
-  final _json = json.decode(response.body); 
+  final responseJson = json.decode(response.body); 
   
-  return new Post.fromJson(_json); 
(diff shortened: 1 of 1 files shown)


brianegan:
Thanks for the update!

cookbook/networking/fetch-data.md
```diff
@@ -83,9 +83,9 @@ we'll need to:
 ```dart
 Future<Post> fetchPost() async {
   final response = await http.get('https://jsonplaceholder.typicode.com/posts/1');
-  final json = json.decode(response.body); 
+  final _json = json.decode(response.body);
```
xster: To avoid names that are too similar, can we just call the local variable responseJson or some such?

xqwzts: Done.

xqwzts:
Updated with the suggested name change.

xster:
Thanks. Will merge on green

PR #925, opened 2018-04-03
Capitalized "Hot Reload"

maguinis:
Since Hot Reload is a feature, it should be capitalized (unless used as a verb).

### DIFF
--- a/hot-reload.md
+++ b/hot-reload.md
@@ -7,16 +7,16 @@ permalink: /hot-reload/
 * TOC
 {:toc}
 
-## Using hot reload
+## Using Hot Reload
 
-Flutter's hot reload feature helps you quickly and easily experiment, build
-UIs, add features, and fix bugs. Hot reload works by injecting updated source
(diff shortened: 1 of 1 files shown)


kwalrath:
I'm not sure we should be doing this - I don't think we necessarily capitalize features. (I'll do a little research to see if I can back up my gut feeling.)

InMatrix:
FWIW, React Native seems to capitalize "Hot Reloading": https://facebook.github.io/react-native/blog/2016/03/24/introducing-hot-reloading.html

InMatrix:
Upon taking a closer look, they don't really capitalize hot reload consistently.

Sfshaza:
My first instinct was also that we don't capitalize features. For example, we didn't introduce "Strong Mode", it was "strong mode." I looked in the [Google Style Guide](https://developers.google.com/style/capitalization) and it doesn't mention capitalization for features, unless patented or trademarked, etc.

mit-mit:
Closing as there doesn't seem to be consensus to change our style here.

PR #928, opened 2018-04-05
Remove developer.apple.com links from link checking

mit-mit:
To work around https://github.com/flutter/flutter/issues/16268


mit-mit:
Build is passing now, merging to get site changes to publish again.

TBR @filiph

PR #938, opened 2018-04-09
Create flutter-for-ios.md

rock3r:
This PR creates the page for the new `Flutter for iOS devs`. It does not contain any content except for the intro (tweaked from the one in the `Flutter for Android devs` page).

Content will come in subsequent PRs, and we aren't adding the page to the sidebar yet — we'll do it when it's ready to go. For now it'll only accessible at `http://flutter.io/flutter-for-ios/` (once the PR is merged).

### DIFF
--- a/flutter-for-ios.md
+++ b/flutter-for-ios.md
@@ -5,7 +5,7 @@ permalink: /flutter-for-ios/
 ---
 
 This document is meant for iOS developers looking to apply their
-existing iOS  knowledge to build mobile apps with Flutter. If you understand
+existing iOS knowledge to build mobile apps with Flutter. If you understand
 the fundamentals of the iOS framework then you can use this document as a
 jump start to Flutter development.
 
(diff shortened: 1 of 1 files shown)


Hixie:
Is there any reason to land this before the content exists?

rock3r:
@Hixie the agreement with @mjohnsullivan is that we're going to do smaller PRs, one per section, as we go instead of one big one when we're done with the document. This way it should be easier to review, and you get content faster (and we can change course faster if need be). If we have the file on your side we can then start doing the section PRs.

Unless, I misunderstood what you meant. If you prefer we can merge this with the first PR which contains the Views section, and then do subsequent sections as previously outlined.

We'd only make the page reachable via sidebar when we're done with the document, still.

mjohnsullivan:
I'm fine with landing this in stages, but we should mark/head the page as WIP to inform users.

LGTM, but with agreement from @Hixie. Otherwise, we can do this in a branch.

rock3r:
Can defo add a "WIP" warning at the very top for now. Will do shortly

rock3r:
I have added the WIP notice, works for you now Matt and Ian?

mjohnsullivan:
Let's go with merging it here and keeping it unlinked until finished.

mjohnsullivan:
LGTM

PR #944, opened 2018-04-09
Fix code validation issue

mit-mit:
Currently failing with `child' is deprecated and shouldn't be used at example/text_input_md_1.dart:31:15 • deprecated_member_use`


cbracken:
lgtm

PR #950, opened 2018-04-11
Update navigation-basics.md

mohamed-aiman:
changed to pass context from the closest instance of the class that encloses the given context.

### DIFF
--- a/cookbook/navigation/navigation-basics.md
+++ b/cookbook/navigation/navigation-basics.md
@@ -133,8 +133,7 @@ class FirstScreen extends StatelessWidget {
         child: new RaisedButton(
           child: new Text('Launch new screen'),
           onPressed: () {
-            Navigator.push(
-              context,
+            Navigator.of(context).push(
               new MaterialPageRoute(builder: (context) => new SecondScreen()),
             );
           },


mohamed-aiman:
I signed it!

brianegan:
Hey, thanks for contributing! Just so I understand: Was the previous code causing trouble for ya? 

`Navigator.push(context, ...)` is just a slightly shorter version of `Navigator.of(context).push(...)`, so I don't think there's any difference between the two versions.

mohamed-aiman:
Ya. With this code it works fine but I was unable to get get context that way inside other widgets that I tried. So for learning purpose I suggest to keep the widest compatible way.

brianegan:
@mohamed-aiman Interesting, did changing from `Navigator.push(context...`) to `Navigator.of(context).push(...)` fix the issue? Or were you perhaps passing the wrong context to the `Navigator`?

It's not a biggie, but if the code in this sample works for ya, I'd prefer to keep the shorter version. It sounds like you might have been running into something else...

Hixie:
@mohamed-aiman I would be interested in seeing what exactly was the problem you were running into; if you could file a bug showing the issue that would be really useful.

In the meantime, I concur with @brianegan that this change doesn't really make sense, so I'm going to close the PR. Again, I would be very interested in hearing about what led you to this change, please don't hesitate to file a bug or e-mail me (ian@hixie.ch) if you can provide any insight into what you ran into!

Thanks for contributing though! Please don't hesitate to submit PRs in the future!

PR #958, opened 2018-04-16
Change OS tabs to default to current users OS

DanTup:
This makes the OS tabs default to the users current OS. It no longer persists OS selection in local storage and the hash fragment in the URL still takes precedence.

Staged here:

https://dantup-flutter-staging.firebaseapp.com/sdk-archive/

I don't know how reliable the `userAgent` check is so LMK if you have any ideas on improving it.

### DIFF
--- a/js/tabs.js
+++ b/js/tabs.js
@@ -1,4 +1,4 @@
-function setupToolsTabs(container, tabIdPrefix, storageName) {
+function setupToolsTabs(container, tabIdPrefix, storageName, defaultTab) {
   var tabs = $('.tabs__top-bar li', container);
   var tabContents = $('.tabs__content', container);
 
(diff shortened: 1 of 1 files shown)


js/tabs.js
```diff
@@ -40,18 +40,30 @@ function setupToolsTabs(container, tabIdPrefix, storageName) {
...
 });
+
+function getOS() {
+  var ua = navigator.userAgent.toLowerCase();
+  if (ua.indexOf("win") !== -1)
+    return "windows";
+  if (ua.indexOf("mac") !== -1)
+    return "macos";
+  if (ua.indexOf("linux") !== -1)
+    return "linux";
```
gspencergoog: The only Unixes we officially support are Linux and macOS, but you could add a test for "X11" to the Linux test to catch any other types of Unix. Even if they're not running on the machine they're downloading it on, someone running on a non-Linux, non-Mac, Unix browser will probably want the Linux package.

Also, it might be better to compare in a case-sensitive way ("Mac", "Win", "Linux"), since "win" and "mac" are really short strings, and might appear in other strings (e.g. "Darwin").

DanTup: Done (and re-staged). It's possible we could could expand Win -> Windows and Mac -> Macintosh but I don't know if there's much value now they're case sensitive (or how consistent that is across browsers).

gspencergoog: No, no need to expand those, IMO, and they might even catch some edge cases if you don't.

js/tabs.js
```diff
@@ -59,11 +59,13 @@ $(document).ready(function () {
...
-  var ua = navigator.userAgent.toLowerCase();
-  if (ua.indexOf("win") !== -1)
+  var ua = navigator.userAgent;
+  if (ua.indexOf("Win") !== -1)
     return "windows";
-  if (ua.indexOf("mac") !== -1)
+  if (ua.indexOf("Mac") !== -1)
     return "macos";
-  if (ua.indexOf("linux") !== -1)
+  if (ua.indexOf("Linux") !== -1)
```
gspencergoog: You could combine these two into one expression, since they return the same thing.

DanTup: I did think of that and wondered if it was better this way in case it changes in future; however, that seems unlikely, so they're merged now.

DanTup:
@mit-mit @gspencergoog Anything else you think needs changing here, or are we ok to merge?

mit-mit:
No additional comments from me

gspencergoog:
32384589-a60f0e74-c078-11e7-9bc1-e5b5287aea9d

DanTup:
Had a quick test on the live site and it all looks good; but let me know if you spot any issues!

PR #968, opened 2018-04-19
Reenable proper .md snippet extraction, fix analysis errors in cookbook

brianegan:
Hey hey :) I played around with a couple techniques for fixing the cookbook analysis errors and went with the following strategy:

  - Skip sections that are meant for human consumption: In order to keep the snippets short and focused, many snippets cannot be analyzed. 
  - "Complete examples" are analyzed for correctness. In most cases, this should capture analysis errors that are relevant, since the snippets of a recipe and the corresponding complete examples should be kept in sync.

Alternative techniques attempted or discussed:

  - Group the snippets together for analysis: I tried this but realized it wouldn't quite work. Why? Some snippets are really "out of context." I.e. we have the code for a `new RaisedButton`, but do not place it inside the `build` method of a `State` class for brevity and focus.
  - Ignore some code before the snippet. We could attempt this, but it would mean repeating the complete example for each snippet, revealing only the part of the snippet that's relevant. That felt cumbersome, but happy to go this route if it's preferred. Right now, it appears you can only include some code before a snippet, rather than "around" a snippet, so we'd need to write that functionality as well.

This is a fix for https://github.com/flutter/flutter/issues/16155

## Update

  - After analysis passed, I saw formatting errors in various files throughout the codebase. These are now fixed.
  - I decided to ignore the project's README.md file, as it contains code that could not be properly analyzed (it contains dart codeblocks within  another codeblock, and the extraction tool was unable to process this)

### DIFF
--- a/catalog/samples/basic_app_bar.md
+++ b/catalog/samples/basic_app_bar.md
@@ -46,7 +46,8 @@ class _BasicAppBarSampleState extends State<BasicAppBarSample> {
   Choice _selectedChoice = choices[0]; // The app's "state".
 
   void _select(Choice choice) {
-    setState(() { // Causes the app to rebuild with the new _selectedChoice.
+    // Causes the app to rebuild with the new _selectedChoice.
+    setState(() {
       _selectedChoice = choice;
     });
   }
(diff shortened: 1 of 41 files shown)


brianegan:
Heya @filiph, @mjohnsullivan or @goderbauer -- When you get a chance, could one of you please have a look at this and let me know if it's good to go?

Thanks!

catalog/samples/basic_app_bar.md
```diff
@@ -58,15 +59,22 @@ class _BasicAppBarSampleState extends State<BasicAppBarSample> {
         appBar: new AppBar(
           title: const Text('Basic AppBar'),
           actions: <Widget>[
-            new IconButton( // action button
+            new IconButton(
+              // action button
```
goderbauer: nit here and elsewhere: should comments like this maybe go before the element they are referring to?

brianegan:
Thanks for the review, @goderbauer. Agreed, keeping them in sync would be the best case scenario. I tried all suggestions mentioned in the original issue as well as a few of my own ideas, but this was simply the best that I could come up with. Do you have any alternative suggestions that might do the trick?

Hixie:
If the small snippets are always exactly subsets of the big snippets we could prefix the small snippets with a code, and then wrap the relevant part of the big snippet with the same code, and the script could verify that (a) a code only appears in one big snippet, and (b) every small snippet with a code matches exactly the relevant part of the big snippet with that code.

brianegan:
Thanks Hixie, I'll look at that option and see if it works out

PR #970, opened 2018-04-24
Update background-parsing.md

AngeeelD:
After decode de JsonArray it's necessary to convert it to List... on the other hand, an exception is thrown in execution time (type 'List<dynamic>' is not a subtype of type 'List<Photo>')


AngeeelD:
I signed it!

brianegan:
Thanks for the contribution! I actually have a fix for this bug in a previous PR: https://github.com/flutter/website/pull/968/files

Therefore, I'll go ahead and close this one down and try to bug folks for a review!

PR #973, opened 2018-04-26
fix type List<dynamic> is not a subtype of type List<Photo> error

xxnjdlys:
fix type List<dynamic> is not a subtype of type List<Photo> error


ArthurYuCN:
I signed it!

ArthurYuCN:
I signed it!

PR #976, opened 2018-04-26
Update widgets-intro.md

wispy316:
Found above code not works in Hot Reload, since there is No class extends Widget. Need a full restart can take change.

### DIFF
--- a/widgets-intro.md
+++ b/widgets-intro.md
@@ -46,6 +46,8 @@ void main() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


wispy316:
I signed it!

Hixie:
Thanks for your contribution.

Since this section is teaching about the widgets, rather than the hot reload/restart feature, I think it's probably better if we omit this line from this particular document. Otherwise we would end up having to put comments like this all over the place and it would start to really detract from the other topics.

PR #980, opened 2018-04-27
Flutter for iOS — ViewControllers section

rock3r:
This PR is for the fifth section of the work to create the `flutter-for-ios.md` document. Writing by @niamh-power on Novoda's behalf, I am just a proxy.

Following the same rough format as the `flutter-for-android.md` document, with iOS details instead.

The following sections have been created:

* What is the equivalent to ViewControllers in Flutter?
* How do I listen to iOS lifecycle events?

⚠️ **This requires #979 to be merged first!** ⚠️

PR #981, opened 2018-04-27
Flutter for iOS — Layouts section

rock3r:
This PR is for the sixth section of the work to create the `flutter-for-ios.md` document. Writing by @niamh-power on Novoda's behalf, I am just a proxy.

Following the same rough format as the `flutter-for-android.md` document, with iOS details instead.

The following sections have been created:

* What is the equivalent of a UITableView or UICollectionView in Flutter?
* How do I know which list item is clicked?
* How do I update ListViews dynamically?
* What is the equivalent of a ScrollView in Flutter?

⚠️ **This requires #980 to be merged first!** ⚠️

### DIFF
--- a/flutter-for-ios.md
+++ b/flutter-for-ios.md
@@ -1415,7 +1415,7 @@ In Flutter the easiest way to do this is using the `ListView` widget. This
 acts as both a `ScrollView` and an iOS `TableView`, as you can layout widgets
 in a vertical format.
   
-<!-- skip -->  
+<!-- skip -->
 {% prettify dart %}  
 @override  
 Widget build(BuildContext context) {  
--- a/flutter-for-ios.md
+++ b/flutter-for-ios.md
@@ -273,7 +273,7 @@ animated widget.
 In Flutter you use an `AnimationController` which is an `Animation<double>`
 that can pause, seek, stop and reverse the animation. It requires a `Ticker`
 which signals when vsync happens and produces a linear interpolation between
-0 and 1 on each frame while it's running.You then create one or more
+0 and 1 on each frame while it's running. You then create one or more
 `Animation`s and attach them to the controller.
 
 For example, you could have a `CurvedAnimation` to have an animation which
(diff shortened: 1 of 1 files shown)


flutter-for-ios.md
```diff
@@ -27,3 +27,1408 @@ that are most relevant to your needs.
...
+{% endprettify %}
+
+## How do I update `ListView`s dynamically?
+
+In iOS, you would update the data for the list view, and notify the table or
+collection view using the `reloadData` method. In Flutter if you were to update
+the list of widgets inside a `setState()`, you would quickly see that your
+data did not change visually.
+
+This is because when `setState()` is called, the Flutter rendering engine will
```
mjohnsullivan: Unsure about this; needs verification

rock3r:
Contents covered in #993, closing this. Will address comments on that one.

flutter-for-ios.md
```diff
@@ -27,3 +27,1408 @@ that are most relevant to your needs.
...
+the user, not responding   to user input, and running in the background. *
+`resumed` — The application is visible and responding to user input. *
+`suspending` — The application will be suspended momentarily. It is not
+triggered on iOS as there is no equivalent   event to map to on iOS
+
+For more details on the meaning of these states, you can check the
+[`AppLifecycleStatus` documentation](https://docs.flutter.io/flutter/dart-ui
+/AppLifecycleState-class.html).
+
+# Layouts
```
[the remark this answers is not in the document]
rock3r: There's no `GridView` in iOS, they use [`UICollectionView`](https://developer.apple.com/documentation/uikit/uicollectionview) instead, and that's covered

flutter-for-ios.md
rock3r: Summoning @takhion the wise

Takhion: ###### _...flash of light, puff of smoke, and magical sounds!_

I believe there might be a fundamental misconception at play here: [`State.setState`] **is** the mechanism though which [`StatefulWidget`]s notify changes, and as a direct consequence the [`State.build`] function is executed again, returning a (potentially) different sub-tree of [`Widget`]s.

A new [`Widget`] instance means nothing by itself, since it's up to the corresponding [`Element`] (aka the [`BuildContext`]) to decide whether the new instance is of the same type and carries new data (if you're curious, check out [`Element.updateChild`] and [`Widget.canUpdate`]).

Now, this specific example is _misleading_ since apparently the [`ListView`] doesn't draw new rows unless the `widgets` list is re-created... But that has nothing to do with the usual way widgets work!

The issue here is that this particular "configuration" of [`ListView`] is really optimised for having a static/immutable list. As such, internally it creates [a specific `childrenDelegate`][`SliverChildListDelegate.init`] ([`SliverChildListDelegate`]) that [when prompted whether to "rebuild"][`SliverChildListDelegate.shouldRebuild`] assumes that the lists are immutable, so it simply compares the old and the new list instances with the equality operator, which for `List` is the default check by reference. In conclusion, by creating our own delegate and forcing `shouldRebuild` to return `true`, we can treat the list as mutable and everything works as expected!

🎉🎉🎉

<details>
<summary><b>example</b></summary>

```dart
import 'package:flutter/material.dart';

void main() {
  runApp(new SampleApp());
}

class SampleApp extends StatelessWidget {
  // This widget is the root of your application.
  @override
    return new MaterialApp(
      title: 'Sample App',
      theme: new ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: new SampleAppPage(),
    );
  }
}

class SampleAppPage extends StatefulWidget {
  SampleAppPage({Key key}) : super(key: key);

  @override
  _SampleAppPageState createState() => new _SampleAppPageState();
}

class _SampleAppPageState extends State {
  // [List] needs a type to make Dart 2 happy
  List<Widget> widgets = [];

  @override
  void initState() {
    super.initState();
    for (int i = 0; i < 10; i++) {
      widgets.add(getRow(i));
    }
  }

  @override
    //// this doesn't work
    //var listView = new ListView(children: widgets);

    // this works!
    var listView = new ListView.custom(
      childrenDelegate: new _MyListDelegate(widgets),
    );

    print("build!\n"
        // context is always the same instance
        "* context=${identityHashCode(context)}\n"
        // listView is always a different instance
        "* listView=${identityHashCode(listView)}");

    return new Scaffold(
      appBar: new AppBar(
        title: new Text("Sample App"),
      ),
      body: listView,
    );
  }

  Widget getRow(int i) {
    return new GestureDetector(
      child: new Padding(
        padding: new EdgeInsets.all(10.0),
        child: new Text("Row $i"),
      ),
      onTap: () {
        setState(() {
          widgets.add(getRow(widgets.length + 1));
          print('row $i (widgets.length=${widgets.length})');
        });
      },
    );
  }
}

class _MyListDelegate extends SliverChildListDelegate {
  _MyListDelegate(List<Widget> children) : super(children);

  @override
  bool shouldRebuild(SliverChildListDelegate oldDelegate) => true;
}
```

</details>

[`BuildContext`]: https://github.com/flutter/flutter/blob/v0.3.5/packages/flutter/lib/src/widgets/framework.dart#L1748-L1811
[`Element`]: https://github.com/flutter/flutter/blob/v0.3.5/packages/flutter/lib/src/widgets/framework.dart#L2425-L2475
[`Element.updateChild`]: https://github.com/flutter/flutter/blob/v0.3.5/packages/flutter/lib/src/widgets/framework.dart#L2637-L2669

[`Widget`]: https://github.com/flutter/flutter/blob/v0.3.5/packages/flutter/lib/src/widgets/framework.dart#L338-L373
[`Widget.canUpdate`]: https://github.com/flutter/flutter/blob/v0.3.5/packages/flutter/lib/src/widgets/framework.dart#L422-L435

[`StatefulWidget`]: https://github.com/flutter/flutter/blob/v0.3.5/packages/flutter/lib/src/widgets/framework.dart#L592-L773
[`State.setState`]: https://github.com/flutter/flutter/blob/v0.3.5/packages/flutter/lib/src/widgets/framework.dart#L1042-L1086
[`State.build`]: https://github.com/flutter/flutter/blob/v0.3.5/packages/flutter/lib/src/widgets/framework.dart#L1191-L1291

[`ListView`]: https://github.com/flutter/flutter/blob/v0.3.5/packages/flutter/lib/src/widgets/scroll_view.dart#L433-L570

[`SliverChildDelegate.shouldRebuild`]: https://github.com/flutter/flutter/blob/v0.3.5/packages/flutter/lib/src/widgets/sliver.dart#L86-L95

[`SliverChildListDelegate`]: https://github.com/flutter/flutter/blob/v0.3.5/packages/flutter/lib/src/widgets/sliver.dart#L214-L243
[`SliverChildListDelegate.shouldRebuild`]: https://github.com/flutter/flutter/blob/v0.3.5/packages/flutter/lib/src/widgets/sliver.dart#L302-L306
[`SliverChildListDelegate.init`]: https://github.com/flutter/flutter/blob/v0.3.5/packages/flutter/lib/src/widgets/scroll_view.dart#L599-L603

rock3r: Thanks Eugy, I'll rephrase this section to match :) ❤️

PR #983, opened 2018-04-27
Trim lines' end when extracting code snippets

rock3r:
~This allows us to avoid trailing whitespace, and fixes cases where, for example, we have empty spaces after comments that prevent the code from finding the end of the comment, causing the code to run to the end of the file and failing with an index-out-of-bounds error.~

~For example:~

```markdown
...

<!-- skip --> 
{% prettify dart %}
void main() {
  runApp(new SampleApp());
}
{% endprettify %}
```

~(notice the extra whitespace at the end of the comment line)~

~Would fail the build.~

~This PR hot-fixes that issue, although a more robust error handling should eventually be built into the code so that whenever a comment is left open there's a proper error message pointing at a line number, and explaining the issue.~

EDIT: see [comment below](https://github.com/flutter/website/pull/983#issuecomment-385158897)

### DIFF
--- a/tool/extract.dart
+++ b/tool/extract.dart
@@ -15,9 +15,9 @@ void main(List<String> args) {
 
   // Traverse markdown files in the root.
   int extractCount = 0;
-  Iterable<FileSystemEntity> files = Directory.current
-      .listSync()
-      .where((FileSystemEntity entity) => entity is File && entity.path.endsWith('.md'));
+  Iterable<FileSystemEntity> files = Directory.current.listSync().where(
+      (FileSystemEntity entity) =>
+          entity is File && entity.path.endsWith('.md'));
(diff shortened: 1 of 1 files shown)


rock3r:
Update: after #968 has been merged, the fix proper is not needed anymore as @brianegan had incorporated it himself. This PR now is basically just running `dartfmt` on `tools/extract.dart`

Sfshaza:
Hi, @rock3r! I'm a little unclear. I can close this PR, since Brian made it unnecessary? Or you want to modify the PR to mention dartfmt?  thx

rock3r:
@sfshaza there is some reformatting left here, but feel free to close it if there's no value for you.

Sfshaza:
OK, thanks. Closing.

PR #989, opened 2018-04-27
Update web-sockets.md

solomon-okolie:


filiph:
Thank you for the fix!

PR #991, opened 2018-04-28
[Building layouts in Flutter] Introduce the imageSection in step5 in isolation

VeskoI:
I think it's easier for the reader to introduce the `imageSection` in Step 5 in isolation, similar to how the other sections are introduced.  

Silently introducing the `ListView` in step 5 is a bit confusing - instead it follows the previous sections. Also the code in Step 6 is a bit cleaner, as all sections are added via variables.


VeskoI:
I signed it!

Hixie:
cc @Sfshaza for review

Sfshaza:
Thanks, @VeskoI! This PR is superseded by another. Closing.

PR #992, opened 2018-04-30
Scan directory for all asset files

fmatosqg:
Updates doc to reflect changes on https://github.com/flutter/flutter/pull/16413

### DIFF
--- a/assets-and-images.md
+++ b/assets-and-images.md
@@ -32,6 +32,14 @@ flutter:
     - assets/background.png
 ```
 
+To include all assets under a folder specify the folder name with the `/` character at the end:
+
+```yaml
+flutter:
+  assets:
+    - assets/
+```
+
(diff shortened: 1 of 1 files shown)


fmatosqg:
Pls don't merge yet, needs to merge code at flutter/flutter repo first

fmatosqg:
@Sfshaza a couple of notes:

- the code associated with this doc update is already in dev but not yet in beta, not sure if it's the right time for rolling the docs live. @tvolkert was following this, pls advise.
- I went to review the docs on the main page, and there's something misleading about "dark" variants which is totally my fault, but needs fixing. 
- It fails to mention no support for subdirectories which is a very popular question. I can address that in a separate PR though.

If this will get reverted until it beta rolls it out I can tackle the edits in one go. However if this PR doesn't get reverted I can make a fix for it but I'll probably take 36h from now to get started, after I start it should done shortly.

PR #1002, opened 2018-05-04
Android obfuscation

swavkulinski:
Basic Android obfuscation and minification.

As suggested in https://github.com/flutter/flutter/pull/16956

### DIFF
--- a/android-release-build.md
+++ b/android-release-build.md
@@ -140,11 +140,9 @@ file.
 Release builds of your app will now automatically be signed.
 
 
-## Obfuscation and minification
+## Minfify and obfuscate
 
-Android build tools allow reduce the size of the APK file and obfuscate classes and methods. By default, Flutter tool won't use these tools. However, if your application uses third-party Java and Android libraries, you may want to reduce the size of the APK. 
-
-There are two steps to enable minification and obfuscation.
+By default, Flutter does not obfuscate or minify Android wrapper. If you intend to use third-party Java or Android libraries, you may want to reduce the size of the APK or protect that code from reverse engineering.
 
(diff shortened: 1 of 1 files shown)


android-release-build.md
```diff
@@ -139,6 +139,54 @@ file.
...
+## Minfify and obfuscate
+
+By default, Flutter does not obfuscate or minify Android wrapper. If you intend to use third-party Java or Android libraries, you may want to reduce the size of the APK or protect that code from reverse engineering.
+
+### Step 1 - Configure obfuscation tool
+
+Create `/android/app/proguard-rules.pro` file and add rules listed below.
+
+```
+#Flutter Wrapper
```
eseidelGoogle: Should this just be part of our template?  @tvolkert @mravn-google ?

swavkulinski: @eseidelGoogle this was the initial idea, however I agree that obfuscation and shrinking is only needed in specific scenarios e.g. hybrid app with large libraries, non-feature requirements etc.

Also running ProGuard slows down build time, which maybe not the big issue for Flutter developers but has it's impact on CI build times.

mravn-google:
I'd prefer to keep it out of the template---at least until we have a template with a clearer separation of files depended on by Flutter tooling vs those that are for the app author to provide/modify (this is work in progress).

Sfshaza:
If we're not putting it into the template, I can add it here. Yes, @mravn-google @eseidelGoogle  ?

Unless I hear otherwise, I'll merge this in the next few.

mravn-google:
@Sfshaza sgtm

Sfshaza:
Thanks @swavkulinski!!

PR #1014, opened 2018-05-08
Fix download links

mit-mit:


mit-mit:
Seems to work in staging now

filiph:
Oh wow, thanks for the fix!

PR #1021, opened 2018-05-13
Update interop list

fmatosqg:
I'm playing the role of the typical keyboard autocorrect. When you say maven do you mean gradle, that's a big ommision? Also why not mention kotlin, when people are getting really engaged?

The other changes are not important ones, but if I'm coming from Android Studio I'd be alarmed if it's not mentioned.


wmleler:
reviewed. LGTM

Sfshaza:
Thanks, @fmatosqg!

PR #1022, opened 2018-05-13
Delete unreachable code

feresr:
Hi, I think I spotted a bug in the flutter docs, this code is unreachable

### DIFF
--- a/cookbook/networking/fetch-data.md
+++ b/cookbook/networking/fetch-data.md
@@ -196,9 +196,6 @@ class MyApp extends StatelessWidget {
               } else if (snapshot.hasError) {
                 return new Text("${snapshot.error}");
               }
-
-              // By default, show a loading spinner
-              return new CircularProgressIndicator();
             },
           ),
         ),


thomas-crane:
This code is reachable in the case where the `snapshot` has not run into an error, but doesn't yet have any data (e.g. the request is still in progress).

Deleting the 'default' return statement actually causes the `FutureBuilder` to return `null` for that code path, and that results in the following error when the code is run.
```
I/flutter ( 4900): The following assertion was thrown building FutureBuilder<Post>(dirty, state:
I/flutter ( 4900): _FutureBuilderState<Post>#9afa8):
I/flutter ( 4900): A build function returned null.
I/flutter ( 4900): The offending widget is: FutureBuilder<Post>
I/flutter ( 4900): Build functions must never return null. To return an empty space that causes the building widget to
I/flutter ( 4900): fill available room, return "new Container()". To return an empty space that takes as little room as
I/flutter ( 4900): possible, return "new Container(width: 0.0, height: 0.0)".
```

You can verify that this is the case by copying the Complete Example from that page into a fresh Flutter app. For whatever reason, the example actually **doesn't** include this default return statement. As a result, the aforementioned error is thrown.

Another pull request should be opened to fix the issue of the Complete Example missing the default return statement.

Sfshaza:
Thank you, @thomas-crane, for jumping in with your thorough explanation and testing! Closing this PR.

Thanks, too, @feresr, for filing a PR!

brianegan:
Hey all, sorry -- missed this one while traveling!

@thomas-crane Is correct -- the code deleted in this PR will be called when the Widget is first built and the Future is waiting for data to load!

Thanks for following up on this @thomas-crane  & @Sfshaza.

PR #1023, opened 2018-05-13
Fix missing BuildContext propagation for Navigator.of

jonasfj:
I was just playing around and figured that `context` probably needs to be propagated for `Navigator.of(context)` to work.


Hixie:
If you're in the `State` of a `StatefulWidget`, then there is a `context` property available from the object scope which should be equivalent to the `context` argument passed to the `build` method. So in theory this shouldn't be needed. Did you try running the code and find it doesn't work?

Sfshaza:
Thanks, @jonasfj for submitting an issue! The code works as is for us, for the codelabs, so I'm closing. If you find otherwise, can you please reopen?

PR #1031, opened 2018-05-16
A few FAQ updates

mit-mit:


faq.md
```diff
@@ -385,11 +390,10 @@ You can compile and deploy your Flutter app to iOS and Android.
 Mobile operating systems: Android Jelly Bean, v16, 4.1.x or newer, and
 iOS 8 or newer.
 
-Mobile hardware: 64-bit iOS devices (starting with iPhone 5S and newer
+Mobile hardware: iOS devices (starting with iPhone 5 and newer
```
eseidelGoogle: @cbracken can give you the data.  I think we run on the 4S just fine. :)

faq.md
```diff
@@ -385,11 +390,10 @@ You can compile and deploy your Flutter app to iOS and Android.
...
 iPhone models), and ARM Android devices.
 
 Note that we currently do not support:
-  * ARM32 iOS devices (iPhone 4, iPhone 5; issue [#2089](https://github.com/flutter/flutter/issues/2089))
   * x86 Android devices (issue [#9253](https://github.com/flutter/flutter/issues/9253))
```
eseidelGoogle: Since this is one item now, we could rephrase this.

Note Flutter currently does not support building for x86 Android (issue #9253) directly, however apps built for ARMv7 or ARM64 run fine (via ARM emulation) on many x86 Android devices.

eseidelGoogle: @cbracken for comment/review.

faq.md
cbracken: Correct. iPhone 4S is our baseline and what we now test in the devicelab. Minor difference, but matters, since it's an older armv7 processor vs the armv7s available in the iPhone 5.

faq.md
cbracken: nit: Either 'Note:' or 'Note that' but other than that, the @eseidelGoogle's proposed phrasing looks great to me!

mit-mit: Yup, went with Eric's phrasing

PR #1035, opened 2018-05-21
fix readme

Distil62:
Change a little misstake in the getting started part write your first application.


Distil62:
I signed it!

Sfshaza:
Hi, @Distil62! Actually, the original code is correct. All even rows in the ListView are word pairings. All odd rows are divider widgets. The index calculation, `final index = i ~/ 2;` , calculates the actual number of word pairings in the ListView, minus the dividers.

PR #1037, opened 2018-05-22
Adding info on landscape view to Android page.

Sfshaza:
Please review, @Hixie.

Staged: https://sz-flutter.firebaseapp.com/flutter-for-android/


Hixie:
I believe that's correct. Someone more familiar with Android may know better; @jonahwilliams, are you familiar with this part of Android?

jonahwilliams:
I believe these are the correct configurations - though it might be worth mentioning somewhere that you almost never want android to restart the activity for a flutter application.  Especially since this goes directly against the advice of the Android documentation.  So supporting split screen, for example, requires you to add screenLayout and probably density too.

Sfshaza:
Please review, @jonahwilliams @willlarche 

I found a number of things that needed correcting, and also need to fix up the language.

PR #1042, opened 2018-05-23
Fix Travis analyze task

mravn-google:
Fixes https://github.com/flutter/flutter/issues/17830


devoncarew:
Thanks @mravn-google!

Sfshaza:
Thanks for fixing, @mravn-google! But it seems strange that this broke so suddenly. Has the tool changed?

Sfshaza:
Thanks!

PR #1045, opened 2018-05-29
[iOS] Language around MD and iOS

willlarche:


Sfshaza:
Oops, I already merged @Hixie  and @filiph! I didn't realize you were also listed as reviewers.

PR #1046, opened 2018-05-29
Wordsmithing the new ios page, also addressing a couple of bugs.

Sfshaza:
Fixes bugs from @mjohnsullivan and @willlarche.

Also some general word smithing and link fixing.

Staged: https://sz-flutter.firebaseapp.com/flutter-for-ios/

### DIFF
--- a/flutter-for-ios.md
+++ b/flutter-for-ios.md
@@ -50,41 +50,41 @@ the hood.
 You'll see that Flutter’s widgets are often styled as ‘Material’ components that
 fit well in a Material Design context such as Android. Material design is just
 one language Flutter UIs can implement, though. On iOS, you can use the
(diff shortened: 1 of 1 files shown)


flutter-for-ios.md
```diff
@@ -426,17 +423,22 @@ class SignatureState extends State<Signature> {
 }
 {% endprettify %}
 
+## Where is the widget's opacity?
+
+On iOS, everything has .opacity or .alpha. In Flutter, most of the time you need to
```
willlarche: It's good but I think it could use an explanation why. It seems like extra composing. Maybe there's a benefit?

Sfshaza: I've asked Ian, and he wasn't sure. I'm going to merge this, for now.

PR #1049, opened 2018-05-31
Tweaks to asset directory docs

mit-mit:
Tweaks to docs added in https://github.com/flutter/website/pull/992 

* Clarify that subdirectories are not included
* Use directory, not folder, to be consistent with rest of doc
* Add sub-heading to make structure clearer


Sfshaza:
Look good, @mit-mit. Thanks!

PR #1061, opened 2018-06-10
 Fix incorrect wordSpacing code markup

daohoangson:
This PR fixes a mixed up between "wordSpacing" and "`properties`" as seen below:

screenshot_20180610-091728


Sfshaza:
Yes, good catch! Thanks for the PR, @daohoangson.

PR #1068, opened 2018-06-16
Update version requirements for VS Code

DanTup:
We often rev the VS Code version we need as they're frequently adding new APIs we need to use. This version has caused confusion (see https://stackoverflow.com/questions/50465219/unable-to-install-dart-code-flutter-in-vs-code) so I think it's better to tell people to just be on latest stable (I don't think there are often good reasons for people to avoid upgrading Code)

### DIFF
--- a/get-started/editor.md
+++ b/get-started/editor.md
@@ -64,7 +64,7 @@ To install these:
 
 ### Install VS Code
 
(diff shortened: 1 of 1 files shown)


Sfshaza:
I'm not @mit-mit, but I LGTM this.  :D

DanTup:
@Sfshaza Thanks!

PR #1069, opened 2018-06-19
Fix typo

thomas-crane:
Change "because no asynnchronous loading needs to take place."
to "because no asynchronous loading needs to take place."


thomas-crane:
I signed it!

Sfshaza:
Excellent catch, @thomas-crane!  Thank you!

PR #1072, opened 2018-06-21
add link to Chinese site in footer

timsneath:
Adds a single link to footer.html which connects to the Chinese site. This should help with PageRank, will give Chinese users a useful link. @chenglu has vetted the Chinese translation.


Sfshaza:
LGTM to me, too, but will wait for @mit-mit.

mit-mit:
LGTM, merging

PR #1073, opened 2018-06-21
add missing bullet between footer links

timsneath:
Despite four of us looking at it, we missed a small bullet glyph between the last two links in the footer!


mit-mit:
Ah, yes missed that!

PR #1078, opened 2018-06-25
Rename Form Widgets for Clarity

brianegan:
This attempts to address this issue: https://github.com/flutter/flutter/issues/18304

Users reading the docs were using `GlobalKey<MyFormState>` instead of `GlobalKey<FormState>`. It's an easy mistake to make, since `MyFormState` and `FormState` look quite similar.

I've changed `MyFormState` to `MyCustomFormState` to hopefully make the two classes more distinct from one another, and added a note to the reader: You must use `GlobalKey<FormState>`.

Please let me know if you think this will help address the issue, or if you can think of something better :)

### DIFF
--- a/cookbook/forms/focus.md
+++ b/cookbook/forms/focus.md
@@ -57,14 +57,14 @@ class. In addition, we need to `dispose` of them when they're no longer needed!
 <!-- skip -->
 ```dart
 // Define a Custom Form Widget
-class MyForm extends StatefulWidget {
+class MyCustomForm extends StatefulWidget {
   @override
-  _MyFormState createState() => _MyFormState();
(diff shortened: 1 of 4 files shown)


Sfshaza:
LGTM, Brian! It's must easier to read given the added length.

jickay:
Looks good to me. Just what happens when things are names similar things. Plus it's always hard to tell what is a name required by the framework and what is a custom one. Comment definitely helps to at least find the problem faster.

brianegan:
Thanks for the feedback @jickay and @Sfshaza!

lordzyx:
I absolutely agree! I really need "readonly" feature.
form validation doesn't work with disabled textformfield, and I need to validate a readonly field, which is editable only by a button. Thank you

PR #1080, opened 2018-06-26
Correcting Swift error in platform-channels.md

mrickard:
Update to receiveBatteryLevel in Swift example to prevent this error on build: "error: call to method 'receiveBatteryLevel' in closure requires explicit 'self.' to make capture semantics explicit."  This affects both simulated and physical iOS devices in Swift-based apps.


mrickard:
I've signed the CLA.

Sfshaza:
Thanks for catching that, @mrickard!

mrickard:
You're welcome, @Sfshaza !

PR #1086, opened 2018-06-29
Remove (optional in Dart 2) `new` keyword

neelkamath:


neelkamath:
I signed it!

Sfshaza:
The edits look good, @neelkamath. Have you tested the code?

neelkamath:
@Sfshaza yup. I deployed the Firebase application [here](https://mit-flutter-staging.firebaseapp.com).

The CI failed though:
```
NoSuchMethodError: Attempted to use type 'File' as a function. Since types do not define a method 'call', this is not possible. Did you intend to call the File constructor and forget the 'new' operator?
```
[This](https://medium.com/flutter-io/flutter-beta-3-7d88125245dc) article says it shouldn't happen:
> The _new_ keyword is now entirely optional: there should be no need to use it any longer, and it’s a bug if the compiler complains. 

CI states the VM being used is:
```
Path to dart is:
../flutter/bin/cache/dart-sdk/bin/dart
Using Dart version:
Dart VM version: 2.0.0-dev.58.0.flutter-f981f09760 (Sat May 26 03:16:14 2018 +0000) on "linux_x64"
```
Perhaps it has something to do with that? Dart 2 [doesn't allow](https://stackoverflow.com/questions/50497121/cannot-omit-new-keyword-despite-using-dart-2-0) omitting the _new_ keyword without a flag, but since Flutter works with Dart 2, we should update the VM being used (if the tests are being run with the server version of Dart instead, and that's possible), or add the command line flag.

Sfshaza:
@RedBrogdon, can you take a look at why this code is failing? Should it be?

RedBrogdon:
@neelkamath has it right -- it's one of the two tooling files involved that's failing. It's server-side dart being run to extract the code snippets for some additional processing. I'm not sure for what, since I'm not that familiar with the engineering odds and ends of the website.

I'd just back out the changes to `tool/extract.dart` and `tool/prdeployer.dart`, land this PR without them, and then file an issue for whoever wrote that code in the first place to update it to Dart 2 and make sure it's running with strong typing.

Sfshaza:
Thanks, @RedBrogdon, excellent suggestion!

Sfshaza:
OK, I scanned through the 64 files and this looks OK. Thanks for backing those tooling changes out.

I'm going to merge this. @neelkamath, you did me SUCH a favor, I'd like to send you a [Dash plushie](https://www.instagram.com/dash_the_dartlang_plushy/). You can contact me directly.

neelkamath:
@Sfshaza the build looks like it's passing ([this](https://travis-ci.org/flutter/website) is what I'm looking at). Is there a different build I'm supposed to be viewing?

Sfshaza:
Yes, @neelkamath, it later passed. Who knows why.

neelkamath:
@Sfshaza a bit late I know, but how do I "contact you directly"?

PR #1088, opened 2018-07-03
Add a note about how to fix too many open files on MacOS.

Sfshaza:
Fixes https://github.com/flutter/flutter/issues/18225

Please review, @goderbauer.

(I was wondering, mightn't Linux have the same problem as MacOS?)

There's only one problem in the file now. Know of any others I can add to make it more of a "real" section?

Staged:  https://sz-flutter.firebaseapp.com/debugging/#common-problems

thx

### DIFF
--- a/debugging.md
+++ b/debugging.md
@@ -887,8 +887,9 @@ The following is a problem that some have encountered on MacOS.
 
 ### "Too many open files" exception (MacOS)
 
-The Mac file system has a default limit on how many files it can have open
-at a time. If you run into this limit, increase the number of available
+The default limit for Mac OS on how many files it can have open at a
+time is rather low.  If you run into this limit,
+increase the number of available
 file handlers using the `ulimit` command:
 
 ```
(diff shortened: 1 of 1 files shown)


goderbauer:
@Sfshaza On Linux the default value for the number of open files is much higher than on Mac OS. Therefore, people are less likely to run into the issue on Linux.

debugging.md
```diff
@@ -877,3 +877,25 @@ You can also overlay such a grid on non-Material applications by using
...
+## Common problems
+
+{% comment %}
+Rewrite the following when we have a larger collection of problems.
+{% endcomment %}
+The following is a problem that some have encountered on MacOS.
+
+### "Too many open files" exception (MacOS)
+
+The Mac file system has a default limit on how many files it can have open
```
goderbauer: My understanding is that this is not a problem with the file system, but rather a safety measure enforced by the shell (https://ss64.com/bash/ulimit.html). 

Also: Other operating system also have a default limit of open files as well. Their limits are just much higher by default so people are unlikely to run into problems.

I would maybe rephrase this to: 

```The default limit for Mac OS on how many files it can have open at a time is rather low. If you run into this limit, ...```

PR #1090, opened 2018-07-03
Adding info about GitLab continuous integration testing.

Sfshaza:
Fixes https://github.com/flutter/flutter/issues/18813

@brianegan, since I'm referencing your package, you clearly know how to do this. Please review and feel free to offer suggestions on improving it, or a package on pub that can be used for an example.

thx

### DIFF
--- a/testing.md
+++ b/testing.md
@@ -10,8 +10,8 @@ permalink: /testing/
 ## Introduction
 
 The more features your app has, the harder it is to test it manually. A
-good set of automated tests will help you make sure your app performs
-correctly before you publish it while retaining your feature and bug fix
+good set of automated tests help you make sure your app performs
+correctly before you publish it, while retaining your feature and bug fix
 velocity.
 
 There are many kinds of automated testing. These are summarized below:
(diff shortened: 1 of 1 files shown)


brianegan:
Hi @Sfshaza! Overall, I think this is a good start, at least as an example.

The docker container this file makes use of is a bit out of date -- I've just pushed up a new version to the docker registry and will update the flutter_redux repo after the Docker registry finishes building it.

In the end, it might be good to have a more in-depth explanation as to what's going on, since Gitlab requires you to work with Docker containers, and the container I've created might not work for everyone. This example also runs code coverage and publishes the results to the gitlab page for the repo, which might not be what folks expect / want.

brianegan:
Ok, the docker container has been updated and the example should be working fine :)

Sfshaza:
Thanks, @brianegan! I chatted with Matt about this and he felt that, at least for now, GitLab is too niche to spend a lot of time on it. 

We welcome PRs on improving it!!!

Thanks for your insights, but I will merge this for now.

PR #1091, opened 2018-07-03
correct folder name to be l10n

mathieujobin:
there is no such thing as i10n, its a typo
i18n means internationalization
l10n means localization

i10n would mean internation

2nd command fails because first folder is wrong


Sfshaza:
Great catch, @mathieujobin! Once you sign the CLA, I will merbe.

Sfshaza:
Sometimes it just takes time. Thanks!

PR #1093, opened 2018-07-04
Add macOS keyboard shortcut too

neelkamath:


using-ide.md
```diff
@@ -104,7 +104,7 @@ has not been able to discover any connected iOS or Android devices or simulators
 You need to connect a device, or start a simulator, to proceed.
 
 1. Locate the **Flutter Target Selector** drop-down button. This shows a
- list of available targets. Select the target you want your app to started on.
+ list of available targets. Select the target you want your app to be started on.
```
Sfshaza: This is definitely more correct than beofre, though the dangling participle hurts me, a bit  ;)

Sfshaza:
LGTM

using-ide.md
neelkamath: *before 😜

PR #1094, opened 2018-07-05
Added "new" operator before MethodChannel and MethodCallHandler

iamSahdeep:
The java code snippet provided in **Step:3a** wont work unless we use `new` operator before `MethodCannel()` and `MethodCallHandler()`. Also, `GeneratedPluginRegistrant.registerWith(this);` is missing from that snippet so, I have added it too. 

Reference taken from  this [example](https://github.com/flutter/flutter/blob/edc16473bfdafc0f5fe546610d3b0fe0eecdd080/examples/platform_channel/android/app/src/main/java/com/example/platformchannel/MainActivity.java#L53)

### DIFF
--- a/platform-channels.md
+++ b/platform-channels.md
@@ -224,9 +224,10 @@ public class MainActivity extends FlutterActivity {
     public void onCreate(Bundle savedInstanceState) {
 
         super.onCreate(savedInstanceState);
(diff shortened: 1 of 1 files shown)


Sfshaza:
LGTM

PR #1096, opened 2018-07-07
fix typing error

vinicentus:
Just a minor typing error...


vinicentus:
Signed it...

Sfshaza:
Thanks for catching that, @vinicentus!  LGTM

PR #1097, opened 2018-07-10
VS Code get-started/test-drive updates

chalin:
- Update steps to match current use of VS Code.
- Copyedits to match Google dev doc guide.

cc @kwalrath @filiph


Sfshaza:
LGTM

PR #1098, opened 2018-07-11
Add more details to the hot reload page

xster:


hot-reload.md
```diff
@@ -44,7 +45,9 @@ app continues to execute from where it was prior to running the hot reload
 command. The code is updated and execution continues.
 
 A code change has a visible effect only if the modified Dart code is run again
-after the change. The next sections describe common situations where the
+after the change. Specifically, a hot reload causes all the existing widgets to rebuild. Code that is not invoked during the rebuilding of the widgets are not automatically re-executed.
```
aam: Last statement is not clear. Do you mean to say that `only code that is involved in rebuilding of the current widgets is automatically re-executed`?

hot-reload.md
```diff
@@ -281,3 +284,11 @@ to:
 
 In these situations, hot reload generates a diagnostic message and fails without
 committing any changes.
+
+## How it works
+
+When hot reload is invoked, the host machine looks at the edited code since the last compilation. Each library with changed code, the application's main library, and the libraries from the main library leading to affected libraries are recompiled. In Dart 2, those libraries' Dart source code are turned into [kernel files](https://github.com/dart-lang/sdk/tree/master/pkg/kernel) and sent to the mobile device's Dart VM.
```
Sfshaza: .."the host machine" -> Do you mean framework?

I think it would be clearer to break down the second sentence:

last compilation, and the following libraries are recompiled:

* Any libraries with changed code.
* The app's main library.
* The libraries from the main library leading to affected libraries.

hot-reload.md
```diff
@@ -281,3 +284,11 @@ to:
...
+
+The Dart VM re-loads all libraries from the new kernel file. So far no code is re-executed.
+
+The hot reload mechanism then causes the Flutter framework to trigger a rebuild/re-layout/repaint of all widgets and render objects.
```
Sfshaza: "of all widgets and.." => "of all affected widgets and..."  ?

hot-reload.md
xster: Thanks. Done

hot-reload.md
xster: Ah, it's clearer this way. Thanks. 

The host machine is the machine with the source code running the flutter tool (as opposed to the mobile device).

hot-reload.md
xster: I don't want to give the impression that only some of the widgets are rebuilt since all widgets are rebuilt. Changed to 'all existing widgets'.

Sfshaza:
LGTM

PR #1100, opened 2018-07-12
Get started codelab: markdown cleanup

chalin:
@Sfshaza - I was working on another PR over the codelab when I noticed that some site-www "What's the point" CSS updates were not part of this site.

Also, on www we've cleaned-up the "What's the point" markdown considerably.

Some other markdown cleanup done in this PR includes:

- Converting `<ul>` and `<ol>` elements to pure markdown
- Showing how we can use standard markdown headers instead of `<p class="h2-like">`
- Using fenced markdown codeblocks when code is indented rather than the prettyprint filter (the resulting code highlighting is the same).

I find that if you look at the resulting markdown, it is much more readable without all that embedded HTML.

Are updates like these helpful? If so, I'll cleanup the rest of the file before doing the other codelab work (since it will help me in my subsequent work too).

cc @kwalrath @filiph

### DIFF
--- a/_sass/_tutorial.scss
+++ b/_sass/_tutorial.scss
@@ -3,14 +3,12 @@
   background-color: $grey-color-light;
   padding: 20px;
   margin-bottom: 20px;
-  padding-top: 10px;
-  margin-top: 0px; 
+  margin-top: 0px;
   h1, h2, h3, h4, h5, h6 {
-    margin: 0;
-    margin-bottom: 5px;
+    color: $text-color;
+    &:nth-child(1) { font-weight: bold; }
   }
-  ul {
-    padding-left: 15px;
-    margin-bottom: 0;
+  > ul, > p {
+    &:last-child { margin-bottom: 0; }
   }
 }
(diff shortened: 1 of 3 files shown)


Sfshaza:
Weird, it looks like Travis failed because of indentation.

Patrice, the cleaned-up markdown looks great. I can't see it staged, though. It still looks good? I drop into an  HTML list when the markdown list won't work - usually when I have code or an image in the list.

Assuming this looks right when generated, I approve. Though you'll have to fix the  build.

Sfshaza:
And yes, these updates are most helpful, @chalin.  🗡 

LGTM given comment above.

chalin:
> Travis failed because of indentation.

Fixed. Travis is green now.

> Patrice, the cleaned-up markdown looks great. I can't see it staged, though.

I can't (currently) submit a PR directly to this repo yet. AFAIK auto-staging doesn't work for PRs originating from a fork.

> It still looks good?

Look the same or better ;)
I've got it staged now: https://ng2-io.firebaseapp.com/get-started/codelab

Sfshaza:
You are correct. Autostaging only works on a branch of the master.

Thanks for staging it. 'Tis a thing of beauty. You rock!

LGTM

PR #1109, opened 2018-07-22
updated "then" method parameter

AlterEgoDev:
updated "then" method parameter to current Dart version
currently throws "The argument type '(Null) → DemoLocalizations' can't be assigned to the parameter type '(bool) → FutureOr<DemoLocalizations>'."


Sfshaza:
Thanks for catching that!

PR #1112, opened 2018-07-24
Update the tutorials to point to the Geolocator plugin

mvanbeusekom:
Updated the chapter "How do I access the GPS sensor?" to point to the "geolocator" plugin instead of the "location" plugin.

### DIFF
--- a/flutter-for-android.md
+++ b/flutter-for-android.md
@@ -2083,7 +2083,7 @@ class _SampleAppPageState extends State<SampleAppPage> {
 
 ## How do I access the GPS sensor?
 
-Use the [`location`](https://pub.dartlang.org/packages/location) community plugin.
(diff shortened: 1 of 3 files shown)


RedBrogdon:
Can you provide a little context for this change? If I understand it correctly, this would recommend one community-developed geo package rather than another?

mvanbeusekom:
@RedBrogdon of course, sorry I should have done so directly.

In cooperation with the Flutter team (Tim, Mikkel and Micheal) we developed the Geolocator plugin as a replacement for the Location plugin. This was done meanly because the Location plugin lacks functionality (which is added in the new Geolocator plugin).

Features like:
- Configure accuracy and update distance;
- Geocoding;
- Calculate distance between two points.

Sfshaza:
Thanks for clarification, @mvanbeusekom. Merging.

PR #1113, opened 2018-07-24
chore: upgrade Jekyll prettifier plugin

chalin:
Copied from https://github.com/dart-lang/site-shared/blob/master/src/_plugins/prettify.rb

@filiph, @Sfshaza - I've ensured that the generated site ([archived in chalin/flutter_site.g](https://github.com/chalin/flutter_site.g)) contains no differences.


_plugins/prettify.rb
```diff
@@ -1,77 +1,112 @@
+# Copied from https://github.com/dart-lang/site-shared/blob/master/src/_plugins/prettify.rb
```
chalin: @filiph - FYI, I've chosen, for now, to just copy the file rather than setup use of site-shared (I don't want to introduce too many changes upfront).

Sfshaza:
Thanks, @chalin!  LGTM!  (Once it passes Travis.)

Sfshaza:
I thought we don't normally check in Gemfile.lock.

chalin:
We checking the Gemfile.lock to ensure that we're running the same gems everywhere (locally and on Travis). We're doing the same for all three of our sites (FYI).

Sfshaza:
LGTM

PR #1115, opened 2018-07-24
chore: use configurable server.sh script

chalin:
Get the local server port from the `_config.yml` file.
Choosing 4002 to avoid clashing with site-www and site-webdev.

cc @Sfshaza @filiph


Sfshaza:
LGTM

PR #1118, opened 2018-07-24
chore: temporarily disable autostaging

chalin:


Sfshaza:
Auto-staging is failing for those of us who commit to master. (@InMatrix had the same problem earlier.)

I asked @chalin to disable it for now.  LGTM.  And thanks!

PR #1121, opened 2018-07-25
Fix the fixes -- tense & new changes

brianegan:
Based on feedback from @Sfshaza


brianegan:
Haha, good call. *Should* do that too :P

PR #1125, opened 2018-07-31
[20030] Added command to run a single test.

manhluong:
Related issue: https://github.com/flutter/flutter/issues/20030

### DIFF
--- a/testing.md
+++ b/testing.md
@@ -308,6 +308,14 @@ void main() {
 }
 ```
 
+### Running only one test
+
+It is possible to run only one test, using `--plain-name`:
+
+```
+flutter test --plain-name "Description passed to testWidgets() or to test()" path/to/test_file.dart 
+```
+
 ### Running integration tests
 
 To run the test on an Android device, connect the device via USB to your


chalin:
Thanks for the PR!

Rather than add special cases for use of the various command-line options, I think that we can more generally suggest readers to consult the command help: ` flutter test --help` (which is what you ended up doing - https://github.com/flutter/flutter/issues/20030#issuecomment-409164967). @Sfshaza WDYT?

manhluong:
@chalin Thanks for the feedback!

Imho running a single test is a very common thing to do, like when you have multiple tests that are failing and you are tackling them one by one, that's why I think is better to add a section in the documentation.

manhluong:
As @zoechi also pointed out in the MR, in the dart docs is stated how to run a single test:

https://github.com/dart-lang/test#running-tests

zoechi:
What I meant is that the Flutter test docs could point out that its test framework is based on dart-lang/test and that the docs found there apply as well and should be consulted to learn about the full range of features.

Sfshaza:
Thanks, @manhluong, for this PR! We love it when devs help fix the docs.

The text you've added is very straightforward, but I tend to agree with @zoechi and @chalin that the Flutter doc should explain that it's based on dart-lang/test and also that they should consult the --help option.

If you'd rather not make those changes, I can do it. And thanks again!

manhluong:
Thanks @Sfshaza !

I will let you do these changes, and I will close this MR, as I don't think that my writing skills are very good! :D

chalin:
#1131 is the followup issue.

porn:
Hey, the `flutter test --help` still doesn't contain any info on how to run single test:
```
$ flutter test --help
Run Flutter unit tests for the current project.

Usage: flutter test [arguments]
-h, --help                      Print this usage information.
    --[no-]pub                  Whether to run "flutter pub get" before executing this command.
                                (defaults to on)

    --name=<regexp>             A regular expression matching substrings of the names of tests to run.
    --plain-name=<substring>    A plain-text substring of the names of tests to run.
    --start-paused              Start in a paused mode and wait for a debugger to connect.
                                You must specify a single test file to run, explicitly.
                                Instructions for connecting with a debugger are printed to the console once the test has started.

    --coverage                  Whether to collect coverage information.
    --merge-coverage            Whether to merge coverage data with "coverage/lcov.base.info".
                                Implies collecting coverage data. (Requires lcov)

    --coverage-path             Where to store coverage information (if coverage is enabled).
                                (defaults to "coverage/lcov.info")

    --update-goldens            Whether matchesGoldenFile() calls within your test methods should update the golden files rather than test for an
                                existing match.

-j, --concurrency=<jobs>        The number of concurrent test processes to run.
                                (defaults to "6")

    --[no-]test-assets          Whether to build the assets bundle for testing.
                                Consider using --no-test-assets if assets are not required.
                                (defaults to on)

    --platform                  The platform to run the unit tests on. Defaults to "tester".
                                [tester (default), chrome]

Run "flutter help" to see global options
```
flutter version:
```
$ flutter --version
Flutter 1.7.8+hotfix.4 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 20e59316b8 (3 weeks ago) • 2019-07-18 20:04:33 -0700
Engine • revision fee001c93f
Tools • Dart 2.4.0
```

in case you are wondering how to test single test, this is how to do it:
```
flutter test test/backend_test.dart
```

lsaudon:
```bash
flutter test --plain-name "name of your test"
```

PR #1128, opened 2018-08-02
Integration testing recipes

brianegan:
Hey hey :)

This adds to the other recipes on testing! The "Test your apps" page has a decent summary of writing integration tests, but I felt like it was missing a lot of important details when I first started writing integration tests. Therefore, I split that example up into smaller chunks and tried to provide a bit more detail for each part.

Three recipes included:

  * Introduction to integration testing
  * Scrolling in integration tests
  * Performance profiling with integration tests

Guidance needed: Should I change up the "Test your apps" page a bit as well? My idea: the page provides a great overview of the testing solutions that Flutter provides. I was thinking of keeping that overview and removing the examples, linking to all of these recipes instead. Thoughts?

### DIFF
--- a/cookbook/testing/integration-test-introduction.md
+++ b/cookbook/testing/integration-test-introduction.md
@@ -1,0 +1,251 @@
+---
+layout: page
+title: "Introduction to Integration testing"
+permalink: /cookbook/testing/integration-test-introduction/
+---
+
+Unit tests and Widget tests are handy for testing individual classes, functions,
+or Widgets. However, they generally don't test how individual pieces work
+together as a whole or capture the performance of an application running on a
+real device. These tasks are performed with integration tests.
+
+Integration tests work as a pair: first, deploy an instrumented application to a
+real device or emulator and then "drive" the application from a separate test
+suite, checking to make sure everything is correct along the way.
+
+To create this test pair, we can use the 
+[flutter_driver](https://docs.flutter.io/flutter/flutter_driver/flutter_driver-library.html) 
+package. It provides tools to create instrumented apps and drive those apps
+from a test suite.
+
+In this recipe, we'll learn how to test a counter app. It will demonstrate
+how to setup integration tests, how to verify specific text is displayed by the 
+app, how to tap on specific Widgets, and how to run integration tests. 
+
+### Directions
+
+  1. Create an app to test
+  2. Add the `flutter_driver` dependency
+  3. Create the test files
+  4. Instrument the app
+  5. Write the integration tests
+  6. Run the integration test
+  
+### 1. Create an app to test
+
+First, we'll create an app that we can test! In this example, we'll test the 
+counter app produced by the `flutter create` command. This app allows
+a user to tap on a button to increase a counter.
+
+Furthermore, we'll also need to provide a
+[`ValueKey`](https://docs.flutter.io/flutter/foundation/ValueKey-class.html) to
+the `Text` and `FloatingActionButton` Widgets. This allows us to identify 
+and interact with these specific Widgets inside the test suite. 
+
+```dart
+import 'package:flutter/material.dart';
+
+void main() => runApp(MyApp());
+
+class MyApp extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return MaterialApp(
+      title: 'Counter App',
+      home: MyHomePage(title: 'Counter App Home Page'),
+    );
+  }
+}
+
+class MyHomePage extends StatefulWidget {
+  MyHomePage({Key key, this.title}) : super(key: key);
+
+  final String title;
+
+  @override
+  _MyHomePageState createState() => _MyHomePageState();
+}
+
+class _MyHomePageState extends State<MyHomePage> {
+  int _counter = 0;
+
+  void _incrementCounter() {
+    setState(() {
+      _counter++;
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: Text(widget.title),
+      ),
+      body: Center(
+        child: Column(
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: <Widget>[
+            Text(
+              'You have pushed the button this many times:',
+            ),
+            Text(
+              '$_counter',
+              // Provide a Key to this specific Text Widget. This allows us
+              // to identify this specific Widget from inside our test suite and
+              // read the text.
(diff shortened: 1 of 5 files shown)


Sfshaza:
This looks amazing, @brianegan! I am OOO today, so will ask for some engineering review.

@chalin, @chinmaygarde

Sfshaza:
@yjbanov, would you mind reviewing?

cookbook/testing/integration-test-scrolling.md
```diff
@@ -0,0 +1,208 @@
...
+of items and verify that a particular item exists in the list. The
+[`FlutterDriver`](https://docs.flutter.io/flutter/flutter_driver/FlutterDriver-class.html)
+class provides three methods for scrolling through lists:
+
+  - The
+  [`scroll`](https://docs.flutter.io/flutter/flutter_driver/FlutterDriver/scroll.html)
+  method allows us to scroll through a specific list by a given amount. 
+  - The
+  [`scrollIntoView`](https://docs.flutter.io/flutter/flutter_driver/FlutterDriver/scrollIntoView.html)
+  method finds a specific Widget that's currently displayed on screen and will
```
yjbanov: nit: "currently displayed on screen" may not be accurate. It may not actually be visible on the screen. It might be rendered offscreen (e.g. clipped by a scrollable viewport). However, the widget we're looking for must be instantiated and rendered into the render object hierarchy, even if its pixels are not painted on screen.

This method is useful when you know that something is eagerly instantiated and all you want is shift to a position such that it's visible. Perhaps you are scrolling a `SingleChildScrollView`. Or perhaps you are testing a scrollable container that renders widgets on-demand (e.g. [ListView.builder](https://docs.flutter.io/flutter/widgets/ListView/ListView.builder.html)) and you expect that at a certain position a certain widget must currently be rendered.

Contrast this to `scrollUntilVisible`, which will scroll "hoping" that eventually the widget we are looking for will be rendered. While `scrollIntoView` will fail instantly if it's not already there, `scrollUntilVisible` may keep scrolling hopelessly until it times out (especially if the list is indefinite).

**Suggestion**: perhaps instead of "displayed" or "visible", we say "rendered", which strongly implies `RenderObject`. A paragraph explaining the difference between being "rendered" and being "visible" might help readers better understand how these methods work (and the framework in general; scrolling is complicated :sweat_smile:).

cookbook/testing/integration-test-scrolling.md
```diff
@@ -0,0 +1,208 @@
...
+While all three methods work for specific use-cases, `scrollUntilVisible` is
+oftentimes the most robust option. Why?
+
+  1. If we use the `scroll` method alone, we might incorrectly assume the height
+  of each item in the list. This could lead to scrolling too much or too little.
+  2. If we use the `scrollIntoView` method, we're assuming the Widget is visible
+  on screen. In order to verify our apps work on a broad range of devices, we
+  might run our integration tests against devices with different screen sizes.
+  Whether or not a particular Widget is visible can depend on the size of the
+  screen.
```
yjbanov: I like this recommendation, but as explained above "visible on screen" may not be accurate enough.

cookbook/testing/integration-test-profiling.md
```diff
@@ -0,0 +1,182 @@
...
+});
+```
+
+### 3. Save the results to disk
+
+Now that we've captured a performance timeline, we need a way to inspect it! 
+The `Timeline` object provides detailed information about all of the events that
+took place. Therefore, it can be a bit difficult to read and understand.
+
+Instead of inspecting the raw data contained within the `Timeline`, we can
```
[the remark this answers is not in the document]
brianegan: Oh wow, I had no idea you could import that into Chrome's tracing tools! I'll definitely add that.

cookbook/testing/integration-test-introduction.md
```diff
@@ -0,0 +1,252 @@
...
+### 6. Run the tests
+
+Now that we have an instrumented app and a test suite, we can run the tests!
+First, be sure to launch an Android Emulator, iOS Simulator, or connect your 
+computer to a real iOS / Android device.
+
+Then, run the following command from the root of the project:
+
+```
+flutter drive --target=test_driver/app.dart
```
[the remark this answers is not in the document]
brianegan: @Sfshaza What are your thoughts here? I could definitely add that info, but I think I've seen other PRs where it's recommended that we don't dive too deeply into the different options the CLI tools provide.

cookbook/testing/integration-test-profiling.md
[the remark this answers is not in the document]
brianegan: Uh, that's awesome. I'll definitely add this info!

cookbook/testing/integration-test-scrolling.md
brianegan: Good call -- I'll definitely add some more info here since, as you can tell, I was a bit confused by looking at the API docs on exactly how that was working :P

Would it be ok to ping you for a review once I put it in? I definitely want to make sure everything is technically correct.

brianegan:
@yjbanov Thanks so much for the thorough review! Really cool to have your input on this topic.

cookbook/testing/integration-test-scrolling.md
yjbanov: Of course! Also, if you can come up with a better wording for our dartdocs, please feel free to send a PR to improve it.

brianegan:
@yjbanov I've pushed on more commit! I decided to forego and entire paragraph on the difference between visible vs rendered since I felt it distracted a bit too much from the main point. Instead, I tried to describe how using a `ListView.builder` will render widgets on-demand and why that could cause problems. Please let me know if these changes are technically accurate and convey the information you'd like to see!

brianegan:
@Sfshaza The CI is failing because `chrome://tracing` is not a valid link. Fair enough, it's not universal to all browsers! However, I think it'd be cool to make it a proper link to convey the intent that you need to navigate to that URL in Chrome. What would your advice be in this case? Remove the link, or is there a way to get around that check for specific links?

chalin:
> ... is there a way to ignore that check for specific links?

Yes: add a regexp to the `options.url_ignore` list in the `website/Rakefile`.

brianegan:
Thanks @chalin! That did the trick.

yjbanov:
lgtm

Sfshaza:
Thanks so much, @brianegan!

Sfshaza:
And thanks for reviewing, @yjbanov!

PR #1130, opened 2018-08-02
Update FAQ entry about app size

goderbauer:
APK size reductions are real!

See https://github.com/flutter/flutter/issues/16833.


Hixie:
LGTM

Sfshaza:
It's so nice to see such enthusiasm over a PR.  :D

PR #1141, opened 2018-08-08
Fixes the Flutter for React Native doc's bug on several Android and IOS devices. 

Samaritan1011001:
I rendered this locally and on the firebase staging website and it seems to solve the problem happening on several devices. But this might not be the complete solution as you can see, the length of a single word being too long is causing the problem and hence I have given the space between these long words. Since they are functions, it does not make sense to do it. I am open for suggestions on what can be done. 

img_b5d12f351a94-1

### DIFF
--- a/flutter-for-react-native.md
+++ b/flutter-for-react-native.md
@@ -1,0 +1,1964 @@
+---
+layout: page
+title: Flutter for React Native Developers
+permalink: /flutter-for-react-native/
+---
+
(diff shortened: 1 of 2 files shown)


Sfshaza:
Hi, @Samaritan1011001! Thanks for bringing attention to a problem when viewing this doc on a device. Unfortunately, breaking up a method call with spaces will lead to confusion. Can you file a bug on this issue? We are currently reworking the website and mobile is an important aspect.  Thanks.

Samaritan1011001:
Sure @Sfshaza. I have created an issue here: https://github.com/flutter/website/issues/1159

PR #1146, opened 2018-08-09
Update comment and tweak retries count

gspencergoog:
Two minor tweaks, but really I just want to try the "deploy" part of the CI setup by submitting another PR.

### DIFF
--- a/tool/build_check_deploy.sh
+++ b/tool/build_check_deploy.sh
@@ -43,7 +43,7 @@ function check_formatting() {
 # Deploys website to Firebase, trying up to total_tries times.
 function deploy() {
   local total_tries="$1"
-  local remaining_tries=$(($total_tries - 1))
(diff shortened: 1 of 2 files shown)


tool/build_check_deploy.sh
```diff
@@ -43,7 +43,7 @@ function check_formatting() {
 # Deploys website to Firebase, trying up to total_tries times.
 function deploy() {
   local total_tries="$1"
-  local remaining_tries=$(($total_tries - 1))
+  local remaining_tries=$(($total_tries))
```
chalin: If we're dropping the arithmetic, then maybe just:

```bash
local remaining_tries=$total_tries
```

gspencergoog: Oh, duh.  Done.

PR #1158, opened 2018-08-15
Added Sketch file that contains a few Cupertino widget catalog images.

matthew-carroll:
Added Sketch file that contains a few Cupertino widget catalog images.


Sfshaza:
Thanks, Matt!!

PR #1166, opened 2018-08-20
fix parse error

ryansong:
Variable 'p' should be  'emailRegexp'.


ryansong:
I signed it!

Sfshaza:
Thanks for catching that, @ryansong! Do you think you can fix the conflicts?

Sfshaza:
Cirrus won't let me force merge, so I'm fixing in another PR:  https://github.com/flutter/website/pull/1191

PR #1169, opened 2018-08-20
Site file reorg: move site sources to src folder

chalin:
No documentation or site formatting changes in this PR. In fact, the only side effect, is to remove files from the deployed site that should never have been there in the first place:

```console
> git diff --summary
 delete mode 100644 _site/AUTHORS
 delete mode 100644 _site/README.md
 delete mode 100644 _site/Rakefile
 delete mode 100644 _site/analysis_options.yaml
 delete mode 100644 _site/firebase.json
 delete mode 100644 _site/firebase_config/firebase_rules.json
 delete mode 100644 _site/pubspec.lock
 delete mode 100644 _site/pubspec.yaml
```

cc @Sfshaza


miksen:
@chalin we seem to have lost all history to the content after this change !? E.g.:

https://github.com/flutter/website/commits/master/src/fastlane.md
https://github.com/flutter/website/commits/master/src/accessiblity.md

mit-mit:
@chalin we seem to have lost all history to the content after this change !? E.g.:

https://github.com/flutter/website/commits/master/src/fastlane.md
https://github.com/flutter/website/commits/master/src/accessiblity.md

chalin:
The history isn't lost, GH just doesn't make it directly accessible -- see _Support `git --follow` to preserve history across renames_ https://github.com/isaacs/github/issues/900 (upvoting this issue might help). Some solutions mentioned in that issue are:

- Using this [chrome extension](https://chrome.google.com/webstore/detail/github-follow/agalokjhnhheienloigiaoohgmjdpned).
- Locally running `git log --follow path/to/file.ext`

E.g.,

```console
> git shortlog --follow src/fastlane.md
7449 (1):
      fix flutter-for-react-native CheatSheet link (#964)

Adam (1):
      Flutter for Xamarin.Forms (#1087)

Adam Barth (110):
      Initial commit
     ...snip...
      Update FractionalOffset references to Alignment (#691)

Adam Williams (1):
      start => starts typo fix (#613)

Aditya Arora (1):
      get-sdk.md: Update installation docs (#774)

Akshay Chordiya (1):
      Fixed image url not working (#854)

Alexandre Ardhuin (11):
      Update getting-started.md
      ...snip...
      output expected formatting on travis (#268)

Alter Ego (1):
      fixed typo (#957)

Alvin Konda (1):
      update (#892)

Andrew Brogdon (1):
      Updating FAQ with APK/ipa details. (#1084)

Arthur Thompson (1):
      Update upgrading.md (#642)

Bema (1):
      Add missing parameter to itemBuilder callback

Benjamin Misell (1):
      Update iOS setup (#918)

Brandon Donnelson (3):
...
```

mit-mit:
Ugh, is there anything we can do to make the GitHub UI work again? If not, I think we should revert this PR (having the GitHub UI work is more valuable than having a few less files in the root).

chalin:
> Ugh, is there anything we can do to make the GitHub UI work again?

Showing interest for https://github.com/isaacs/github/issues/900 might be our best bet.

> If not, I think we should revert this PR (having the GitHub UI work is more valuable than having a few less files in the root).

This is just the beginning of structural changes to that repo. As you know a new version of the site is in the works. This is a first step along that path. (Btw, all of our other sites are organized in this way. This new organization helps separate concerns: localizing Jekyll source files to `/src`, while separating them from other resources like code, tooling, etc.)

I believe that the root of the problem is that source code was embedded "deep" inside the Jekyll website source structure. It is usually best to keep sources at the root below, say, `examples`.

That being said, I'll let @Sfshaza and @filiph weigh in.

Sfshaza:
I know that @mitmit is concerned that we preserve the history of the repo,
but when we reworked www.dartlang, for example, we started a brand new
repo. We preferred not to drag all the old history around, which had
greatly bloated the repo.

So, it's not my decision, but I'm OK with it. In the case of www.dartlang,
we saved the old repo for reference.

Filip?

filiph:
I'm strongly in favor of moving the files. As Patrice says, this is just a beginning of a larger refactor — it's not just a one-time cosmetic change. It unlocks the ability to grow the site, and bring in valuable tools and shared content from dartlang and webdev.

While I understand the need to check history/blame for a piece of documentation, I think that regular users don't do it that often, if ever. Power users can use `--follow` from the command line.

mit-mit:
If you all think the move has value, and is aligned with future changes, then just go ahead. I was just reacting to a PR that was listed as having "no side effects" that did have an affect on something that I care about.

I would also add that -- as shown by the contribution history of this repo -- a lot of this repo was created by non-"power users", so it's important to consider our work flows as-well.

chalin:
> If you all think the move has value, and is aligned with future changes, then just go ahead.

Ok, great.

> "no side effects" that did have an affect on something that I care about.

Sorry about that. The "no side effects" was meant to imply that the site would look the same.

Since the git history isn't lost, it's just not currently accessible from GH, making noise on https://github.com/isaacs/github/issues/900 might be a good thing to do.

As you probably know, while you can't view the _complete_ history of the relocated files all in one go, you can use GH view the history of files before they were relocated. For example, here is the history for `fastlane.md` (as viewed from the last commit before the file reorg): https://github.com/flutter/website/commits/7871e94d289189f860fbe069dd8729770cc73667/fastlane.md

If this is something that you'd like to do often enough, I can create a branch referring to the last commit before the file reorg. It will make viewing history easier. Let me know if that might be helpful.

mit-mit:
It's OK really, let's just focus on getting all those awesome future changes done!

chalin:
SG!

xster:
Can you update the asset references from the main repo like https://github.com/flutter/flutter/blob/master/README.md too? :)

chalin:
Yes, thanks for pointing that out! Done: https://github.com/flutter/flutter/pull/20935.

PR #1178, opened 2018-08-20
Fix broken links caused by src/ dir move in 4ed3f9f from #1169

slightfoot:
@chalin The src dir change caused broken links on the site, and we've been seeing complaints in the Gitter chat. Do you think you can review this and get this merged ASAP?

### DIFF
--- a/src/_layouts/page.html
+++ b/src/_layouts/page.html
@@ -7,7 +7,7 @@ layout: default
   {% unless page.hide_title %}
   <header class="post-header">
       <div class="btn-group contribute" role="group">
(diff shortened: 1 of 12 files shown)


slightfoot:
Ref #1169

chalin:
> The CI runs checks to ensure there aren't any broken links on the site, but the links weren't broken until the change was merged.

Strange indeed. I'll see if that can be remedied.

chalin:
Given that the docs link directly into the repo, we can't avoid this, but we can mitigate it.

PR #1180, opened 2018-08-21
Fix spelling of fastlane

KrauseFx:


Sfshaza:
They do seem to use it as lower case.

KrauseFx:
Yes we do 😃

Sfshaza:
lol!  :)

PR #1187, opened 2018-08-24
Fix letter capitalization

philippgille:


philippgille:
I signed it!

Sfshaza:
Thanks for catching that, @philippgille!

PR #1209, opened 2018-08-30
Migrate to bootstrap 4

### LINKED ISSUE #1194: Migrate to bootstrap 4
- [ ] Drop `.visible*` classes, they no longer exist; replace with use of appropriate `.d-*` classes or media breakpoints.
- [ ] Bump up grid tiers, but not for buttons sizes (`btn-sm` stays as is).
- ...

chalin:
Closes #1194

cc @filiph @fertolg


Sfshaza:
Wow! What a lot of work, @chalin. I scanned the file changes and played with the staged site. It looks great!

Sfshaza:
LGTM from me, but waiting for @filiph to weigh in.

src/_includes/head.html
```diff
@@ -27,7 +27,7 @@
   {% assign og_image_path = page.image | default: layout.image | default: '/images/flutter-logo-sharing.png' %}
   <meta property="og:image" content="{{og_image_path | absolute_url}}">
 
-  <link href="https://fonts.googleapis.com/icon?family=Material+Icons+Extended" rel="stylesheet">
+  <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
```
filiph: Are we sure we don't need the Extended set? I _think_ we don't, but I'd rather err on the side of caution.

src/_includes/header.html
```diff
@@ -1,31 +1,29 @@
...
-              <input type="hidden" name="hl" value="en">
-              <input type="search" name="q" id="q" autocomplete="off" placeholder="Search">
-            </form>
-          </div>
-        </nav>
+    <div class="site-nav-horizontal">
+      {% include main-nav.html class="nav-item" %}
+    </div>
+
+    <div class="site-nav-vertical dropdown">
```
filiph: [nit] The current version shows the drop down on mouse-over, and it also shows `cursor: hand`.

src/_layouts/default.html
```diff
@@ -10,20 +10,20 @@
...
+        <div id="side-nav-container" class="col-md-3">
           {% include sidebar.html %}
         </div>
 
         {% if page.url == '/' %}
         {% comment %} Homepage {% endcomment %}
 
         <!-- Homepage Content Column -->
-        <div class="col-sm-12 main-contents">
+        <div class="col-md-12 main-contents">
```
filiph: I don't understand why the change from `sm` to `md` here and elsewhere. Should we always use `sm` by default and then only override for bigger screen? (I thought that's the philosophy of bootstrap's responsive design.)

chalin: Because a [new tier was added to BS 4](https://getbootstrap.com/docs/4.0/migration/#grid-system), so every tier has been bumped up one level:`sm` -> `md` -> `lg` -> `xl`.

chalin: (Thanks for the feedback. I'm heading out, so I'll respond to you other comments later.)

chalin: Just to be clear: the media breakpoints of both the current and this new version of the site are the same (e.g., 710px is one breakpoint for the home page). It is just that break point names have changed.

src/_includes/head.html
chalin: AFAIK, this is how one might use an icon from the extended collection:

- Reference the font family (`font-family: 'Material Icons Extended';`) directly, or
- Use the CSS class `.material-icons-extended`, or
- `<span class="mdi mdi-file-upload-outline"></span>`, or
- Use a direct link: https://petershaggynoble.github.io/MDI-Sandbox/?icon=file-upload-outline, or
- Use a local image of the icon, but in that case we don't need to load the font.

(If you can think of other ways, let me know.)

The site source contains no such uses.

src/_includes/header.html
chalin: The hover behavior is non-standard for menus/dropdowns, and doesn't work on mobile. I'll address the cusor in a followup PR.

chalin: > I'll address the cursor in a followup PR.

Fixed in #1220.

src/_includes/head.html
filiph: Ah, thanks for the due diligence on this.

src/_layouts/default.html
filiph: Thanks for clarifying!

PR #1217, opened 2018-09-03
Update fetch-data example to not call API in build

mogest:
It's recommended in the FutureBuilder documentation to not call an API
from the build method, so cookbook documentation should stick with best
practice.  A couple of strategies for taking API calls out of the build
method are added in this commit, and the complete example updated.


mogest:
Hi, I've signed the CLA.

Hixie:
Seems reasonable. cc @brianegan

PR #1218, opened 2018-09-03
updated ExampleWidget

vemarav:


Sfshaza:
Good catch, @vemarav. That snippet had a bad mix of tabs and spaces. I won't merge this PR to master, because of the conflicts, but I fixed it on the dash branch.

PR #1220, opened 2018-09-04
Set cursor for main nav dropdown

chalin:
Followup to #1209, specifically https://github.com/flutter/website/pull/1209#discussion_r214196038. /cc @filiph (I would have expected the anchor element to always have a pointer cursor, but apparently not.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ branches:
   only: [master, /^\d\.x$/, /^\d-dev$/, /travis-build/]
 
 cache:
-  timeout: 300
+  timeout: 1000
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -41,7 +41,7 @@ branches:
   only: [master, /^\d\.x$/, /^\d-dev$/, /travis-build/]
 
 cache:
-  timeout: 300
+  timeout: 1000
```
chalin: Cache operations were failing, e.g., see https://travis-ci.org/flutter/website/jobs/424196357.

PR #1226, opened 2018-09-04
New sidenav

Sfshaza:
I've added the new side nav to _data/sidenav.yml, which is how we implement it for www.dartlang.

@chalin


chalin:
Oops, this is over master. Closing since the commit is on the dash branch already.

PR #1242, opened 2018-09-06
[dash] Adds default and landing layouts

fertolg:
- default layout contains the sidebar, main content and table of contents
- landing layout contains just the main content, no sidebar or toc, and is centered in the page at the default max-width

### DIFF
--- a/src/_layouts/page.html
+++ b/src/_layouts/page.html
@@ -1,6 +1,0 @@
----
-layout: default
----
-
-{% comment %}TODO: drop this layout{% endcomment %}
-{{ content }}
--- a/src/index.md
+++ b/src/index.md
@@ -1,5 +1,6 @@
 ---
 title: Flutter - Beautiful native apps in record time
+layout: landing
 homepage: true
 hide_title: true
 description: >
--- a/src/_includes/sidebar.html
+++ b/src/_includes/sidebar.html
@@ -1,77 +1,87 @@
-<ul id="mysidebar">
+<ul class="nav flex-column">
 
-  <li class="sidebar-title">Get started</li>
+  <li class="nav-item">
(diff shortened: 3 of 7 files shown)


fertolg:
The build failed, but I don't think that's because of my changes. It seems jekyll is missing from the install commands or something like that: https://travis-ci.org/flutter/website/jobs/425505475#L744

I'll hold off submitting this until we can catch up tomorrow.

src/_layouts/page.html
```diff
@@ -1,6 +0,0 @@
----
```
chalin: There are still a few pages referring to `layout: page`, you might want to strip those out now.

src/index.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Flutter - Beautiful native apps in record time
+layout: landing
 homepage: true
 hide_title: true
```
chalin: - `page.hide_title` isn't used anywhere AFAICT, so I'd suggest removing it.
- `page.homepage` is used in `head.html`, but we might want to trigger on `page.layout == "landing"` instead.

chalin:
As I mentioned elsewhere, you'll need to rebase to get the build to pass (and resolve your zsh issues).

After both your and Sham's PRs are merged, I'll look into populating the sidenav and toc with actual content (though that might only happen next week).

Sfshaza:
Sorry about the conflicts to sidebar.html, @fertolg. I had to update that file to fix broken links.

src/_layouts/page.html
fertolg: I can't find any references to `layout: page` actually.

chalin: I thought that I had seen some in this PR (although I thought that I had taken them all out). If there are none left, then good!

src/index.md
fertolg: the way it's being used right now is to control the HTML title, I think that's a specific thing for the homepage, since other pages could theoretically use landing.html (like the showcase page).

PR #1244, opened 2018-09-07
[dash] A little help for zsh users, and ensure bundle install is always run

chalin:


chalin:
@fertolg - this should fix the zsh issues you've been having.
@Sfshaza - rebase from master should fix the Travis build issues you are having with your PR (and it might fix the issues you are having locally when trying to setup your environment).

PR #1245, opened 2018-09-07
Removing unused examples.

Sfshaza:
Patrice, the rest of the examples are being used, in that the code is inside the layout tutorial. I'd like to keep the code here to make sure it's valid.

The tapbox-* examples are the one exception. I wanted to add those to the interactive tutorial, but never got to it.

@chalin


chalin:
Ok, so the code is used, but not directly included, right? Is there a reason for not using the include feature? (In anycase, this code would make it a good candidate for the code-excerpt feature.)

chalin:
Thanks for taking a look and cleaning out the unused files.

PR #1246, opened 2018-09-07
i18n: Add iOS-specific instructions

cbracken:
In addition to the Dart code changes to support additional locales, iOS
apps require supported locates to be declared in their Info.plist file.


cbracken:
Fixes flutter/flutter#14128 on `master` branch. I'll send a followup for `dash`.

cbracken:
This is the `master` equivalent to #1247, which has now landed on `dash`.

chalin:
@cbracken - thanks for the clarification! (At first I thought that #1246 and #1247 were just followup PRs over master.)

PR #1253, opened 2018-09-09
[dash] Moving more files around and fixing broken links that result

Sfshaza:
@chalin


src/_data/catalog/index.json
```diff
@@ -71,7 +71,7 @@
   },
   {
     "name": "Animation and Motion",
-    "description": "Bring animations to your app. Check out <a href=\"/animations/\">Animations</a> in Flutter for an overview.",
+    "description": "Bring animations to your app. Check out <a href=\"/development/ui/animations\">Animations</a> in Flutter for an overview.",
```
chalin: Hmm, is this meant to be the page description? I don't think that we should have links in our descriptions.

chalin:
Hey Shams: Travis is still reporting some broken links.

src/_data/catalog/index.json
Sfshaza: No, this was the source of great annoyance. It took me forever to find this broken link. I don't know who put this there but I'd be fine with removing it.

Sfshaza:
Patrice and I have both signed the CLA.

PR #1258, opened 2018-09-10
[dash] Add base styles for doc pages

fertolg:


fertolg:
Travis CI doesn't show up in checks, but it did pass, here it is: https://travis-ci.org/flutter/website/jobs/426911074

chalin:
Very nice update!

Btw, the main nav used to fold into a hamburger, but it doesn't anymore. Is that temporary?

src/_includes/header.html
```diff
@@ -1,5 +1,5 @@
-<header class="site-header">
-  <nav class="navbar navbar-expand-lg navbar-light justify-content-between">
+<header class="flutter-header">
```
chalin: Could we stick with `site-foo` instead of `flutter-foo`. For main site classes like this, it would encourage reuse across sites. (While the site theme might change across sites, I'd expect, for example, `site-header` to be used everywhere.)

fertolg:
I merged this in already but just for future reference, the hamburger is still there but the button is all white unless a bootstrap nav modified is added (like navbar-light). This is fine since we need to add more custom styles to make it look like the mocks so I left it alone for now.

PR #1267, opened 2018-09-12
[dash] Refine style for navbar and sidebar

fertolg:
- added material icons
- matched font sizes and spacing with sketch files

### DIFF
--- a/src/_assets/css/_header.scss
+++ b/src/_assets/css/_header.scss
@@ -6,7 +6,41 @@
   top: 0;
   z-index: $flutter-z-header;
 
-  .nav-link {
-    color: $flutter-color-nav-links;
+  .navbar {
+    font-size: 20px;
+
+    @include media-breakpoint-up(md) {
+      font-size: 14px;
+    }
+
+    .navbar-toggler {
+      color: $flutter-color-body;
+      margin-right: 16px;
+      padding: 0;
+    }
+
+    .btn {
+      font-size: 14px;
+      padding: 10px 40px;
+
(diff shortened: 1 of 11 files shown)


chalin:
There is some strange subitem text overlap when a sidenav entry opens up, have you noticed that?

src/_assets/css/_header.scss
```diff
@@ -6,7 +6,41 @@
   top: 0;
   z-index: $flutter-z-header;
 
-  .nav-link {
-    color: $flutter-color-nav-links;
+  .navbar {
+    font-size: 20px;
```
chalin: Should we not follow Bootstrap and use `rem` units for things like font sizes?

For example, we could use [$font-size-lg](https://github.com/twbs/bootstrap/blob/6e2a823a44063af96749d54560808656e007c2a8/scss/_variables.scss#L257) here (that is at 1.25 rem).

Similar remarks apply to padding and margins; e.g., maybe using [$spacers](https://github.com/twbs/bootstrap/blob/6e2a823a44063af96749d54560808656e007c2a8/scss/_variables.scss#L131-L141) increments when appropriate.

fertolg: I changed all hardcoded values for $font-size-xx but regarding $spacers I'll need to look into that a bit more. The spacers provided by default are not really close to the increments used on multiple parts of the site.

chalin: I'd suggest giving custom spacers a try (even if we roll that back out). WDYT?

PR #1268, opened 2018-09-12
[dash] Adds barebones footer to the site

fertolg:
This footer is meant to be a starting point to have a conversation with the designers about what to do with it, since the mocks show the footer on developers.google.com but that doesn't apply to this site.


src/_includes/footer.html
```diff
@@ -21,6 +22,10 @@
             Commons Attribution 4.0 International License</a>,
             and code samples are licensed under the BSD License.
           </p>
+        </div>
+        <div class="site-footer__gdevs">
+          <img src="/images/google-developers-logo.svg" alt="Flutter Logo" width="233" height="34">
+        </div>
```
chalin: I understand that the latest mocks contains this logo, but I believe that that is likely to have been a "copy-paste" error. I'd vote to drop this logo for now. We can add it back if it _must_ be here, but in the meantime, it is only throwing off the styling on narrow displays (at least as compared to the current site's footer).

fertolg: If I had to guess I think this is the only addition they would actually make to the footer, based on other third party sites I've worked on.

fertolg: I'll take it out, and merge, we can then ask the design team to weigh in.

Issue #1259, opened 2018-09-11
[dash] Support autogenerated toc using jekyll-toc


chalin:
- [x] Setup use of `jekyll-toc`.
  - [x] Base setup and use of `jekyll-toc` - #1260
  - [x] Create path to ensure `no_toc` is supported - https://github.com/toshimaru/jekyll-toc/pull/50.
  - [x] Once the path is merged, revert to using gem (not from git).
- [x] [in progress] Strip out uses of manually inserted Kramdown TOCs: `* TOC {:toc}`. Related: #1337. Closed by #1261, #1341.
- [x] Ensure that pages that shouldn't have a TOC don't (by setting `toc: false` in the frontmatter) - #1261, ...
- [x] Ensure all pages with a TOC have:
  - Appropriate heading levels: e.g., there should be no \<h1>s, see #2126
  - [x] Headings (e.g., inside `card`s or \<aside>s) are marked `no_toc`.
- Ensure that link targets bring the page in view adjusted in a way that compensates for the main nav. See #1488
- [x] Setup use of a "scroll spy" to highlight TOC heading entry as section is scrolled to - #1271
- [x] Setup **Page Top** button in side TOC - #1271
- [x] Ensure that TOC entry links work (that the targets lead to the appropriate anchors) - #1264

Also see followup issue: TOC issues #1272.

cc @fertolg @Sfshaza @filiph

fertolg:
I'm removing myself from this ticket since I don't think this applies to me, but please let me know if I'm wrong.

chalin:
All done!

PR #1271, opened 2018-09-13
[dash] Added side- and inlined-TOC styling and scroll spy

chalin:
Contributes to #1259:

- Added a scroll spy for the side TOC.
- Added styling for TOC. Note that side vs. inline TOC styling is different.
- Page layout is adjusted for when there is no TOC.
- Added styling for page title and page GitHub buttons.
- Added a custom spacer for key `4_5` (representing 4.5), though it isn't really used yet.

Note: TOC styling isn't fully done. Some cleanup is still necessary (see comment in _toc.scss).

Staged, e.g., see:
- https://ng2-io.firebaseapp.com/docs for a page with a TOC
- https://ng2-io.firebaseapp.com/widgets for a page w/o a TOC

cc @filiph

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,22 +98,22 @@ create your own Firebase project (e.g. 'mit-flutter-staging')
 1. Tell Firebase about that project with the firebase
 [`use` command](https://firebase.googleblog.com/2016/07/deploy-to-multiple-environments-with.html):
 	```
-	$ firebase use --add
+	$ npx firebase use --add
 	? Which project do you want to add? <select the project you created>
 	? What alias do you want to use for this project? (e.g. staging) staging
 	```
 
 1. Tell Firebase that you want to deploy to staging:
 	```
-	$ firebase use staging
+	$ npx firebase use staging
 	Now using alias staging (<your project name>)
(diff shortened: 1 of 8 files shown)


fertolg:
Thanks for taking care of this! I see a few issues:

- [ ] The 'scroll to top' button wasn't in the mocks, I'm not entirely sure it's needed? It's a bit weird to put it in the TOC, normally a scroll to top is available when you reach the end of the article.
- [ ] The anchors in the TOC are not working properly as they are covered by the navbar. We will either have to set 'fake' anchors (empty divs before the actual titles) that are positioned x pixels higher than they should be so that the anchors work, or we will need to do the scroll with javascript (similar to how the scroll to top button works).
- [ ] The TOC has no hierarchy at the moment, I personally feel we should only generate links to the h2s, because otherwise it becomes really confusing (see screenshot). If we are going to keep all the headers, then we definitely need some hierarchy like the sidebar has.
uzcts8qtj78

chalin:
Thanks for the feedback! I've created #1272 so that we can track the points you've made there.

PR #1278, opened 2018-09-16
Update "Google AdWords" in showcase per rebrand

alanrussian:
"AdWords" was rebranded to "Google Ads"

https://www.blog.google/technology/ads/new-advertising-brands/


Sfshaza:
Also fixed in the Dash branch.

PR #1284, opened 2018-09-20
Update to RP2

Hixie:
TBR=@timsneath


Hixie:
cc @chalin do I have to do anything special to make this deploy?

mit-mit:
@Hixie thanks for doing this while I was sleeping!

mit-mit:
Change LGTM

chalin:
> @chalin do I have to do anything special to make this deploy?

Travis autodeploys after every successful build, so it is live now.

PR #1287, opened 2018-09-20
[dash] Fix sidebar and toc scrolling issues

fertolg:
- sidebar and toc will now be fixed instead of sticky
- added javascript code to adjust fixed columns heights while scrolling


chalin:
I have a few comments about this PR (based on our site-www experiences), but I'll get back to you a little later.

PR #1308, opened 2018-09-25
[dash] Hide TOC page-top button

chalin:
This is based on feedback from Daniel et all.

### DIFF
--- a/src/_assets/css/_site_overrides.scss
+++ b/src/_assets/css/_site_overrides.scss
@@ -1,0 +1,3 @@
+// Flutter-site specific style overrides
+
+.site-toc--button__page-top { display: none; }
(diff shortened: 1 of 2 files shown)


chalin:
@Sfshaza, @fertolg, @filiph: Btw, the Firebase docs put a Material Icon 'arrow_upward' floating to the right of every header:

screen shot 2018-09-25 at 06 55 48

src/_assets/css/_site_overrides.scss
```diff
@@ -0,0 +1,3 @@
+// Flutter-site specific style overrides
```
chalin: @fertolg - note this include is to be used for site-specific overrides (the plan is to share the rest of the Sass include files across sites, and place all site-specific overrides in this file -- or other includes, but imported here). /cc @filiph

PR #1311, opened 2018-09-25
Fixing a couple issues filed against master.

Sfshaza:
@brianegan  @mogest  (Brian, Ian thought this change was reasonable, so I proceeded.)

Please review.


README.md
```diff
@@ -190,14 +190,14 @@ it with triple backticks followed by the language.
...
-      Widget build(BuildContext context) {
-        return new Container();
-      }
-    }
-	```
+```dart
+class ExampleWidget extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return new Container();
```
chalin: We can drop `new` now.

src/cookbook/networking/fetch-data.md
```diff
@@ -145,6 +145,60 @@ FutureBuilder<Post>(
 );
 ```
 
+## 5. Moving the fetch call out of the `build` method
+
+Although it's convenient, it's not recommended to put a call to an API in your
+`build` method.
```
chalin: Per writing guidelines method names should be followed by parentheses: `build()`. (This should be applied on this line and a few below here.)

brianegan:
Hi @Sfshaza! Yep, this looks like a great change. My initial goal was to keep that section of the recipe short based on earlier feedback, but I think it's really important to provide that information. It's a common mistake.

README.md
Sfshaza: Good catch!

src/cookbook/networking/fetch-data.md
Sfshaza: Yeah, we're kinda sorta consistent about this.  :)  Fixed.

Sfshaza:
Thanks for your quick reviews, @brianegan and @chalin! Feedback incorporated.

Sfshaza:
Sorry about that, @mogest. We are in a fairly peculiar state right now, working on getting out a new version of the website. We aren't making changes to master unless they are P0. If you go to one of the flutter.io pages, and hover over the Edit source text, it will say that we aren't accepting changes at this time. You may have noticed that I put those edits into a branch that isn't master.

Ordinarily, I'd be thrilled to accept a community-provided PR, so long as it meets requirements. (It can be difficult when someone submits a PR involving many files, because I have no way to cherry pick and not all edits are valid.) This situation was complicated by the fact that @brianegan (the cookbook author) is a contractor and was on a sabbatical, delaying his review of your original PR.

I'm sorry that the timing worked out this way. Once the new website is out, I'd be happy to return to accepting community PRs. I am a lone Flutter writer, so I welcome the help.  :)

PR #1314, opened 2018-09-25
Updates ‘Getting started’ pages styles

fertolg:
- styling code and pre-formatted code blocks
- styling figures and images

### DIFF
--- a/src/_assets/css/_base.scss
+++ b/src/_assets/css/_base.scss
@@ -9,6 +9,7 @@ h1 {
 
 h2 {
   margin: bs-spacer(12) 0 bs-spacer(8);
+  clear: both;
 }
 
 h3 {
@@ -25,10 +26,6 @@ dd {
   margin-bottom: bs-spacer(2);
 }
 
-.btn {
-  font-family: $site-font-family-alt;
-}
-
 .fixed-col {
   @include media-breakpoint-up(md) {
     bottom: 0;
@@ -40,4 +37,27 @@ dd {
 
 .material-icons {
   font: $site-font-icon;
+}
+
+.site-image-right {
+  display: block;
+  margin: 0 auto;
+  max-height: 60vh;
+  max-width: 100%;
+
+  @include media-breakpoint-up(md) {
+    float: right;
+    margin: 0 0 bs-spacer(10) bs-spacer(10);
+    max-height: none;
+    max-width: calc((100%/2.5) - #{bs-spacer(10)});
+  }
+}
+
+.site-figure {
+  display: flex;
+  justify-content: center;
+
+  figcaption{
+    margin-top: bs-spacer(4);
+  }
 }
--- a/src/_assets/css/_base.scss
+++ b/src/_assets/css/_base.scss
@@ -59,5 +59,6 @@ dd {
 
   figcaption{
     margin-top: bs-spacer(4);
+    text-align: center;
   }
 }
--- a/src/_assets/css/_base.scss
+++ b/src/_assets/css/_base.scss
@@ -57,8 +57,16 @@ dd {
   display: flex;
   justify-content: center;
 
-  figcaption{
+  .site-figure-container {
+    flex: 0 1 auto;
+
+    img {
+      width: 100%;
+    }
+
+    figcaption {
     margin-top: bs-spacer(4);
     text-align: center;
+  }
   }
 }
--- a/src/_assets/css/_bootstrap_adjust.scss
+++ b/src/_assets/css/_bootstrap_adjust.scss
@@ -19,3 +19,21 @@ a.card {
     }
   }
 }
+
+.btn {
+  font-family: $site-font-family-alt;
+}
+
+code {
+  background-color: $flutter-color-light-blue;
+  padding: 0 3px;
+}
+
+pre {
+  background-color: $site-color-codeblock-bg;
+  padding: bs-spacer(7) bs-spacer(8);
+
+  code {
+    background-color: transparent;
+  }
+}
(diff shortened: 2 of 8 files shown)


src/_assets/css/_base.scss
```diff
@@ -9,6 +9,7 @@ h1 {
 
 h2 {
   margin: bs-spacer(12) 0 bs-spacer(8);
+  clear: both;
```
chalin: Hmm, I'd prefer to not apply this to h2 generally. I think that in individual cases, we can add use of the BS `clearfix` class in the markup/markdown file. WDYT?

src/get-started/_android-ios-figure-pair.md
```diff
@@ -10,15 +10,15 @@ Sadly, Jekyll doesn't seem to supported nested includes, so the following won't
...
-<figure class="figure">
-  {% asset 'get-started/ios/{{include.image}}'
-      alt='{{include.alt}} on iOS'
-      class='figure-img img-fluid' %}
-  <figcaption class="figure-caption text-center">iOS</figcaption>
+<figure class="site-figure">
+  <div>
+      {% asset 'get-started/android/{{include.image}}'
+          alt='{{include.alt}} on Android' %}
+      <figcaption class="figure-caption text-center">Android</figcaption>
```
chalin: So you've chosen to keep `text-center` here (and below)?

src/_assets/css/_base.scss
```diff
@@ -41,3 +38,26 @@ dd {
 .material-icons {
   font: $site-font-icon;
 }
+
+.site-image-right {
```
chalin: Is this a general purpose "site image" placement class, or has it been tailored to use in the get-started page?

src/_assets/css/_base.scss
```diff
@@ -41,3 +38,26 @@ dd {
...
+    max-width: calc((100%/2.5) - #{bs-spacer(10)});
+  }
+}
+
+.site-figure {
+  display: flex;
+  justify-content: center;
+
+  figcaption{
+    margin-top: bs-spacer(4);
```
chalin: Center the caption text? (If so, remove the `text-center` occurrences in the page sources.)

src/_assets/css/_bootstrap_adjust.scss
```diff
@@ -19,3 +19,21 @@ a.card {
     }
   }
 }
+
+.btn {
```
chalin: Thanks for moving this here.

src/get-started/_app-figure.md
```diff
@@ -14,9 +14,11 @@ TODO: Fernando, should we set all `figure-caption` to have centered text,
...
-<figure class="figure {{include.class}}">
-  {% asset '{{path}}/{{include.image}}' alt='{{alt}}' class='figure-img img-fluid' %}
-  {% if caption -%}
-  <figcaption class="figure-caption text-center">{{caption}}</figcaption>
-  {% endif -%}
+<figure class="site-figure {{include.class}}">
+  <div>
+      {% asset '{{path}}/{{include.image}}' alt='{{alt}}' %}
+      {% if caption -%}
+      <figcaption class="figure-caption text-center">{{caption}}</figcaption>
```
chalin: Keep/remove `text-center`?

src/get-started/codelab.md
```diff
@@ -539,7 +535,7 @@ at the following link to get back on track.
 
 * [lib/main.dart]({{site.codelab-code-url}}/startup_namer/4_end_of_infinite_list/lib/main.dart)
 
-{% include_relative _app-figure.md image="startup-namer.gif" caption="The app from part 2" class="float-right" %}
+{% include_relative _app-figure.md image="startup-namer.gif" caption="The app from part 2" %}
```
chalin: We've lost the `class="float-right"`, so the image isn't placed as it was before.

chalin:
Also note that the app-screenshot pairs aren't properly formatted at certain widths, e.g.:

screen shot 2018-09-25 at 18 40 28

src/_assets/css/_base.scss
fertolg: it's a general purpose class for any image that needs to be floated to the right of a paragraph.

src/_assets/css/_base.scss
fertolg: I added the centered style to the figcaption now.

fertolg:
I fixed the overflowing issue on my latest commit.

src/_assets/css/_base.scss
fertolg: I didn't do it for h3s and below because it's possible an image would be floating in that section. We can certainly use 'clearfix' we just have to educate the writers that whenever they use a 'site-image-right` they should add a clearfix to the next section.

chalin: > The problem is that if we do it as a class we would need to add it every time there is a floating image.

I don't expect that it will happen that often.

> I thought adding to the h2 top level would be ok since when using an h2 in the docs we are indicating a new high level section, and it's very unlikely the h2 should continue floating next to an image in the previous section.

Ok, let's try it this way for now.

chalin:
Thanks for the fixes. It is looking great!

The code styling seems a bit off, but I didn't have time to investigate further -- I'll do that tomorrow. Go ahead a merge once Travis is happy.

src/_assets/css/_bootstrap_adjust.scss
```diff
@@ -19,3 +19,21 @@ a.card {
...
+  font-family: $site-font-family-alt;
+}
+
+code {
+  background-color: $flutter-color-light-blue;
```
chalin: Why use blue? More often on other sites (like developers and www), I've see gray used, such as #f7f7f7. /cc @Sfshaza

Issue #1323, opened 2018-09-26
[dash] Sept 26 walkthrough followup


chalin:
- [x] [FG] Vertically center GitHub buttons in h1 line
- [x] [FG] Use standard (400) font weight for body font - #1329
- [x] [FG] Docs landing page: Card padding looks like too much - #1329
- [x] [FG] Sidebar width remains constant regardless of TOC being present or not - #1329

---

- Clicking a TOC entry doesn't position the page at the identified header (the main nav is hiding). FG & PC have discussed alternative solutions. See #1488
- [x] `get-started/install`: use cards for OS-specific install pages
- [x] [PC] Add support for page navigation via previous-page, next-page links. #1322
- [x] [PC] Get Started button, support show/hide per page #1325

chalin:
The last issue will be tracked via #1488. Closing this issue group since everything else is done.

PR #1329, opened 2018-09-26
[dash] Addresses part of #1323

fertolg:
- [x] Use standard (400) font weight for body font
- [x] Reduces card deck padding
- [x] Sidebar width remains constant regardless of TOC being present

Other improvements:
- Main content div fills the screen when there is not enough content (aligning the footer with the bottom of the page).
- TOC highlighting uses medium instead of the most bold available.

### DIFF
--- a/src/_assets/css/_bootstrap_config.scss
+++ b/src/_assets/css/_bootstrap_config.scss
@@ -9,13 +9,13 @@ $code-color: $site-color-body;
 // Font config
 $font-family-sans-serif: $site-font-family-base;
 $font-family-monospace: $site-font-family-monospace;
-$font-weight-base: 300;
+$font-weight-base: 400;
 $headings-font-family: $site-font-family-alt;
 $headings-font-weight: 400;
 $h1-font-size: 3rem;
 $h2-font-size: 2.25rem;
 $h3-font-size: 1.125rem;
-$dt-font-weight: 300;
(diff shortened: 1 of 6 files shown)


src/_assets/css/_bootstrap_config.scss
```diff
@@ -9,13 +9,13 @@ $code-color: $site-color-body;
...
 $font-family-monospace: $site-font-family-monospace;
-$font-weight-base: 300;
+$font-weight-base: 400;
 $headings-font-family: $site-font-family-alt;
 $headings-font-weight: 400;
 $h1-font-size: 3rem;
 $h2-font-size: 2.25rem;
 $h3-font-size: 1.125rem;
-$dt-font-weight: 300;
+$dt-font-weight: 400;
```
chalin: Are these expected to be in sync, if so set it to `$font-weight-base`?

src/_layouts/default.html
```diff
@@ -3,13 +3,12 @@
 ---
 
 {% comment %} TODO(chalin): The following should probably just be encoded in Sass. {% endcomment -%}
```
chalin: You can drop this comment (since we've discussed this already).

chalin:
I'd like to build atop this PR, so I'll merge and we can address the minor issues in a followup PR.

Issue #1338, opened 2018-09-27
[dash] Generate sidenav from YAML data file


chalin:
- [x] Handle cases where nav group has a permalink - #1339
- [ ] Handle cases where a nav group has items that don't share a common permalink prefix - #1429

PR #1339, opened 2018-09-27
[dash] Sidenav from yaml, iteration 1

chalin:
Contributes to #1338

While the entire sidenav is being generated, about 1/4 of the entries "don't work": i.e., visiting the page doesn't cause the corresponding top-level sidenav group to open. This will be fixed in the next iteration.

Styling might need to be tweaked too.

Staged, e.g., see https://ng2-io.firebaseapp.com/get-started/codelab

### DIFF
--- a/src/_assets/css/_bootstrap_config.scss
+++ b/src/_assets/css/_bootstrap_config.scss
@@ -44,7 +44,6 @@ $alert-padding-x: bs-spacer(6);
 $border-radius: 0;
 $border-radius-lg: 0;
 $border-radius-sm: 0;
-// $btn-font-weight: 400; // Default is already normal (400). TODO: drop?
 $card-border-radius: 4px;
 $card-group-margin: bs-spacer(2);
 $grid-gutter-width: 50px;
--- a/src/_assets/css/_bootstrap_config.scss
+++ b/src/_assets/css/_bootstrap_config.scss
@@ -9,13 +9,13 @@ $code-color: $site-color-body;
 // Font config
 $font-family-sans-serif: $site-font-family-base;
 $font-family-monospace: $site-font-family-monospace;
-$font-weight-base: 400;
+$font-weight-base: 400; // assert: matches Bootstrap default
 $headings-font-family: $site-font-family-alt;
-$headings-font-weight: 400;
+$headings-font-weight: $font-weight-base;
 $h1-font-size: 3rem;
 $h2-font-size: 2.25rem;
 $h3-font-size: 1.125rem;
-$dt-font-weight: 400;
+$dt-font-weight: $font-weight-base;
 
 // Spacing (extra spacers beyond Bootstrap's defaults)
 $site-spacer: 1rem; // assert: should match Bootstrap $spacer
@@ -44,7 +44,7 @@ $alert-padding-x: bs-spacer(6);
 $border-radius: 0;
 $border-radius-lg: 0;
 $border-radius-sm: 0;
-$btn-font-weight: 400;
+// $btn-font-weight: 400; // Default is already normal (400). TODO: drop?
 $card-border-radius: 4px;
 $card-group-margin: bs-spacer(2);
 $grid-gutter-width: 50px;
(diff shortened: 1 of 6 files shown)


src/_assets/css/_bootstrap_config.scss
```diff
@@ -44,7 +44,7 @@ $alert-padding-x: bs-spacer(6);
 $border-radius: 0;
 $border-radius-lg: 0;
 $border-radius-sm: 0;
-$btn-font-weight: 400;
+// $btn-font-weight: 400; // Default is already normal (400). TODO: drop?
```
fertolg: This can be dropped, I had it as 400 because the site used to default to 300

src/_assets/css/_sidebar.scss
```diff
@@ -52,22 +54,20 @@
     }
   }
 
-  // Second nav level
-  > .nav > .nav > .nav-item {
+  // Second-level nav item
+  > .nav > .nav-item > .nav > .nav-item {
     font-size: $font-size-sm;
 
     .nav-link {
```
fertolg: To be consistent, we should probably add `> ` here as well since you added it to the top level one.

src/_assets/css/_sidebar.scss
```diff
@@ -52,22 +54,20 @@
...
     }
   }
 
-  // Third nav level
-  > .nav > .nav > .nav > .nav-item {
+  // Third-level nav item
+  > .nav > .nav-item > .nav > .nav-item > .nav > .nav-item {
     font-size: $font-size-sm;
 
     .nav-link {
```
fertolg: And here as well.

src/_includes/sidebar.html
```diff
@@ -1,94 +1,120 @@
...
-  <ul class="nav flex-column flex-nowrap collapse" id="detailsMenu">
-    <li class="nav-item"><a class="nav-link" href="/resources/faq">FAQ</a></li>
-    <li class="nav-item"><a class="nav-link" href="/resources/technical-overview">Technical overview</a></li>
-    <li class="nav-item"><a class="nav-link" href="https://docs.google.com/presentation/d/1B3p0kP6NV_XMOimRV09Ms75ymIjU5gr6GGIX74Om_DE/edit?usp=sharing">Magic of Flutter slides</a></li>
-    <li class="nav-item"><a class="nav-link" href="https://docs.google.com/presentation/d/1cw7A4HbvM_Abv320rVgPVGiUP2msVs7tfGbkgdrTy0I/edit?usp=sharing">Architecture diagrams</a></li>
-    <li class="nav-item"><a class="nav-link" href="https://www.youtube.com/watch?v=dkyY9WCGMi0">Framework's layered design <i class="fa fa-video-camera" aria-hidden="true"></i></a></li>
-    <li class="nav-item"><a class="nav-link" href="https://www.youtube.com/watch?v=UUfXWzp0-DU">Flutter's rendering pipeline <i class="fa fa-video-camera" aria-hidden="true"></i></a></li>
-  </ul>
+    <li class="nav-item">
+      <a class="nav-link {{class}}" data-toggle="collapse" href="#{{id}}" role="button"
```
fertolg: Instead of `href="#{{id}}"` I would recommend using `data-target="#{{id}}"` so that the `a` tag doesn't render as a link, since it's just an interactive element, but it doesn't take the user anywhere.

chalin: If you use the `data-target` then when you hover over a top-level entry, the cursor is a `crosshair`. IMHO, that is unintuitive for something that is clickable. That was why I gave the anchor a `href`.

I can drop the `href` attribute, but then we'd need to add a change in cursor. If we keep the `href`, then we'd need to change the color. Which do you prefer? I'd favor the latter. You?

src/_assets/css/_bootstrap_config.scss
chalin: Done

src/_assets/css/_sidebar.scss
chalin: Done here a below

src/_includes/sidebar.html
chalin: @fertolg - merging now so that I can work on iteration 2 of this feature. We can adjust the styles later as needed.

PR #1349, opened 2018-10-01
[dash] Changed "fat arrow" to "arrow".

Sfshaza:
@mit-mit

### DIFF
--- a/src/get-started/codelab.md
+++ b/src/get-started/codelab.md
@@ -133,8 +133,8 @@ where the Dart code lives.
   [Material](https://material.io/guidelines/) is a visual design language
   that is standard on mobile and the web. Flutter offers a rich set
   of Material widgets.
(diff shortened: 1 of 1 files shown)


src/get-started/codelab.md
```diff
@@ -133,8 +133,8 @@ where the Dart code lives.
   [Material](https://material.io/guidelines/) is a visual design language
   that is standard on mobile and the web. Flutter offers a rich set
   of Material widgets.
-* The main method specifies fat arrow (`=>`) notation.
-  Use fat arrow notation for one-line functions or methods.
+* The main method specifies arrow (`=>`) notation.
```
chalin: - Per guidelines, it should read: "The `main()` method ..."
- Instead of "specifies", how about the simpler "uses":
  > The `main()` method <ins>uses the</ins> arrow (`=>`) notation.

PR #1350, opened 2018-10-01
[dash] Initial port of prev/next subnav, added to top & bottom on page

### LINKED ISSUE #1322: [dash] Add support for page navigation via previous-page, next-page links.
Add support for page navigation (when a collection of pages are related), via previous-page, next-page links. On site-www and -webdev we have previous/next-page links at the top and bottom of pages, something like this:

> screen shot 2018-09-26 at 15 24 24

On webdev, we also have support for (auto-)eliding of headings on narrow displays:

> screen shot 2018-09-26 at 15 25 21

chalin:
For Shams! With this initial port, you need to provide the `prev` and `next` links.

Fixes #1322

Staged, e.g., see: https://ng2-io.firebaseapp.com/get-started/install

cc @fertolg 

--- 

This is what it looks like:

screen shot 2018-10-01 at 13 35 18

Notice how long titles are shortened on narrow displays:

screen shot 2018-10-01 at 13 34 56


src/_includes/subnav.html
```diff
@@ -0,0 +1,16 @@
+{% if page.next or page.prev -%}
+<nav id="subnav">
```
fertolg: this will output two navs with the same ID which is not good practice, we may want to change this to a class?

src/_includes/subnav.html
```diff
@@ -0,0 +1,16 @@
...
+  <ul>
+    <li class="previous">
+      {% if page.prev.path -%}
```
fertolg: can we move the if statement to wrap the <li> so that it won't output anything when there's no previous or next?

src/_assets/css/_subnav.scss
```diff
@@ -0,0 +1,63 @@
+//
+// Subnav styles (from site-www)
+//
+// TODO: As a first cut, the following styles were copied from site-www. Review,
+// in particular, adjust dimensions (that are currently in px).
+//
+
+#subnav {
```
fertolg: we should add `clear: both` to the subnav so that it doesn't overlap with floating images like in step 4.

Sfshaza:
Magnifique!!!  (What's LGTM in French?)

src/_assets/css/_subnav.scss
chalin: Done

src/_includes/subnav.html
chalin: Yes, thanks! Done

src/_includes/subnav.html
chalin: I've fixed the styling issues and now I can move the `<li>` inside the if.

chalin:
> Magnifique!!! (What's LGTM in French?)

Magnifique is much more poetic! :)

src/get-started/test-drive.md
```diff
@@ -186,5 +186,3 @@ You should see the updated string in the running app almost immediately.
 ## Next step
 
 Let's learn some core Flutter concepts, by creating a small app.
-
-[Next step: Write your first app](/get-started/codelab)
```
chalin: @Sfshaza FYI, I dropped the hardcoded "Next step" links (since each Get Started page now has next/prev links at the top and bottom of the pages.

chalin:
@fertolg - I reworked the styles and removed the link icon when no next/prev link is present. Travis is happy so I'll merge now, but comments on the latest changes are still welcome. Thanks.

PR #1353, opened 2018-10-01
[dash] Update text of security policy, based on group feedback.

Sfshaza:
@chalin


Sfshaza:
(Ray, Tim, and Ian LGTM'd in email.) Ray may have further changes later on.

PR #1373, opened 2018-10-02
[dash] Progress on landing page styles

fertolg:
- Global elements style overrides for the homepage (h2s, h3s, etc)
- Adds styles for intro card, image cards, news cards and the cta card
- Adds features icons
- Adds provisional images from design reference

There is still work left to do for special types of cards, and the hero illustration, but I wanted to share this sooner rather than later so we can review in pieces.

_Note:_ images and SVGs are optimized using ImageOptim and svgo.


src/_assets/css/_bootstrap_config.scss
```diff
@@ -17,7 +17,13 @@ $h2-font-size: 2.25rem;
 $h3-font-size: 1.125rem;
 $dt-font-weight: $font-weight-base;
 
-// Spacing (extra spacers beyond Bootstrap's defaults)
+// Layout config
+$container-max-widths: (
+  sm: 540px,
+  md: 720px,
+  lg: 960px,
+  xl: 1330px
```
chalin: Why deviate from the default xl 1140px?
Maybe define xxl instead as the key for 1130px?

src/_layouts/base.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
   {% include head.html -%}
 
-  <body>
+  <body {% if page.body_class %}class="{{ page.body_class }}"{% endif %}>
```
chalin: nit: use 
```
<body {%- if page.body_class %} class="{{ page.body_class }}" {%- endif %}>
```

chalin:
Two other points:

 1. The page "title" seems squished against the main nav. I'm assuming that that should be easy to fix?

    > screen shot 2018-10-03 at 02 42 18

 2. In the "Find others" section:

    > screen shot 2018-10-03 at 02 40 57

    - Could the input placeholder be "Enter Zip or Postal Code" instead, that'll make it in the spirit of the new [USMC Trade accord ;)](https://ustr.gov/about-us/policy-offices/press-office/fact-sheets/2018/october/united-states%E2%80%93mexico%E2%80%93canada-trade-fact). If that is the case, then the input type should not be `number`.
    - Is the button label final? Should it really be "Get Started", or it is meant to be the submit button of the form -- and so should be something link "Find"?

chalin:
I'd like to add the infrastructure in support of the banner on the landing page, so I'm going to merge this now (since you've already made a pass following my first round of comments).

PR #1385, opened 2018-10-04
Fix docs on performance overlay

GaryQian:
See https://github.com/flutter/flutter/issues/22663


Sfshaza:
I also fixed this on the new website.

PR #1387, opened 2018-10-04
Progress on homepage styling

fertolg:
- Styled partners and learn from developers cards
- Added video wrapper element to display youtube embedded videos in 16:9
- updated icon colors and revised spacing per new mocks


chalin:
> The image files need to be moved, and src paths need to use `{% asset ... %}`

**Edit**: The image files need to be moved, and `` element replace by `{% asset ... %}` tags. (Again, I can do this once you are mostly done with the homepage.)

PR #1408, opened 2018-10-08
Initial style for Showcase page

fertolg:
- Moved shared landing styles to a new _landing-page.scss file
- Updated markup for showcase page
- Added hero element to landing shared styles for a full width block within the fixed width container
- Added basic custom styles for showcase page

_Note:_ This is just a first pass at the styles, but there’s already a lot going on, so I want to start the review.

### DIFF
--- a/src/_assets/css/_base.scss
+++ b/src/_assets/css/_base.scss
@@ -1,6 +1,7 @@
 body {
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+  overflow-x: hidden;
 }
 
 h1 {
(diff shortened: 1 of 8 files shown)


chalin:
Very nice!

One stylistic comment: I much prefer the less busy Doc page cards that look like this:

> screen shot 2018-10-09 at 03 08 28

As opposed to this:

> screen shot 2018-10-09 at 03 08 40

Could we try the Doc-page style cards?

chalin:
This PR impacts the Sass files, so I'll go ahead and merge so that I can build on top of these changes. Followup PRs can address any subsequent issues that arise. Hopefully that's ok with you @fertolg.

galeyang:
I believe there are more links for showcase apps, like Learn More, Download, while the entire Doc-page style cards are clickable. (I'm not sure what it looks like in the latest mockup)

Sfshaza:
I agree with Patrice that the docs pages are much more attractive.

PR #1427, opened 2018-10-10
[dash] By default collapse sidenav level 2 entries with children

### LINKED ISSUE #1398: [dash] left nav's default expand/collapse status
- [x] From UX research, we suggested to expand the "Development" and "APIs & References" because these are frequent used sections - #1426

- [ ] Also we initially suggested to collapse all sub-menu items to avoid information overload. For example, when users expand the menu "Development", they will first see "User interface, Data & backend, Accessibility & Internationalization, etc" without expanding the sub entries.

chalin:
Closes #1398

Note that sidenav level 2 entries with children don't have a caret (per the current mocks).

### DIFF
--- a/src/_data/side-nav.yml
+++ b/src/_data/side-nav.yml
@@ -45,7 +45,6 @@
   permalink: /development
   children:
   - title: User interface
-    # permalink: /development/ui
     children:
     - title: Tour the framework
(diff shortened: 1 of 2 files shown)


chalin:
I'll merge since my next PR builds atop this one. We can always address any necessary changes in a followup PR.

Issue #1428, opened 2018-10-10
External links should be visually distinctive


chalin:
Www postfixes external links with an icon. For example, see https://www.dartlang.org/guides/libraries:

> <img width="419" alt="screen shot 2018-10-10 at 09 59 22" src="https://user-images.githubusercontent.com/4140793/46741564-496d3800-cc73-11e8-9ca7-9557852c04fc.png">

cc @filiph @fertolg

galeyang:
Yes! I also noticed it but have not filed the issue. Can we open external links in a new tab (<a target="_blank") and add the same icon shown in your screenshot?

chalin:
#1429 implements the "external link" icon for sidenav entries. PTAL

> Can we open external links in a new tab

Surely. In that case I'd suggest using `target="_blank" rel="noopener"`.

Sfshaza:
Yes, I had noticed the lack of icons for external links (but hadn't filed either) and am glad to see them back!

galeyang:
From the staged site, it seemed to be solved.

chalin:
We have external-link icons for the navs only, not in-page links.

sfshaza2:
Definitely fixed in the sidenav.

PR #1429, opened 2018-10-10
[dash] Support proper active-entry status-indicator for all sidenav entries

### LINKED ISSUE #1338: [dash] Generate sidenav from YAML data file

chalin:
The previous implementation of the sidenav template relied on `permalink` paths to determine if a sidenav entry lineage was active. E.g., if the page at `/foo/bar` was active, then its parent entry at `/foo` would be active too. For the new IA, that worked only for 80% of the cases.

This PR switches to using the tree structure of the sidenav YAML file to determine which sidenav entry lineage is (which page, and which of its ancestor sidenav entries are) active, regardless of the presence (or not) of entry permalink fields.

Also:

- Rename `testing/debugging/oem.md` (back) to `testing/oem-debuggers.md`
- Contribute to #1428 (for the sidenav entries only)
- Make visually distinct sidenav entries not referring to a page (yet)

Here is a screenshot illustrating all of these features:

> screen shot 2018-10-10 at 12 08 27

Closes #1338 

cc @fertolg @filiph


galeyang:
LGTM

Issue #1423, opened 2018-10-10
[dash] invalid_use_of_protected_member in sample code


chalin:
We've started getting the following analyzer message:

```nocode
ANALYZING _includes/code/*:
Analyzing 3 directories...

   info • The member 'lerp' can only be used within instance members of subclasses of 'Tween' • ../src/_includes/code/animation/radial_hero_animation_animate_rectclip/main.dart:67:49 • invalid_use_of_protected_member

1 issue found. (ran in 9.6s)
```

The line in question is this one: https://github.com/flutter/website/blob/01ca050b6a59a8f107bb995864324b9a33e276ac/src/_includes/code/animation/radial_hero_animation_animate_rectclip/main.dart#L67

I know that this is just an analyzer "info" message, but it still points to the fact that the code is illustrating invalid use of an API. @filiph, what do you suggest as a fix?

cc @Sfshaza

chalin:
We're getting this error on master as well (FYI).

PR #1430, opened 2018-10-10
Fix build by temporarily ignoring analyzer issue when checking example code

chalin:
For details, see #1423. cc @filiph @Sfshaza


Sfshaza:
LGTM

Issue #1420, opened 2018-10-09
[dash] Add hover and selected design for the top nav item


galeyang:
Current screenshot:
![image](https://user-images.githubusercontent.com/5102864/46700393-91d41980-cbd1-11e8-843c-652d89e800a2.png)

Mockup for selected status
![image](https://user-images.githubusercontent.com/5102864/46700471-ca73f300-cbd1-11e8-9d15-e5c93540de95.png)

chalin:
See also #1435

chalin:
Actually, #1434 brings in the basic functionality so I'll mark this issue as closed. We'll track styling via #1435.

PR #1434, opened 2018-10-10
[dash] Visually distinguish active main nav entries

chalin:
Contributes to #1420. I've done rudimentary styling, @fertolg will need to make it prettier :).

Screenshot:

> screen shot 2018-10-10 at 13 47 44

### DIFF
--- a/src/_assets/css/_header.scss
+++ b/src/_assets/css/_header.scss
@@ -42,6 +42,11 @@
       @include media-breakpoint-up(md) {
         padding: 0 bs-spacer(4);
       }
+
+      &.active {
+        border-bottom: 0.25rem solid $site-color-primary;
+        // font-weight: $font-weight-bold;
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/_assets/css/_header.scss
```diff
@@ -42,6 +42,11 @@
       @include media-breakpoint-up(md) {
         padding: 0 bs-spacer(4);
       }
+
+      &.active {
+        border-bottom: 0.25rem solid $site-color-primary;
+        // font-weight: $font-weight-bold;
```
chalin: We don't make active top-level sidenav entries bold, so I don't think that we should do so for the main nav. Also, making it bold makes the entries shift and it is visually distracting.

chalin: That sounds good. I'll merge this PR (which has to code setting the class for the active entry). @fertolg will follow through with style changes.

chalin: Cf. #1435

PR #1439, opened 2018-10-10
Update Showcase page style

fertolg:
- Implements new app cards layout
- Implements static hero slide (*note:* this will be made dynamic, with multiple slides, in a future PR)
- Adds new showcase apps data


fertolg:
@chalin merging this PR based on your suggestion via ping, but feel free to review and leave comments for me to address in the future.

src/showcase/index.md
```diff
@@ -6,20 +6,24 @@ show-nav-get-started-button: true
...
+                <div class="col-md-7 offset-md-5 col-lg-5 offset-lg-6 col-xl-4 offset-xl-7">
+                    <div class="card">
+                        <div class="card-body">
+                            <h2 class="card-category">Featured</h2>
+                            <h3>Hamilton Musical</h3>
+                            <p>
+                            Official app of the hit Broadway musical, Hamilton. Includes daily lotteries, 
exclusive news and
+                            videos, a trivia game, merchandise store, and more.
+                            </p>
```
chalin: If we're using the same data as that found in the `showcases.yml`, then there is a way keep this DRY by identifying the hero/featured case in the data file and using the entry fields here. I'll address this (via #1443), but not before the demo.

src/showcase/index.md
```diff
@@ -6,20 +6,24 @@ show-nav-get-started-button: true
...
-                        <a class="btn btn-link" href="#">Learn more</a>
-                        <button class="btn btn-link btn-icon" type="button" data-toggle="modal" data-target="#videoModal"><i class="material-icons">play_circle_filled</i> Watch the video</button>
+<section class="landing-page__hero showcase__carousel">
+    <div class="showcase__carousel__slide" style="background-image: url('{% asset 'showcase/hero-hamilton.jpg' @path %}') ">
+        <div class="container">
+            <div class="row">
+                <div class="col-md-7 offset-md-5 col-lg-5 offset-lg-6 col-xl-4 offset-xl-7">
+                    <div class="card">
+                        <div class="card-body">
+                            <h2 class="card-category">Featured</h2>
```
chalin: As a hero, it seems obvious to me that this is a featured app. Having the "Featured" label breaks the hero card aesthetics IMHO. Could we drop it?

src/showcase/index.md
```diff
@@ -42,29 +46,51 @@ toc: false
...
+                                <span class="dropdown">
+                                    <a href="#" class="dropdown-toggle" id="dropdownMenuButton{{ forloop.index }}" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                                        Download
+                                    </a>
+                                    <div class="dropdown-menu" aria-labelledby="dropdownMenuButton{{ forloop.index }}">
+                                        {% if case.play_store_link %}<a class="dropdown-item" href="{{ case.play_store_link }}" target="_blank">Android</a>{% endif %}
+                                        {% if case.app_store_link %}<a class="dropdown-item" href="{{ case.app_store_link }}" target="_blank">iOS</a>{% endif %}
+                                    </div>
+                                </span>
+                            {% endif %}
```
chalin: Regarding the "Download" feature: could we ...

- Right-justify it in the card?
- Replace the "Download" text with a download icon?

Actually, another idea would be to replace the dropdown with platform specific icon(s), floated to the right. WDYT?

src/showcase/index.md
```diff
@@ -42,29 +46,51 @@ toc: false
...
+                        </div>
                         <div class="card-body">
                             <h3>{{ case.name }}</h3>
                             <p>{{ case.description }}</p>
                         </div>
                         <div class="card-footer">
-                            <a class="btn btn-link" href="{{ case.learn_more_link }}">Learn more</a>
+                            {% if case.learn_more_link %}<a href="{{ case.learn_more_link }}">Learn more</a>{% endif %}
+                            {% if case.play_store_link or case.app_store_link %}
```
chalin: nit: Btw, _generally_, you should use `-%}` to close a Liquid tag. The hyphen instructs Liquid to eat up (in this case trailing) whitespace.

src/showcase/index.md
```diff
@@ -42,29 +46,51 @@ toc: false
...
             {% capture modulo3 %}{{ forloop.index0 | modulo:3 }}{% endcapture %}
             {% if modulo3 == '0' %}
-                <div class="showcase__apps__cards text-left card-deck">
+                <div class="showcase__apps__cards text-left card-deck flex-column flex-lg-row">
             {% endif %}
                     <div class="card">
+                        <div class="card-header">
+                            {% if case.logo_src %}
+                                {% asset '{{ case.logo_src }}' class='showcase__apps__logo' %}
+                            {% endif %}
```
chalin: It looks strange to have the logo above the name, esp. when the logo is a lockup containing the same text as the name. In those cases, it might be more visually appealing to drop the text name and show a larger lockup.

src/showcase/index.md
```diff
@@ -42,29 +46,51 @@ toc: false
...
+        </div>
     </div>
 </section>
 
 <section class="landing-page__cta card text-center">
     <div class="card-body">
         <h2 class="landing-page__cta__headline">Try Flutter today.</h2>
         <p class="landing-page__cta__body">Getting started is easy.</p>
-        <a class="landing-page__cta__button btn btn-primary" href="/get-started/install">Get started</a>
+        <a class="landing-page__cta__button btn btn-primary btn-cta" href="/get-started/install">Get started</a>
```
chalin: Compare these two buttons from the page:

- screen shot 2018-10-11 at 02 36 58
- screen shot 2018-10-11 at 02 37 03

The text sizes are quite similar but the "Get started" button itself is much larger. IMHO, the "Get started" text should be proportionally larger too.

Issue #1458, opened 2018-10-11
[dash] Set a maximum width to avoid the unbearable line length in the docs


galeyang:
The line length can be unbearable when the window is maximized.

![image](https://user-images.githubusercontent.com/5102864/46837253-74d24e80-cd69-11e8-9a43-2952ecd883fa.png)

The best practice from the readability standpoint is 45-75 (or up to 85) characters. Even on a 15-inch laptop, we can sometime get up to 148 characters per line.

----
**Suggestion**

- **Option 1: set a `max-width` only to the text content. That is, code blocks and images can still make the best use of the space.**

Example: [Bootstrap](https://getbootstrap.com/docs/4.1/layout/grid/) sets a max-width only to the text/paragraph html tag.

![image](https://user-images.githubusercontent.com/5102864/46837916-d5628b00-cd6b-11e8-92cf-5691a8199d66.png)
![image](https://user-images.githubusercontent.com/5102864/46838128-bb757800-cd6c-11e8-979b-d2325434e8f1.png)

- **Option 2: set a `max-width` to the entire center panel**

 Example: [Material.io](https://material.io/design/motion/choreography.html) sets a max-width to the center container 

![image](https://user-images.githubusercontent.com/5102864/46837746-64bb6e80-cd6b-11e8-9550-4f8905003576.png)

fertolg:
This was fixed in #1459

sfshaza2:
@fertolg, there's a followup request to make the box smaller

PR #1459, opened 2018-10-11
[dash] Updates for widgets pages

fertolg:
- adds bootstrap cards for widget catalog and the widget index
- removes “flutter by google” nav logo
- reorganizes styles to keep cards more consistent across different pages (showcase, widgets, etc).
- adds a max width container to the default layout to address #1458

### DIFF
--- a/src/_assets/css/_base.scss
+++ b/src/_assets/css/_base.scss
@@ -21,10 +21,42 @@ dd {
   margin-bottom: .75rem;
 }
 
-.card-title {
-  font-family: $site-font-family-alt;
-  font-size: $font-size-lg;
-  margin-bottom: bs-spacer(2);
+.card-deck.card-deck--responsive {
+  .card {
+    flex: 0 0 calc((100% / 3) - #{bs-spacer(4)});
+    margin-bottom: bs-spacer(4);
+  }
+}
+
+.card-footer.card-footer--transparent {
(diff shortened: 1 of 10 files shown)


src/development/ui/widgets.md
```diff
@@ -9,10 +9,21 @@ platform, and interactive widgets.
...
+<div>
+{% for section in site.data.catalog.index %}
+    {% capture modulo3 %}{{ forloop.index0 | modulo:3 }}{% endcapture %}
+    {% if modulo3 == '0' %}
+        <div class="card-deck">
+    {% endif %}
+            <a class="card" href="/development/ui/widgets/{{section.id}}">
+                <div class="card-body">
+                    <header class="card-title">{{section.name}}</header>
+                    <p class="card-text">{{section.description | escape}}</p>
```
fertolg: If I don't escape here, one of the descriptions contains html (a link), which should be fine, but for some reason when it outputs it doesn't seem to be outputting correctly and it totally messes up the second card.

Sfshaza:
@fertolg, this seems to be failing because it can't find the widgets. They are currently in /api-and-reference/widgets.

fertolg:
Thanks @Sfshaza ! Fixed the broken links and the build is now green. I'll submit this PR so that I can move along with other things.

PR #1502, opened 2018-10-18
[dash] fix deep linking to URL with fragment - compensate for top nav

### LINKED ISSUE #1488: [dash] TOC entry doesn't position the page at the identified header
From #1323: Clicking a TOC entry doesn't position the page at the identified header (the main nav is hiding).

chalin:
Following @fertolg's suggestion, this fixes #1488 

Staged, e.g., see:

- https://ng2-io.firebaseapp.com/get-started/codelab#step-1-create-the-starter-flutter-app
- https://ng2-io.firebaseapp.com/development/tools/ide/vs-code#running-and-debugging

### DIFF
--- a/src/_assets/css/_content.scss
+++ b/src/_assets/css/_content.scss
@@ -1,3 +1,10 @@
+@mixin header-clearance {
+  $clearance: 100px;
+
+  margin-top: -$clearance;
+  padding-top: $clearance;
+}
+
 .site-content {
   padding-bottom: $site-content-top-padding;
   padding-top: $site-content-top-padding;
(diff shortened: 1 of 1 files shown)


src/_assets/css/_content.scss
```diff
@@ -1,3 +1,10 @@
+@mixin header-clearance {
```
fertolg: I wonder if this mixin should be at a higher level file, because we may use this in the homepage or showcases page as well, which aren't content pages.

chalin: I suppose that we can move it when the need arises (though `_landing-page` comes after `_content` in `main.scss` so that would still work.

chalin:
I'll address the fact that some TOC links aren't working separately, but thanks for pointing out that specific case!

PR #1516, opened 2018-10-19
Adds carousel infrastructure

fertolg:
- Update homepage and showcases hero to a carousel
- Customize bootstrap carousel to have no ‘slide’ animation
- Add support for custom animations by adding a ‘transition-out’ class to the outgoing slide.

*Note:* Custom animation for both heroes is not implemented in this PR, only the infrastructure to support it.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -15,9 +15,23 @@ description: >
   {% include banner.html -%}
 </div>
 
-<section class="homepage__hero text-center">
-    {% asset 'homepage/hero-beautiful_apps.png' class='homepage__hero__image' %}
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -15,9 +15,23 @@
   {% include banner.html -%}
 </div>
 
-<section class="homepage__hero text-center">
-    {% asset 'homepage/hero-beautiful_apps.png' class='homepage__hero__image' %}
-    <h1 class="homepage__hero__text">Build <span class="color-white">Beautiful</span> Apps</h1>
+<section class="homepage__hero text-center carousel slide" data-ride="carousel" data-interval=3000>
```
chalin: I think that it is generally preferable to put all argument values in quotes.

PR #1537, opened 2018-10-24
Update scrolling.md - add "await" for driver.close

AlexTaran:
driver.close() returns future, so it should be "awaited" during teardown.


AlexTaran:
I signed it!

PR #1538, opened 2018-10-24
Update scrolling.md - add "await" for driver.close

AlexTaran:
driver.close() returns future, so it should be "awaited" during teardown.

### DIFF
--- a/src/cookbook/testing/integration/scrolling.md
+++ b/src/cookbook/testing/integration/scrolling.md
@@ -164,7 +164,7 @@ void main() {
     // Close the connection to the driver after the tests have completed
     tearDownAll(() async {
       if (driver != null) {
-        driver.close();
+        await driver.close();
       }
     });
 


chalin:
@brianegan - I see both in the docs (i.e., a call to `driver.close()` with and w/o an awaiting. Which is the recommended best practice? /cc @sfshaza2

chalin:
@kevmoo @Hixie @gspencergoog - can one of you review this. Note that we aren't consistent in the docs. We should agree on a "best practice / idiom" and apply it to all examples.

kevmoo:
In general, you should `await` futures, unless the API claims otherwise. This has been the cause of a bunch of hard-to-debug hangs/crashes where we don't know why the app is not terminating – and there is nothing blocking.

Hixie:
LGTM.

brianegan:
Thanks for helping out here @chalin & sorry about the radio silence -- busy travel schedule the past few  days!

chalin:
np

PR #1572, opened 2018-10-29
Update README.md

jonasbp:


sfshaza2:
LGTM. Thanks, @jonasbp!

README.md
```diff
@@ -1,4 +1,4 @@
-# [Flutter][]'s website
+# [Flutter][]'s website <img src="https://flutter.io/images/flutter-mark-square-100.png" width="5%">
```
chalin: @sfshaza2 - I don't think that we need this kind of branding on the repo. The main problem with this addition is that it is going to break once we move dash into master. Neither of www or webdev make use of such images. I'd suggest reverting this change.

cc @filiph - any opinion about this?

PR #1578, opened 2018-10-29
Making widget card image clickable

### LINKED ISSUE #1485: Widget catalog pages should have better hyperlinks
From URL: https://flutter.io/widgets/basics/

All of the widget catalog pages list widgets with the following structure:

> [Image]
> *Widget Name*
> *Description*
> Documentation

Only the "Documentation" link navigates anywhere.  The image and widget name are not hyperlinks despite being bigger and more intuitive click targets.  Meanwhile all the "Documentation" links are not great for accessibility since they all have the same name.

At a minimum, the image should be a hyperlink to the documentation page.  I also think that we should remove the "Documentation" links and make the widget names hyperlinks instead.

fertolg:
Fixes #1485

### DIFF
--- a/src/_includes/catalogpage.html
+++ b/src/_includes/catalogpage.html
@@ -21,9 +21,11 @@
         {% for comp in site.data.catalog.widgets %}
             {% if comp.categories contains include.category %}
                 <div class="card">
-                    <div class="card-image-holder">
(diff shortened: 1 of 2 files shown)


chalin:
I haven't tested the changes, but it LGTM.

PR #1579, opened 2018-10-29
Incorporating Andrew's feedback on samples submenu.

sfshaza2:
Please review @RedBrogdon.

### DIFF
--- a/src/_data/sidenav.yml
+++ b/src/_data/sidenav.yml
@@ -28,16 +28,12 @@
 
 - title: Samples & codelabs
   children:
+    - title: Sample apps on GitHub
+      permalink: https://github.com/flutter/samples/blob/master/INDEX.md
     - title: Cookbook
       permalink: /docs/cookbook
-    - title: UI samples catalog
-      permalink: /docs/catalog/samples
     - title: Codelabs
       permalink: /docs/codelabs
-    - title: GitHub samples
-      permalink: https://github.com/flutter/samples/blob/master/INDEX.md
-    - title: Flutter gallery - demo app
-      permalink: https://github.com/flutter/flutter/tree/master/examples/flutter_gallery#flutter-gallery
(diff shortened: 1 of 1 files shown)


src/_data/sidenav.yml
```diff
@@ -28,16 +28,12 @@
 
 - title: Samples & codelabs
   children:
+    - title: Sample apps on GitHub
+      permalink: https://github.com/flutter/samples/blob/master/INDEX.md
     - title: Cookbook
       permalink: /docs/cookbook
-    - title: UI samples catalog
-      permalink: /docs/catalog/samples
```
chalin: So now we won't have any sidenav links into the "Sample App Catalog". Is this really what we want?

It seem like there are lots of pages under `/docs/catalog/samples`, but I'm unsure that they are all being references (from anywhere else in the site):

> screen shot 2018-10-29 at 14 34 22

sfshaza2: It's what Andrew wants.

sfshaza2: @RedBrogdon, do you plan to incorporate those old examples into your GitHub examples? Can we delete those examples?

RedBrogdon: It's possible the folks on the engineering team might feel differently since they're the ones that originally wrote those samples, but I believe we'd be better served by either creating similar content in the cookbook or creating a "basic Flutter UI components" sample app in flutter/samples that covered the same Widgets.

It's not that the content they contain is incorrect or anything -- more that they're off DevRel's radar in their current form and location, and wouldn't be getting the same maintenance from us or exposure in our get-the-word-out efforts.

sfshaza2:
Thanks, @RedBrogdon. I vote for leaving these here for now. But I would like you to consider adding similar functionality in our samples repo.

PR #1593, opened 2018-10-29
Constrain max width to 960px on docs content pages

### LINKED ISSUE #1458: [dash] Set a maximum width to avoid the unbearable line length in the docs

fertolg:
Fixes #1458

### DIFF
--- a/src/_assets/css/_content.scss
+++ b/src/_assets/css/_content.scss
@@ -17,6 +17,10 @@
   padding-bottom: $site-content-top-padding;
   padding-top: $site-content-top-padding;
 
+  .container {
+    max-width: $site-content-max-width;
+  }
+
   @include media-breakpoint-up(md) {
     border-left: 1px solid $site-color-light-grey;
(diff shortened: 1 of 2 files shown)


fertolg:
Merging this since I am building on top of it but let me know if you have comments and I'll address in a follow up PR.

Issue #1462, opened 2018-10-12
[dash] Windows displays unicode missing characters


fertolg:
From a stakeholder review:

> seeing a number of these Unicode placeholders (NB These are not visible on Mac)

![lsxyttdbrbe](https://user-images.githubusercontent.com/2965170/46897802-d530c100-ce39-11e8-9ce9-b572a038e224.png)

sfshaza2:
Duplicate of bug filed by @Hixie.

fertolg:
Fixed in #1594

PR #1594, opened 2018-10-29
Removes hidden character on first news item

### LINKED ISSUE #1585: [dash] - LSUP before "Pixel" in "Flutter Release Preview 2:  Pixel-Perfect on iOS" on home page
Linux-only, apparently. Probably the other platforms just hide that character.

fertolg:
Fixes #1585, #1462


fertolg:
Merging this since I am building on top of it but let me know if you have comments and I'll address in a follow up PR.

PR #1600, opened 2018-10-29
Remove content that moved to the style guide

Hixie:
I moved a bunch of stuff to the style guide.

cc @sfshaza2


sfshaza2:
Where is our style guide? Shouldn't it be on flutter.io?

Hixie:
It's contributor-facing, rather than developer-facing, so it's on the wiki:
https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo

I plan to slowly move most of this doc into the wiki, this is just the first step.

sfshaza2:
I see. Thanks.

PR #1604, opened 2018-10-30
Moving info on China to our site.

sfshaza2:
Moving the developer info about China from the wiki to flutter.io.


src/community/china.md
```diff
@@ -0,0 +1,78 @@
+---
+title: Using Flutter in China
+description: Where to a version of flutter.io localized to Mandarin.
```
chalin: This description needs copy editing.

sfshaza2: Ee gads.

src/community/china.md
```diff
@@ -0,0 +1,78 @@
...
+toc: false
+---
+
+The Flutter community has made a Mandarin version of the flutter.io
+website available at [https://flutter-io.cn](https://flutter-io.cn.).
```
chalin: - [ ] There is an **error in the link**: note the extraneous period at the end of the URL.
- [ ] IMHO we should drop the `https://` from the link text.

sfshaza2: I do now.  ;)

sfshaza2:
Approved offline. Gale filed a bug against the broken GDG link (which I deleted): https://github.com/flutter/flutter/issues/23754

PR #1605, opened 2018-10-30
Adding a contributing page.

sfshaza2:
@Hixie, because of all the repo restructuring, this was moved to a new PR.

All feedback is incorporated.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,47 @@
+
+If you would like to contribute to the Flutter project,
+we’re happy to have your help! Anyone can contribute, whether you’re new to
+the project or you’ve been around a long time, and whether you self-identify
+as a developer, an end user, or someone who just can’t stand seeing typos.
+
+For information on contributing code to the Flutter SDK, see
+[Contributing to Flutter](https://github.com/flutter/flutter/blob/master/CONTRIBUTING.md)
+in the [Flutter SDK](https://github.com/flutter/flutter) repo.
+
+We are also happy to accept your PRs to the
+[Flutter website](https://github.com/flutter/website/) repo.
+
+For ways to get involved in the Flutter community or to learn about us,
+visit the [Flutter community](/community) page.
+
+# Ways to contribute
+
+We encourage you to reach out and join the conversation.
+
+An easy way to send feedback is to "thumbs up" issues important to you
+in either the issue tracker for the [Flutter SDK and API docs][issues],
+or the [flutter.io website][doc-issues].
+
+Other ways you can contribute:
+
+* [Ask HOWTO questions that can be answered with specific solutions][so]
+* [Live chat with Flutter engineers and users][gitter]
+* [Discuss Flutter, best practices, app design, and more on our
+   mailing list][mailinglist]
+* [Report bugs and request features][issues]
+* [Report API docs bugs][issues]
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,47 @@
+
```
Hixie: Looks like this lost the header, does that mean it's not an actual page on the Web site?

CONTRIBUTING.md
```diff
@@ -0,0 +1,47 @@
...
+in the [Flutter SDK](https://github.com/flutter/flutter) repo.
+
+We are also happy to accept your PRs to the
+[Flutter website](https://github.com/flutter/website/) repo.
+
+For ways to get involved in the Flutter community or to learn about us,
+visit the [Flutter community](/community) page.
+
+# Ways to contribute
+
```
Hixie: the "Ways to contribute" section doesn't include the ways that are mentioned in the introduction, such as fixing typos or sending PRs.

Hixie:
Can you clarify the role of this page relative to the /community page and the CONTRIBUTING.md page on the flutter repo?

CONTRIBUTING.md
sfshaza2: Correct. I moved it up one level.

CONTRIBUTING.md
sfshaza2: Addressed.

sfshaza2:
The CONTRIBUTING page is part of the repo, so doesn't need front matter.

The community page is being designed by an web designer. It will have pretty icons that link to different communities and sites. It will link to this page in the github repo.  There is a top-level link in the website banner that links to the community page.

sfshaza2:
Please take another look, @Hixie

CONTRIBUTING.md
Hixie: if it's not on the web site, who's the target audience? It's probably fine to just point to the flutter/flutter CONTRIBUTING.md doc, at that point. (I'm rewriting that one now, FWIW.)

Hixie:
I think if the community page is the web site page then we're better off just having the community page point to the contributing doc on the main github repo.

chalin:
FYI, we have distinct `CONTRIBUTING.md` files for AngularDart vs. the AngularDart website because the contributing guidelines are different:

- https://github.com/dart-lang/angular/blob/master/CONTRIBUTING.md
- https://github.com/dart-lang/site-webdev/blob/master/CONTRIBUTING.md

I believe that we should maintain a similar kind of separation.

kwalrath:
Also, there's special support in GitHub for the top-level CONTRIBUTING.md file. Every repo should have one, even if it mostly just points to another page.

sfshaza2:
I also feel we should have a CONTRIBUTING page for the website repo. We actually receive many PRs from the community, and the website is quite separate from the SDK. 

Unless you have problems with this page, @Hixie, can I merge it? I'd like to have it in place for the website launch, which is very soon.  Th

sfshaza2:
OK, we lack consensus on the need for a contributing page, but the folks I have spoken to feel that every website needs a contributing page.

@Hixie, I am merging this for now, but we can open a bug, if you like, to review this decision.

Robe21:
> @Hixie, because of all the repo restructuring, this was moved to a new PR.
> 
> All feedback is incorporated.

PR #1606, opened 2018-10-30
Open search results links in same tab.

### LINKED ISSUE #1483: Search results should not open in a new tab
From URL: https://flutter.io/search/

If doing a search from https://flutter.io/, clicking on each search search result opens that result in a new browser tab.  I think that this is very annoying, particularly because it is not expected.  This is not how https://docs.flutter.io/ behaves nor how Google searches normally behave.

If a user wants to open a new tab, they can choose to do that themselves via a number of means; we should not forcing that choice on them.

fertolg:
Fixes #1483

### DIFF
--- a/src/search.html
+++ b/src/search.html
@@ -16,6 +16,6 @@ title: Search
         })();
     </script>
     <script type="text/javascript">
-        document.write("<gcse:search></gcse:search>");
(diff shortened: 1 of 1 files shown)


sfshaza2:
LGTM

PR #1613, opened 2018-10-31
[dash] Add homepage hero carousel animation

fertolg:
_Note:_ Reusing the same assets for all 3 slides temporarily


sfshaza2:
LGTM

PR #1614, opened 2018-10-31
Adding a page on state management and cleaning up the sidenav.

sfshaza2:
@filiph, can you review this very short state management page?

@chalin, I also cleaned up the side nav.

### DIFF
--- a/src/_data/sidenav.yml
+++ b/src/_data/sidenav.yml
@@ -36,7 +36,6 @@
       permalink: /docs/codelabs
 
 - title: Development
-  permalink: /docs/development
   children:
     - title: User interface
       permalink: /docs/development/ui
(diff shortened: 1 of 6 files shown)


filiph:
LGTM. When this lands, I might have a quick follow up PR.

src/_data/sidenav.yml
```diff
@@ -36,7 +36,6 @@
       permalink: /docs/codelabs
 
 - title: Development
-  permalink: /docs/development
```
chalin: This shouldn't have been removed. It will break breadcrumbs, among other things.

src/_data/sidenav.yml
```diff
@@ -51,40 +50,41 @@
           permalink: /docs/development/ui/assets-and-images
         - title: Navigation & routing
           permalink: /docs/development/ui/navigation
-        # Commenting out the following entry per
-        # https://github.com/flutter/website/issues/1496#issuecomment-431455211
         # - title: Inspect an app's UI
         #   permalink: /docs/development/ui/inspector
         - title: Animations
-          permalink: /docs/development/ui/animations
```
chalin: Ditto: this shouldn't have been removed since it will break breadcrumbs, among other things -- and it likely the cause of #1617.

src/_data/sidenav.yml
```diff
@@ -134,7 +134,6 @@
           permalink: /docs/development/tools/inspector
 
 - title: Testing & optimization
-  permalink: /docs/testing
```
chalin: Ditto: should not be removed.

src/docs/development/state-mgmt.md
```diff
@@ -0,0 +1,37 @@
+---
+title: State management
+description: How to structure an app to manage the state of the data flowing through it.
+---
```
chalin: This file was added under `src/docs/development/state-mgmt.md`, but the URL path used in the sidenav is `permalink: /docs/development/data-and-backend/state-mgmt` (see https://github.com/flutter/website/pull/1614/files#diff-17a8fae72c56a55f7908c522f5afc95aR82), so this file will need to be moved.

chalin:
As was mentioned elsewhere, this PR never actually made it into the dash branch (or it might have made it but was overwritten). The state management page is being (re-)added via #1622.

chalin:
The nonbreaking changes of this PR are now completely covered by #1622, #1623, and #1624. (The only things not ported, are some of the sidenav.yml edits that: removed needed permalinks, and added/removed comments.)

PR #1616, opened 2018-10-31
[dash] Resizing some overly large pics.

sfshaza2:
@chalin, I've resized the pics that I could. I went through every page of the cookbook, but I was only able to re-size the static GIF files. There are a number of animated GIF files that were over 1000 pixels wide that I was unable to re-size. I'm filing a bug separately for those.

(This occurred because the old website honored setting a width on an image, but the new site ignores that attribute.)


src/docs/cookbook/design/themes.md
```diff
@@ -1,5 +1,7 @@
 ---
 title: Using Themes to share colors and font styles
+short-title: Themes
+description: how to share colors and font styles throughout an app using Themes.
```
chalin: how -> How

PR #1622, opened 2018-10-31
[dash] Add state management page

chalin:
This PR contains:

- The **State management** page of #1614, and only that page. (Note that #1614 never actually made it onto the dash branch.)
- Moves the page into the `src/docs/development/data-and-backend` folder so as to match the path used in the sidenav.
- Fixes links

cc @sfshaza2 @filiph


chalin:
Travis is green and the page content was approved by @filiph in #1614, so I'll merge this now.

PR #1623, opened 2018-10-31
[dash] Docs page copyedits from Shams

chalin:
This is a change that was a part of #1614, that is being resubmitted. cc @sfshaza2


chalin:
This was approved as a part of #1614, and Travis is green so I'll merge now.

PR #1624, opened 2018-10-31
[dash] Animations: sidenav entry cleanup and page misc edits by Shams

chalin:
Changes in this PR were originally a part of #1614, which didn't make it into dash. cc @sfshaza2


chalin:
This was approved as a part of #1614, and Travis is green so I'll merge now.

PR #1639, opened 2018-11-01
Augmenting the state management page.

sfshaza2:
Filip already LGTM'd these edits in another PR.

### DIFF
--- a/src/docs/development/data-and-backend/state-mgmt.md
+++ b/src/docs/development/data-and-backend/state-mgmt.md
@@ -10,12 +10,11 @@ Until we provide more docs/examples, learn more at the
 following links, some of which have been
 contributed by the Flutter community:
 
-## State management
+## setState
 
 * [Build reactive mobile apps with Flutter](https://www.youtube.com/watch?v=RS36gBEp8OI&feature=youtu.be), a video from Google I/O 2018 [Accompanying article on Medium]({{site.flutter-medium}}/build-reactive-mobile-apps-in-flutter-companion-article-13950959e381)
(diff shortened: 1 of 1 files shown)


src/docs/development/data-and-backend/state-mgmt.md
```diff
@@ -6,32 +6,39 @@ description: How to structure an app to manage the state of the data flowing thr
...
 Flutter app?" is a question we often hear.
 This is an complex topic with strong, and differing, opinions.
-Until we provide more docs/examples, find
-information at the following links, some of these are
+Until we provide more docs/examples, learn more at the
+following links, some of which have been
 contributed by the Flutter community:
 
-## State management
+## setState
```
chalin: Capitalization?

sfshaza2:
No, it's a method name and we don't use codefonts in headers.

src/docs/development/data-and-backend/state-mgmt.md
chalin: > No, it's a method name and we don't use codefonts in headers.

Right, but it should at least be something like:

> ## Using setState()

From the [guide](https://developers.google.com/style/code-in-text?hl=th#method-names): "Put an empty pair of parentheses after a method name to indicate that it's a method."

sfshaza2:
But that loses parallelism. We don't say "Using BLoC" or "Using scoped
model"

src/docs/development/data-and-backend/state-mgmt.md
chalin: > But that loses parallelism ...

Right, but we might not have a choice. IMHO, a heading of just "setState" isn't proper. How about:

> ## The setState() approach

@kwalrath, any opinion about this?

Issue #1371, opened 2018-10-02
Add breadcrumbs to cookbook pages


chalin:

chalin:
Breadcrumbs are implemented except for the cookbook categories (i.e., right now it looks as if the cookbook is flat, with all recipes at the same level), e.g.:

<img width="545" alt="screen shot 2019-01-10 at 16 39 56" src="https://user-images.githubusercontent.com/4140793/50998749-6b868880-14f6-11e9-9d11-657435ede9ff.png">

For the example above, the breadcrumbs should be: Docs > Cookbook > List > Basic List.

Issue #1370, opened 2018-10-02
Add next/prev links to cookbook pages


chalin:
@Sfshaza is the order of the entries in the Cookbook significant? If they were presented in alphabetical order would that be ok? /cc @filiph

Sfshaza:
Are you talking about alphabetizing the sections themselves, and then the entries within the sections? That might not be optimal.

@brianegan, what do you think?

chalin:
> Are you talking about alphabetizing the sections themselves, and then the entries within the sections? 

Right: I'm asking if it would be ok to alphabetically sort:

- Groups (like "Design basics", "Images") based on the group title
- Group items within a group, based on the item titles

brianegan:
Hey there -- sorry, somehow missed this notification!

  > That might not be optimal.

Agreed.

In general, I try to keep each recipe independent of the others, but that doesn't work for all topics, such as Forms and Testing. In those cases, I present the recipes in a specific order. 

I think it would be better for users if the articles could appear in such a curated order, rather than in a forced Alphabetical order.

brianegan:
Hrm, it's a good question. 

I've generally tried to order those from "Basics" to "More Advanced," e.g. "Design Basics" comes before "Testing" -- and I think that makes good sense :) But as you said -- lists vs Animations? That ordering is quite subjective.

chalin:
> e.g. "Design Basics" comes before "Testing" 

That respects an alphabetical ordering :). Btw, here are the topics in alphabetical order:

- Animation
- Design
- Forms
- Gestures
- Images
- Lists
- Maintenance
- Navigation
- Networking
- Persistence
- Testing

> But as you said -- lists vs Animations? That ordering is quite subjective.

There are a lot of recipes (and more are likely to come) so I'm exploring schemes for the autogeneration of next/prev links between recipes -- I'd like to avoid having to manually setup (and then maintain) such next/prev links.

Let's see if we can find a win-win solution concerning the ordering of recipe groups (we'll address intra-group recipe ordering afterwards).

The cookbook's opening paragraph reads (emphasis mine):

> This cookbook contains recipes that demonstrate how to solve common problems while writing Flutter apps. **Each recipe is self-contained** and can be used as a reference to help you build up an app.

[...]

PR #1646, opened 2018-11-01
[dash] Cookbook: a first cut at breadcrumbs; base for next/prev links

chalin:
Note that with this PR, the cookbook recipe list is autogenerated from the list of actual cookbook pages.

Also:
- Preliminary implementation of breadcrumbs (#1371). Group names (like "Testing") still don't appear in the breadcrumb list.
- Base work for next/prev links among cookbook recipes (#1370)

Staged at https://ng2-io.firebaseapp.com/docs/cookbook


sfshaza2:
LGTM, once the broken anchors are fixed.

chalin:
Travis is green, merging.

Issue #1644, opened 2018-11-01
[dash] Make the left nav feel less deep


InMatrix:
We heard some feedback that the hierarchy in the left nav feels deep. I'd like to suggest a few ideas to mitigate this perception:

1. We could let specific second-level sections to be auto-expand upon opening their parent section to make better use of the desktop screen real estate. For example, `From another platform?` can be auto-expanded when the user clicks `Get started`. I'd also suggest auto-expanding `User interface` when `Development` is opened, since we expect the UI docs to be the most frequently used section on the site. 
2. We could consider making the the `Development` section non-collapsable and effectively elevating the second-level sections under `Development` to the top-level. I believe this is actually suggested in the original IA proposal. 
3. Alternatively, we could elevate the `User interface` section, since this section has most content and the deepest hierarchy. We can have a `UI Development` top-level section followed by a `Beyond UI` section for everything else in the current `Development` section.

sfshaza2:
Before Monday, we can do the following:

- Open "From another platform" by default, but still let it be collapsible.
- Remove the IDE level under Tools.
- I'd prefer that we not make Development auto-expanded (it's kind of long), but @chalin says he can turn it on today to see how it goes.
- Re: elevating the UI section. This is a major change to the IA and kind of late in the game. I don't want to introduce this before launch.

chalin:
@galeyang, that is an interesting alternative. In the meantime, please take a look at the staged solution of #1651, e.g., at https://ng2-io.firebaseapp.com/docs.

chalin:
Note that opting for Gale's latest proposal effectively means promoting all current 2nd-level entries under Development to the 1st level. Given that the font is bigger for 1st-level entries, that is going to take up a lot more vertical space. cc @fertolg

galeyang:
> (cited myself) If we like this option, it would be great to have designer's help on the visual design.

> (@chalin) Given that the font is bigger for 1st-level entries, that is going to take up a lot more vertical space.

Yeah, in that proposal we'll need to revisit the visual hierarchy...otherwise it could be awkward.

InMatrix:
The updated left nav looks good to me. @Sfshaza Can you get folks who felt the left nav was too deep to take another look? Thanks!

chalin:
If we are ok with **Development** being open by default, I think that we should cater to @galeyang's early request of having **Reference** be open by default too. Thoughts?

chalin:
> expand Reference by default, too

[...]

PR #1651, opened 2018-11-01
[dash] Have some sidenav entries open by default

chalin:
The following entries are expanded by default:

- **From another platform**?
- **Development**
- ** IDEs**

**Staged**, e.g., see https://ng2-io.firebaseapp.com/docs

Contributes to #1644.

@InMatrix @galeyang @sfshaza2 - PTAL


sfshaza2:
LGTM!

PR #1668, opened 2018-11-02
Sort state management links roughly by complexity

filiph:
Also adds a “General overview” section.


sfshaza2:
Thanks for cleaning it up, @filiph! It looks great! LGTM

PR #1670, opened 2018-11-02
Update _get-sdk-win.md

timsneath:
A few minor tweaks to the Windows instructions. Make the language more Windows user-centric; update the environment variable process to be more resilient to Windows 10 Control Panel / Settings changes; remove unnecessary mention of Dart SDK.

### DIFF
--- a/src/docs/get-started/install/_get-sdk-win.md
+++ b/src/docs/get-started/install/_get-sdk-win.md
@@ -9,7 +9,7 @@
     archive](/docs/development/tools/sdk/archive) page.
  1. Extract the zip file and place the contained `flutter` in the desired
     installation location for the Flutter SDK (eg. `C:\src\flutter`; do not
-    install flutter in a directory like `C:\Program Files\` that requires
+    install Flutter in a directory like `C:\Program Files\` that requires
     elevated privileges).
  1. Locate the file `flutter_console.bat` inside the `flutter` directory. Start
     it by double-clicking.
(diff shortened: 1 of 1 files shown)


sfshaza2:
LGTM

src/docs/get-started/install/_get-sdk-win.md
```diff
@@ -20,32 +20,33 @@ To update an existing version of Flutter, see [Upgrading Flutter](/docs/developm
 
 ### Update your path
 
-If you wish to run Flutter commands in the regular Windows command prompt, take
+If you wish to run Flutter commands in the regular Windows console, take
```
jamesderlin: I don't have a Windows 10 box handy.  Did they rename "Command Prompt" to "Console" in the Start Menu?  If so, ok.  If not, then I think you should leave this alone.

src/docs/get-started/install/_get-sdk-win.md
```diff
@@ -20,32 +20,33 @@ To update an existing version of Flutter, see [Upgrading Flutter](/docs/developm
...
+* From the Start search bar, type 'env' and select **Edit environment
+  variables for your account**
 * Under **User variables** check if there is an entry called **Path**:
   * If the entry does exist, append the full path to `flutter\bin` using `;` as
     a separator from existing values.
   * If the entry does not exist, create a new user variable named `Path` with
     the full path to `flutter\bin` as its value.
 
-Reboot Windows to fully apply this change.
+Note that you will have to close and reopen any existing command prompt windows
```
jamesderlin: Note that you use "command prompt windows" here but "console" above and below.  Whatever terminology we choose, we should be consistent.

sfshaza2: @jamesderlin, Tim tested this on a Windows box, so I figured the terminology was right. I agree with your point about consistency, though.

timsneath:
Good comments, @jamesderlin - thanks! My general goal was to eradicate Unix-y terms like terminal and shell, in favor of Windows terms (console and command prompt). Both are somewhat interchangeable amongst Windows developers, though I agree that we can further improve the consistency in this page. 

Referencing the Microsoft style guide, I'd recommend we generally gravitate towards the following:
- "[at a command prompt](https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/c/command-prompt)" for referencing a command you type in at the CLI;
- "[console window](https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/c/console)" to reference the application that contains a command prompt. 

In practice, the number of places where the distinction is critical is fairly limited -- although this page is one of them :)

src/docs/get-started/install/_get-sdk-win.md
timsneath: See comments below. You're right -- `console windows` would be better here.

src/docs/get-started/install/_get-sdk-win.md
timsneath: See comments below. In this case, `console` is more accurate because there are many Start menu icons that could lead to the Windows console. And more obvious since the alternative is the `Flutter Console`.

PR #1699, opened 2018-11-05
Apply `toc: true` by default to doc pages only

chalin:
This in particular removes the empty TOC from the search page.

Staged, e.g., see https://ng2-io.firebaseapp.com/search?q=apple


sfshaza2:
LGTM!

PR #1700, opened 2018-11-05
fix name

Rahiche:
fix channel name from Tensorflow to Tensor Programming 
source :
https://www.youtube.com/channel/UCYqCZOwHbnPwyjawKfE21wg


Rahiche:
CC @Sfshaza 
i think this is a typo you should fix

PR #1701, opened 2018-11-05
Update libimobiledevice installation instructions

tvolkert:
HEAD libimobiledevice depends on HEAD usbmuxd.

https://github.com/flutter/flutter/issues/22595

### DIFF
--- a/src/docs/get-started/install/_ios-setup.md
+++ b/src/docs/get-started/install/_ios-setup.md
@@ -46,6 +46,9 @@ To deploy your Flutter app to a physical iOS device, you’ll need some addition
 
     ```terminal
     $ brew update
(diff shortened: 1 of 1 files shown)


tvolkert:
@Sfshaza is updating `master` the right thing to do?  Should I be also (or instead) updating the `dash` branch?

chalin:
Yes, `master` is now the right branch to target (`dash` is going away).

chalin:
I think that there should be a note linking back to https://github.com/flutter/flutter/issues/22595 and pointing out that this is a temporary workaround.

sfshaza2:
Yes, @tvolkert!  We've deleted the dash branch today, once we pushed it to master.

PR #1704, opened 2018-11-05
SDK Archive page: use BS tabs, drop "Show all" in favor of scrollable list

### LINKED ISSUE #1632: [dash] "Show all" link on SDK archive page isn't hiding anything
See e.g.:
https://flutter-io-staging-1.firebaseapp.com/docs/development/tools/sdk/archive#linux

chalin:
Fixes #1632

- Upgrade page so that it uses Bootstrap (BS) tabs like the other site pages
- Better style table entries using BS stripped table styling
- Use scrollable table entries rather than a "Show all..." button (which sometimes would result if a very long page.

Staged at https://ng2-io.firebaseapp.com/docs/development/tools/sdk/archive

cc @Hixie 

---

### Old

> screen shot 2018-11-05 at 15 14 04

### New

> screen shot 2018-11-05 at 15 14 38


sfshaza2:
@Hixie, can you take a look? (Since you filed the issue.)

src/_assets/js/archive.js
```diff
@@ -1,7 +1,7 @@
 // TODO(chalin): package archive.js as its own top-level js so that it can be loaded independently
 
 // Number of releases to show by default (rest will be behind a "show all" link).
-var releasesToShow = 5;
+var releasesToShow = 99999;
```
Hixie: ?

src/_assets/js/archive.js
```diff
@@ -11,12 +11,9 @@ var fetchFlutterReleases = function (os, callback, errorCallback) {
...
     url: url,
     dataType: "json",
-    success: function (data) {
-      callback(data, os);
-    },
+    success: function (data) { callback(data, os); },
     error: function (xhr, textStatus, errorThrown) {
-      if (errorCallback)
-        errorCallback(os);
+      if (errorCallback) errorCallback(os);
```
Hixie: this looks less clear than before

src/_assets/js/archive.js
```diff
@@ -58,7 +55,7 @@ function updateTable(releases, os) {
 
 function updateTableFailed(os) {
   var tab = $("#tab-os-" + os);
-  tab.find(".loading").text("Failed to load releases");
+  tab.find(".loading").text("Failed to load releases. Refresh page to try again.");
```
Hixie: Should we just automatically refresh?

Hixie:
RSLGTM

The nested scroll views are a bit weird, I'm not a huge fan of that approach, but I don't feel strongly about it.

src/_assets/js/archive.js
chalin: I didn't want to remove the logic used to partition the entries (in case the approach used in this PR wasn't approved).

src/_assets/js/archive.js
chalin: We could do that in a followup PR.

chalin:
Thanks for the RSLGTM, I'll go ahead and merge. We can address comments in a followup PR.

PR #1705, opened 2018-11-06
Add backtick to title

ChangJoo-Park:
This change is same as https://github.com/flutter/website/blob/master/src/docs/get-started/install/_get-sdk-win.md#run-flutter-doctor


chalin:
Thanks, for this PR, but it should actually be the other way around. We shouldn't be using the code style in heading. cc @sfshaza2

sfshaza2:
Thanks for the PR, @ChangJoo-Park! But @chalin is right. You can see our style guidelines here:  https://developers.google.com/style/headings

PR #1706, opened 2018-11-06
Fix references in markdown and web

ChangJoo-Park:
image

This is broken syntax


sfshaza2:
Hi, @ChangJoo-Park! Thanks for catching this bug! Unfortunately your fix will work, but is not quite right, When you see a markdown link like this: ` [text][]` it means that the value is set up at the bottom of the file. In this case, the markdown says:

```
Apple Developer Program is required to distribute your app to the App Store. View the [differences between Apple membership][].
```

And, if you look at the bottom of the file, it's set up like this:

```
[differences between Apple membership types]: https://developer.apple.com/support/compare-memberships
```

Do you see the problem? The text in the square brackets needs to match and does not. To fix this bug, all you need to do is to remove "types" from the square brackets below, or add "types" to the square brackets above.

ChangJoo-Park:
I agree with you @sfshaza2 in markdown format.  but can you see this page? 

image

https://flutter.io/docs/get-started/install/macos

Inside this page, It seems to be incorrect.

sfshaza2:
Fixed in PR https://github.com/flutter/website/pull/1952.  Closing.

PR #1708, opened 2018-11-06
Accessibility

tBoccinfuso:
New PR with same changes, last one seemed to be closed when master was changed.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,7 +10,7 @@ description: >
   around the world, and is free and open source.
 ---
 
-<section class="homepage__hero text-center carousel slide" data-ride="carousel" data-interval=6500 data-pause="false">
+<section class="homepage__hero text-center carousel slide" data-ride="carousel" data-interval=6500 data-pause="false" aria-label="Home carousel">
     <div class="carousel-inner">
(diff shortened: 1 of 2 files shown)


tBoccinfuso:
@chalin Opened a new PR for the accessibility changes.

sfshaza2:
@fertolg, can either you or @c-machado weigh in?

chalin:
A few remarks:

- @c-machado does it really add value to have some sections labelled in this way?
  - For example, there is no element in the carousel that can get a focus, so maybe this ARIA lable isn't useful.

If we do decide to add some labels, here are some more questions:

- @c-machado: do we need `role="contentinfo"`?
- Not all `section` elements have a label:
  - The section containing the CTA doesn't have a label but it probably should.
  - Again, I'm unsure that the carousel should have a label.
- The text for some labels might need revision - @sfshaza2 maybe you'd like to take a look at that?

sfshaza2:
@fertolg, I don't see to have @c-machado's email. Can you ping and ask her to look at this?

sfshaza2:
I am closing this PR. We can reopen if we get more definitive info on making the page more accessible. (See Patrice's (@chalin) questions above.

PR #1741, opened 2018-11-16
chore: update Gem versions, in particular for `rack`

chalin:
This addresses a vulnerability reported in an earlier version of `rack`.


chalin:
Since this fixes a vulnerability, I'll merge now.

PR #1751, opened 2018-11-18
Platform channels and generated plugins

dnfield:
Fixes https://github.com/flutter/flutter/issues/24171

Docs are fine for Android, but were missing these lines for iOS.

/cc @jrmarkham


mit-mit:
Sorry, I'm not sure I'm a good reviewer of this

Issue #1492, opened 2018-10-17
[dash] Deprecated/watermark format option


Sfshaza:
Needed for future. Please implement. Thx!

Issue #1612, opened 2018-10-30
[dash] Samples catalog does not appear in the sidenav


chalin:
Should it be added to the **Reference** section?

Relevant links:

- flutter.io: https://flutter.io/catalog/samples/
- dash: https://flutter-io-staging-1.firebaseapp.com/docs/catalog/samples

galeyang:
The team recently decided to take it down for now, since it's out of date and overshadows new content we've recently added in the Samples github repo. We'll figure a more comprehensive replacement of its function in the future proposal that @InMatrix will be working on.

chalin:
Ok. In the meantime, what should I redirect [/catalog/samples/](https://flutter.io/catalog/samples/) to under the new site?

galeyang:
> what should I redirect /catalog/samples/ to under the new site?

@RedBrogdon Do you have any suggestions?

chalin:
Btw, while the samples are not linked to from the sidenav, the pages still exist under dash. For examples, see https://flutter-io-staging-1.firebaseapp.com/docs/catalog/samples.

Two choices come to mind:

(1) Keep the samples catalog, and link to it from the sidenav section on "Samples & codelabs", but mark the samples as "Obsolete". This is what we've done, for example, on the [dartlang.org/articles](https://www.dartlang.org/articles) index page and, e.g., this individual article page: https://www.dartlang.org/articles/language/await-async.

(2) Eliminate the sample catalog pages entirely from the site. As a redirect target for [/catalog/samples/](https://flutter.io/catalog/samples/), I could create an index page for the sidenav "Samples & codelabs" group that would simply list the nav entries of this group.

WDYT?

InMatrix:
My preference is (1), we can even consider showing the link to the samples catalog in a lighter shade of gray and mark it as "Deprecated" or "Obsolete". In the page itself, we can state that we are working on a new visual index of UI samples and this page won't receive updates, in the mean time, please visit the Github repo and the cookbook. In addition, the link should go below the link to the samples repo on GitHub, since @RedBrogdon was concerned about showing the samples catalog will direct attention away from the samples repo, which has a lot more content. 

I'd be interested in hearing others' opinions on this.

sfshaza2:
I like Tao's suggestion, especially if making the text in the sidenav grey is not too hard.

sfshaza2:
[...]

PR #1760, opened 2018-11-21
Snackbar component, deprecated pages

fertolg:
- Implements snackbar component and styles
- Adds deprecated metadata to catalog sample pages
- Adds a snackbar when a doc is deprecated

Staging links:
- https://flutter-web-staging.firebaseapp.com/docs/catalog/samples
- https://flutter-web-staging.firebaseapp.com/docs/catalog/samples/animated-list

Fixes: #1492

### DIFF
--- a/.firebaserc
+++ b/.firebaserc
@@ -5,7 +5,8 @@
     "staging-1": "flutter-io-staging-1",
     "staging-2": "flutter-io-staging-2",
     "pc2": "ng2-io",
-    "sz" :  "sz-flutter",
-    "sz2" : "sz-flutter-2"
+    "sz": "sz-flutter",
+    "sz2": "sz-flutter-2",
+    "ft": "flutter-web-staging"
   }
 }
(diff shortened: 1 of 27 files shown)


fertolg:
@chalin I don't see the deprecated pages in the navigation structure, so I'm unsure how to tackle the part where the nav should show them greyed out. Should I add these pages into the side nav? If so, do you know where they are supposed to live under?

chalin:
> I don't see the deprecated pages in the navigation structure ...

Right, that's what issue #1612 is about. Feel free to address #1612 in this PR or a followup PR (I'd be ok with the latter).

Personally, I find the snackbar too discrete. @sfshaza2, WDYT?

chalin:
> My first reaction was similar to Patrice's - surprise that the note was not the width of the window. But as I scrolled around, it was very noticeable. So I'm OK with it.

I'm ok with the width. Maybe we could:

- Change the color of the snackbar to an alert color
- Also, change the page background color (a light tint of yellow or orange) -- that would catch the readers attention and lead them to read the snackbar message.

Given @InMatrix's stated preference in https://github.com/flutter/website/issues/1612#issuecomment-434756324, maybe you have an opinion about this PR? /cc @galeyang

InMatrix:
Thanks for making this PR! I agree with [the comments](https://github.com/flutter/website/pull/1760#issuecomment-440908897) @Sfshaza made earlier. I don't think it's necessary to make the snackbar appear more prominently than it is.

PR #1761, opened 2018-11-22
Correct VS Code shortcut for quick-fixes

DanTup:
The quick-fix shortcut seems to be wrong on the site, it's `Cmd`(/`Ctrl`)+`.`.

https://code.visualstudio.com/docs/getstarted/keybindings#_rich-languages-editing

screen shot 2018-11-21 at 12 59 22 pm

Fixes https://github.com/Dart-Code/Dart-Code/issues/1317.

### DIFF
--- a/src/docs/development/tools/ide/vs-code.md
+++ b/src/docs/development/tools/ide/vs-code.md
@@ -143,7 +143,7 @@ To open Observatory:
 Assists are code changes related to a certain code identifier. A number of these
 are available when the cursor is placed on a Flutter widget identifier, as
 indicated by the yellow lightbulb icon. The assist can be invoked by clicking
-the lightbulb, or by using the keyboard shortcut `Ctrl`+`.` (`Cmd`+`.` on macOS), as
+the lightbulb, or by using the keyboard shortcut `Ctrl`+`.` (`Cmd`+`.` on Mac), as
 illustrated here:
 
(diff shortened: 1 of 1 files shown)


src/docs/development/tools/ide/vs-code.md
```diff
@@ -143,8 +143,8 @@ To open Observatory:
 Assists are code changes related to a certain code identifier. A number of these
 are available when the cursor is placed on a Flutter widget identifier, as
 indicated by the yellow lightbulb icon. The assist can be invoked by clicking
-the lightbulb, or by using the keyboard shortcut `Ctrl`+`Enter`, as illustrated
-here:
+the lightbulb, or by using the keyboard shortcut `Ctrl`+`.` (`Cmd`+`.` on macOS), as
```
chalin: macOS -> Mac

(in accordance with the [Doc Style Guide section on UI elements](https://developers.google.com/style/ui-elements?hl=th#keyboard))

chalin: The guidelines also suggest using the full modifier key name (e.g. `Command`), but I can fix that later, since there are occurrences in other places too.

DanTup: I've changed this, though I think there are other instances of `macOS` too (I think I copied this one from somewhere else for consistency) so may be worth searching them all if you're going to fix up the `Cmd`s etc.

Issue #1764, opened 2018-11-25
reduce the level of nesting for IDE pages


devoncarew:
Currently to view information about IDE support for Flutter, from the nav menu you have to open 'Development' > 'Tools & Techniques' > 'IDEs' > 'IntelliJ'. This is highly nested and could make these topics harder to find.

I'd like to reduce the nested here. That could involve:
- changing the 'Tools & techniques' topic title to something like 'Tools, techniques, and IDEs'
- removing the 'IDEs' topic category, and moving the IntelliJ and VS Code pages up one level

cc @Sfshaza @sfshaza2

PR #1765, opened 2018-11-25
remove the IDE topics category; move IDE pages up one level

devoncarew:
- remove the IDE topics category and move IDE pages up one level
- update redirects and url references
- fix https://github.com/flutter/website/issues/1764


sfshaza2:
I thought we addressed this by auto-opening the IDE directory when opening the tab above, but this is fine, too.

devoncarew:
> I thought we addressed this by auto-opening the IDE directory when opening the tab above, but this is fine, too.

Ah, I didn't know about that change. I think this is nice in that there's less nesting overall.

sfshaza2:
True!

chalin:
See #1775, _Sidenav: active-entry highlight broken by #1765 (IDEs)_

PR #1773, opened 2018-11-27
correctly attribute which chart is which thread

maryx:
The text under the image contradicts the details in the bullet points about which graph shows the UI thread and GPU thread.

### DIFF
--- a/src/docs/testing/ui-performance/index.md
+++ b/src/docs/testing/ui-performance/index.md
@@ -143,7 +143,7 @@ The following screenshot shows the performance overlay running on the Flutter
 Gallery example:
 
 <center><img src="/docs/testing/ui-performance/images/performance-overlay-green.png" alt="screenshot of performance overlay showing zero jank"></center>
-<center>Performance overlay showing the GPU thread (top), and UI thread (bottom).
+<center>Performance overlay showing the UI thread (top), and GPU thread (bottom).
 	The vertical green bars represent the current frame.</center><br>
 
 Flutter uses several threads to do its work. All your Dart code runs on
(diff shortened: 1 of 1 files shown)


maryx:
[Line 170 ](https://github.com/flutter/website/pull/1773/files#diff-301cf703797391d62864d401a9febe39R170 )says the UI thread is on the top and line 179 says the GPU thread is on the bottom.

chalin:
Ah! Thanks for spotting the inconsistency, but it's the prose that needs to be adjusted, not the figure caption. Can you make the change?

Actually, since the caption is clear, I'd suggest dropping the two sentences "Shown in the [top|bottom] row of the performance overlay". @sfshaza2 WDYT?

sfshaza2:
That sounds fine. They changed this. Originally it was correct.  :)

maryx:
I personally did like the explicitness of how each thread's last sentence was whether the performance overlay showed the thread or not (and if it did, where it is shown). But I can remove those two particular lines (for UI and GPU) if you want :)

chalin:
I'm ok either way. @sfshaza2?

sfshaza2:
If you liked it, @maryx, then why remove it? Did I miss something?

chalin:
She hasn't removed it yet, that was my suggestion for an edit to this PR. The question is: do you prefer having the "top/bottom" distinction made in the figure caption only, or in the prose as well?

sfshaza2:
Given that I wrote that prose, I'm fine with having it there.  :)

chalin:
> Ok cool, let me know if you all LGTM!

Surely, as soon as you make the changes. Thanks.

maryx:
Ah! Apologies, forgot that they were switched.

maryx:
@chalin @sfshaza2 I'm on https://docs.flutter.io/flutter/widgets/PerformanceOverlay-class.html and notice that it says `The overlay show two time series. The first shows how much time was required on this thread to produce each frame. The second shows how much time was required on the GPU thread to produce each frame. `. Is this inconsistent with https://flutter.io/docs/testing/ui-performance (`Performance overlay showing the GPU thread (top), and UI thread (bottom).`)?

Issue #1775, opened 2018-11-27
Sidenav: active-entry highlight broken by #1765 (IDEs)


chalin:
**PR #1765 breaks the sidenav active-entry highlighting**? Ie.:

- Visit https://flutter.io/docs/development/tools/android-studio and
- Look at the sidenav

You won't see any active entry being highlighted.

> > (@sfshaza2) I thought we addressed this by auto-opening the IDE directory when opening the tab above, but this is fine, too.
> 
> (@devoncarew) Ah, I didn't know about that change.

As far as I knew, when the auto-open-IDE-group PR was merged, it was judged as acceptable.

> I think this is nice in that there's less nesting overall.

We have deep nesting in other parts of the site, so IMHO we shouldn't be trying to flatten the site navigational hierarchy.

---

Anyhow, moving forward, should we simply revert #1765 @sfshaza2 @devoncarew

devoncarew:
> We have deep nesting in other parts of the site

Good to know, but I think a flatter structure is overall better. This does also remove one row from the table of contents as well.

> You won't see any active entry being highlighted.

Do you know the what the issue is - is there a mismatch between the TOC and a page ID for instance?

chalin:
> Do you know the what the issue is - is there a mismatch between the TOC and a page ID for instance?

Yes. So I gather that you'd rather not revert #1765? :). In that case, I can fix this issue.

devoncarew:
I'd prefer to keep it :)

I think https://github.com/flutter/website/pull/1777 should fix the issue (and thanks for finding and reporting it!)

PR #1777, opened 2018-11-27
fix the IDE side nav highlights

devoncarew:
- fix the IDE side nav highlights
- fix https://github.com/flutter/website/issues/1775

@chalin


devoncarew:
(I wonder if it would be possible to validate the `permalink` items in the .yml files as part of a build process?)

chalin:
> (I wonder if it would be possible to validate the permalink items in the .yml files as part of a build process?)

#1778

Issue #1738, opened 2018-11-15
Drop `.html` suffix from canonical page link


chalin:
Firebase for the new website is configured to redirect to **clean URLs** (w/o `.html` suffices) and URLs with **no trailing suffixes**, canonical page URLs should be updated to reflect this. E.g.:

- https://flutter.io/docs/get-started/install is currently set to
  `<link rel="canonical" href="https://flutter.io/docs/get-started/install/">`
- https://flutter.io/docs/get-started/learn-more is currently set to
  `<link rel="canonical" href="https://flutter.io/docs/get-started/learn-more.html">`

This change will impact analytics, but it is best that we do this as early as possible (since the new site hasn't been put through its paces that much yet).

Are you ok with such a change @sfshaza2?

sfshaza2:
Yes, as you said, earlier is better than later.

PR #1782, opened 2018-11-27
Drop self-referential canonical links in site page heads

chalin:
Followup to #1738 and #1739.


chalin:
Note that self-referential canonical links, though redundant, aren't an issue for SEO, but linkcheck seems to run into problems. So this PR removes them. cc @filiph

sfshaza2:
Thanks for fixing, @chalin!

PR #1812, opened 2018-12-01
cookbook: Added Navigation Named Routes Drawer recipe

pdxjohnny:
Going through the examples I thought this might be a good cookbook for others.
However, this might be more suitable as an update to `Add a Drawer to a screen`


src/docs/cookbook/navigation/named-routes-drawer.md
```diff
@@ -0,0 +1,236 @@
...
+In order to share Navigation logic used in the drawer between screens we'll
+create a `NavigationDrawer` widget. Within the `onTap` of our `ListTitle`
+widgets we'll use the
+[`Navigator.pushReplacementNamed`](https://docs.flutter.io/flutter/widgets/Navigator/pushReplacementNamed.html)
+function. This associates the list items in our `Drawer` with the screens we'll
+define our `routes` table.
+
+<!-- skip -->
+```dart
+class NavigationDrawer extends StatelessWidget  {
```
brianegan: Just to ensure we're creating the same NavigationDrawer every time, it might be good to provide a `const` Constructor here. That way you can always instantiate this with `const NavigationDrawer()`, which will ensure the same copy will always be re-used.

Otherwise, it might be good to accept a `key` here, and pass the same key on each page. If you were to provide a different key to each page, or if the structure of the pages was a bit different, Flutter might think these are two different things and create a new instance. Using a consistent `key` or `const` constructor should protect against that!

src/docs/cookbook/navigation/named-routes-drawer.md
```diff
@@ -0,0 +1,236 @@
...
+          DrawerHeader(
+            child: Text('Drawer Header'),
+            decoration: BoxDecoration(
+              color: Colors.blue,
+            ),
+          ),
+          ListTile(
+            title: Text('First'),
+            onTap: () {
+              // Update the state of the app
```
brianegan: Rather than only keeping these comments, perhaps we could say something about why you're using `pushReplacement` here and what effect that will have?

src/docs/cookbook/navigation/named-routes-drawer.md
```diff
@@ -0,0 +1,236 @@
...
+Second, we'll need two screens to navigate to. Both just contain Text widgets.
+
+<!-- skip -->
+```dart
+class FirstScreen extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      // Reuse our NavigationDrawer between widgets
+      drawer: NavigationDrawer(),
```
brianegan: `const NavigationDrawer` to ensure it really is only a single instance.

src/docs/cookbook/navigation/named-routes-drawer.md
```diff
@@ -0,0 +1,236 @@
...
+    );
+  }
+}
+
+class SecondScreen extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      // Reuse our NavigationDrawer between widgets
+      drawer: NavigationDrawer(),
```
brianegan: `const NavigationDrawer()`

pdxjohnny:
Thank you for the review! I'll make these changes ASAP.

pdxjohnny:
Haha not yet! Sorry for the slow turnaround. I'm hoping to make those
changes within the next two weeks.

sfshaza2:
This has languished for months. If you can get back to it, @pdxjohnny, please reopen. @brianegan had some feedback to incorporate.

PR #1820, opened 2018-12-02
Update react-native-devs.md

rknell:
Shared preferences library referencing old version that is incompatible with latest version of Flutter.

Just a version bump to the latest version.

### DIFF
--- a/src/docs/get-started/flutter-for/react-native-devs.md
+++ b/src/docs/get-started/flutter-for/react-native-devs.md
@@ -1507,7 +1507,7 @@ file then import the package in your Dart file.
 dependencies:
   flutter:
     sdk: flutter
-  shared_preferences: ^0.3.0
+  shared_preferences: ^0.4.3
 ```
 
 <!-- skip -->


rknell:
I signed it!

src/docs/get-started/flutter-for/react-native-devs.md
```diff
@@ -1507,7 +1507,7 @@ file then import the package in your Dart file.
 dependencies:
   flutter:
     sdk: flutter
-  shared_preferences: ^0.3.0
+  shared_preferences: ^0.4.3
```
chalin: @Hixie @gspencergoog @filiph: gent's. Is this update necessary and/or desirable? /cc @sfshaza2

rknell: My two cents - and the reason I made the pull request - is because its part of the onboarding process for new devs. Following the guide the app will not build.

Perhaps it could be policy to provide a link to the library in the docs instead of an "Add this here" kind of instruction - at least that would create less maintenance, but that is not my call.

I'll leave it with you guys to decide from here.

PR #1863, opened 2018-12-04
Fixed link to the page with differences between Apple membership

fobo66:


fobo66:
Closed because it was already fixed in #1952

chalin:
Thanks for the PR, but it turned out that I also wanted to tweak the prose.

PR #1872, opened 2018-12-05
Replace old homepage news item

chalin:
Staged at https://flutter-io-staging-2.firebaseapp.com


chalin:
@maguinis, are you ok with the new news-item image introduced in this PR? If not, can you get Daniel to create a new one? (I did the best to put together an image with the resources I have access to, especially fonts.)

PR #1884, opened 2018-12-05
SDK archive page: show all channels

chalin:


chalin:
@Hixie @sfshaza2 - This PR reintroduces the beta channel to the archive page. I'll address other concerns (lack of description) in a followup PR.

PR #1899, opened 2018-12-05
Add Expanded widget as a "Layout helper"

### LINKED ISSUE #1856: Expanded widget missing from Layout page of widget catalog
@animeshjain commented on [Fri Aug 25 2017](https://github.com/flutter/flutter/issues/11782)

From URL: https://flutter.io/widgets/layout

I feel this catalog should also include the Expanded widget. It took me a while to discover that, and it is a valuable one.

---

@sethladd commented on [Mon Aug 28 2017](https://github.com/flutter/flutter/issues/11782#issuecomment-325462707)

Thanks for the feedback! Patches definitely welcome :)

---

@Hixie commented on [Wed Dec 13 2017](https://github.com/flutter/flutter/issues/11782#issuecomment-351246324)

Fixing the widget catalog is not on our near-term priority list.

chalin:
Fixes #1856 

@Hixie - do you agree with the `subcategories` attribution? This results in two helpers:

> screen shot 2018-12-05 at 14 30 41

cc @sfshaza2 @kwalrath

### DIFF
--- a/src/_data/catalog/widgets.json
+++ b/src/_data/catalog/widgets.json
@@ -939,7 +939,7 @@
     "categories": [
     ],
     "subcategories": [
-      "Layout helpers"
+      "Multi-child layout widgets"
     ],
     "link": "https://docs.flutter.io/flutter/widgets/LayoutBuilder-class.html",
     "image": "<img alt='' src='/images/catalog-widget-placeholder.png'>"
@@ -1634,9 +1634,9 @@
     "categories": [
     ],
     "subcategories": [
-      "Layout helpers"
+      "Multi-child layout widgets"
     ],
-    "description": "A widget that expands a child of a Row, Column, or Flex..",
+    "description": "A widget that expands a child of a Row, Column, or Flex.",
     "link": "https://docs.flutter.io/flutter/widgets/Expanded-class.html",
     "image": "<img alt='' src='/images/catalog-widget-placeholder.png'>"
   },
--- a/src/_data/catalog/widgets.json
+++ b/src/_data/catalog/widgets.json
@@ -1630,6 +1630,17 @@
     "image": "<img alt='' src='/images/catalog-widget-placeholder.png'>"
   },
   {
+    "name": "Expanded",
+    "categories": [
+    ],
+    "subcategories": [
+      "Layout helpers"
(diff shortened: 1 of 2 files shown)


src/_data/catalog/widgets.json
```diff
@@ -1629,6 +1629,17 @@
...
     "image": "<img alt='' src='/images/catalog-widget-placeholder.png'>"
   },
+  {
+    "name": "Expanded",
+    "categories": [
+    ],
+    "subcategories": [
+      "Layout helpers"
+    ],
+    "description": "A widget that expands a child of a Row, Column, or Flex..",
```
Hixie: double period

src/_data/catalog/widgets.json
```diff
@@ -1629,6 +1629,17 @@
...
+    "link": "https://docs.flutter.io/flutter/widgets/Expanded-class.html",
+    "image": "<img alt='' src='/images/catalog-widget-placeholder.png'>"
```
Hixie: needs an icon

Hixie:
LayoutBuilder and Expanded should probably not be in the same category.

LayoutBuilder belongs with the other platonic widgets, or alternatively with MultiChildCustomLayout (or whatever it's called). Expanded belongs with Flexible (it's almost identical), and with either Positioned and other parent data widgets, or with Row/Column/Flex.

src/_data/catalog/widgets.json
chalin: Oops! Fixed

src/_data/catalog/widgets.json
chalin: Are you saying that there is an image that exists already that I can use? (Or is it a general remark that we should be avoiding the use of the placeholder image as much as possible?)

chalin:
> LayoutBuilder belongs with the other platonic widgets, or alternatively with MultiChildCustomLayout ...

Ok, I've moved both `LayoutBuilder` into the "Multi-child layout widgets" subcategory.

> Expanded belongs with ... with Row/Column/Flex.

`Row` is under the "Multi-child layout widgets" subcategory, so I'll move `Expanded` there too.

Thanks for your feedback!

Issue #1923, opened 2018-12-06
Rework visual samples catalog


chalin:
This is a followup to https://github.com/flutter/website/issues/1198#issuecomment-444932383 where @RedBrogdon wrote:

> One of the items on my Q1 todo list, though, is to come up with something more visual, similar to the old samples page in terms of visual content, but linking people to the samples created by DevRel.
> 
> I'd be only happy to take pointers from y'all about how to best work something like that into the site.

chalin:
We could work together on this. I can think of ways to autogenerate the samples page based on the samples found in the https://github.com/flutter/samples (which we'd access as a submodule, so that we know exactly which version the website is build from / referring to). We might just need to agree on conventions such as where to find an app screenshot or thumbnail to show on the samples page. WDYT @RedBrogdon?

cc @sfshaza2 @kwalrath

InMatrix:
> When we rework the page (or maybe before), we probably want to remove the old samples, which live at URLs like https://flutter.io/docs/catalog/samples/animated-list.

Yes, we should evaluate if those samples are worth keeping. If they are, we can create a copy in the fluttter/samples repo.

RedBrogdon:
Working together on this sounds great. I spoke with @galeyang about her plan to design a new visual samples index yesterday, so we're on the same page there.

I'd love to get more information about how content for it can be generated, though. Can I do something as simple as include a test file designed to move the app through a series of screens while saving screenshots, for instance? If it's all right with you, @chalin, I may reach out next week to try and put a GVC on the calendar.

johnpryan:
Does anyone mind if I mark this as fixed? https://flutter.github.io/samples

InMatrix:
I think it's fine to close this bug, so we can file more specific improvements request against the new samples catalog. 

One thing I'd like to get fixed sooner is the name of this page in the left navigation:

![image](https://user-images.githubusercontent.com/348942/77208655-8882f800-6ab9-11ea-8ab5-4cc676a7b491.png)

The `Flutter Gallery [New]` link takes the user to the https://flutter.github.io/samples, but I would assume this link was more specifically for https://flutter.github.io/gallery/.

sfshaza2:
SGTM

johnpryan:
Added a fix here: https://github.com/flutter/website/pull/3837

PR #1924, opened 2018-12-06
Gut /docs/catalog/samples

### LINKED ISSUE #1198: examples not linked in the examples catalog
There are many examples in the catalog/samples folder, but only four are linked at https://flutter.io/catalog/samples/

Why?

kwalrath:
And remove its link from the sidenav.

Fixes #1198.

/cc @RedBrogdon @sfshaza2


sfshaza2:
Yes, we did.

chalin:
I think that originally we wanted a samples page so that it would act as a target for redirects. In PR keeps the page (for redirects), but instead correctly directs folks to the samples repo. Is that ok @sfshaza2?

chalin:
Out-of-band, @sfshaza2 said that if @InMatrix says this is ok, then we can merge.

InMatrix:
I asked the OP in #1198 to weigh in. If we don't hear from him/her in a day or two, I'm Okay with removing the deprecated samples catalog as long as we're committed to a proper replacement as tracked in #1923.

chalin:
Good plan, thanks.

sfshaza2:
sgtm. Thanks, Tao.

PR #1942, opened 2018-12-07
 Add a link to the Flutter App Highlight Reel

### LINKED ISSUE #1929: Add app highlight reel to our flutter.io/showcase page
Either via a button or a new card - would be great to have this video accessible on flutter.io/showcase -> https://youtu.be/REJDzio_h7o

chalin:
How's this as a fix for #1929?

(**Edit: updated screenshot to reflect latest PR commits**)
screen shot 2018-12-07 at 18 49 21

Staged at https://flutter-io-staging-2.firebaseapp.com/showcase

Closes #1929

### DIFF
--- a/src/showcase/index.html
+++ b/src/showcase/index.html
@@ -48,7 +48,9 @@ body_class: landing-page showcase
                 <h2>Made with Flutter</h2>
                 <p>
                 Flutter is used by Google and developers and organizations around the world to deliver beautiful native apps on
-                iOS and Android. Here is a small selection of apps made with Flutter.
+                iOS and Android. Watch the
+                <a href="https://youtu.be/REJDzio_h7o">Flutter App Highlight Reel</a>,
+                or see below for a small selection of apps made with Flutter.
                 </p>
             </div>
         </div>
--- a/src/showcase/index.html
+++ b/src/showcase/index.html
@@ -48,8 +48,10 @@ body_class: landing-page showcase
                 <h2>Made with Flutter</h2>
                 <p>
                 Flutter is used by Google and developers and organizations around the world to deliver beautiful native apps on
-                iOS and Android. Watch the
-                <a href="https://youtu.be/REJDzio_h7o">Flutter App Highlight Reel</a>,
+                iOS and Android.
+                </p>
+                <p>
+                Watch the <a href="https://youtu.be/REJDzio_h7o">Flutter App Highlight Reel</a>,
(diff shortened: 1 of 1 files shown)


chalin:
An alternative would be to keep the sentences together in the same paragraph:

screen shot 2018-12-07 at 17 44 45

kwalrath:
I slightly prefer one paragraph. It de-emphasizes the highlight reel link a little bit, but it avoids having multiple centered paragraphs, which looks untidy to me.

I'd like to tweak the wording a bit. (Shocking, I know!)

chalin:
Thanks for your feedback. I separated the paragraphs precisely to emphasize the reel, which is what I thought @maguinis wanted (but I could be wrong).

> I'd like to tweak the wording a bit. (Shocking, I know!)

Please tweak away! :)

src/showcase/index.html
```diff
@@ -48,7 +48,11 @@ <h3>{{ case.name }}</h3>
                 <h2>Made with Flutter</h2>
                 <p>
                 Flutter is used by Google and developers and organizations around the world to deliver beautiful native apps on
-                iOS and Android. Here is a small selection of apps made with Flutter.
+                iOS and Android.
+                </p>
+                <p>
+                Watch the <a href="https://youtu.be/REJDzio_h7o">Flutter App Highlight Reel</a>,
+                or see below for a small selection of apps made with Flutter.
```
kwalrath: How about:

...iOS and Android. Here's a small selection of apps made with Flutter. Also see the <a href="...">Flutter app highlight reel.</a>

(That puts the link at the end, where I think it's more likely to be noticed/clicked, and also makes the wording a little less formal.)

chalin: I like it! (I'd considered something like that, but decided against it because we write "Here ...", but then deliver the goods only after an intervening sentence. :)

chalin:
Update done. Staging server refreshed: https://flutter-io-staging-2.firebaseapp.com/showcase. PTAL

chalin:
The old screenshot is given below. I've updated the image in the opening comment so that it doesn't lead to confusion.

screen shot 2018-12-07 at 17 51 40

chalin:
@maguinis - Since Kathy has approved, and under the assumption that you at least consider this PR as a step in the right direction, I'm going to merge so that we can let folks have access to the reel URL. As usual, we can make iterative improvements later as needed.

PR #1946, opened 2018-12-08
Update the FAQ

### LINKED ISSUE #1937: "Will Apple reject my app" could be stronger
@eseidelGoogle commented on [Wed Aug 08 2018](https://github.com/flutter/flutter/issues/20309)

https://flutter.io/faq/#will-apple-reject-my-flutter-app

"We can’t speak for Apple, but Apple’s policies have changed over the years, and they have allowed apps built with systems like Flutter. We are aware of apps built with Flutter that have been reviewed and released via the App Store.

Of course, Apple is ultimately in charge of their ecosystem, but our goal is to continue to do everything we can to ensure Flutter apps can be deployed into Apple’s App Store."

We could offer stronger examples, such as the fact that Flutter-built apps have even been featured on the Apple App store (e.g. Hamilton).  We could also cite that technologies like Flutter (e.g. Unity) are used to produce *many* apps on the app store.

Thoughts @mjohnsullivan @mit-mit?

---

@mit-mit commented on [Thu Aug 09 2018](https://github.com/flutter/flutter/issues/20309#issuecomment-411739639)

Yes, would be great to rephrase this!

kwalrath:
Fixes #1937.

Staged: https://kw-flutter-2.firebaseapp.com/docs/resources/faq#will-apple-reject-my-flutter-app

/cc @sfshaza2 @chalin


src/docs/resources/faq.md
```diff
@@ -884,11 +883,13 @@ We are focused on three things:
...
-apps built with Flutter that have been reviewed and
-released via the App Store.
+years, and their App Store contains many apps built with technologies
+like Unity and Flutter.
+Apple has even featured Flutter apps such as [Hamilton.][Hamilton for iOS]
 
 Of course, Apple is ultimately in charge of their ecosystem, but our goal
-is to continue to do everything we can to ensure Flutter apps can be
+is to continue to do everything we can to ensure that Flutter apps can be
 deployed into Apple's App Store.
```
chalin: deployed into -?-> deployed to

src/docs/resources/faq.md
```diff
@@ -84,8 +84,7 @@ apps are in the works, or "inside the firewall".
 
 Teams outside of Google are also using Flutter.
 For example, the official app for the musical Hamilton
-([Android](https://play.google.com/store/apps/details?id=com.hamilton.app),
-[iOS](https://itunes.apple.com/us/app/hamilton-the-official-app/id1255231054?mt=8))
+([Android][Hamilton for Android], [iOS][Hamilton for iOS])
```
chalin: I know that the change to this question was outside the scope of the issue that you've fixed, but ... IMHO this question should merely be a segway to the Showcase. The entire answer could be:

> Developers both inside and outside Google are using Flutter to build beautiful native business-critical apps for iOS and Android. Discover some of these apps by visiting the Flutter [Showcase](/showcase).

Or, if you prefer, we can borrow some of the introductory Showcase text:

> Flutter is used by Google and developers and organizations around the world to deliver beautiful native apps on iOS and Android. Learn about some of these apps from the Flutter [Showcase](/showcase).

(What originally got me thinking about this change is the following: We don't provide any download links in the first paragraph, so why do so here?)

pedromorgan:
whaw.. there we go..

To publish and before..

U need a mac and xcode.. = dev account...

Guess next step is to do like nativescript.org and do cloud builds ;-))))))))))

src/docs/resources/faq.md
mit-mit: I really like the idea of linking to the Showcase here.

src/docs/resources/faq.md
sfshaza2: "is to continue to do.." => "is continuing to do..."

And, yes, "to be deployed TO Apple's App Store."

PR #1956, opened 2018-12-09
Fix images in Image widget of Sizing widgets example not fit full width

oopsguy:
[https://flutter.io/docs/development/ui/layout#sizing-widgets](https://flutter.io/docs/development/ui/layout#sizing-widgets)

In this example, the middle widget is not twice as wide as the other two widgets because of `BoxFit.scaleDown`.

### DIFF
--- a/src/_includes/code/layout/row-expanded-2/main.dart
+++ b/src/_includes/code/layout/row-expanded-2/main.dart
@@ -45,13 +45,22 @@ class _MyHomePageState extends State<MyHomePage> {
           crossAxisAlignment: CrossAxisAlignment.center,
           children: [
             Expanded(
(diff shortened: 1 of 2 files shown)


oopsguy:
I signed it!

chalin:
Hi @oopsguy. Thanks for the PR. Are you saying that the sample code given in the [Sizing widgets](/docs/development/ui/layout#sizing-widgets) section doesn't render as illustrated in each of the corresponding images?

oopsguy:
Yes, that is what I want to say.

oopsguy:
@chalin @sfshaza2 I am very sorry, this is a careless mistake. The size of images I downloaded are not the same as the size of images in the Sizing widgets section. I should use the cloned images directly.

PR #1957, opened 2018-12-09
JSON and serialization edits

pedromorgan:
renamed  to avoid confuse, and small tweaks..

### DIFF
--- a/src/docs/development/data-and-backend/json.md
+++ b/src/docs/development/data-and-backend/json.md
@@ -78,10 +78,13 @@ encoding, see
 
 The simple answer is no.
 
-Such a library would require using runtime reflection, which is disabled in
-Flutter. Runtime reflection interferes with _tree shaking_, which Dart has
+Such a library would require using runtime 
+[reflection](https://en.wikipedia.org/wiki/Reflection_(computer_programming)) 
+which is disabled in
+Flutter. Runtime reflection interferes with 
+[tree shaking](https://en.wikipedia.org/wiki/Tree_shaking), which Dart has
 supported for quite a long time. With tree shaking, you can “shake off” unused
-code from your release builds. This optimizes the app's size significantly.
+code from the release builds and this optimizes the app's size significantly.
 
 Since reflection makes all code implicitly used by default, it makes tree
 shaking difficult. The tools cannot know what parts are unused at runtime, so
@@ -127,13 +130,13 @@ with the JSON string as the method argument.
 
 <!-- skip -->
 ```dart
-Map<String, dynamic> user = json.decode(json);
+Map<String, dynamic> user = json.decode(user_data_str);
 
 print('Howdy, ${user['name']}!');
 print('We sent the verification link to ${user['email']}.');
 ```
 
-Unfortunately, `json.decode()` merely returns a `Map<String, dynamic>`, meaning
+Unfortunately, `json.decode()` returns a `Map<String, dynamic>`, meaning
 that you do not know the types of the values until runtime. With this approach,
 you lose most of the statically typed language features: type safety,
 autocompletion and most importantly, compile-time exceptions. Your code will
@@ -151,6 +154,7 @@ class, called `User` in this example. Inside the `User` class, you'll find:
 * A `User.fromJson` constructor, for constructing a new `User` instance from a
   map structure.
 * A `toJson` method, which converts a `User` instance into a map.
+* Note: the `from*` `to*` naming is recommended but optional eg create a `fromMyFormat` etc
 
 With this approach, the _calling code_ can have type safety,
 autocompletion for the `name` and `email` fields, and compile-time exceptions.
@@ -167,9 +171,9 @@ class User {
 
   User(this.name, this.email);
 
-  User.fromJson(Map<String, dynamic> json)
-      : name = json['name'],
-        email = json['email'];
+  User.fromJson(Map<String, dynamic> data_map)
+      : name = data_map['name'],
+        email = data_map['email'];
 
   Map<String, dynamic> toJson() =>
     {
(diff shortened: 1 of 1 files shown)


pedromorgan:
I signed it! cymro bach yn cymru

chalin:
Thanks for your PR, but use of `json` as the formal parameter name is the convention (idiom?) followed when using, https://pub.dartlang.org/packages/json_serializable based code.

@kevmoo - do you agree with the statement above, or have any other comments about this?

cc @kwalrath

kevmoo:
> Thanks for your PR, but use of json as the formal parameter name is the convention (idiom?) followed when using, https://pub.dartlang.org/packages/json_serializable based code.

SGTM

src/docs/development/data-and-backend/json.md
```diff
@@ -127,13 +130,13 @@ with the JSON string as the method argument.
 
 <!-- skip -->
 ```dart
-Map<String, dynamic> user = json.decode(json);
+Map<String, dynamic> user = json.decode(user_data_str);
```
chalin: @kwalrath @kevmoo - should we be promoting use of `jsonDecode()` here instead? /cc @Hixie @sfshaza2

chalin:
After some discussion with @kwalrath, I'm warming up to some of the changes made in this PR. Let's see if we can make it work.

src/docs/development/data-and-backend/json.md
```diff
@@ -78,10 +78,13 @@ encoding, see
...
-Such a library would require using runtime reflection, which is disabled in
-Flutter. Runtime reflection interferes with _tree shaking_, which Dart has
+Such a library would require using runtime 
+[reflection](https://en.wikipedia.org/wiki/Reflection_(computer_programming)) 
+which is disabled in
+Flutter. Runtime reflection interferes with 
+[tree shaking](https://en.wikipedia.org/wiki/Tree_shaking), which Dart has
 supported for quite a long time. With tree shaking, you can “shake off” unused
-code from your release builds. This optimizes the app's size significantly.
+code from the release builds and this optimizes the app's size significantly.
```
chalin: @kwalrath - are you ok with the changes to this paragraph?

src/docs/development/data-and-backend/json.md
```diff
@@ -184,7 +188,7 @@ itself. With this new approach, you can decode a user easily.
 
 <!-- skip -->
 ```dart
-Map userMap = json.decode(json);
+Map userMap = json.decode(string_with_json);
```
chalin: In the [Library Tour](https://www.dartlang.org/guides/libraries/library-tour#dartconvert---decoding-and-encoding-json-utf-8-and-more), we use the variable name `jsonString`. Could you change it to that name?

src/docs/development/data-and-backend/json.md
```diff
@@ -197,7 +201,7 @@ already does it for you.
 
 <!-- skip -->
 ```dart
-String json = json.encode(user);
+String data_string = json.encode(user);
```
chalin: `data_string` -> `jsonString`

src/docs/development/data-and-backend/json.md
kwalrath: I like the original text better.

src/docs/development/data-and-backend/json.md
kwalrath: I personally like `jsonDecode()`, but not everyone agrees. @kevmoo?

src/docs/development/data-and-backend/json.md
chalin: How about the added links to Wikipedia?

kwalrath: Ah, missed those. I like them, but we should move the punctuation inside the brackets, so the external link markers don't look weird. So:

[reflection] -> [reflection,]
[tree shaking](...), -> [treeshaking,](...)

src/docs/development/data-and-backend/json.md
```diff
@@ -167,9 +171,9 @@ class User {
 
   User(this.name, this.email);
 
-  User.fromJson(Map<String, dynamic> json)
-      : name = json['name'],
-        email = json['email'];
+  User.fromJson(Map<String, dynamic> data_map)
+      : name = data_map['name'],
+        email = data_map['email'];
```
chalin: Use of `json` here is ok. Revert these changes.

src/docs/development/data-and-backend/json.md
```diff
@@ -151,6 +154,7 @@ class, called `User` in this example. Inside the `User` class, you'll find:
 * A `User.fromJson` constructor, for constructing a new `User` instance from a
   map structure.
 * A `toJson` method, which converts a `User` instance into a map.
+* Note: the `from*` `to*` naming is recommended but optional eg create a `fromMyFormat` etc
```
chalin: I'd suggest dropping this note. @kwalrath - WDYT?

src/docs/development/data-and-backend/json.md
chalin: Also, move the link definitions to the end of the file, rather than inlining them. E.g.,

    Such a library would require using runtime [reflection][] ...
    
    [reflection]: https://en.wikipedia.org/wiki/Reflection_(computer_programming)

kwalrath: Of course, if you have the commas in there, you'll probably want to use something like `[reflection,][reflection]`.

chalin: Oops, yes, thanks.

pedromorgan:
@chalin NO.. why not do edits yourself. make a decision, and rip the good parts, ignore others ;-))

One thing I dislike is the `your can, we can one can.`
cos this is crap when u speak+understand welsh, japanese and english.. doth not translate...

The language = manual needs to me more "commanding"..
```you can create a new app by.....```
to
```Create a new app, by...```

src/docs/development/data-and-backend/json.md
kevmoo: Ditto! Avoid confusion w/ local var/param named `json`

chalin: FYI, I've created: _Code cleanup: `json.decode()` -> `jsonDecode()`_ #2021

PR #1958, opened 2018-12-09
Remove factory constructor

passsy:
I don't see why a `factory` constructor is useful here. It behaves like a normal constructor:
1. It does always return a new instance
2. It returns always the same type (`Photo`)

The factory constructor also prevents subclasses from overriding `fromJson`. But it's not documented why this would be helpful here.

Another reason could be to write code in the factory body rather than the initializer list. Esthetically a good idea.

### DIFF
--- a/src/docs/cookbook/networking/fetch-data.md
+++ b/src/docs/cookbook/networking/fetch-data.md
@@ -56,7 +56,7 @@ convert the `http.Response` into our own Dart object.
 ### Create a `Post` class
 
 First, we'll need to create a `Post` class that contains the data from our
-network request. It will also include a factory constructor that allows us to
+network request. It will also include a named constructor that allows us to
 create a `Post` from json.
 
 Converting JSON by hand is only one option. For more information, please see the
@@ -72,7 +72,7 @@ class Post {
 
   Post({this.userId, this.id, this.title, this.body});
 
-  factory Post.fromJson(Map<String, dynamic> json) {
+  Post.fromJson(Map<String, dynamic> json) {
     return Post(
       userId: json['userId'],
       id: json['id'],
@@ -91,7 +91,7 @@ we'll need to:
   1. Convert the response body into a json `Map` with the `dart:convert`
   package.
   2. If the server returns an "OK" response with a status code of 200, convert
-  the json `Map` into a `Post` using the `fromJson` factory.
+  the json `Map` into a `Post` using the `fromJson` constructor.
   3. If the server returns an unexpected response, throw an error
 
 <!-- skip -->
@@ -236,7 +236,7 @@ class Post {
 
   Post({this.userId, this.id, this.title, this.body});
 
-  factory Post.fromJson(Map<String, dynamic> json) {
+  Post.fromJson(Map<String, dynamic> json) {
     return Post(
       userId: json['userId'],
       id: json['id'],
--- a/src/docs/cookbook/networking/fetch-data.md
+++ b/src/docs/cookbook/networking/fetch-data.md
@@ -72,14 +72,11 @@ class Post {
 
   Post({this.userId, this.id, this.title, this.body});
 
-  Post.fromJson(Map<String, dynamic> json) {
-    return Post(
-      userId: json['userId'],
-      id: json['id'],
-      title: json['title'],
-      body: json['body'],
-    );
-  }
+  Post.fromJson(Map<String, dynamic> json)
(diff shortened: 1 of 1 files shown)


rrousselGit:
Another solution is to introduce asserts:

```dart
class Photo {
  final int id;
  final String title;
  final String thumbnailUrl;

  Photo({ @required this.id, this.title, this.thumbnailUrl})
    assert(id != null);

  factory Photo.fromJson(Map<String, dynamic> json) {
    return Photo(
      id: json['id'] as int,
      title: json['title'] as String,
      thumbnailUrl: json['thumbnailUrl'] as String,
    );
  }
}
```

chalin:
@kevmoo any comments about this? cc @kwalrath

chalin:
> Sadly, we don't have any guidance here - https://www.dartlang.org/guides/language/effective-dart/design

@kwalrath - do you want to open an issue for this over www? cc @munificent

rrousselGit:
@kevmoo While I agree that this is a good thing, the factory is still not needed:

```dart
class Foo {
  final String foo;
  final int bar;

  Foo({this.foo, this.bar});

  Foo.fromJson(Map<String, dynamic> json)
      : this(
          foo: json['foo'] as String,
          bar: json['bar'] as int,
        );
}
```

kevmoo:
> the factory is still not needed

Sure. I like that more. Then it's a matter of how much we want to lead folks toward `json_seriailzable`.

chalin:
@brianegan - thoughts?

chalin:
@filiph - maybe you have some thoughts about this too?

passsy:
I do not expect this PR to be merged. I just wanted to start a discussion that the tutorial shouldn't throw new patterns at newcomers which aren't required for the problem to solve.

At first, I couldn't find any reason for the facoty constructor. Now I would most likely use the facoty constructor as well because the initializer list is uncomfortable to use, especially for beginners. Both solutions aren't ideal.

Here's a way to make use of the factory constructor which can't be replaced with the initializer list:

```dart
class Post {
  final int userId;
  final int id;
  final String title;
  final String body;

  Post({this.userId, this.id, this.title, this.body});

  factory Post.fromJson(String rawJson) {
    // intermediate variables for parsing aren't possible with initializer lists
    final json = json.decode(rawJson);

    return Post(
      userId: json['userId'],
      id: json['id'],
      title: json['title'],
      body: json['body'],
    );
  }
}
```

Although it might not be a good idea to add the the json parsing into the PODO.

Alternatively parse the json in a nice way:

```dart
  factory Post.fromJson(Map<String, dynamic> json) {
	final userId = json['userId'] as int;
	final userId = json['id'] as int;
	
	return Post(
      userId: userId,
      id: id,
      title: json['title'],
      body: json['body'],
    );
  }
```

brianegan:
Sorry, travel day for me! Overall, I do not feel strongly one way or another about this change. If folks really feel strongly that this is confusing, I'm happy to go that route.

Overall, I prefer factory constructors for the reasons folks have stated above, and I personally don't think it's more complicated than a named constructor. But again -- I don't feel strongly.

pedromorgan:
Ok a newbie and @passsy makes sence.. Got no factory in my v1 code..

sfshaza2:
Assigning to @brianegan. Brian, can you beef up the recipe a bit? Perhaps offer both options? First without factory, then with?

brianegan:
Hey all -- sorry about the delay, just back from vacation! 

@Sfshaza I could definitely do that. However, I'm wondering if it might make more sense to simply choose one path and stick with it while adding a link to more info? The reason: I would like to focus this recipe on the task at hand: making network requests and Parsing JSON, and I'm worried that diving into the different ways to construct an object feels tangential to that goal.

I'm totally sympathetic to the reason for creating this PR: "I just wanted to start a discussion that the tutorial shouldn't throw new patterns at newcomers which aren't required for the problem to solve." That's a great point! But overall I'm not sure adding both options to this recipe solves that problem.

sfshaza2:
No, I don't feel strongly and you make good points, @brianegan.

chalin:
> I would like to focus this recipe on the task at hand

I agree with that.

Given the full discussion here, I'd say that there isn't anything that needs to be changed. (If we had a link to more info, where would it point to? :)

mjohnsullivan:
As we're tidying up our outstanding PRs, I'm going to go and close this out. We can continue the discussion via another channel

PR #2027, opened 2018-12-13
Get started codelab: use code excerpter to sync code

chalin:
I've tried to minimize codelab markdown diffs, e.g., by using the more verbose `[[highlight]]...[[/highlight]]` tags.

**Not** yet implemented (but to be added in a followup PR):

- Checking code freshness as part of the build.
- Using the pretty code diff plugin

Staged at https://flutter-io-staging-2.firebaseapp.com/docs/get-started/codelab

### DIFF
--- a/src/docs/get-started/codelab.md
+++ b/src/docs/get-started/codelab.md
@@ -12,6 +12,7 @@ next:
 {% assign code-url = 'https://raw.githubusercontent.com/flutter/codelabs/master' -%}
 
 {% asset get-started/startup-namer-part-1 alt="The app that you'll be building" class='site-image-right' %}
+<?code-excerpt path-base="codelabs/startup_namer"?>
 
 This is a guide to creating your first Flutter app. If you
 are familiar with object-oriented code and basic programming
@@ -88,6 +89,7 @@ where the Dart code lives.
     Replace with the following code, which displays "Hello World" in the center
     of the screen.
 
+    <?code-excerpt "1-base/lib/main.dart"?>
     ```dart
     import 'package:flutter/material.dart';
 
@@ -173,12 +175,12 @@ packages, on [the Package site](https://pub.dartlang.org/flutter/).
     to the dependencies list.
     Add the highlighted line below:
 
+    <?code-excerpt "2-use-package/pubspec.yaml (dependencies)" indent-by="2" replace="/english_words.*/[[highlight]]$&[[\/highlight]]/g"?>
     {% prettify yaml %}
       dependencies:
         flutter:
           sdk: flutter
-
-        cupertino_icons: ^0.1.0
+        cupertino_icons: ^0.1.2
         [[highlight]]english_words: ^3.1.0[[/highlight]]
     {% endprettify %}
 
@@ -199,6 +201,7 @@ packages, on [the Package site](https://pub.dartlang.org/flutter/).
  3. In **lib/main.dart**, import the new package:
 
     <!-- skip -->
+    <?code-excerpt "2-use-package/lib/main.dart" retain="/^import/" replace="/import.*?english.*/[[highlight]]$&[[\/highlight]]/g" indent-by="2"?>
     {% prettify dart %}
       import 'package:flutter/material.dart';
       [[highlight]]import 'package:english_words/english_words.dart';[[/highlight]]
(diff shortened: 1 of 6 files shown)


src/docs/get-started/codelab.md
```diff
@@ -326,40 +331,44 @@ a child inside the existing `MyApp` stateless widget.
...
 
     <!-- skip -->
+    <?code-excerpt "3-stateful-widget/lib/main.dart (RandomWordsState)" indent-by="2" replace="/(\n  )(.*)/$1[[highlight]]$2[[\/highlight]]/g"?>
     {% prettify dart %}
       class RandomWordsState extends State<RandomWords> {
         [[highlight]]@override[[/highlight]]
         [[highlight]]Widget build(BuildContext context) {[[/highlight]]
-          [[highlight]]final wordPair = WordPair.random();[[/highlight]]
-          [[highlight]]return Text(wordPair.asPascalCase);[[/highlight]]
+        [[highlight]]  final wordPair = WordPair.random();[[/highlight]]
```
kwalrath: I think Shams would prefer the indentation spaces to not be highlighted. Is that possible?

src/docs/get-started/codelab.md
```diff
@@ -326,40 +331,44 @@ a child inside the existing `MyApp` stateless widget.
...
-                //child: [[highlight]]Text(wordPair.asPascalCase),[[/highlight]] // Change the highlighted text to...
-                child: [[highlight]]RandomWords(),[[/highlight]] // ... this highlighted text
-              ),
-            ),
-          );
-        }
-      }
-    {% endprettify %}
+    <?code-excerpt "2-use-package/lib/main.dart" diff-with="3-stateful-widget/lib/main.dart" to="}"?>
+    ```diff
```
kwalrath: Without diff formatting, this looks worse than the currently published version. Would it be possible to add the diff formatting in this PR?

src/docs/get-started/codelab.md
```diff
@@ -427,104 +437,96 @@ lazily, on demand.
...
-            [[highlight]]appBar: AppBar([[/highlight]]
-              [[highlight]]title: Text('Startup Name Generator'),[[/highlight]]
-            [[highlight]]),[[/highlight]]
-            [[highlight]]body: _buildSuggestions(),[[/highlight]]
-          [[highlight]]);[[/highlight]]
-        }
-        ...
+      @override
+      Widget build(BuildContext context) {
+        return Scaffold(
```
kwalrath: Perhaps highlight the contents? It seems OK to me to not highlight completely new methods, but when you're replacing the innards of a method, highlights seem appropriate.

src/docs/get-started/codelab.md
```diff
@@ -427,104 +437,96 @@ lazily, on demand.
...
 
-    Replace the original method with the highlighted build method below:
+    Replace the original method with the highlighted method below:
 
     <!-- skip -->
+    <?code-excerpt "4-infinite-list/lib/main.dart (MyApp)" replace="/(\n  )(.*)/$1[[highlight]]$2[[\/highlight]]/g" indent-by="2"?>
     {% prettify dart %}
       class MyApp extends StatelessWidget {
-        @override
+        [[highlight]]@override[[/highlight]]
```
kwalrath: The highlights here are inconsistent with the (lack of) highlights above. If the build() method already existed in this class (did it?), then highlighting only the method contents (the stuff that's changed) seems right to me.

src/docs/get-started/codelab.md
chalin: This will be turned into a diff soon.

src/docs/get-started/codelab.md
chalin: Pretty diffs are in the next PR, which is almost done:

screen shot 2018-12-13 at 13 04 48

src/docs/get-started/codelab.md
chalin: Yes, this will be changed to a pretty diff soon.

src/docs/get-started/codelab.md
chalin: This will also be changed to a pretty diff soon.

src/docs/get-started/codelab.md
chalin: Hmm, I can't use a diff against the code of the previous step because then it shows all of the class as new.

Personally, I find the aligned left margin highlight nicer (it seems to strengthen the idea that that "block" of code is what is being added.

src/docs/get-started/codelab.md
chalin: I've added the highlighting because the simple diff plugin can't handle this case.

src/docs/get-started/codelab.md
kwalrath: I don't have a strong feeling, but I'm pretty sure @sfshaza2 does.

PR #2032, opened 2018-12-13
Use non-breaking whitespace before units

goderbauer:
This will hopefully avoid awkward line breaks like this:

screen shot 2018-12-12 at 6 14 09 pm

### DIFF
--- a/src/docs/resources/faq.md
+++ b/src/docs/resources/faq.md
@@ -352,14 +352,14 @@ In December 2018, we measured the download size of a
 [minimal Flutter app](https://github.com/flutter/flutter/tree/75228a59dacc24f617272f7759677e242bbf74ec/examples/hello_world)
 (no Material Components, just a single `Center` widget, built with
 `flutter build apk`), bundled and compressed as a release APK, to be
-approximately 4.06 MB.
+approximately 4.06 MB.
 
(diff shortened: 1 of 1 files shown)


src/docs/resources/faq.md
```diff
@@ -352,14 +352,14 @@ In December 2018, we measured the download size of a
...
-the core engine is approximately 2.7 MB (compressed),
-the framework + app code is approximately 820.6 KB (compressed),
-the LICENSE file is 53.5 KB (compressed),
-necessary Java code (classes.dex) is 61.8 KB (compressed),
-and there is approximately 450.4 KB of (compressed) ICU data.
+the core engine is approximately 2.7 MB (compressed),
+the framework + app code is approximately 820.6 KB (compressed),
+the LICENSE file is 53.5 KB (compressed),
+necessary Java code (classes.dex) is 61.8 KB (compressed),
+and there is approximately 450.4 KB of (compressed) ICU data.
```
chalin: How about using `&nbsp;` instead. That would make it obvious (for this and future PRs) that we've got the right kinda space character here w/o having to resort to a character dump of the file.

goderbauer: Done.

PR #2048, opened 2018-12-15
copy change

pedromorgan:
manwel said basil fawlty..


pedromorgan:
in fact reveiwing this oneself..

Needs to have
- what is a package
- instaling a using package
- createing a package
- deleting forever a package. .coss dead, no perms or alike.. ie BURY and Unpublish

pedromorgan:
itd the wrong way around atmo,, need consumer first and then extend ahem.. + publish

sfshaza2:
Replaced by PR:  https://github.com/flutter/website/pull/2711

PR #2090, opened 2018-12-18
Fixed the get-started urls by add docs/

chenglu:
Please consider this PR, since not everyone will deploy Flutter docs by using Firebase (could set 301 to custom url / path), trying to use the full URL (add "docs/" before "get-started") will be great, thanks!


chalin:
> not everyone will deploy Flutter docs by using Firebase

Yes, good point. (Regardless, our in-doc links should point to actual pages, and not count on redirects.)

chenglu:
Great, thanks!

PR #2106, opened 2018-12-21
Update ios-devs.md

JulianMelzig:
Removed doubled "In" words from the docs.
Also changed the file extension from `jpeg` to `png`


JulianMelzig:
I signed it!

PR #2130, opened 2018-12-28
Remove duplicated widgets on catalog

wurikiji:


kwalrath:
I bet these duplicate entries were added because the widgets didn't appear in the expected (sub)category pages. I wonder if we should combine the entries, rather than just eliminating the duplicate entries.

/cc @sfshaza2

sfshaza2:
Thanks for your PR, @wurikiji, but we recently fixed this!

PR #2152, opened 2019-01-02
Adding info on branding.

sfshaza2:
Staged: 
https://sz-flutter.firebaseapp.com/tos
https://sz-flutter.firebaseapp.com/brand

Please review, @maguinis.


maguinis:
Scott and I JUST made a few tweeks to the copy -- could you please update then it will be good. @Sfshaza

PR #2163, opened 2019-01-04
App code for examples/layout/lakes in 6 steps

chalin:
cc @sfshaza2 - I'm adding the code in a separate PR so that we can validate links into the code. Such links will be added to the updated "Building Layouts" page, which will be the subject of my next PR.

This PR does not yet enable the execution of the app tests. That will also be done in a followup PR.


chalin:
Just to be clear, this PR only adds code to the examples directory. There are no changes to the site content.

Issue #2154, opened 2019-01-02
Improve the readability of full sample code listed in the tutorials


galeyang:
From recent new users' study, some participants found the raw source code listed on our tutorials (e.g. Building layout, Adding interactivity)  was a little hard to read when there is **no syntax highlighting**.

**_Some links to raw source code: main.dart_**

![image](https://user-images.githubusercontent.com/5102864/50613978-3cd43680-0e94-11e9-8455-52f861a8cee0.png)

**_Source code without syntax highlighting:_**
![image](https://user-images.githubusercontent.com/5102864/50614709-9b021900-0e96-11e9-882a-8c6c793063e4.png)

### **Proposal**
We can consider using links with Github webpage view (such as [https://github.com/flutter/website/blob/master/src/_includes/code/layout/widgets-only/main.dart](https://github.com/flutter/website/blob/master/src/_includes/code/layout/widgets-only/main.dart))
![image](https://user-images.githubusercontent.com/5102864/50614858-15cb3400-0e97-11e9-9892-a016fb2cbdcf.png)

cc: @jayoung-lee @sfshaza2

PR #2164, opened 2019-01-04
Update the Building Layouts page to use the new app sources

chalin:
Makes use of the code excerpter and code diff plugin in the Building Layouts page, steps 0 to 6.

Code changes:

- While the prose mentioned the need for a `MaterialApp` `theme` element, the actual sample code did not have any. It seems that the default primary color is blue, so I dropped the theme code setting from the prose.
- I promoted the local helper function `buildButtonColumn()` to be a private class method. This makes the code easier to read.

Contributes to #2154

Staged at https://flutter-io-staging-2.firebaseapp.com/docs/development/ui/layout

cc @filiph 

---

Left to do:
- Clean out the src/_includes/code/layout/lakes files once I've ensured that they are not used elsewhere.
- Enable the execution of the smoke tests for the 6 app versions of the Lakes app.

### DIFF
--- a/src/docs/development/ui/layout/index.md
+++ b/src/docs/development/ui/layout/index.md
@@ -1,11 +1,20 @@
 ---
 title: Building Layouts
 description: Learn how Flutter's layout mechanism works and how to build a layout.
+diff2html: true
 ---
+
+{% comment %}
+  TODO: address https://github.com/flutter/website/issues/2154 by adjusting the
+  definitions of the code and file Liquid variables below. Eventually consolidate
+  with examples and exFile.
+{% endcomment -%}
 
 {% assign api = 'https://docs.flutter.io/flutter' -%}
 {% assign code = 'https://raw.githubusercontent.com/flutter/website/master/src/_includes/code' -%}
 {% assign file = 'https://github.com/flutter/website/tree/master/src/_includes/code' -%}
+{% assign examples = 'https://github.com/flutter/website/tree/master/examples' -%}
+{% assign exFile = 'https://raw.githubusercontent.com/flutter/website/master/examples' -%}
 {% assign demo = 'https://github.com/flutter/flutter/blob/master/examples/flutter_gallery/lib/demo' -%}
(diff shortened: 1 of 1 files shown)


chalin:
Btw, just to be clear: we now have 6 fully functional apps (corresponding to each of the 6 steps) + tests. The code was added to the examples folder via #2163 (I added it first so that we could test links back into the source).

sfshaza2:
I know I sent you some comments via IM, but here's a summary:

* I love the automatic code diffs feature.
* I don't especially love the /*1*/ notation, which looks to me like a numbered footnote, but it isn't.
* You implemented the new coding feature on the first (big) example, but not on all the small examples later in the tutorial. Is that intentional? Or are you just putting it off until later?
* Now that we're updating this, we should migrate the (non-code) changes to the other version here:  https://codelabs.developers.google.com/codelabs/first-flutter-app-pt1/index.html?index=..%2F..index#0  (probably by me)
*  We've been asked to link to non-raw code files on GitHub, because the raw files suppress syntax highlighting. But we can do that in another PR, if you prefer.  :)

I don't see anything to prevent you from merging this!  LGTM

chalin:
> * I love the automatic code diffs feature.

\o/

> * I don't especially love the /*1*/ notation, which looks to me like a numbered footnote, but it isn't.

It is a numbered note commenting on a specific point in the code. I can change the appearance if you'd like but it would be nice if we could use the same notation as on www.

> * You implemented the new coding feature on the first (big) example, but not on all the small examples later in the tutorial. Is that intentional? Or are you just putting it off until later?

Yes.

> * Now that we're updating this, we should migrate the (non-code) changes to the other version here:  ...

We agreed over IM that this particular app doesn't have a corresponding codelab.

> *  We've been asked to link to non-raw code files on GitHub, because the raw files suppress syntax highlighting. But we can do that in another PR, if you prefer.  :)

Yes, that was only partly addressed here. It will be the subject of another PR.

PR #2211, opened 2019-01-15
CI: run `examples/**` app tests

chalin:
cc @sfshaza2 @filiph


examples/layout/lakes/step5/test/widget_test.dart
```diff
@@ -1,15 +1,24 @@
...
+import 'package:flutter/foundation.dart';
 import 'package:flutter_test/flutter_test.dart';
 
 import 'package:layout/main.dart';
 
 void main() {
   testWidgets('Codelab smoke test', (WidgetTester tester) async {
+    var exception;
+
+    FlutterError.onError = (FlutterErrorDetails details) async {
```
filiph: A comment here would be helpful, since this is not "normal" -- only codelabs will want to do this.

examples/layout/lakes/step5/test/widget_test.dart
```diff
@@ -1,15 +1,24 @@
...
+      exception = details.exception;
+    };
+
     await tester.pumpWidget(new MyApp());
     expect(find.text('Oeschinen Lake Campground'), findsOneWidget);
     expect(find.text('ROUTE'), findsOneWidget);
     // FIXME: the following fails
     // expect(find.text('Oeschinen'), findsOneWidget);
+
+    expect(exception, contains('RenderFlex overflowed by 16 pixels'));
```
filiph: There could potentially be an exception _before_ this one that we're not catching/expecting. A simple counter could solve this. But no biggie either way.

filiph:
Also, for the avoidance of confusion, I'd say we want to add a README.md file (or symlink?) to all the directories that currently (in the repo) don't contain all the necessary files. 

If I imagine myself landing on such a sample and there is just the few files that are left, I think something must be broken. My first instinct is _not_ to go up the directory tree to read READMEs (although, rationally, it should).

examples/layout/lakes/step5/test/widget_test.dart
chalin: Fixed

examples/layout/lakes/step5/test/widget_test.dart
chalin: @filiph, how about this:

```dart
  testWidgets('Codelab smoke test', (WidgetTester tester) async {
    // We don't usually expect the Flutter framework to throw a FlutterError
    // during a smoke test but exceptionally, the app under test here does. In
    // this particular case, the error is expected because we know that the app
    // represents an intermediate step towards the development of the final
    // version.
    var exceptions = [];
    FlutterError.onError = (FlutterErrorDetails details) async {
      exceptions.add(details.exception);
    };
   ...
}
```

chalin:
> Also, for the avoidance of confusion, I'd say we want to add a README.md file (or symlink?)

I created a master README at the top level of `examples` and symlinked to it from each of the app folders.

chalin:
@filiph - done with changes, PTAL

examples/layout/lakes/step5/test/widget_test.dart
kwalrath: Comment lgtm. It could be slightly simpler:

// A FlutterError shouldn't normally occur during a smoke test, but it
// is expected for this not-quite-finished app.

PR #2226, opened 2019-01-16
Don't use raw github repo images in examples

chalin:
Some code referred to an image in the repo (via github.com or raw.githubusercontent.com), but that image doesn't exist anymore. Instead, we're now using an image from https://picsum.photos.

cc @brianegan


chalin:
> I hope we don't have to redo screenshots. :)

Good point, but there are no images in those pages.

Issue #2233, opened 2019-01-17
Code excerpts: use new tooling and infra


chalin:
We currently have two mechanisms for dealing with code fragments: the old approach (using Jekyll include on individual fragments) and the new (where fragments are excerpt from fully tested apps). 

The Get Started codelab and the Building Layouts tutorial use the new mechanism. This issue will be used to track the updates of remaining pages.

- [x] Get Started codelab
- [x] Layouts tutorial
- [x] Layouts in Flutter 
- [ ] Animations
- [ ] i18n
- [ ] development/platform-integration/c-interop.md

chalin:
Done with all the Development > UI > Layout pages.

chalin:
With @sfshaza2's blessing, promoting to P1 to match the companion P1 #3076.

PR #2235, opened 2019-01-17
Rework of "Lay out a widget" + use of code-excerpts

chalin:
This PR has the reworked "Lay out a widget" section only. Note that I changed the numbered list to section headings. I did this because the numbered list just didn't work anymore: item 4 was a page and a half long. At that length readers lose the context of the numbered list.

Contributes to #2233 

https://flutter-io-staging-2.firebaseapp.com/docs/development/ui/layout#lay-out-a-widget


src/docs/development/ui/layout/index.md
```diff
@@ -72,170 +72,158 @@ the children should occupy.
 
 ## Lay out a widget
 
-{{site.alert.secondary}}
-  <h4 class="no_toc">What's the point?</h4>
```
chalin: @sfshaza2 - I dropped the "What's the point" alert for this section. Most points didn't seem that helpful, except for one -- concerning `Scaffold` -- which I moved into the main text because it wasn't even discussed there. WDYT?

sfshaza2: OK

chalin:
> I was a little surprised to see the copyright notice on the code removed.

We discussed this offline.

PR #2294, opened 2019-01-28
Layout row app + tests

chalin:
Contributes to #2233


chalin:
Btw, this is the code as is from `src/_includes/code/layout/row`. (Changes are coming in a followup PR.)

PR #2295, opened 2019-01-28
Layout row example - fix code excerpt

chalin:
Contributes to #2233 

Before (code inclusion used to be line base, but somehow the line numbers must have changed, since the excerpt is incomplete):

> screen shot 2019-01-28 at 12 22 45

After (using named code regions ensures that we dont' run into this problem again):

> screen shot 2019-01-28 at 12 22 56

cc @sfshaza2

### DIFF
--- a/examples/layout/row/lib/main.dart
+++ b/examples/layout/row/lib/main.dart
@@ -6,42 +6,27 @@ void main() {
   runApp(MyApp());
 }
 
-// #docregion MyApp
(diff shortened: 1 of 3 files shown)


chalin:
Btw, I simplified the code, changing it from a stateless `StatefulWidget` to an actual `StatelessWidget`.

chalin:
Note that the images don't correspond to the app screenshot, but I'm not going to worry about that for now.

PR #2301, opened 2019-01-28
Layout sizing example app

chalin:
This PR contains edits to the "Sizing widgets" section. The main edits to the prose are:

- Move the alert from the previous section (which talked about overflows), into the Sizing section.
- Then present the fix to the overflow example, which is to use `Expanded`.
- Finally show the "expanded" example that uses a flex factor of 2.

Contributes to #2233.

Staged at: https://flutter-io-staging-1.firebaseapp.com/docs/development/ui/layout#sizing-widgets

---

The updated section looks like this:
> screen shot 2019-01-28 at 16 33 07


src/docs/development/ui/layout/index.md
```diff
@@ -370,65 +368,83 @@ space evenly between, above, and below each image.
...
 ### Sizing widgets
 
-Perhaps you want a widget to occupy twice as much space as its siblings.
-You can place the child of a row or column in an
-[Expanded]({{api}}/widgets/Expanded-class.html)
-widget to control widget sizing along the main axis.
-The Expanded widget has a `flex` property, an integer that determines
-the flex factor for a widget. The default flex factor for an Expanded
-widget is 1.
+When a layout is too large to fit a device, a yellow and black striped pattern
```
sfshaza2: Wow, the yello and black striping is new!

PR #2310, opened 2019-01-30
Update Android editing docs

mklim:
Right now we don't explain how to actually open the Android files in a
Flutter app in Android Studio so that the IDE understands them and knows
how to edit them. (Sees the gradle scripts, can run automatic
refactoring, etc.)

These instructions are correct to the best of my knowledge and testing.
They've also been adapted from [the
flutter-intellij docs on editing Android code](https://github.com/flutter/flutter-intellij/blob/81b4b05c88a7daf912cc9ca9171e42650445cfa3/docs/android.md).


stevemessick:
Your instructions are usable but they do seem incomplete. Most people only see the Welcome screen when they start Android Studio for the first time.

If you already have Android Studio open on a Flutter app the easiest way to open a new window to edit the Android module is to select the module ("android" subdirectory) in the Project view, right click and choose: `Flutter > Open Android module in Android Studio`. If you are already looking at the source for a specific file and that file is selected in the Project view then the menu option will be: `Flutter > Open for editing in Android Studio`. Choosing it will open Android Studio with that file open in the editor.

Long-term, we'd prefer people not have to do anything special. To that end, we added a new option in the Flutter plugin preferences/settings. Under "Experiments" is an item with the cumbersome name "Enable code completion, navigation, etc. for Java/Kotlin". It does what it says, and eliminates the need to open another window just to edit Java code. You do have to restart the IDE to trigger indexing. The reason this is still experimental is that Android (i.e. Gradle) indexing can take a long time. People who are not working in the Android module don't need to be inconvenienced.

mklim:
> That looks good. I'd still like to see a reference to the Experimental option, but if you don't want to put that on the website I won't insist.

Thanks. I thought about it, but I'd rather not point people towards an experimental option yet on the official flutter.io docs. Especially when there's already two other options that are about as good. I think it would be confusing for people to get three suggestions at once and also have one of them be not totally stable yet. I'm happy to follow up on this and add it in as the suggested way once it's no longer experimental though.

sfshaza2:
I had a few grammatical edits . :)  But thanks for this PR!

Issue #2279, opened 2019-01-25
Reduce doubles to ints


xster:
Since Dart 2.1 supports double-to-ints.

For instance ` final _biggerFont = const TextStyle(fontSize: 18.0);` could be `final _biggerFont = const TextStyle(fontSize: 18);`

etc

chalin:
@Hixie @filiph @gspencergoog, any comment about this? /cc @sfshaza2

chalin:
@sfshaza2 @filiph et all: note that

1. I've applied this change to all example apps under the `examples` folder.
1. We **can't just do a global search-and-replace** for code snippets that are unchecked (those with `<-- skip -->` before them). There are situations where type inference is at play and using `int` arguments might require adding a `<double>` type argument.

Here is a concrete example of 2 that I encountered:

- Original code: `Tween(begin: 0.0, end: 300.0)`

- Just dropping `.0` leads to a type error:
  `Tween(begin: 0, end: 300)` --> is inferred as a `Tween<int>`, which is ok, but it wasn't what was expected in the context that the change was made.

- Fix (add `<double>` type argument): `Tween<double>(begin: 0, end: 300)`

atsansone:
@gspencergoog : What exactly would we need to do to resolve this issue? The request is vague. Thanks!

gspencergoog:
If the example code has the `.0` removed from most of the double args, I think this is resolved.  The places where a type would have to be added because the compiler can't infer the correct type (like the `Tween` example above), I think it's fine to just leave the `.0`, as that is enough to infer the right type and still is pretty readable. Although it might help with comprehension in some cases to be explicit about the types, adding `<double>` could also be too verbose for some short examples, so I'd probably just leave those to case by case judgement.

PR #2314, opened 2019-01-30
Layout example: Card and Stack

chalin:
Contributes to #2233, #2279


examples/layout/lakes/step5/lib/main.dart
```diff
@@ -105,11 +105,11 @@ class MyApp extends StatelessWidget {
       children: [
         Icon(icon, color: color),
         Container(
-          margin: const EdgeInsets.only(top: 8.0),
+          margin: const EdgeInsets.only(top: 8),
```
mshibanami: Hi @chalin, I think this code makes a compile error that the type is wrong. The same applies to other changes.

filiph: Hi @mshibanami, this should not give you a compile error anymore. Which version of Flutter are you using?

mshibanami: Oops, sorry. As you said, this problem was caused by the old version of flutter/dart. It's fixed after upgrading. Thanks @filiph!

PR #2320, opened 2019-01-31
Animation tutorial: create full example apps and use code excerpter

### LINKED ISSUE #2319: Animated­Widget example: prose doesn't match code
In https://flutter.io/docs/development/ui/animations/tutorial#animation-examples, the prose mentions:

> `LogoApp` now derives from `AnimatedWidget` instead of `StatefulWidget`.

That may have been the case at some point, but it isn't anymore. `LogoApp` still extends `StatefulWidget`. Instead a new class, `AnimatedLogo` is created that extends `AnimatedWidget`.

cc @sfshaza2 @filiph

chalin:
Fixes #2319
Contributes to #2233, #2279

Staged at https://flutter-io-staging-1.firebaseapp.com/docs/development/ui/animations/tutorial

(The indentation of code excerpts is off in a few cases. I'll file an issue over the code excerpter / code-excerpt renderer tool to address this.)


filiph:
Ouch, thanks for fixing this!

LGTM, although I did not fully review the code — let me know if you expect me to, and/or direct me to the important parts.

chalin:
There are no major (functional) changes to the code, so we should be good.

src/docs/development/ui/animations/tutorial.md
```diff
@@ -714,50 +705,44 @@ to opaque.
...
+```
 
 You can get the size with `sizeAnimation.value` and the opacity
 with `opacityAnimation.value`, but the constructor for `AnimatedWidget`
 only takes a single `Animation` object. To solve this problem,
-the example creates its own Tween objects and explicitly calculates the
+the example creates its own `Tween` objects and explicitly calculates the
 values.
 
-`LogoApp` widget was changed to encapsulate its own `Tween` objects.
```
chalin: Here is the part of the prose where it mismatches with the code. It's not the `LogoApp` that's impacted, its the `AnimatedLogo` class.

Issue #2321, opened 2019-01-31
'MacOS install' page issue: this missing detail would've helped me


romanmirov:
Page URL: https://flutter.io/docs/get-started/install/macos.html
Page source: https://github.com/flutter/website/tree/master/src/docs/get-started/install/macos.md

Description: Hi! Just some feedback. I had some trouble figuring out this error I got after following every instruction of this page up to step 4 of the section "Set Up the iOS Simulator." Basically, you guys should include how to properly create a flutter project. I just made a directory in my Dev folder where the flutter folder is located, so I have flutter and then projectname, if you will. What I needed to do was type `flutter create "projectname"`. So yeah. Save someone else 30 minutes of debugging by including that little bit. Thank you Flutter! 

PS: I resolved my issue after stumbling upon this question: [https://stackoverflow.com/questions/50375407/error-no-pubspec-yaml-file-found-when-running-flutter-for-the-first-time](https://stackoverflow.com/questions/50375407/error-no-pubspec-yaml-file-found-when-running-flutter-for-the-first-time)

sfshaza2:
Fixed in PR https://github.com/flutter/website/pull/2328

Closing

PR #2323, opened 2019-01-31
iOS hint: Create a flutter project before running flutter run.

kangwang1988:
iOS hint: Create a flutter project before running flutter run.

This will avoid users who meet errors when running `flutter run` without create a flutter project using `flutter create`

See: https://github.com/flutter/website/issues/2321


sfshaza2:
Thanks for this PR, @kangwang1988! This bug was *just* fixed. I'm closing, but reopen if we've missed anything.

Issue #2322, opened 2019-01-31
Android dev's ListView sample throws exception


anolivetree:
https://flutter.io/docs/get-started/flutter-for/android-devs

The first sample code of 'How do I update ListView’s dynamically?' throws exception.

```
I/flutter (23342): The following assertion was thrown building SampleAppPage(dirty, state: _SampleAppPageState#02255):
I/flutter (23342): type 'List<dynamic>' is not a subtype of type 'List<Widget>'
```

Replacing List widgets = []; with List&lt;Widget&gt; widgets = []; fixed the problem on my PC.

chalin:
Could be fixed in the context of #2233.

mjohnsullivan:
Fix added in https://github.com/flutter/website/pull/2324

PR #2324, opened 2019-01-31
Fixed typing of lists in sample code

mjohnsullivan:
Added list type info to two samples


mjohnsullivan:
Fixes https://github.com/flutter/website/issues/2322

Issue #2285, opened 2019-01-25
Links with target="_blank" should also have rel="noopener"


chalin:

sfshaza2:
I thought we were doing this now.

chalin:
I've noticed some places where it's missing.

mjohnsullivan:
Added rel="noopener" where missing

PR #2327, opened 2019-01-31
Added rel="noopener" where missing

mjohnsullivan:
Added rel="noopener" where missing from target="_blank"


mjohnsullivan:
Fixes https://github.com/flutter/website/issues/2285

PR #2330, opened 2019-01-31
Updated package:test to package_flutter_test

mjohnsullivan:
Changed test imports from package:test to package:flutter_test

### DIFF
--- a/src/docs/cookbook/testing/unit/introduction.md
+++ b/src/docs/cookbook/testing/unit/introduction.md
@@ -108,7 +108,7 @@ from the `test` package.
 <!-- skip -->
 ```dart
 // Import the test package and Counter class
-import 'package:test/test.dart';
+import 'package:flutter_test/flutter_test.dart';
 import 'package:counter_app/counter.dart';
 
 void main() {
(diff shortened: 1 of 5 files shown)


src/docs/cookbook/testing/unit/introduction.md
```diff
@@ -108,7 +108,7 @@ from the `test` package.
 <!-- skip -->
 ```dart
 // Import the test package and Counter class
-import 'package:test/test.dart';
+import 'package:flutter_test/flutter_test.dart';
```
brianegan: Ah, good catch! One Q: Instead of changing this line, should we encourage folks to import the `test` package directly? Since this is related to unit tests, I was hoping to keep it pure Dart!

src/docs/testing/index.md
```diff
@@ -69,7 +69,7 @@ Example:
 Add this file to `test/unit_test.dart`:
 
 {% prettify dart %}
-import 'package:test/test.dart';
+import 'package:flutter_test/flutter_test.dart';
```
brianegan: Overall, I think the current code is fine since we ask folks to import the test package. That said, I'm working on some updates to this page so it references the cookbook a bit more rather than having some duplicated content!

src/docs/cookbook/testing/unit/introduction.md
chalin: > this is related to unit tests, I was hoping to keep it pure Dart!

That was my feeling too.

sfshaza2:
Based on @brianegan's and @chalin's feedback, I'm closing this. But thanks, @mjohnsullivan!

PR #2332, opened 2019-02-01
State management docs update

filiph:
Makes the state management page into a full-blown section. The current link-page is just the last part of it. The first 3 parts are a gentle introduction to the problem and an explanation of a simple/starter solution.

Needs some additional love. Adding here as PR to let people to comment on lines.

### DIFF
--- a/src/docs/development/data-and-backend/state-mgmt/simple.md
+++ b/src/docs/development/data-and-backend/state-mgmt/simple.md
@@ -41,6 +41,7 @@ In Flutter, it makes sense to keep the state above the widgets that use it.
 
 Why? In declarative frameworks like Flutter, if you want to change the UI, you have to rebuild it. There is no easy way to have `MyCart.updateWith(somethingNew)`. In other words, it's hard to imperatively change a widget from outside, by calling a method on it. And even if you could make this work, you would be fighting the framework instead of letting it help you.
 
+<!-- skip -->
 ```dart
 // BAD: DO NOT DO THIS
 void onTapHandler() {
@@ -51,6 +52,7 @@ void onTapHandler() {
 
 Even if you get the above code to work, you will then have to deal with the following in the `MyCart` widget:
 
+<!-- skip -->
 ```dart
 // BAD: DO NOT DO THIS
 void build(BuildContext context) {
@@ -68,6 +70,7 @@ You would need to take into consideration the current state of the UI and apply 
 
 In Flutter, you construct a new widget every time its contents change. Instead of `MyCart.updateWith(somethingNew)` (a method call) you use `MyCart(contents)` (a constructor). Because you can only construct new widgets in the build methods of their parents, if you want to change `contents`, it needs to live in `MyCart`'s parent or above.
 
+<!-- skip -->
 ```dart
 // GOOD
 void onTapHandler() {
@@ -78,6 +81,7 @@ void onTapHandler() {
 
 Now `MyCart` has only one code path for building any version of the UI.
 
+<!-- skip -->
 ```dart
 // GOOD
 void build(BuildContext context) {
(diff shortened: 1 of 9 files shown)


src/docs/development/data-and-backend/state-mgmt/simple.md
```diff
@@ -0,0 +1,327 @@
...
+
+We could use `ScopedModelDescendant<CartModel>` for this, but that would be wasteful. We'd be asking the framework to rebuild a widget that doesn't need to be rebuilt. 
+
+For this use case, we can use `ScopedModel.of`. 
+
+```dart
+ScopedModel.of<CartModel>(context).add(item);
+```
+
+Using the above line in a build method will not lead to any rebuilding.
```
brianegan: I think this line could be misinterpreted -- it will lead to the connected ScopedModelDescendants to rebuild (some, some rebuilding is happening!), but it will not rebuild the *current* Widget that is accessing the scoped model in this way.

Maybe something like: "Using the above line in a build method will not cause this Widget to rebuild when `notifyListeners` is called"

src/docs/development/data-and-backend/state-mgmt/simple.md
```diff
@@ -0,0 +1,327 @@
...
+  title: Ephemeral versus app state
+  path: /docs/development/data-and-backend/state-mgmt/ephemeral-vs-app
+next:
+  title: List of approaches
+  path: /docs/development/data-and-backend/state-mgmt/options
+---
+
+Now that you know about [declarative UI programming](/docs/development/data-and-backend/state-mgmt/declarative) and the difference between [ephemeral and app state](/docs/development/data-and-backend/state-mgmt/ephemeral-vs-app), you are ready to learn about simple app state management.
+
+This page describes an approach particularly appropriate to those new to Flutter, without a pre-existing preference for a particular state management approach. Developers with strong background in state management from other reactive frameworks, such as Redux or Rx, will find packages and tutorials listed on the [following page](/docs/development/data-and-backend/state-mgmt/options).
```
RedBrogdon: If your intent with doc is (at least in part) to give people a "default" approach to state management via ScopedModel, consider being more explicit about it right up front:

* You're about to see us use a package called ScopedModel
* It uses Inherited Widgets and a Model object
* If you don't have a strong reason to choose Redux/Rx/BLoC, this is probably the approach you should start with.
  - Because it's easy to understand, but flexible and powerful
  - And it doesn't use much code

You've already got part of that in place (and cover the rest later, I think), but a little more detail right up front might help ground the reader before they progress.

src/docs/development/data-and-backend/state-mgmt/simple.md
filiph: ack

src/docs/development/data-and-backend/state-mgmt/simple.md
filiph: ack

src/docs/development/data-and-backend/state-mgmt/simple.md
```diff
@@ -0,0 +1,327 @@
...
+ScopedModelDescendant<CartModel>(
+  builder: (context, child, cart) {
+    return Text("Total price: ${cart.totalPrice}");
+  },
+);
+```
+
+We must specify the type of the model that we want to access. In this case, we want `CartModel`, so we write `ScopedModelDescendant<CartModel>`. If you don't specify the generic (`<CartModel>`), the `scoped_model` package won't be able to help you. As mentioned above, `scoped_model` is based on types, and without the type, it doesn't know what you want.
+
+The only required argument of the `ScopedModelDescendant` widget is the builder. Builder is a function that will be called whenever the model changes. (In other words, when you call `notifyListeners()` in your model, all the builder methods of all the corresponding `ScopedModelDescendant` widgets are called.)
```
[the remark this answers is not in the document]
filiph: ack

PR #2335, opened 2019-02-02
Integration testing: Add test dependency

ened:
The pubspec.yaml was missing the `test` dependency.


chalin:
Related #2330 (which should be merged first after the requested edits are made). cc @sfshaza2

chalin:
By the way, thanks for the PR and the followup comment @ened. I'll let @brianegan take it from here.

PR #2336, opened 2019-02-02
Fix missing import in intent example

sparky005:
Hello! It looks like there is a missing import in the java code on the [page describing how to receive an Android intent from an external application.](https://flutter.io/docs/get-started/flutter-for/android-devs#how-do-i-handle-incoming-intents-from-external-applications-in-flutter) This PR simply adds the missing import to the example code, making it a readily working example.


sparky005:
I signed it!

PR #2347, opened 2019-02-05
[cookbook] fixed equality assertion in example

monetant:
Code snippet typo fix.


brianegan:
LGTM, thanks for catching this!

sfshaza2:
@monetant, thanks for this PR! I'd love to accept it, but first you have to sign the CLA.

brianegan:
Ack, my bad -- this code looks funny, but it's actually correct. 

In this case, we only want to assign `isDebug` to `true` if we're in a development environment. Since `assert` statements ONLY run in Dev mode, `isDebug` will only be set to true if we're in the dev environment.

PR #2363, opened 2019-02-08
Trim VSCode install instructions in editor.md

timsneath:
Visual Studio Code no longer needs to be reloaded to activate extensions as of the January 2019 release (v1.31), so removing that step.


timsneath:
@patridge beat me to the punch with #2359, so merging his PR instead.

PR #2375, opened 2019-02-12
Rephrasing and replacing URLs with macros.

sfshaza2:
@mklim, please review. I've removed the absolute URLs with macros but I've also tightened up the text. Please make sure I haven't changed the meaning of anything.

### DIFF
--- a/src/docs/development/packages-and-plugins/androidx-compatibility.md
+++ b/src/docs/development/packages-and-plugins/androidx-compatibility.md
@@ -1,5 +1,6 @@
 ---
 title: AndroidX compatibility
+description: How to fix AndroidX incompatibilities that have been detected by the Flutter framework.
 ---
 
 {{site.alert.note}}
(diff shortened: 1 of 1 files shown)


src/docs/development/packages-and-plugins/androidx-compatibility.md
```diff
@@ -1,87 +1,85 @@
...
-stay backwards compatible. There's two conflicting sets of support libraries
-technically available today, but the
-original libraries are deprecated and no longer supported.
-[AndroidX](https://developer.android.com/jetpack/androidx/) is the only current
-version of the support libraries. It has feature parity with the originals and
-some additional capabilities.
+Android code often uses the
+[`android.support`]({{site.android-dev}}/topic/libraries/support-library/)
+libraries to ensure backwards compatibility. The `android.support`
+libraries were deprecated as of Android 9.0, and replaced with
```
mklim: This is true but the wording may come across as a little confusing. I think the libraries were deprecated at the time 9.0 was announced, but their deprecation is independent of Android SDK version. That is devs supporting Android 7.0 max could still use these libraries to add backwards compatibility all the way down to Android 4.0 (arbitrary example). And they would still be affected by this because it's the libraries themselves that are deprecated, no matter what Android SDK they're used on.

PR #2376, opened 2019-02-13
State Management Options Docs Update for Bloc Package

felangel:
@filiph I compiled a list of resources for the bloc package. Let me know what you think?

### DIFF
--- a/src/docs/development/data-and-backend/state-mgmt/options.md
+++ b/src/docs/development/data-and-backend/state-mgmt/options.md
@@ -50,8 +50,8 @@ Learn more at the following links, many of which have been contributed by the Fl
 
 ## BLoC / Rx
 
-* [Bloc Library](https://felangel.github.io/bloc), by Felix Angelov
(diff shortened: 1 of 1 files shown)


filiph:
Hi @felangel, thanks so much for the pull request!

To keep the page relatively balanced, can you select at most 2 new bullet points and add only them? Alternatively, you can delete one Bloc bullet point and replace it with another.

felangel:
@filiph do you prefer one over the other? I feel bad about removing an existing bullet point haha

filiph:
Ah, this is much better, thanks! 

I think that you can keep Saowen's article since you're only adding one link. Also, I'd like to start with conceptual articles (explain the pattern itself, without concrete implementation).

How about:

* Saowen's article
* Sagar's article
* Your Bloc page

This means removing the link to reddit (which I like, but it's probably the least "lasting" url in the set).

felangel:
@filiph just applied your feedback, thanks!

PR #2407, opened 2019-02-20
Update deploy script to publish to flutter.dev

kf6gpe:
Change YAML macro to point to api.flutter.dev as part of the domain move.


kf6gpe:
Fix with input from @filiph --- this should correctly redirect for the new YouTube channel and change the domain. Thanks!

PR #2453, opened 2019-02-28
Fix Flutter Create Byte Count Command Code

rnitta:
`$ find . -name "*.dart" | xargs cat | wc -c` is correct.
But now, a trailling period is included in `<code>`.
It will be confusing, so I fixed it.


rnitta:
signed.

PR #2469, opened 2019-03-04
Recipe: Play and Pause video

brianegan:
Adds a recipe demonstrating how to play and pause video using the video_player plugin!

Note: This recipe builds on the Camera recipe and that code should be ignored (had to do this because this recipe depends on the `plugin` directory existing). 

I have a couple more recipes ready to go that also add to the plugin directory, but want to hold off on submitting all of them at once so we can review them one at a time!


brianegan:
> I see some minor things editorial things that I'm not going to worry about for now, like an overuse of "will". ;) LGTM

Haha, noooooooo! I'll submit a PR tomorrow AM fixing those up.

PR #2496, opened 2019-03-11
Recipe: Navigate with arguments. Fixes #2296.

### LINKED ISSUE #2296: Cookbook should have example about passing arguments to named route
Once https://github.com/flutter/flutter/pull/27058 lands, we should have an example in https://flutter.io/docs/cookbook#navigation for pushing arguments to a named route.

It should show:
* pushing a route with arguments (https://github.com/flutter/flutter/pull/27058/files#diff-b4db74ec348d48e72da369c2bc858c7bR266)
* retrieving arguments from route settings in `MaterialApp`'s onGenerateRoute (https://github.com/flutter/flutter/pull/27058/files#diff-3dce29bc9fe4645a415193a989325ce9R88)
* retrieving them via `ModalRoute.of(context).settings.arguments` from within a route.

/cc @brianegan

brianegan:
Provides a recipe on navigating with arguments. Overall, this one was a little bit of a struggle for me!

I'd split this into two recipes, then combined it back into one. I'm not sure which is best, so feedback on the format in addition to the normal grammar mistakes is most welcome :)

@goderbauer -- I'd particularly like your input on this one. I have to say, from a user standpoint, the API is getting a bit weird. You now have two ways to pass arguments to a non-named route: Either directly or via the arguments. I couldn't really think of a reason why you'd use the arguments, honestly, but I've included it here. However, I'm not really sure if we should encourage folks to navigate in that way?

For me, the API totally makes sense for named routes, especially for tools that want to convert strings into "arguments," e.g. `/user/{id}` could be parsed in `onGenerateRoute` and the `id` could be passed as an argument. That is really handy for extracting information from deeplinks.

Anyhow, would love your feedback on this one as well to let me know if I'm misunderstanding the new api!


src/docs/cookbook/navigation/pass-arguments.md
```diff
@@ -0,0 +1,401 @@
...
+    // When the user taps the button, navigate to the specific route
+    // and provide the arguments as part of the RouteSettings.
+    Navigator.push(
+      context,
+      MaterialPageRoute(
+        builder: (context) => ExtractArgumentsScreen(),
+        // Pass the arguments as part of the RouteSettings. The
+        // ExtractArgumentScreen reads the arguments from these
+        // settings.
+        settings: RouteSettings(
```
goderbauer: I think I would have the cookbook focus on pushing named routes with arguments. That's what the API was meant for. The example shown here where a non-named route is pushed doesn't really benefit from the new API. Instead of using route setting in this case the arguments should just be passed directly to ExtractArgumentsScreen().

My suggestion would be to either remove the non-named example completely and center this cook book entry around named routes - or move the non-named case to the bottom of the entry and refactor it to pass the arguments directly to the route bypassing the new API.

brianegan: > The example shown here where a non-named route is pushed doesn't really benefit from the new API. Instead of using route setting in this case the arguments should just be passed directly to ExtractArgumentsScreen().

Gotcha. That was my feeling as well, but this API opened up the possibility of passing arguments in this new way and wasn't sure if you wanted that documented. IMO, it might be a good idea to document your advice in the `arguments` docs as well: https://docs.flutter.io/flutter/widgets/RouteSettings/arguments.html

  > My suggestion would be to either remove the non-named example completely and center this cook book entry around named routes

Sounds good. I'll figure out how to split this recipe up. It now feels like there are two distinct paths:

  1. Handle the route using the `routes` map and extract the arguments using `ModalRoute`
  2. Handle the route and extract arguments using `onGenerateRoute`

I might split these up into two separate recipes for clarity.

brianegan:
Thanks for the feedback, @goderbauer! 

To keep the recipe focused, I've given clear instructions on how to use `ModalRoute.of` to extract the arguments by navigating to a named route. I've removed any mention of non-named routes, which actually let me clean up the `onGenerateRoute` section quite a bit. It's now mentioned as an alternative and all of the supporting code can be seen in the "complete example."

Please let me know whatcha think!

src/docs/cookbook/navigation/navigate-with-arguments.md
```diff
@@ -0,0 +1,332 @@
+---
+title: Navigate to a named route with arguments
+prev:
+  title: Animating a Widget across screens
+  path: /docs/cookbook/navigation/hero-animations
+next:
+  title: Navigate to a new screen and back
+  path: /docs/cookbook/navigation/navigation-basics
```
goderbauer: About ordering: It feels like "navigate with arguments" is more specific than "navigation basics". Shouldn't the basics come before this new entry?

sfshaza2:
@brianegan, @goderbauer suggested a slight reordering. Do you want to do that or do you want it merged?

src/docs/cookbook/navigation/navigate-with-arguments.md
brianegan: Oh the curse of alphabetical ordering! 

I'll try to come up with a different title that makes sense and is alphabetically below "Navigation Basics"

brianegan: Renamed to "Pass arguments to a named route," which comes directly after "Navigate with named routes."

I'm not sure how long this balancing act will work, haha, but the problem is fixed for this PR.

brianegan:
@Sfshaza Renamed! I think it's better this way. Happy to merge unless there's more feedback :)

PR #2505, opened 2019-03-12
Correcting Swift error in platform-channel.md

Kiboom:
Correct example code in bottom of the section [**'Step 4b: Add an iOS platform-specific implementation using Swift'**](https://flutter.dev/docs/development/platform-integration/platform-channels#step-4b-add-an-ios-platform-specific-implementation-using-swift), to prevent this error on build: 

> Chain the optional using '?' to access member 'receiveBatteryLevel' only for non-'nil' base values 

This affects both simulated and physical iOS devices in Swift-based apps.

### DIFF
--- a/src/docs/development/platform-integration/platform-channels.md
+++ b/src/docs/development/platform-integration/platform-channels.md
@@ -591,7 +591,7 @@ batteryChannel.setMethodCallHandler({
     result(FlutterMethodNotImplemented)
     return
   }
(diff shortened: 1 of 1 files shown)


sfshaza2:
This looks good to me, @xster. Can you verify? (It's a one character change in the Swift code.)

xster:
LGTM. Seems reasonable.

sfshaza2:
Thanks, @Kiboom!

Issue #2518, opened 2019-03-14
'Pass arguments to a named route' page issue


Twinber:
Page URL: https://flutter.dev/docs/cookbook/navigation/navigate-with-arguments.html
Page source: https://github.com/flutter/website/tree/master/src/docs/cookbook/navigation/navigate-with-arguments.md

Found a typo? You can fix it yourself by going to the page source and clicking the pencil icon. Or finish creating this issue.

Description of issue:

// If you push the the PassArguments route ¿?

sfshaza2:
@brianegan, just remove the double "the the" . :)

rubensdemelo:
@sfshaza2 done.

PR #2520, opened 2019-03-14
Fix 2518: Remove "the the" from comments

rubensdemelo:
Remove "the the" from comments

Fix https://github.com/flutter/website/issues/2518


brianegan:
Good catch, thanks so much!

PR #2527, opened 2019-03-15
Moving SDK archive page and renaming to releases.

sfshaza2:
Staged:
https://sz-flutter.firebaseapp.com/releases
https://sz-flutter.firebaseapp.com/docs/development/tools/sdk/archive (redirects)

We still have to iron out a couple issues. Currently in an email thread.


sfshaza2:
This plan got some pushback, so we're taking another approach. Closing.

PR #2547, opened 2019-03-26
Add `FlutterDebugger` to sidenav

lijy91:


lijy91:
I signed it!

sfshaza2:
Hi, @lijy91! Thanks for this PR. We keep a tight limit on the side navigation menu, so I can't accept this PR. But, what about adding your debugging tool info to the /docs/testing/debugging page? Maybe add a section for third party debugging. This approach will scale as other tools come online.

lijy91:
@sfshaza2 Can you add this part?

sfshaza2:
Sure, @lijy91, but I can't get to it immediately.

lijy91:
@sfshaza2 If you have any news, please let me know.

PR #2552, opened 2019-03-28
The example result for i ~/ 2 is wrong

vshy108:


vshy108:
misunderstand the operator is modulo.

PR #2557, opened 2019-03-29
Add instructions for pre-downloading binaries

cbracken:
We recently changed Flutter's behaviour such that binary artifacts are
downloaded on an as-needed basis. While this is an improvement for many
users, in cases such as hermetic build environments with no network
access, or in cases where a user expects to have
intermittent/slow/expensive internet access, provide instructions to
optionally pre-download binaries.


Hixie:
consider "availability" or "reliability" instead of "accessibility".

LGTM

cbracken:
Good call. Done.

PR #2559, opened 2019-03-30
changed the order

sohilww:
after I read the navigate to a new screen and back, below of doc the next tutorial is "Navigate with named routes" 
because of that, I changed the order


sohilww:
> I signed it!

PR #2584, opened 2019-04-10
Add CupertinoScrollbar to the catalog

xster:


src/_data/catalog/widgets.json
```diff
@@ -586,6 +586,16 @@
     "link": "https://docs.flutter.io/flutter/cupertino/CupertinoPopupSurface-class.html",
     "image": "<img alt='' src='/images/catalog-widget-placeholder.png'>"
   },
+  {
+    "name": "CupertinoScrollbar",
+    "description": "An iOS-style scrollbar that appears while scrolling lists",
```
RedBrogdon: What about "An iOS-style scrollbar that indicates which portion of a Scrollable widget is currently visible?" That's more in line with the API docs and avoids implying that the scrollbar is just for use with lists.

xster: yup, your wording is better. Changed.

src/_data/catalog/widgets.json
```diff
@@ -586,6 +586,16 @@
...
     "image": "<img alt='' src='/images/catalog-widget-placeholder.png'>"
   },
+  {
+    "name": "CupertinoScrollbar",
+    "description": "An iOS-style scrollbar that indicates which portion of a scrollable widget is currently visible.",
+    "categories": [
+      "Cupertino (iOS-style widgets)"
+    ],
+    "subcategories": [],
+    "link": "https://docs.flutter.io/flutter/cupertino/CupertinoScrollbar-class.html",
```
sfshaza2: Please use api.flutter.dev

xster: replaced everywhere in the file

PR #2597, opened 2019-04-17
Prevent new Flutter versions from breaking the build

filiph:
This allows a little bit more slack for the flutter tests and their dependencies.


xster:
LGTM

PR #2600, opened 2019-04-18
Update the i18n tutorial, 52 languages supported!

HansMuller:
A few small updates to the i18n tutorial, mostly to highlight the fact that we now support 52 languages.

Added a reference to the Chinese locale ("zh") in two examples because https://github.com/flutter/flutter/issues/27134. Removed the country codes from the same examples to keep things simple.


mjohnsullivan:
LGTM

PR #2602, opened 2019-04-19
New internationalization tutorial example: add_language

HansMuller:
This will be referred to by a new [i18n tutorial section](https://github.com/flutter/website/blob/master/src/docs/development/accessibility-and-localization/internationalization.md) called "Adding support for new languages".


kwalrath:
@mjohnsullivan LGTM if it LGTY.

mjohnsullivan:
LGTM with a suggestion that a couple of comments are added to the code to let the user know that the app fails if the requested translations are not included.

mjohnsullivan:
LGTM

PR #2603, opened 2019-04-19
Added a section to the i18n tutorial

HansMuller:
Added a section to the i18n tutorial: Adding support for a new language.

This material was based on https://github.com/flutter/flutter/issues/15439#issuecomment-480462059

### DIFF
--- a/src/docs/development/accessibility-and-localization/internationalization.md
+++ b/src/docs/development/accessibility-and-localization/internationalization.md
@@ -400,7 +400,7 @@ as factory for the GlobalMaterialLocalizations subclass,
 must also be defined.
 
 Here's [the source code for a complete example](
-{{site.github}}/flutter/website/tree/master/examples/internationalization/add_language), 
+{{site.github}}/flutter/website/tree/master/examples/internationalization/add_language/lib/main.dart), 
 less the actual Belarusan translations, of an app that includes support for a new language.
 
 The locale-specific GlobalMaterialLocalizations subclass is called
@@ -432,7 +432,7 @@ Belarusan string.
 
 The getters return "raw" Dart strings that have an r prefix, like
 `r'About $applicationName'`, because sometimes the strings contain
-`variables` with a `$` prefix. The variables are expanded by parameterized 
+variables with a `$` prefix. The variables are expanded by parameterized 
 localization methods: 
 {% prettify dart %}
 @override
@@ -449,9 +449,12 @@ For more information about localization strings, see the
 [flutter_localizations README](
 {{site.github}}/flutter/flutter/blob/master/packages/flutter_localizations/lib/src/l10n/README.md).
 
-Finally, add Belarusan to the app's `supportedLocales` list, and
-add the `BeMaterialLocalizations` delegate instance to the app's
-`localizationsDelegates` list:
+Once you've implemented your language-specific subclasses of 
+GlobalMaterialLocalizations and LocalizationsDelegate, you just 
+need to add the language and a delegate instance to your app. 
+Here's some code that sets the app's language to Belarusan and 
+adds theBeMaterialLocalizations delegate instance to the app's
+localizationsDelegates list:
 
 {% prettify dart %}
 MaterialApp(
--- a/src/docs/development/accessibility-and-localization/internationalization.md
+++ b/src/docs/development/accessibility-and-localization/internationalization.md
@@ -382,6 +382,91 @@ class DemoLocalizationsDelegate extends LocalizationsDelegate<DemoLocalizations>
 }
 {% endprettify %}
 
+<a name="adding-language"></a>
+## Adding support for a new language
+
+An app that needs to support a language that's not included in
+[GlobalMaterialLocalizations]({{site.api}}/flutter/flutter_localizations/GlobalMaterialLocalizations-class.html)
+will have to do some extra work. It's not difficult, however to
+do the complete job one must provide about 70 translations
+("localizations") for words or phrases.
+
+As an example, we'll show how to add support for the Belarusan
+language.
+
+A new GlobalMaterialLocalizations subclass will define the
+localizations that the Material library depends on. Additionally one
+must create a LocalizationsDelegate subclass that essentially just
+constructs an instance of the new GlobalMaterialLocalizations
+subclass.
+
+There's a complete example, less actual Belarusan translations, here:
+LINK for https://github.com/flutter/website/pull/2602 TBD.
+
+The locale-specific GlobalMaterialLocalizations subclass is called
+`BeMaterialLocalizations` and the LocalizationsDelegate subclass is
+`_BeMaterialLocalizationsDelegate`. The value of
+`BeMaterialLocalizations.delegate` is an instance of the delegate, and
+it's all that's needed by an app that uses these localizations.
+
+The delegate class includes basic date and number format
+localizations. All of the other localizations are defined by String
+valued property getters in BeMaterialLocalizations. Like this:
+
+{% prettify dart %}
+@override
+String get backButtonTooltip => r'Back';
+
+@override
+String get cancelButtonLabel => r'CANCEL';
+
+@override
+String get closeButtonLabel => r'CLOSE';
+
+// etc..
+{% endprettify %}
+
+These are the English translations of course. To complete the job one must change the return value of each getter to an appropriate Belarusan string.
+
+The getters return "raw" Dart strings that have an r prefix, like
+`r'About $applicationName'`, because sometimes the strings contain
+`$variables` that are expanded by methods with parameters, like
+`aboutListTileTitle(String applicationName)`:
+{% prettify dart %}
+@override
+String get aboutListTileTitleRaw => r'About $applicationName';
+
+@override
+String aboutListTileTitle(String applicationName) {
+  final String text = aboutListTileTitleRaw;
+  return text.replaceFirst(r'$applicationName', applicationName);
+}
+{% endprettify %}
+
+More information about the localization strings can be found in the
+flutter_localizations package's
+[README]({{site.github}}/flutter/flutter/blob/master/packages/flutter_localizations/lib/src/l10n/README.md)
+file.
+
+The example app adds support for Belarusan by including it in the
+app's `supportedLocales` list and by adding the new
+`BeMaterialLocalizations` delegate to the app's
+`localizationsDelegetes` list:
+
+{% prettify dart %}
+MaterialApp(
+  localizationsDelegates: [
+    GlobalWidgetsLocalizations.delegate,
+    GlobalMaterialLocalizations.delegate,
+    BeMaterialLocalizations.delegate,
+  ],
+  supportedLocales: [
+    const Locale('be', 'BY')
+  ],
+  home: ...
+)
+{% endprettify %}
+
 <a name="dart-tools"></a>
 ## Appendix: Using the Dart intl tools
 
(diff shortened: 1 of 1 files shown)


kwalrath:
I'll stage this if you haven't already, @mjohnsullivan.

kwalrath:
Ah, I see this depends on #2602. I'll wait until Hans fixes the TBD.

HansMuller:
@kwalrath I've added a link to the example that landed in #2602. This PR should be ready to go.

kwalrath:
Staged: https://kw-flutter-1.firebaseapp.com/docs/development/accessibility-and-localization/internationalization#adding-language

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -382,6 +382,92 @@ class DemoLocalizationsDelegate extends LocalizationsDelegate<DemoLocalizations>
 }
 {% endprettify %}
 
+<a name="adding-language"></a>
+## Adding support for a new language
+
+An app that needs to support a language that's not included in
+[GlobalMaterialLocalizations]({{site.api}}/flutter/flutter_localizations/GlobalMaterialLocalizations-class.html)
+will have to do some extra work. It's not difficult, however to
+do the complete job one must provide about 70 translations
```
kwalrath: We don't use "one" to refer to a person. Instead, we generally use "you" (unless something else is more appropriate). Also, we avoid unnecessary future tense ("will").

How about (modify these words as needed to be clear and fit in with the rest of the doc):

An app...has to do some extra work: it must provide about 70 ...

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -382,6 +382,92 @@ class DemoLocalizationsDelegate extends LocalizationsDelegate<DemoLocalizations>
...
+("localizations") for words or phrases.
+
+As an example, we'll show how to add support for the Belarusan
+language.
+
+A new GlobalMaterialLocalizations subclass will define the
+localizations that the Material library depends on. Additionally one
```
kwalrath: one -> you
OR
Additionally, a new ...subclass creates an instance of...

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -382,6 +382,92 @@ class DemoLocalizationsDelegate extends LocalizationsDelegate<DemoLocalizations>
...
+An app that needs to support a language that's not included in
+[GlobalMaterialLocalizations]({{site.api}}/flutter/flutter_localizations/GlobalMaterialLocalizations-class.html)
+will have to do some extra work. It's not difficult, however to
+do the complete job one must provide about 70 translations
+("localizations") for words or phrases.
+
+As an example, we'll show how to add support for the Belarusan
+language.
+
+A new GlobalMaterialLocalizations subclass will define the
```
kwalrath: will define -> defines

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -382,6 +382,92 @@ class DemoLocalizationsDelegate extends LocalizationsDelegate<DemoLocalizations>
...
+must create a LocalizationsDelegate subclass that essentially just
+constructs an instance of the new GlobalMaterialLocalizations
+subclass.
+
+Here's [the source code for a complete example](
+{{site.github}}/flutter/website/tree/master/examples/internationalization/add_language), 
+less the actual Belarusan translations, of an app that includes support for a new language.
+
+The locale-specific GlobalMaterialLocalizations subclass is called
+`BeMaterialLocalizations` and the LocalizationsDelegate subclass is
```
kwalrath: and the -> , and the
OR
; the

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -382,6 +382,92 @@ class DemoLocalizationsDelegate extends LocalizationsDelegate<DemoLocalizations>
...
+`_BeMaterialLocalizationsDelegate`. The value of
+`BeMaterialLocalizations.delegate` is an instance of the delegate, and
+it's all that's needed by an app that uses these localizations.
+
+The delegate class includes basic date and number format
+localizations. All of the other localizations are defined by String
+valued property getters in BeMaterialLocalizations. Like this:
```
kwalrath: . Like -> , like
(to avoid the sentence fragment)

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -382,6 +382,92 @@ class DemoLocalizationsDelegate extends LocalizationsDelegate<DemoLocalizations>
...
+@override
+String get cancelButtonLabel => r'CANCEL';
+
+@override
+String get closeButtonLabel => r'CLOSE';
+
+// etc..
+{% endprettify %}
+
+These are the English translations of course. To complete the job one must change the return value of each getter to an appropriate Belarusan string.
```
kwalrath: one must -> , you need to 

(also consider reducing the line length)

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -382,6 +382,92 @@ class DemoLocalizationsDelegate extends LocalizationsDelegate<DemoLocalizations>
...
+
+The getters return "raw" Dart strings that have an r prefix, like
+`r'About $applicationName'`, because sometimes the strings contain
+`$variables` that are expanded by methods with parameters, like
```
kwalrath: `$variables` seems questionable, since it's literal and yet not. Maybe this (which you could probably improve):

variables with a `$` prefixes. Methods replace these variables with parameter values,
as in the following example:

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -382,6 +382,92 @@ class DemoLocalizationsDelegate extends LocalizationsDelegate<DemoLocalizations>
...
+String get aboutListTileTitleRaw => r'About $applicationName';
+
+@override
+String aboutListTileTitle(String applicationName) {
+  final String text = aboutListTileTitleRaw;
+  return text.replaceFirst(r'$applicationName', applicationName);
+}
+{% endprettify %}
+
+More information about the localization strings can be found in the
```
kwalrath: Let's [switch this around](https://developers.google.com/style/clause-order) and make the link text even clearer:

For more information about location strings, see the [flutter_localizations README](...).

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -382,6 +382,92 @@ class DemoLocalizationsDelegate extends LocalizationsDelegate<DemoLocalizations>
...
+flutter_localizations package's
+[README]({{site.github}}/flutter/flutter/blob/master/packages/flutter_localizations/lib/src/l10n/README.md)
+file.
+
+The example app adds support for Belarusan by including it in the
+app's `supportedLocales` list and by adding the new
+`BeMaterialLocalizations` delegate to the app's
+`localizationsDelegetes` list:
```
kwalrath: getes -> gates

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -382,6 +382,92 @@ class DemoLocalizationsDelegate extends LocalizationsDelegate<DemoLocalizations>
...
+  return text.replaceFirst(r'$applicationName', applicationName);
+}
+{% endprettify %}
+
+More information about the localization strings can be found in the
+flutter_localizations package's
+[README]({{site.github}}/flutter/flutter/blob/master/packages/flutter_localizations/lib/src/l10n/README.md)
+file.
+
+The example app adds support for Belarusan by including it in the
```
kwalrath: Let's generalize this. Maybe:

Finally, add the language to the app's ..., and add the language's delegate instance to...

HansMuller:
@kwalrath Thanks for all the feedback. I think I've got them all.

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -382,6 +382,91 @@ class DemoLocalizationsDelegate extends LocalizationsDelegate<DemoLocalizations>
...
+// etc..
+{% endprettify %}
+
+These are the English translations of course. To complete the job you 
+need to change the return value of each getter to an appropriate 
+Belarusan string.
+
+The getters return "raw" Dart strings that have an r prefix, like
+`r'About $applicationName'`, because sometimes the strings contain
+`variables` with a `$` prefix. The variables are expanded by parameterized
```
kwalrath: `variables` -> variables
(no special font, since you don't actually see the string `variables` in the code, right?)

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -382,6 +382,91 @@ class DemoLocalizationsDelegate extends LocalizationsDelegate<DemoLocalizations>
...
+  final String text = aboutListTileTitleRaw;
+  return text.replaceFirst(r'$applicationName', applicationName);
+}
+{% endprettify %}
+
+For more information about localization strings, see the 
+[flutter_localizations README](
+{{site.github}}/flutter/flutter/blob/master/packages/flutter_localizations/lib/src/l10n/README.md).
+
+Finally, add Belarusan to the app's `supportedLocales` list, and
```
kwalrath: This could still be more general. Also, I wouldn't say it _adds_ Belarusan, because that appears to be the only supported language. So... how about something like this:

Once you've implemented your language-specific subclasses of GlobalMaterialLocalizations and LocalizationsDelegate, you just need to add the language and a delegate instance to your app. Here's some code that sets the app's language to Belarusan and adds the`BeMaterialLocalizations` delegate instance to the app's`localizationsDelegates` list:

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -382,6 +382,91 @@ class DemoLocalizationsDelegate extends LocalizationsDelegate<DemoLocalizations>
...
+language.
+
+A new GlobalMaterialLocalizations subclass defines the
+localizations that the Material library depends on.
+A new LocalizationsDelegate subclass, which serves
+as factory for the GlobalMaterialLocalizations subclass, 
+must also be defined.
+
+Here's [the source code for a complete example](
+{{site.github}}/flutter/website/tree/master/examples/internationalization/add_language),
```
kwalrath: How about linking directly to the lib/main.dart file, since it seems to be the only file there?

src/docs/development/accessibility-and-localization/internationalization.md
HansMuller: Yes, good point.

src/docs/development/accessibility-and-localization/internationalization.md
HansMuller: Sounds good, I substituted your version.

src/docs/development/accessibility-and-localization/internationalization.md
HansMuller: Done.

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -449,9 +449,12 @@ For more information about localization strings, see the
...
 {{site.github}}/flutter/flutter/blob/master/packages/flutter_localizations/lib/src/l10n/README.md).
 
-Finally, add Belarusan to the app's `supportedLocales` list, and
-add the `BeMaterialLocalizations` delegate instance to the app's
-`localizationsDelegates` list:
+Once you've implemented your language-specific subclasses of 
+GlobalMaterialLocalizations and LocalizationsDelegate, you just 
+need to add the language and a delegate instance to your app. 
+Here's some code that sets the app's language to Belarusan and 
+adds theBeMaterialLocalizations delegate instance to the app's
```
kwalrath: theBe -> the Be

sfshaza2:
@HansMuller, Eric Windmill's article on InheritedWidget has disappeared, so the linkcheck fails, so the build fails. Given that this PR has nothing to do with that, I will merge it manually and fix the link in another PR.

PR #2604, opened 2019-04-19
Update sample with more nuanced locale definitions

GaryQian:


GaryQian:
Abandoned in favor of a separate section to explain nuances of locale definition

PR #2614, opened 2019-04-25
Added link to Rollbar

istvanerdo:
I would really appreciate if our brand mention had a link back to our website so I've added it in the PR. 

Thanks in advance!


istvanerdo:
I signed it!

PR #2615, opened 2019-04-25
Fix research sign-up link in CONTRIBUTING.md

InMatrix:
The current relative path goes to a 404 page.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,7 +46,7 @@ Other ways you can contribute:
 * [Submit PRs to flutter.dev][doc-PRs]
 * [Follow us on Twitter: @flutterdev](https://twitter.com/flutterdev/)
 * [Read the Flutter Publication on Medium](https://medium.com/flutter-io)
-* [Sign up to Future UX Studies on Flutter](/research-signup)
+* [Sign up to Future UX Studies on Flutter](https://flutter.dev/research-signup)
 * [Join the subreddit to keep up with the latest in the Flutter
    community][reddit]
 


CONTRIBUTING.md
```diff
@@ -46,7 +46,7 @@ Other ways you can contribute:
 * [Submit PRs to flutter.dev][doc-PRs]
 * [Follow us on Twitter: @flutterdev](https://twitter.com/flutterdev/)
 * [Read the Flutter Publication on Medium](https://medium.com/flutter-io)
-* [Sign up to Future UX Studies on Flutter](/research-signup)
+* [Sign up to Future UX Studies on Flutter](https://flutter.dev/research-signup)
```
sfshaza2: Tao, was the link not working before?

At any rate, we avoid hard coded links like this. You can use:

```
({{site.url}}/research-signup)
```

But... woah! I just searched for research-signup on the site and can't find it:

```
website [master] $ find . -name "*signup*"
website [master] $ find . -name "*research*"
website [master] $ 
```

InMatrix: The link doesn't work when you click on it from https://github.com/flutter/website/blob/master/CONTRIBUTING.md

It's normal that you couldn't find the page. It's not actually a page, but a URL redirect I set up some time ago. It goes to a Google Form.

InMatrix: Is CONTRIBUTING.md getting converted to a Webpage on flutter.dev or it's a doc meant to be viewed as Markdown on Github?

sfshaza2: You are correct. At first, I wanted it to be a visible page, so I put it under src. Then I received feedback that it shouldn't be visible, so I moved it up one level. But you're right, that means it won't be converted.

So, I remove all objections.  :D

LGTM

PR #2624, opened 2019-05-01
Fix documentation on JavaScript to Dart comparison

jpaquim:
Remove unnecessary arrow syntax on the Promise-based `_getIPAddress` function.
Fix formatting on `url` variable declaration.
Remove unnecessary `await` in accessing object field in javascript.
Rename `json` to `responseJson` to better match the previous Promise-based example, since they have the same semantics.
Rename `data` to `ip` to better match the Dart example.

### DIFF
--- a/src/docs/get-started/flutter-for/react-native-devs.md
+++ b/src/docs/get-started/flutter-for/react-native-devs.md
@@ -261,10 +261,14 @@ used to wait for a `Promise`.
 class Example {
   async _getIPAddress() {
     const url = "https://httpbin.org/ip";
-    const response = await fetch(url);
-    const responseJson = await response.json();
-    const ip = responseJson.origin;
-    console.log(ip);
+    try {
+      const response = await fetch(url);
+      const responseJson = await response.json();
+      const ip = responseJson.origin;
+      console.log(ip);
+    } catch (error) {
+      console.error(error);
+    }
   }
 }
 
@@ -285,9 +289,13 @@ import 'package:http/http.dart' as http;
 class Example {
   _getIPAddress() async {
     final url = 'https://httpbin.org/ip';
-    final response = await http.get(url);
-    String ip = json.decode(response.body)['origin'];
-    print(ip);
+    try {
+      final response = await http.get(url);
+      String ip = json.decode(response.body)['origin'];
+      print(ip);
+    } catch (error) {
+      print(error);
+    }
   }
 }
 
--- a/src/docs/get-started/flutter-for/react-native-devs.md
+++ b/src/docs/get-started/flutter-for/react-native-devs.md
@@ -212,17 +212,18 @@ class Example {
     return fetch(url)
       .then(response => response.json())
       .then(responseJson => {
-        console.log(responseJson.origin);
+        const ip = responseJson.origin;
+        return ip;
       })
       .catch(error => {
         console.error(error);
       });
-  };
+  }
 }
 
 function main() {
   const example = new Example();
-  example._getIPAddress();
+  example._getIPAddress().then(ip => console.log(ip));
 }
 
 main();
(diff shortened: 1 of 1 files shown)


jpaquim:
I just signed the CLA, please give this a look.

src/docs/get-started/flutter-for/react-native-devs.md
```diff
@@ -246,11 +246,11 @@ used to wait for a `Promise`.
 ```js
 // JavaScript
 async _getIPAddress() {
```
cymarechal: Isn't it invalid JavaScript ? I would rather use `async function _getIPAddress`.

src/docs/get-started/flutter-for/react-native-devs.md
```diff
@@ -206,8 +206,8 @@ objects to handle this.
 
 ```js
 // JavaScript
-_getIPAddress = () => {
-  const url="https://httpbin.org/ip";
+_getIPAddress() {
```
cymarechal: `function _getIPAddress()`.

src/docs/get-started/flutter-for/react-native-devs.md
jpaquim: Please consider the fact that most of the code here is designed to be inside a class, both in the case of JS and Dart. And inside a class, this is valid syntax.

sfshaza2:
@legalcodes, can you take a look at this? I don't know JS well enough to evaluate.

src/docs/get-started/flutter-for/react-native-devs.md
legalcodes: @jpaquim Thanks for these changes. Indeed this code is valid, [even with plain objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions).

Ideally we should refactor the examples on this page so that they run without any changes (assuming the fetch dependency is present), e.g. :

```
const example = {
  async _getIPAddress() {
    const url = "https://httpbin.org/ip";
    const response = await fetch(url);
    const responseJson = await response.json();
    const ip = responseJson.origin;
    console.log(ip);
  }
}

example._getIPAddress();
```
Any chance you can take a stab at this?

legalcodes:
@jpaquim Are you able to make [these changes]( https://github.com/flutter/website/pull/2624/files#r286156794)?

jpaquim:
@legalcodes yes, sure! I should be able to take care of it later today.

jpaquim:
@legalcodes ok, I rebased with master and pushed some commits with the new changes:
- First using just a wrapper object around the JS methods
- Then using also a wrapper class around the Dart methods
- Then using a wrapper class instead of object around the JS methods (which I think makes the most sense)
Let me know which one you prefer.

jpaquim:
@legalcodes I added the calling examples you mentioned, could you check now?

legalcodes:
LGTM, many thanks @jpaquim

sfshaza2:
Have we lost you, @jpaquim?

jpaquim:
@sfshaza2 @legalcodes I'm sorry for not getting back to you earlier, I had a pretty busy week.

I pushed some additional commits, hopefully they will make things more consistent and easy to compare between the two languages. Please let me know if something in the Dart examples is not very idiomatic, as I'm relatively new to the language.

However, I still have some questions:
- The section on [fetching data from API calls](https://flutter.dev/docs/get-started/flutter-for/react-native-devs#how-do-i-fetch-data-from-api-calls) should also be updated to reflect these changes, right? Is the content using HttpClient still up to date, or should it also be changed to use the http package?
- The examples on DartPad should also be updated. However, I don't see a way to specify the dependency on the http package, so I can't build the examples that use it... How do we solve this?

legalcodes:
@jpaquim No problem at all -- many thanks for this update! I'm going to verify these examples then we should be in good shape to merge. 

> The section on fetching data from API calls should also be updated to reflect these changes...
Yep, I'm double checking to make sure that the `http` is what we want to use for all of our examples, but once confirmed we can create an issue and start tracking this need.
> The examples on DartPad should also be updated. 
We'll track DartPad work separately -- it doesn't currently support imports.

src/docs/get-started/flutter-for/react-native-devs.md
```diff
@@ -206,27 +206,51 @@ objects to handle this.
...
+    });
+  }
+}
+
+main() {
+  final example = new Example();
+  example
+      ._getIPAddress()
+      .then((ip) => print(ip))
+      .catchError((error) => print(error));
```
legalcodes: Wherever we're using non-DartPad-runnable examples, let's remove the "Try it out in DartPad" link just below the example

sfshaza2:
@jpaquim, you rock! Thanks for your diligence on this!

jpaquim:
@legalcodes @sfshaza2 thanks for your kind words!

I've pushed the changes you requested, removing the DartPad links to the examples that are using the http package. Please let me know if there's anything else I can do.

PR #2630, opened 2019-05-02
Fix documentation on 'Using packages'

YoungChulShin:
1. Fix example code indentation
2. Modify first example code content. (Content is duplicated)
   - Before : It is valid for second example
      ```
      Run the app. When you click the 'Show Flutter homepage' you should see the phone's default browser open, and the Flutter homepage appear.
      ```
   - After
      ```
      Run the app. You can see app's background color is orange.
      ```

### DIFF
--- a/src/docs/development/packages-and-plugins/using-packages.md
+++ b/src/docs/development/packages-and-plugins/using-packages.md
@@ -196,33 +196,32 @@ To use this package:
 
 1. Open `lib/main.dart` and replace its full contents with:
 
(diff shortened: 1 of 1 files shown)


Germannf:
https://twitter.com/Acer?s=17

sfshaza2:
Thanks, @YoungChulShin, for catching this!! I've made a lot of other edits to this file, so it's fixed in PR https://github.com/flutter/website/pull/2711

PR #2669, opened 2019-05-14
Summarized 'Run' command

yoshi1125hisa:
In Dockerfile, it is not good to write `RUN` command repeatedly. I fixed it.


sfshaza2:
@gspencergoog, can you review and confirm?

gspencergoog:
This is fine.  It results in fewer sub containers, which means they sometimes get rebuilt more often than before, but since we don't do this very often that's not a big deal. It will probably build slightly faster overall too.

sfshaza2:
Thanks for the review, Greg! Much appreciated.

Thanks for the PR, @yoshi1125hisa!

PR #2675, opened 2019-05-20
Translate docs index into Korean

techhtml:


techhtml:
Sorry this is wrong PR!

PR #2680, opened 2019-05-21
Change `flutter packages` to `flutter pub`

mit-mit:
Related to https://github.com/flutter/flutter/pull/33041

### DIFF
--- a/src/community/china.md
+++ b/src/community/china.md
@@ -53,7 +53,7 @@ $ flutter doctor
 
 After these steps, you should be able to continue
 [setting up Flutter](/docs/get-started/editor) normally.
-From here on, packages fetched by `flutter packages get` are
+From here on, packages fetched by `flutter pub get` are
 downloaded from `flutter-io.cn` in any shell where `PUB_HOSTED_URL`
 and `FLUTTER_STORAGE_BASE_URL` are set.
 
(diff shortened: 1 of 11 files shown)


mit-mit:
This was related to the overall branding change to pub.dev. It's all pub now!

sfshaza2:
Yes, it makes sense. I just hadn't thought about it. Thanks for taking care of it!

Btw, can you add me as an admin on the flutter/website repo?

MaikuB:
@mit-mit from what I've seen, the code changes to enable the use of the `flutter pub` command only landed in master recently and aren't in the stable channel. Therefore, if i'm not mistaken, those that follow the Flutter installation guide on the website will be on the stable channel and will find that running the commands will fail to run. If that's the case then the change would either need to be reverted until the changes have flowed through to stable, or both `flutter packages` and `flutter pub` is listed with a note that the latter isn't on stable yet.

mit-mit:
@MaikuB the changes landed over two different PRs:

  * https://github.com/flutter/flutter/pull/30115: This added support in `flutter packages/flutter pub` (those two are aliases) for all existing pub commands (downgrade, deps, run, activate, etc.). This landed on March 28th, and is in all Flutter channels incl. stable.
  * https://github.com/flutter/flutter/pull/33041: This did indeed just land, but all that that is to "advertize" `flutter pub` over `flutter packages` (and note that `flutter packages` is still a valid alias, so it's non-breaking)

MaikuB:
@mit-mit Thanks for the fast reply and clarification. In that case there may an issue and I've logged it on the flutter/flutter repo [here](https://github.com/flutter/flutter/issues/33302)

PR #2682, opened 2019-05-21
Reordering the codelab list to be newbie friendly

efortuna:
Open to discussion whether some of the Material codelabs should be moved up (above Firebase for example) (or if cupertino should be moved to be nearer the Material codelabs)


sfshaza2:
I wish you had held off, @efortuna and @johnpryan . I have a (much larger) PR that also affects this page. The labs were organized alphabetically.

PR #2686, opened 2019-05-22
Switch simple state management page to provider

filiph:
This is switching the example from ScopedModel to Provider.


filiph:
Staged: https://flutter-website-filiph-staging.firebaseapp.com/docs/development/data-and-backend/state-mgmt/simple

This is, in large part, a search/replace. ScopedModel and Provider are very similar. That said, there were some paragraphs I had to change, and I'd love to have that reviewed.

PR #2691, opened 2019-05-22
Remove last mention of Model

filiph:
Also make the code correspond to the gif by assigning each item a price of $42 (instead of $1).


filiph:
Staged: https://flutter-website-filiph-staging.firebaseapp.com/docs/development/data-and-backend/state-mgmt/simple

Issue #2689, opened 2019-05-22
Suggestion: move package interdependencies conflict resolution section to page on using packages


MaikuB:
Currently there is a page that deals with handling conflicts within various packages [here](https://flutter.dev/docs/development/packages-and-plugins/developing-packages#conflict-resolution). Whilst some of the content there may be applicable to package developers, the content is likely more relevant to consumers of packages. The [using packages](https://flutter.dev/docs/development/packages-and-plugins/using-packages) page doesn't have a section that would provide that kind of guidance to such developers. One suggestion would be that the conflict resolution could be moved to the using packages page, or the using packages page could include a section and link that would take the user to https://flutter.dev/docs/development/packages-and-plugins/developing-packages#conflict-resolution

I'd be happy to submit a PR for this but wanted to start a discussion to get some thoughts on this beforehand

sfshaza2:
@MaikuB, this is an excellent suggestion. I just reviewed the using-packages page, and it's a bit rough, in general. If you'd like to file a PR to move the conflict info to using-packages, that would be great.

sfshaza2:
Thanks, again, @MaikuB!

MaikuB:
No worries though I noticed an issue as part of my amendments. This should be addressed as part of a new PR I've created [here](https://github.com/flutter/website/pull/2698). Apologies about that.

PR #2695, opened 2019-05-23
Move package interdependencies conflict resolution section to using packages

MaikuB:
This is to move the conflict resolution section as per #2689. I've changed the text according so it focuses more on using packages that may have conflicts within a user's app whilst indicating it's applicable to a package the user is developing


MaikuB:
I signed it!

PR #2700, opened 2019-05-24
Add section to localization about fully defining locales.

GaryQian:
With the addition of more advanced localization capability, it should be explained that for languages like Chinese, more localization work should be done to differentiate between things like simplified and traditional Chinese.

### DIFF
--- a/src/docs/development/accessibility-and-localization/internationalization.md
+++ b/src/docs/development/accessibility-and-localization/internationalization.md
@@ -73,9 +73,9 @@ MaterialApp(
    GlobalWidgetsLocalizations.delegate,
  ],
  supportedLocales: [
-    const Locale('en'), // English
-    const Locale('he'), // Hebrew
-    const Locale('zh'), // Chinese
+    const Locale.fromSubtags(languageCode: 'en'), // English
+    const Locale.fromSubtags(languageCode: 'he'), // Hebrew
+    const Locale.fromSubtags(languageCode: 'zh'), // Chinese *See Advanced Locales below*
     // ... other locales the app supports
   ],
   // ...
@@ -84,6 +84,9 @@ MaterialApp(
 
 Apps based on WidgetsApp are similar except that the
 `GlobalMaterialLocalizations.delegate` isn't needed.
+
+We prefer the full `Locale.fromSubtags` constructor as it supports scriptCode,
+though the `Locale` default constructor is still fully valid.
 
 The elements of the `localizationsDelegates` list are factories that produce
 collections of localized values. `GlobalMaterialLocalizations.delegate`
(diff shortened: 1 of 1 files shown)


src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -73,9 +73,9 @@ MaterialApp(
    GlobalWidgetsLocalizations.delegate,
  ],
  supportedLocales: [
-    const Locale('en'), // English
-    const Locale('he'), // Hebrew
-    const Locale('zh'), // Chinese
+    const Locale.fromSubtags(languageCode: 'en'), // English
```
HansMuller: It would be simpler to leave English and Hebrew as just simple `Locale` constructor calls, since the advanced Locale features aren't needed.

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -96,6 +99,42 @@ More information about these app properties, the types they
...
+{% endprettify %}
+
+This explicit full definition will ensure your app can distinguish between and provide
+the fully nuanced localized content to all combinations of these country codes. If a
+user's preferred locale is not specified, then the closest match will be used instead,
+which will likely contain differences to what the user expects. See
+[`Localizations`]({{site.api}}/flutter/widgets/WidgetsApp/supportedLocales.html) for
+how the supported locales and the preferred locales are resolved.
+
+Although Chinese is a primary example, other languages like French (FR_fr, FR_ca, etc)
```
HansMuller: Do we have a comprehensive list of supported locales that would benefit from being "fully differentiated"?

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -96,6 +99,42 @@ More information about these app properties, the types they
...
+  const Locale.fromSubtags(languageCode: 'zh', scriptCode: 'Hans'), // generic simplified Chinese 'zh_Hans'
+  const Locale.fromSubtags(languageCode: 'zh', scriptCode: 'Hant'), // generic traditional Chinese 'zh_Hant'
+  const Locale.fromSubtags(languageCode: 'zh', scriptCode: 'Hans', countryCode: ), // 'zh_Hans_CN'
+  const Locale.fromSubtags(languageCode: 'zh', scriptCode: 'Hant', countryCode: ), // 'zh_Hant_TW'
+  const Locale.fromSubtags(languageCode: 'zh', scriptCode: 'Hant', countryCode: ), // 'zh_Hant_HK'
+],
+{% endprettify %}
+
+This explicit full definition will ensure your app can distinguish between and provide
+the fully nuanced localized content to all combinations of these country codes. If a
```
HansMuller: Will the app also benefit from script-code-specific localized content provided by the Material and Cupertino libraries? In other words, should apps do this as a matter of course, even if they're not providing any localizations of their own?

src/docs/development/accessibility-and-localization/internationalization.md
GaryQian: We may be able to throw one together, but the total combinations of all possible countries/tags/etc may be extremely large and not particularly useful. This basically applies to anything with more than just a language code, but only becomes relevant when the variations are significant.

src/docs/development/accessibility-and-localization/internationalization.md
GaryQian: Yes, if the script-code is specified (off the top of my head I recall for Chinese and Serbian), we will provide script-specific content. However, since different countries/regions express certain characters slightly differently, these nuances are lost without the country code.

Generic traditional chinese is currently using zh_TW, due to matching platforms the best.

PR #2714, opened 2019-05-30
Add Provider to state management options

filiph:
Provider was notably missing. Some people arrive straight to https://flutter.dev/docs/development/data-and-backend/state-mgmt/options, never seeing any of the previous pages.


filiph:
Hmm, I'm not sure. You mean in /options? Isn't that a little on the nose?

PR #2726, opened 2019-06-03
fix: fix shopping list example setState not working as expect

fatelei:
The original example, the default value of `inCart` is `false`, so the callback not working as expect.


fatelei:
I signed it!

SValab:
Hello.
With the change, unfortunately it's still not working. I noticed the onTap function passed to the ListTile  is calling _onCartChanged(product, **!inCart**);_ . This should be _onCartChanged(product, **inCart**);_

PR #2733, opened 2019-06-05
Fix example validator functions on "Building a form with validation" page

badver:
According to FormFieldValidator docs, validator function should return an error string to display if the input is invalid, or null otherwise.


badver:
I signed it!

Issue #2734, opened 2019-06-06
Fix build


kwalrath:
As of https://travis-ci.org/flutter/website/builds/539735108 (#2716), both jobs in the build have started failing.

One cause is that https://ericwindmill.com/posts/using-flutter-inherited-widgets-effectively/ 404s:

```
http://localhost:4002/docs/development/data-and-backend/state-mgmt/options
- (943:0) 'Using Fl..' => https://ericwindmill.com/posts/using-flutter-inherited-widgets-effectively/ (HTTP 404)
- (943:0) 'Using Fl..' => https://ericwindmill.com/posts/using-flutter-inherited-widgets-effectively/ (HTTP 404)
```
The other job appears to be failing due to a failed assertion.

kwalrath:
#2736 fixed the 404 that caused one job to fail. Now we need to fix the other job...

legalcodes:
Looking into this...

kwalrath:
#1840 seems to say that the site build didn't used to fail due to external 404s. Instead, you'd get reports. I think that makes more sense than breaking the site because an external link changed.

@sfshaza2 can you tell us more?

sfshaza2:
No, on flutter.dev, Travis has pretty much always failed, even if an external link was broken. Not my favorite feature.  :)  I'm ok if we turn that off and use the weekly report.

chenglu:
I think the problem may be caused the CI issue is because of the new released flutter_test package since I passed the CI build by commenting the testing code at widget_test.dart.

> examples/layout/lakes/step5/test/widget_test.dart 
and 
> examples/layout/pavlova/test/widget_test.dart

FYI, wish it will help.

sfshaza2:
Thanks, @chenglu!

legalcodes:
@chenglu can you link to your passing CI in this issue? Many thanks.

legalcodes:
We've temporarily disabled the tests for flutter examples: 
https://github.com/flutter/website/pull/2738

The `./tool/check-code.sh` Travis tasks are now passing:
https://travis-ci.org/flutter/website/jobs/542483725
https://travis-ci.org/flutter/website/jobs/542707775

PRs are unblocked. Outstanding TODOs for this issue: 

1. Investigate whether flutter examples should be tested as a potential blocker for incoming PRs to flutter/website
2. Investigate error thrown by testing package.

sfshaza2:
Now that the immediate crisis is over, downgrading the severity of this issue.

[...]

PR #2736, opened 2019-06-06
Fix a bad link that was breaking the build.

kwalrath:
Contributes to #2734


legalcodes:
Let's do a retrospective on this so we can figure out whether broken links of this sort *should* break builds

kwalrath:
This fixed half of the build.

legalcodes:
Sure taking a look

PR #2739, opened 2019-06-07
Add Information for Simulator (Mac Os)

larsroettig:
If start with Mobile development you need to know about all requirements of starting to develop with flutter.


larsroettig:
But think it should be covered also on this site because new User start there maybe don't know about what is need to start with the development.

PR #2750, opened 2019-06-12
Update widgets-intro.md

tomo-ms:
The parameter of onCartChanged method seems to be wrong.
By removing "!" from inCart, the sample worked correctly.


tomo-ms:
I signed it!

PR #2765, opened 2019-06-15
Update stateful hot reload => Stateful Hot Reload

legalcodes:


src/docs/development/tools/hot-reload.md
```diff
@@ -69,7 +69,7 @@ Dart code to keep using hot reload.
 
 ## Previous state is combined with new code
 
-Flutter's hot reload feature, sometimes described as _stateful hot reload_,
+Flutter's hot reload feature, sometimes described as _Stateful Hot Reload_,
```
sfshaza2: This is one of those sort of tricky situations. The guideline says to use Stateful Hot Reload when referencing the official Flutter feature. But this sentence says, "sometimes described as..." which makes it a more colloquial reference, rather than an Official Feature Reference.  I think case, lower case is fine.  (I'd be interested to know whether K agrees with this.)

src/docs/resources/faq.md
```diff
@@ -194,7 +194,7 @@ work.
 
 Debug mode builds use a virtual machine (VM) to run Dart code
 (hence the "debug" banner they show to remind people that they're
-slightly slower) in order to enable stateful hot reload.
+slightly slower) in order to enable Stateful Hot Reload.
```
sfshaza2: Here it does seem like it's talking about an Official Feature.

P.S. I would change "they show" to "shown".  And I'd probably change "that they're slightly slower" to "that these builds are slightly slower."

src/docs/resources/faq.md
```diff
@@ -208,7 +208,7 @@ work.
 
 Debug mode builds use a virtual machine (VM) to run Dart code (hence the
 "debug" banner they show to remind people that they're slightly slower)
-in order to enable stateful hot reload.
+in order to enable Stateful Hot Reload.
```
sfshaza2: Upper case is fine, but I could go either way here.

Basically, when using "English", it's ok to use lower case. So, you would write "enable the widget...", not "enable the Widget".  But if you are, say, writing about a specific class or instance or method, then you would code font and use proper case.

src/docs/resources/faq.md
```diff
@@ -281,7 +281,7 @@ and met all our requirements and criteria.
 
 Dart runtimes and compilers support the combination of two critical features for
 Flutter: a JIT-based fast development cycle that allows for shape changing and
-stateful hot reloads in a language with types, plus an Ahead-of-Time compiler
+Stateful Hot Reloads in a language with types, plus an Ahead-of-Time compiler
```
sfshaza2: This is definitely just using English, not hyping a feature. I'd leave this lower case.

src/docs/development/tools/hot-reload.md
legalcodes: For consistency's sake, how about :

`Flutter's Stateful Hot Reload` 
and drop the `sometimes described` (as I'm not sure what value that's adding here)

legalcodes: We should avoid colloquial reference per the style guide:
https://developers.google.com/style/tone

legalcodes:
I don't think it's tricky to prevent the exceptions from killing the rule. 

If we provide guidance on using a term like hot reload, my vote is that we craft our descriptions to conform to one of the two ways that we recommend using the term (this PR represents my stab at doing this). I think there's a value add to taking some effort to avoid colloquialism for the sake of remaining consistent in our usage.

We can also avoid "colloquial references"  and "using "English" colloquially" per the style guide: 
https://developers.google.com/style/tone

That's my two cents anyway!

src/docs/resources/faq.md
legalcodes: Great, I'll leave this as-is, I want to keep this PR focused our use of `stateful hot reload`

src/docs/resources/faq.md
legalcodes: Cool, will leave this as upper case here.

src/docs/resources/faq.md
legalcodes: Agreed that this isn't meant to reference the official feature, but "stateful" is still needed as a descriptor in this context. Will leave it as is.

src/docs/development/tools/hot-reload.md
sfshaza2: I think they refer to "colloquial" as in idioms or slang. And, yes, that is to be avoided. But what I mean is different. It's better to use English than to always use code fonted words. Maybe I'm not describing it well.

But I agree that deleting "sometimes described as" here would make the upper case version acceptable.

src/docs/resources/faq.md
sfshaza2: Right, it's fine to leave "stateful". Just because those three words are used together does not mean that it should be automatically made upper case. It has more to do with referring to the Official Feature (TM), or just describing.

sfshaza2:
As I said, there's colloquial and there's colloquial. Idiom, slang, yes, to be avoided. But using a phrase as descriptive doesn't require branding.

sfshaza2:
LGTM with a minor change. Go ahead and merge when done.  :)

legalcodes:
@sfshaza2 I added just a few more cases I think are referencing the official feature and should be capitalized.

PR #2767, opened 2019-06-16
Consistent wording in Simple app state management

SamuelDebruyn:
I think `argument` is more appropriate and consistent than `attribute` here.


SamuelDebruyn:
I signed it!

PR #2787, opened 2019-06-25
travis ci: make link checker skip codelab

johnpryan:
There doesn't seem to be anything wrong with https://codelabs.developers.google.com/codelabs/google-photos-sharing but this seems to be the culprit so I think we should skip it when we test external links


sfshaza2:
John, I forced the merge, and assume that this will fix the build.

PR #2799, opened 2019-07-02
Adjust the step by step to the same file name

soonsam123:
Doesn't make more sense to keep using `my_icon` for the file name since we are following a step by step?


soonsam123:
I signed it!

PR #2807, opened 2019-07-07
Remove confusing provider article

cbenhagen:
The article article describes the use of the Provider package with a ChangeNotifier while calling this the BLoC pattern. As this will confuse a lot of people I would prefer to remove the article while we wait for it to be fixed (https://github.com/developer-school/flutter_bloc_provider/issues/2).


sfshaza2:
@filiph, please review.

filiph:
LGTM. Merging.

PR #2819, opened 2019-07-10
Added details about invoking platform channel APIs on the main thread.

matthew-carroll:
Added details about invoking platform channel APIs on the main thread.

Staging site here:
https://flutter-website.firebaseapp.com/docs/development/platform-integration/platform-channels

### DIFF
--- a/src/docs/development/platform-integration/platform-channels.md
+++ b/src/docs/development/platform-integration/platform-channels.md
@@ -47,7 +47,8 @@ to ensure the user interface remains responsive.
 {{site.alert.note}} 
   Even though Flutter sends messages to and from Dart asynchronously,
   whenever you invoke a channel method, you must invoke that method on the
-  platform's main thread.
+  platform's main thread. See the 
+  [section on threading](#channels-and-platform-threading) for more information.
 {{site.alert.end}}
 
 On the client side, `MethodChannel` ([API][MethodChannel]) enables sending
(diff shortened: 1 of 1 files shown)


src/docs/development/platform-integration/platform-channels.md
```diff
@@ -44,6 +44,12 @@ channels as illustrated in this diagram:
 Messages and responses are passed asynchronously,
 to ensure the user interface remains responsive.
 
+{{site.alert.note}} 
+  Even though Flutter sends messages to and from Dart asynchronously,
+  whenever you invoke a channel method, you must invoke that method on the
+  platform's main thread.
+{{site.alert.end}}
```
dnfield: Maybe some examples of how to do that, or a link to the place with examples?

mklim: It may also be worth explaining what "main thread" really means in relationship to the different platforms. The term is a little overloaded.

xster:
LGTM

src/docs/development/platform-integration/platform-channels.md
matthew-carroll: @dnfield @mklim good suggestions. I pushed an update that links to official docs about each platform's main thread, and I added examples in each language. Please let me know if it looks good.

src/docs/development/platform-integration/platform-channels.md
```diff
@@ -646,3 +661,66 @@ classes, or create your own codec.
 [BinaryCodec]: {{site.api}}/flutter/services/BinaryCodec-class.html
 [StringCodec]: {{site.api}}/flutter/services/StringCodec-class.html
 [JSONMessageCodec]: {{site.api}}/flutter/services/JSONMessageCodec-class.html
+
+## Channels and Platform Threading
+
+When writing code on the platform side, which is the Android/iOS side using
```
mklim: subjective/optional style nit: This could be rewritten to use [active voice](https://developers.google.com/style/voice). Also, leading with prepositional phrases like this makes the structure of the sentence kind of hard to follow since the subject is lurking at the end.

I suggest an alternative like:

> Invoke all channel methods on the platform's main thread when writing code on the platform side.

(Not sure what the value is of specifying that "platform" means "Android/iOS.")

This applies to a couple other sentences below.

src/docs/development/platform-integration/platform-channels.md
```diff
@@ -646,3 +661,66 @@ classes, or create your own codec.
...
+Java/Kotlin/Objective-C/Swift, all channel methods must be invoked on the
+platform's main thread. Please see official documentation about [the main
+thread on Android] and [the main thread on iOS].
+
+[the main thread on Android]: https://developer.android.com/guide/components/processes-and-threads#Threads
```
mklim: On Android "main thread" is an overloaded term. The documentation linked here mentions that colloquially the "main thread" often means the "UI Thread" and then proceeds to talk about the UI Thread, but points out that "main thread" is actually a distinct term and links to another page about the annotations where `@MainThread` is a separate thing.

I'm pretty sure on Android we really mean `@UiThread` whenever we refer to "main thread," and not `@MainThread`, so I think it's worth pointing out the discrepancy. This is mostly what I was thinking of when I was talking about term overloading earlier.

matthew-carroll:
@sfshaza2 @dnfield CI is failing on `bundle exec jekyll build`, but when I run it locally it works fine and when I deploy to firebase it seems fine, any idea why CI might be failing?

src/docs/development/platform-integration/platform-channels.md
matthew-carroll: I gave it another pass. Let me know what you think.

mklim: LG!

src/docs/development/platform-integration/platform-channels.md
sfshaza2: Very good catch, @mklim.

sfshaza2:
Yes, I suspect it’s because of some broken API links that resulted from the 1.7 release. Let me address those.

matthew-carroll:
@sfshaza2 if the fix needs a rebase of this branch, can you let me know when to run that so the tests try again?

sfshaza2:
My recent PR (earlier today) fixes some of the problems with the build, but the one problem is doesn’t fix is this broken API link: https://api.flutter.dev/objcdoc/Classes/FlutterMethodChannel.html

@dnfield has fixed it on master, but it’s not reflected on the website. Ian asked for it to be pushed to stable, but it hasn’t happened yet.

matthew-carroll:
@sfshaza2 I rebased against the latest on master and it looks like there is still an issue...

sfshaza2:
Yes, that’s why I filed this bug:
https://github.com/flutter/flutter/issues/36018

matthew-carroll:
Ok, so let me see if understand the issue. Even though the right docs have been pushed to master, and this branch is based on master, the root cause of the issue is that master links are evaluated against stable instead of master, and therefore we're broken?

dnfield:
I've pushed up the missing docs to the stable docs site - hopefully we're good now.

matthew-carroll:
@dnfield is there a way to kickstart a new CI build for this PR? Or do I need to push something to cause it to try again?

dnfield:
I just kicked it - click on the "Details" link by the `continuous-integration/travis-ci/pr`, and on that page there's a button to restart.

matthew-carroll:
Ah, yeah, I forgot about that restart button. Thanks.

dnfield:
Well now it looks like it's failing because it can't get to twitter.com/flutterdev

matthew-carroll:
Spoke with Shams offline. I'm going to merge this in because the failures are unrelated to this change, and probably won't be corrected in the immediate future.

PR #2823, opened 2019-07-11
Added splash screen guide for Android.

matthew-carroll:
Add android splash screen guide

I can't stage this one yet because I'm still waiting to merge the platform channel threading guide update.


src/docs/development/platform-integration/add-to-app-android/add-splash-screen.md
```diff
@@ -0,0 +1,221 @@
+---
+title: Adding a splash screen and launch screen to an Android app
```
xster: Most of this page is more relevant for the full-app case (once we make the FullFlutterApp which I assume will have the same functionality) such as https://flutter.dev/docs/development/ui/assets-and-images#updating-the-launch-screen than add to app. 

Can we update that section so full-app developers can find instructions too and for the add to app specific things such as FlutterFragment, we can leave them here or let them be in the Fragment guide?

matthew-carroll: I replaced the description in the guide you linked with a cross-reference to this new guide, which explains the same details and much more.

matthew-carroll:
@sfshaza2 I used the Android site macro instead of the URL, and I replaced the one instance of "will" that I could find.

src/_data/sidenav.yml
```diff
@@ -126,6 +126,8 @@
           children:
             - title: Add a single Flutter screen
               permalink: /docs/development/platform-integration/add-to-app-android/add-flutter-screen
+            - title: Add a Splash Screen
```
xster: This is ok to merge for now. @Sfshaza do you have any thoughts on how big this subtree can get? @matthew-carroll at this rate I assume this subtree will have something like 5-10 articles? That seems ok to me. Otherwise, we can also generalize into a "Performance implications" type page and this could be a solution type to the general 'perceived latency' challenge, to which there are many solutions. 

Anyway, just food for thought. Nothing we need to do now.

src/docs/development/platform-integration/add-to-app-android/add-splash-screen.md
```diff
@@ -0,0 +1,221 @@
...
+[launch screen]: {{site.android-dev}}/topic/performance/vitals/launch-time#themed
+
+Flutter provides support for displaying an Android launch screen before showing
+a `FlutterActivity`. Here are instructions to display an Android launch screen:
+
+### Define a launch theme
+
+In `styles.xml`, define a theme whose `windowBackground` is set to the
+`Drawable` that should be displayed as the launch screen.
+
```
xster: This is still in the default flutter create template right? If so, mention it

src/docs/development/platform-integration/add-to-app-android/add-splash-screen.md
```diff
@@ -0,0 +1,221 @@
...
+```xml
+<activity
+    android:name=".MyActivity"
+    android:theme="@style/LaunchTheme"
+    // ...
+    >
+    <meta-data
+        android:name="io.flutter.embedding.android.NormalTheme"
+        android:resource="@style/NormalTheme"
```
xster: I have the same meta point as above. While reading this, I'm not sure which lines I'm expected to add and which ones have defaults in the create templates and that I just need to modify (and what is the default behavior)

matthew-carroll: It makes sense that once the template has this code that we'd want to reference it here. Is it a good idea for us to predict what will be in those templates and attempt to document it now before that code exists? I'm concerned that it's a great opportunity to have these docs mismatch with the eventual template implementation...if we wait until the template exists to update these guides, then they would be guaranteed to be correct. What do you think?

xster: Agree. 

My bad for not realizing this isn't in the tools template yet. 
Let's document what's going to be on the stable we cut for the new android embedding. Until then, let's keep this branch correct for master as the tooling/framework changes.

src/docs/development/platform-integration/add-to-app-android/add-splash-screen.md
```diff
@@ -0,0 +1,221 @@
...
+</activity>
+```
+
+The Android app now displays the desired launch screen while the app
+initializes.
+
+## Flutter Splash Screen
+
+Each Flutter experience in an app requires a few moments to initialize the Dart
+isolate that runs the code. This means a user momentarily sees a blank screen.
```
xster: momentarily sees a blank screen while Flutter renders its first frame

src/docs/development/platform-integration/add-to-app-android/add-splash-screen.md
```diff
@@ -0,0 +1,221 @@
...
+#### In a FlutterActivity
+
+To display a `Drawable` as a Flutter splash screen in a `FlutterActivity`, add
+the following meta-data to the associated `FlutterActivity` in
+`AndroidManifest.xml`.
+
+```xml
+<meta-data
+    android:name="io.flutter.embedding.android.SplashScreenDrawable"
+    android:resource="@drawable/my_splash"
```
xster: Should this be by default (or otherwise document that they can): make this drawable the same drawable that we set as the LaunchTheme windowBackground's drawable (something most people will want to do I assume)?

src/docs/development/platform-integration/add-to-app-android/add-splash-screen.md
```diff
@@ -0,0 +1,221 @@
...
+
+TODO(mattcarroll)
+
+### Creating a custom SplashScreen
+
+Splash screens are a great branding opportunity. As such, many teams choose to
+implement unique, highly-customized splash experiences. To facilitate this,
+Flutter allows developers to display an arbitrary Android `View` as a splash
+screen, and even allows developers to control how that `View` transitions to
+Flutter.
```
xster: transitions to Flutter once Flutter has its first frame ready and rendered.

src/_data/sidenav.yml
sfshaza2: I think 5-10 articles is OK. But we’d probably not want it to grow a lot more than that.

xster: ok, let's leave the structure as is until we run into the size issue.

src/docs/development/platform-integration/add-to-app-android/add-splash-screen.md
matthew-carroll: done

src/docs/development/platform-integration/add-to-app-android/add-splash-screen.md
matthew-carroll: done

src/docs/development/platform-integration/add-to-app-android/add-splash-screen.md
matthew-carroll: done

src/docs/development/platform-integration/add-to-app-android/add-splash-screen.md
matthew-carroll: done

PR #2825, opened 2019-07-11
Update brand guidelines to include Event rule

maguinis:


src/brand.md
```diff
@@ -67,7 +67,11 @@ rules governing the proper use of the Flutter trademarks.
   trademarks of Google LLC. We are not endorsed by or affiliated with
   Google LLC."
 * Keep some distance between the Flutter trademarks and any other trademarks,
-  logos, or icons that are displayed on the webpage.
+  logos, or icons that are displayed on the webpage.  
+* <b>[Unofficial Flutter Events]</b> You may use the Flutter trademarks as part of the name
```
timsneath: This feels like it might be better under "Community Use Exceptions", since it's a similar rule to the one about local user groups. WDYT?

sfshaza2: I’m waiting for you and @timsneath to resolve.

sfshaza2:
@dnfield pushed the objective C API docs live, which fixed one problem. However, the icon property page for the Icon class is now missing and breaking the build.  I’ve filed:  https://github.com/flutter/flutter/issues/36036. Finally, the build is complaining because we’re getting 429 errors (too many requests) from a number of our youtube embeds. Not sure what is up with that.

src/brand.md
maguinis: Ok, initially thought to include in DO as it's more visible, but if we bold the [Unofficial Flutter Events] then the community section works. I'll adjust

src/brand.md
```diff
@@ -119,4 +115,9 @@ below are specific exceptions to the rules described above:
    Where the Flutter trademarks are displayed on a social media account
    as part of a community site name, you should include a disclaimer that
    clarifies that it is not an official Google account.
+   
+3. <b>[Unofficial Flutter Events]</b> You may use the Flutter trademarks as part of the name 
+   of a community event (e.g. conference), but please make sure to include the following disclaimer
```
sfshaza2: e.g. => for example, a

We don’t use e.g.: https://developers.google.com/style/abbreviations#dont-use

Also, we break our source lines up at 80 chars or less. These lines seem longer.

PR #2848, opened 2019-07-21
normalize the json files in the widget catalog

devoncarew:
Normalize the json files in the widget catalog:
- read `src/_data/catalog/index.json` in and write out w/ Dart's json encoder
- read `src/_data/catalog/widgets.json` in and write out w/ Dart's json encoder
- in a separate commit (for better reviewability), sort the entries in `widgets.json` lexically by class name

These changes should be entirely semantics free. We'er considering generating these metadata files from the package:flutter sources, and these normalizations will help identify any actual semantic differences between the pre-generated catalog and the post-generated catalog.


devoncarew:
I'm not sure who the best reviewers would be for these changes; cc'ing @sfshaza2, @kwalrath, @xster, @chalin, @InMatrix. Looks like the history for the file was lost at some point.

sfshaza2:
@devoncarew, this PR looks good. However, what about missing widgets? For example, RangeSlider.

Also, it looks good, as far as I can tell. Did you stage these changes to verify that it looks good?

devoncarew:
> @devoncarew, this PR looks good. However, what about missing widgets? For example, RangeSlider.

This PR is just whitespace and ordering changes for the files - the changes shouldn't have any effect on behavior.

I am working on a tool to generate widget catalog info from the package:flutter sources. That's still in progress; it will have some benefits in terms of maintenance and missing widgets. This catalog currently has 154 widgets, and the in-progress generated one has 338 ones.

> Also, it looks good, as far as I can tell. Did you stage these changes to verify that it looks good?

No, how do I stage?

PR #2851, opened 2019-07-22
Updating info on size of Flutter engine

sfshaza2:
Thanks for this info, @blasten! However, I thought that we were going to publish info on how to make the APK smaller. Are there techniques for that?


src/docs/resources/faq.md
```diff
@@ -345,37 +345,44 @@ Flutter should be able to run most Dart code that does not import
...
-which is also built into [Android Studio][].
+which is also [built into Android Studio][].
 
-On iOS, a release IPA of the same app has a download
-size of 10.8&nbsp;MB on an iPhone X, as reported by Apple's
-App Store Connect. The IPA is larger than the APK
-mainly because Apple encrypts binaries within the IPA,
-making the compression less efficient
+On iOS, a release IPA of the same app has a download size of
+10.8 MB on an iPhone X, as reported by Apple’s App Store Connect.
```
goderbauer: While Android APKs increased in size between December and July, the size of an iOS IPA stayed the same?

PR #2858, opened 2019-07-23
Update link to the sample with newly landed url

filiph:
Thanks to https://github.com/flutter/samples/pull/87, we can now link to flutter/samples instead of filiph/samples.

### DIFF
--- a/src/docs/development/data-and-backend/state-mgmt/simple.md
+++ b/src/docs/development/data-and-backend/state-mgmt/simple.md
@@ -420,7 +420,7 @@ rebuild when `notifyListeners` is called.
 ## Putting it all together
 
 You can [check out the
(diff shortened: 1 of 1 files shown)


src/docs/development/data-and-backend/state-mgmt/simple.md
```diff
@@ -420,7 +420,7 @@ rebuild when `notifyListeners` is called.
 ## Putting it all together
 
 You can [check out the
-example]({{site.github}}/filiph/samples/tree/provider-shopper/provider_shopper)
+example]({{site.github}}flutter/samples/tree/master/provider_shopper)
```
sfshaza2: There should be a slash before "flutter". But I'm glad to have this in a more official repo!

filiph: Thanks for noticing! Fixed.

PR #2862, opened 2019-07-25
add physics simulation article to cookbook 

johnpryan:
This also changes the headers for the other animations article to `h2` headers, and removes the h1 header since it's already included


src/docs/cookbook/animation/physics-simulation.md
```diff
@@ -0,0 +1,393 @@
...
+  );
+  var simulation = SpringSimulation(spring, 0, 1, -unitVelocity);
+
+  _controller.animateWith(simulation);
+},
+```
+
+{{site.alert.note}}
+Now that the animation controller uses a simulation it's `duration` argument is
+no longer required.
```
johnpryan: @goderbauer is there anything you feel should be explained here? Is this a best-practice?

src/docs/cookbook/animation/page-route-animation.md
```diff
@@ -27,7 +25,7 @@ To create a custom page route transition, this recipe uses the following steps:
 4. Use a `CurveTween`
 5. Combine the two `Tween`s
 
-# 1. Set up a PageRouteBuilder
+## 1. Set up a PageRouteBuilder
```
sfshaza2: Thanks for taking care of this! I noticed it after we published your last page, but hadn't gotten around to fixing it.

src/docs/cookbook/animation/physics-simulation.md
```diff
@@ -0,0 +1,393 @@
...
+  }
+}
+```
+
+Make the `_DraggableCardState` class extend from
+`SingleTickerProviderStateMixin`.  Then construct an `AnimationController` in
+`initState` and set `vsync` to `this`.
+
+{{site.alert.note}}
+Extending `SingleTickerProviderStateMixin` allows the state object to be a
```
sfshaza2: For future reference. The convention is to indent the note content by 2 spaces per line.

src/docs/cookbook/animation/physics-simulation.md
```diff
@@ -0,0 +1,393 @@
...
+`TickerProvider` for the `AnimationController`. For more information, see the
+documentation for
+[TickerProvider](https://api.flutter.dev/flutter/scheduler/TickerProvider-class.html).
```
sfshaza2: Please use the URL macros. So, 

[TickerProvider]({{site.api}}/flutter/scheduler/TickerProvider-class.html).

src/docs/cookbook/animation/physics-simulation.md
```diff
@@ -0,0 +1,393 @@
...
+
+```dart
+Alignment _dragAlignment = Alignment.center;
+```
+
+Add a `GestureDetector` that handles the `onPanDown`, `onPanUpdate`, and
+`onPanEnd` callbacks. To adjust the alignment, use a `MediaQuery` to get the
+size of the widget, and divide by 2. (This converts units of "pixels dragged" to
+coordinates that
+[Align](https://api.flutter.dev/flutter/widgets/Align-class.html) uses.) Then,
```
sfshaza2: Please use the {{site.api}} macro.

src/docs/cookbook/animation/physics-simulation.md
```diff
@@ -0,0 +1,393 @@
...
+}
+```
+
+## Step 3: Animate the widget
+
+When the widget is released, it should spring back to the center.
+
+Add an `Animation<Alignment>` field and an `_updateAnimation()` method. This
+method defines a `Tween` that interpolates between the point the widget was
+dragged to, to the point in the center .
```
sfshaza2: Delete the space before the period.

src/docs/cookbook/animation/physics-simulation.md
sfshaza2: Intent the note content by 2 spaces per line.

src/docs/cookbook/animation/physics-simulation.md
```diff
@@ -0,0 +1,393 @@
...
+
+## Step 4: Calculate the velocity to simulate a springing motion
+
+The last step is to do a little math, to calculate the velocity of the widget
+after it's finished being dragged. This is so that the widget realistically
+continues at that speed before being snapped back. ( `_updateAnimation()`
+already sets the direction  by setting the animation's start and end alignment.)
+
+The `onPanEnd` callback provides a `DragEndDetails` object with the same
+velocity that the user was dragging their finger at. This value is in pixels per
```
sfshaza2: Normally, I'm not so picky, but I can't have this sentence end with a preposition.  :)
=> ... object with the same velocity as the user's finger drag.

Also removes "their" and the whole gender issue.  Not that I love this sentence either, but I think it's less confusing.

src/docs/cookbook/animation/physics-simulation.md
```diff
@@ -0,0 +1,393 @@
...
+ _controller.reset();
+ _controller.forward();
+},
+```
+
+## Step 4: Calculate the velocity to simulate a springing motion
+
+The last step is to do a little math, to calculate the velocity of the widget
+after it's finished being dragged. This is so that the widget realistically
+continues at that speed before being snapped back. ( `_updateAnimation()`
```
sfshaza2: I try to avoid starting sentences this way. Also note the empty space at the beginning of the sentence.

=> (The `_updateAnimation()` method already sets...)

src/docs/cookbook/animation/physics-simulation.md
```diff
@@ -0,0 +1,393 @@
...
+/// released.
+class DraggableCard extends StatefulWidget {
+  final Widget child;
+  DraggableCard({this.child});
+
+  @override
+  _DraggableCardState createState() => _DraggableCardState();
+}
+
+/// A draggable card that moves back to [Alignment.center] when it's
```
sfshaza2: I find it confusing for both the widget and the state classes to have identical comments. I'd leave the comment for the widget class as-is, but I'd change the comment for the state class to something like:

/// Holds the state for the DraggableCard widget.

src/docs/cookbook/animation/physics-simulation.md
johnpryan: Yeah this sentence is a bit problematic... I've reworked it to try to match the docs more closely:

```
The `onPanEnd` callback provides a [DragEndDetails][] object. This object
provides the velocity of the pointer when it stopped contacting the screen. The
velocity is in pixels per second, but the `Align` widget doesn't use pixels. It
uses coordinate
```

johnpryan:
@sfshaza2 I took an axe to a few of the sentences in question, and added more links to some classnames. (The links are reference-style so they got moved to the bottom.)

Comments should be addressed - Thanks for the feedback!

PR #2865, opened 2019-07-26
Updated Form-validation Documentation according to screenshot

rkishan516:
In Screenshot the error shown is "Please enter some text" while the cookbook code returns "Enter some text". So, changed that according to screenshot.


rkishan516:
I signed it.

PR #2901, opened 2019-08-12
Add /go/ links section

Hixie:
cc @Sfshaza 

This adds a redirect as discussed. The idea is that in due course we would have many of these.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -112,7 +112,7 @@
       { "source": "/youtube", "destination": "https://youtube.com/flutterdev", "type": 301 },
       { "source": "/desktop", "destination": "https://github.com/flutter/flutter/wiki/Desktop-shells", "type": 302 },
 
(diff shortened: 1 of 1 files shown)


Hixie:
I don't understand why the CI failed.

Please feel free to check this in if it's ok.

sfshaza2:
@Hixie, I kicked off the job again, just to make sure it wasn't a glitch. It's failing on the link check, and I'm wondering if:
a) The script doesn't like the comma at the end of the newly added line (which is the last line in the list), or
b) The script doesn't like that the redirect for /go/template doesn't have a corresponding physical file.

I'm also wondering if adding go/template to tool/config/linkcheck-skip-list.txt might be useful here.

Hixie:
I've tried removing the comma.

sfshaza2:
That seemed to fix it! Thanks!

PR #2904, opened 2019-08-13
SomeWidget in examples not clear

pashute:
I clarified where we are in the code, what SomeWidget() is and how you get the current state in the GOOD example. 

Perhaps add a line at line 123 with
    
    // No need for a function like updateWith(Item item)...

### DIFF
--- a/src/docs/development/data-and-backend/state-mgmt/simple.md
+++ b/src/docs/development/data-and-backend/state-mgmt/simple.md
@@ -81,13 +81,13 @@ with the following in the `MyCart` widget:
 <!-- skip -->
 ```dart
 // BAD: DO NOT DO THIS
-Widget build(BuildContext context) {
-  return SomeWidget(
+Widget build(BuildContext context) { // MyCart widget-builder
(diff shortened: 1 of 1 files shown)


pashute:
@googlebot I signed it!

filiph:
Hi @pashute, thanks for the PR! I like the last change (adding clarification to the line comment, "... using cartModel that has the current state of the cart ...") but I'm not sure about the rest. To me the advantage of explicitly stating where we are is more than offset by the more verbose sample. How strongly do you feel about those other changes?

pashute:
I needed to re-read a few times to understand where I was. To a newb like
me, it was not immediately clear what I'm looking at and why.
A good example should be intuitive (without need for references)
interactive and attractive.

sfshaza2:
Have we lost you, @pashute? I will close if so.

pashute:
I'm not sure I understand what has happened with this request. Did anyone correct the example, or merge my work into the main branch?  If so please close again...

PR #2915, opened 2019-08-15
Update internationalization.md

hamersun:
missing countryCode target for CN, TW and HK in describing how to express every variant of Chinese.


hamersun:
@googlebot I signed it!

PR #2918, opened 2019-08-18
The command line of build_runner build

chenglu:
We've found an issue here cfug/flutter.cn/issues/425 from the Flutter China community about the command line of build_runner build, we tried `flutter pub run build_runner build` and got errors, but `flutter packages pub run build_runner build` got works, I was wondering if there are any issues about the document, I just add `packages` in the command line.

cc/ @t496971418 @Vadaski @Realank, let me know if you have any updates.
cc/ @sfshaza2

### DIFF
--- a/src/docs/development/data-and-backend/json.md
+++ b/src/docs/development/data-and-backend/json.md
@@ -321,7 +321,7 @@ There are two ways of running the code generator.
 
 #### One-time code generation
 
-By running `flutter pub run build_runner build` in the project root,
+By running `flutter packages pub run build_runner build` in the project root,
 you generate JSON serialization code for your models whenever they are needed.
 This triggers a one-time build that goes through the source files, picks the
 relevant ones, and generates the necessary serialization code for them.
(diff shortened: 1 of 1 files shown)


sfshaza2:
@tvolkert, can you confirm this change?

zanderso:
That's correct, `pub` is an alias for `packages`. Unfortunately, `flutter packages run` and `flutter packages pub run` both exist and do something slightly different. `flutter packages run` was added by @mit-mit recently. It sets the environment variable `PUB_ENVIRONMENT` to `forward`. `flutter packages pub run` is what we've had for awhile. It sets `PUB_ENVIRONMENT` to `interactive`. I'm not sure what the effects of that difference are. There might be other differences in the actual pub invocation that I couldn't see from a cursory reading.

@mit-mit could you help us understand why we have both of these and what the difference is? Thanks!

mit-mit:
I looked at the code. The only difference I can tell between the short-hand forwards I added, and the old general `packages/pub pub` command is indeed that the latter runs interactively. I have a potential fix for that in https://github.com/flutter/flutter/pull/39052.

mit-mit:
Hi @chenglu, I've landed what I think is a fix for this. Can you try to update to latest flutter master and see if works now?

chenglu:
Hi @mit-mit:
Yes, we have confirmed this, it works now.
Thanks for your time to all of you, I will close this PR, thanks!

PR #2921, opened 2019-08-19
Updating the news to 1.7.

sfshaza2:
@timsneath, should we also update the second news item? That iOS one has been there awhile.


sfshaza2:
Tim is on vacay for a couple weeks. Vectoring to @maguinis...

maguinis:
Yes, this LGTM. Thanks.

PR #2929, opened 2019-08-22
troubleshooting possible vscode debug failure

pashute:
In cases where some other use of this emulator was done (e.g. in React Native etc.)

### DIFF
--- a/src/docs/get-started/test-drive/_vscode.md
+++ b/src/docs/get-started/test-drive/_vscode.md
@@ -44,4 +44,8 @@ contains a simple demo app that uses [Material Components][].
 [Quickly switching between Flutter devices]: https://dartcode.org/docs/quickly-switching-between-flutter-devices
 [status bar]: {% asset tools/vs-code/device_status_bar.png @path %}
 
+
+## Troubleshooting
+
+If you get a ```adb: error: failed to copy...``` go to the Android Studio ADV Device manager, select the device, and clear the cache by selecting the small arror on the right and choosing "Clear cache". 
 </div>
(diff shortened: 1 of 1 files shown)


src/docs/get-started/test-drive/_vscode.md
```diff
@@ -44,4 +44,9 @@ contains a simple demo app that uses [Material Components][].
 [Quickly switching between Flutter devices]: https://dartcode.org/docs/quickly-switching-between-flutter-devices
 [status bar]: {% asset tools/vs-code/device_status_bar.png @path %}
 
+
+## Troubleshooting
+
+If you get an ```adb: error: failed to copy...``` go to the Android Studio AVD Android Virtual Device manager, select your device, and clear the cache by selecting the dropdown on the right and choosing "Wipe data".
```
domesticmouse: I feel like telling a VSCode user that the solution to a misbehaving Android device is to open Android Studio is less than ideal. Any chance of raising a bug against [VSCode Flutter plugin](https://github.com/Dart-Code/Flutter) that covers this situation? 

/cc @DanTup for visibility

DanTup: I'm not sure in what circumstances you would need to do this, but VS Code doesn't directly work with the emulators so we'd need to expose a helper to do this in the Flutter tools daemon first (which would then call something like `emulator -avd flutter_emulator -wipe-data`), then VS Code could call that.

That said, I don't think we want to start trying to reproduce all the AVD functionality into Flutter and VS Code as it's a ton of work and VS Code is supposed to only be "an editor". I think it may only make sense to try and wire this up if it's a common issue, but I don't have enough information to know if that's the case.

Right now, there are several things that VS Code users have to use Android Studio for (for ex. things like the asset manager). It's not ideal, but I don't think it's something that is likely to change in the near future.

sfshaza2: There are also some errors with this edit. For example:
* Line length exceeds 80 chars
* Inline code fonting uses only one tick mark.
* UI text is shown in bold.

Issue #1676, opened 2018-11-02
'Inside Flutter' page issue


Rahiche:
Page URL: https://flutter.io/docs/resources/inside-flutter.html
Page source: https://github.com/flutter/website/tree/dash/src/docs/resources/inside-flutter.md

in Inside Flutter page  i see this 

Notice that the **_child’s_** layout cannot depend on the **_child’s_** position because the **_child’s_** position is not determined until after the child returns from layout

i am not an English expert but maybe this can be delivered better ?

Purukitto:
1. `Note that the child's layout cannot depend on its position, as the position is described by the parent only after it returns from the layout.`
  Rephrased the sentence to be grammatically correct, understandable and proper use of pronouns and voice. Should be fine as the page had several blocks of text following similar layout.

2. `Note that the child's layout cannot depend on the child's position, as the position is described by the child's parent only after it returns from the layout.`
  Similar to the above but kept the nouns instead of using pronoun's, might result in being more legible in some case. Is still grammatically correct.

Purukitto:
I have gone ahead and made a pull request using the first option

Rahiche:
I guess this should be closed now

PR #2933, opened 2019-08-24
#1676

Purukitto:


Purukitto:
@googlebot I signed it!

src/docs/resources/inside-flutter.md
```diff
@@ -66,10 +66,10 @@ are a min and max width and a min and max height. During layout,
 the child determines its geometry by choosing a size within these bounds.
 After the child returns from layout, the parent decides the child's
 position in the parent's coordinate system<sup><a href="#a3">3</a></sup>.
-Notice that the child's layout cannot depend on the child's position
-because the child's position is not determined until after the child
-returns from layout. As a result, the parent is free to reposition
-the child without needing to recompute the child's layout.
+Note that the child's layout cannot depend on its position, as the position
```
sfshaza2: Hi, @Purukitto! You are completely right that the word "child" is overused in the original sentence. However, I think your edit could add some confusion as to what "it" is. It's better to be clear than to eject both extra uses of "child." How about:

Note that the child's layout can't depend on its position,
as the position is not determined until after the child
returns from layout. As a result, the parent is free to reposition
the child without needing to recompute its layout.

Purukitto:
@sfshaza2  hope this works

PR #1237, opened 2018-09-06
[dash] Travis: use `language:dart`, reinstate link checking

chalin:
Fixes #1239
Contributes to #1185

PR #2938, opened 2019-08-27
Add DartPad to flutter.dev

legalcodes:
Last staged 9/26/19: 

https://jt-flutter-website-3.firebaseapp.com (Scroll all the way to the bottom)


maguinis:
This has my approval - I sat next to Jon yesterday as we made these
changes. I proposed adding it to the Install page since that is where our
"Get Started" buttons all lead to and hence it is the path that devs
wanting to try Flutter take. I found this to be a cleaner option than to
add 'Try Flutter' since that could be confused with 'Get Started' (unless
we changed 'Get Started' to 'Download'.

One thing I did mention is that having DartPad on bottom of homepage and
Install page could help us A/B test and if we feel more users are using the
Homepage version then we can alter buttons to point to that one.

Also, since the current time on the install page is very short, it will be
a good measure to see how much longer on average users spend on that page
when we implement DartPad there.

Martin Aguinis | Developer Marketing | Flutter

galeyang:
Talked to @legalcodes  offline. Recommend to divide this into several PRs/issues

- Landing page:
We'd suggest to go ahead --- putting DartPad at the bottom of flutter.dev and monitor the usage. We can track the funnel of (1) page visit, (2) to seeing it, (3) and try it. For the "seeing it part", we need to add extra tracking code.

- Get started and installation parts can be in different PRs.

---
Detailed feedback about Dartpad on the landing page
- Heading "Try Flutter in your browser": text color should be `#4A4A4A`. Same as other headings.
- iframe border color should be: `#cccccc`. Same as other [use cases](https://jt-flutter-website-2.firebaseapp.com/docs/codelabs/layout-basics) on flutter.dev
- It'd good if the sample code (at least the first one) is visually compelling. In the past UX study, new users may raise concerns if UIs shown on the landing page and showcase page are not very polished.
- Maybe add a little comments to explain the code since most users are very new to Flutter/Dart when they see the code on the landing page.
- "Want more practice? Try a codelab." maybe link to the [layout codelab](https://flutter.dev/docs/codelabs/layout-basics) since other codelabs require users to install Flutter SDK first.

maguinis:
LGTM from Marketing. Let's ship!

src/_packages/dartpad_picker/web/dartpad_picker_main.dart
```diff
@@ -0,0 +1,243 @@
...
+  if (isMobile()) {
+    querySelector('.dash-dartpad').style.display = 'none';
+    return;
+  }
+
+  var dartPadHost = querySelector('#dartpad-host');
+  var select = querySelector('#dartpad-select');
+
+  var snippets = [
+    Snippet('Spinning flutter', spinning_logo),
```
johnpryan: nit: "Flutter"?

legalcodes: Thanks!

sfshaza2:
I think the last I saw about this PR, Jon was suggesting that we back out changes to the Install page and only update the home page, by adding DP, with no pointers, to the bottom of the page.

It doesn't feel like there's much point to the PR if that's all it does. We would track how many people *find* the DP instance at the bottom of the page, and that's it.

Thoughts?

legalcodes:
@Sfshaza  Thank you for your thoughts!

The original goal of this PR was to do just that -- add DartPad to the homepage as a standalone PR. The proposal to add it to the install came as a later addition, and we have had extensive discussion about the benefits of modifying the install page in a separate PR from this one @galeyang @RedBrogdon please feel free to chime in!

Of the ~550 lines of code in this PR, only ~30 pertain to modifying `src/docs/get-started/install/index.md` so IMHO this is PR is valid on its own, but I am open to hearing other thoughts.

sfshaza2:
LGTM! Ship it! (I'll let you merge it.)

.gitignore
```diff
@@ -10,7 +10,6 @@ node_modules
 .flutter-plugins
 .packages
 .dart_tool
-# pubspec.lock
```
chalin: Hi @legalcodes @johnpryan et all. Why was it decided that we'd start committing the `pubspec.lock` files contained in the repo?

chalin: Hey all, I created a fresh issue to discuss this (#3107) rather than having only a comment in an old PR. (Sorry for the email notification noise.)

_config.yml
```diff
@@ -116,6 +116,17 @@ alert:
     <i class="fas fa-exclamation-triangle"></i> **Warning:**
   end: </aside>
 
+custom:
```
chalin: We should probably drop this `custom` field/tag at some point.

.gitignore
chalin: Oups, my bad, I wasn't looking at this carefully. You just deleted a commented out line! I'm the one who commented it out in #1237 last year. I didn't realize that it had been that long. I thought that it was a change that had been done in the last 6 months or so. Apologies! 

FYI, I'll still go ahead a remove the `pubspec.lock` files.

PR #2950, opened 2019-08-30
Adding survey

sfshaza2:
Staged: https://sz-flutter-2.firebaseapp.com/docs/codelabs/layout-basics

Gale, I'm wondering if we should add a "too easy" option for the last question?

Also, I don't see the open ended question here.

cc: @kwalrath

### DIFF
--- a/src/docs/codelabs/layout-basics.md
+++ b/src/docs/codelabs/layout-basics.md
@@ -4,13 +4,15 @@ description: "A codelab that teaches basic Flutter layout concepts through DartP
 toc: true
 --- 
 Welcome to the Flutter layout codelab, 
-where you learn how to build a Flutter UI without downloading and installing Flutter or Dart!
+where you learn how to build a Flutter UI without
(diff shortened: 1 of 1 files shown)


galeyang:
@sfshaza2 
- Could you share more what you think about the "too easy" option? 
- Open ended question: If your browser is smaller, you may need to scroll down in the blue box to see the full question set.

sfshaza2:
OOOH! Once I scrolled down in the blue box, I did see the other options for the last question. And I saw the open ended question.

The box gave zero indication that I needed to scroll it.

So, nevermind!

galeyang:
This looks good to me! 

One limitation is that we don't know how many people actually see the survey (exposure). I was trying to figure out how to track it last Friday but apparently it will take a little longer time. So I plan to file another issue to track this. Same for the previous future codelab survey.

With this caveat in mind,  I'm fine with publishing this survey now.

PR #2964, opened 2019-09-05
Add go/performance-best-practices

filiph:


filiph:
I'll merge it, TBR. It's a small change with little implication, and it passed linkcheck. I also need this change to publish the linked article.

PR #2979, opened 2019-09-08
fix typo: react native example async await

wkwksama:
the `_getIPAddress` function must return `data`, isn't it ?


wkwksama:
@googlebot I signed it!

Issue #2941, opened 2019-08-27
Add C interop page


kwalrath:
The FFI docs (see https://github.com/dart-lang/site-www/pull/1849) should point to flutter.dev for Flutter-specific issues.

In the first go-round, dart.dev will point to https://github.com/flutter/flutter/wiki/Binding-to-native-code-via-FFI. But we'd love to transfer that doc to flutter.dev and link to that.

mit-mit:
fyi @sjindel-google

sfshaza2:
@kwalrath, the Flutter. ffi info is now on flutter.dev:
https://flutter.dev/docs/development/platform-integration/c-interop

kwalrath:
Thanks, @sfshaza2, I'll link to your page.

It's a little strange that your page says C/C++ and ours says just C. (Ours used to say C++, but we removed that.) Do you know why yours says C++?

/cc @mjohnsullivan

sjindel-google:
Can we please leave the wiki page for a while? I've definitely pasted that link a bunch of places...

kwalrath:
PR is up: dart-lang/site-www#1954

sjindel-google:
Also, I don't know whether it should say "C" or "C/C++", but the point is that the FFI can bind to a C *interface*, regardless of whether that interface is implemented by C, C++, Swift, Objective-C, Rust or even Gol

kwalrath:
> Also, I don't know whether it should say "C" or "C/C++", but the point is that the FFI can bind to a C _interface_, regardless of whether that interface is implemented by C, C++, Swift, Objective-C, Rust or even Gol

We should be clearer about this, somehow.

mit-mit:
I suggest we just use 'C interop' everywhere

PR #2994, opened 2019-09-13
Adding a page on FFI

sfshaza2:
Staged: https://sz-flutter.firebaseapp.com/docs/development/platform-integration/c-interop

Fixes https://github.com/flutter/website/issues/2941

I was asked to migrate a page from the Flutter wiki to our site. This is not a topic I know much about, so please help me clarify any confusion, errors, or omissions. Once this page is live, we can forward the old page.

Another question: Will this also work with web?

Thanks!

cc @kwalrath @mit-mit

### DIFF
--- a/src/docs/development/platform-integration/c-interop.md
+++ b/src/docs/development/platform-integration/c-interop.md
@@ -92,7 +92,7 @@ the application binary of any downstream app.
 ### Closed-source third-party library
 
 Add a `vendored_frameworks` field to the `.podspec`
-of the Flutter plugin. See the [CocoaPods Example][].
+of the Flutter plugin. See the [CocoaPods example][].
 
 Note that binary code should not be uploaded to
 Pub directly but rather downloaded from a trusted
@@ -138,11 +138,11 @@ not included in your Flutter package.
 
 [Add C and C++ code to your project]: {{site.android-dev}}/studio/projects/add-native-code
 [Android NDK Native APIs]: {{site.android-dev}}/ndk/guides/stable_apis
-[CocoaPods Example]: {{site.github}}/CocoaPods/CocoaPods/blob/master/examples/Vendored%20Framework%20Example/Example%20Pods/VendoredFrameworkExample.podspec
-[Dart API reference documentation]: {{site.dart-api}}/dev/
-[`DynamicLibrary.executable`]: {{site.dart-api}}/dev/dart-ffi/DynamicLibrary/DynamicLibrary.executable.html
-[`DynamicLibrary.open`]: {{site.dart-api}}/dev/dart-ffi/DynamicLibrary/DynamicLibrary.open.html
-[`DynamicLibrary.process`]: {{site.dart-api}}/dev/dart-ffi/DynamicLibrary/DynamicLibrary.process.html
+[CocoaPods example]: {{site.github}}/CocoaPods/CocoaPods/blob/master/examples/Vendored%20Framework%20Example/Example%20Pods/VendoredFrameworkExample.podspec
+[Dart API reference documentation]: {{site.dart.api}}/dev/
+[`DynamicLibrary.executable`]: {{site.dart.api}}/dev/dart-ffi/DynamicLibrary/DynamicLibrary.executable.html
+[`DynamicLibrary.open`]: {{site.dart.api}}/dev/dart-ffi/DynamicLibrary/DynamicLibrary.open.html
+[`DynamicLibrary.process`]: {{site.dart.api}}/dev/dart-ffi/DynamicLibrary/DynamicLibrary.process.html
 [FFI]: https://en.wikipedia.org/wiki/Foreign_function_interface
 [ffi issue]: {{site.github}}/dart-lang/sdk/issues/34452
 [Upgrading Flutter]: /docs/development/tools/sdk/upgrading
--- a/src/docs/development/platform-integration/c-interop.md
+++ b/src/docs/development/platform-integration/c-interop.md
@@ -3,7 +3,7 @@ title: "Binding to native code using dart:ffi"
 description: "To use C code in your Flutter program, use the dart:ffi library (currently in preview)."
 ---
 
-Flutter mobile can use the dart:ffi library
+Flutter mobile can use the [dart:ffi][] library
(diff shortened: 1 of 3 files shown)


sjindel-google:
Thanks for doing this! It does not work on the Web yet.

I forwarded you some feedback from users who are actually very confused by the documentation, and I think we should consider improving it before publishing it on the website.

/cc @mkustermann @dcharkes

sfshaza2:
@sjindel-google, I would agree with the assessment that the wiki page is highly confusing. It appears that there are many variables when connecting to different kinds of libraries, and on different target platforms. Also, you can connect directly or by creating a plugin.

I would suggest that we narrow the options down and select a few recommended ways to do this - typical scenarios. With your input/feedback/guidance, we can restructure the page completely, and make it more focused on the steps required. But I'm not quite sure yet, where to begin.

sjindel-google:
I tried to organize it by use cases originally, but I didn't really describe what those are. Here's what I had in mind:

Use case 0:

Binding against a "system" library which is shipped with the target device.
Examples include binding against SQLite (on iOS) or GLES (on Android).

Use case 1:

First-party library. This is a library which the App developer has and wants to include in the App. I think we should not address this use case directly since it's a special case of (2) and (3).

Use case 2:

Open-source third-party library. The developer wants to create a Flutter plugin which includes both C/C++/Obj-C and Dart source code.

Use case 3:

Closed-source third-party library. The developer wants to create a Flutter plugin which includes Dart source code (necessary) but distribute the C/C++ library in binary form.

We don't want to encourage this use case in general, and we definitely don't want users putting binary artifacts on Pub. However, we should still provide guidance on doing this with the binary artifacts published through a third-party website, because some important users (e.g. Realm) may need it.

sfshaza2:
Hi, @sjindel-google! I've taken a stab at making the doc more clear. Can you take a look? I've embedded some questions for you in bold text.  Thanks!!

src/docs/development/platform-integration/c-interop.md
```diff
@@ -0,0 +1,202 @@
+---
+title: "Binding to native code using dart:ffi"
+description: "To use C code in your Flutter program, use the dart:ffi library (currently in preview)."
+---
+
+Flutter mobile can use the [dart:ffi][] library
+to call native C APIs. The _ffi_ acronym stands for
```
sjindel-google: _FFI_

src/docs/development/platform-integration/c-interop.md
```diff
@@ -0,0 +1,202 @@
...
+  $ flutter upgrade
+  ```
+  For more information on Flutter's channels,
+  see [Upgrading Flutter][].
+{{ site.alert.end }}
+
+API documentation is available from the Dart dev channel:
+[Dart API reference documentation][].
+
+## iOS and macOS
```
sjindel-google: I think a forward is needed, like:

```
Before your library or program can use the FFI to bind to native code, you must ensure that the native code is loaded and its symbols are visible to Dart.

Thus this page focuses on compiling, packaging and loading native code within a Flutter plugin or app.
```

src/docs/development/platform-integration/c-interop.md
```diff
@@ -0,0 +1,202 @@
...
+  ```terminal
+  $ flutter channel dev
+  $ flutter upgrade
+  ```
+  For more information on Flutter's channels,
+  see [Upgrading Flutter][].
+{{ site.alert.end }}
+
+API documentation is available from the Dart dev channel:
+[Dart API reference documentation][].
```
sjindel-google: I think we need a brief introduction to linking at be start:

A native library can be linked into an app either dynamically or statically. A statically linked library is embedded into the app's executable image, and will be loaded when the app starts.

Symbols from a statically linked library can be loaded via `DynamicLibrary.executable` or `DynamicLibrary.process`.

A dynamically linked library, by contrast, is distributed in a separate file or folder within the app, and loaded on-demand. On Android, a dynamically linked library is distributed as a set of `.so` (ELF) files, one for each architecture. On iOS, it's distributed as a `.framework` folder.

A dynamically linked library can be loaded into Dart via `DynamicLibrary.open`.

src/docs/development/platform-integration/c-interop.md
kwalrath: Technically it's an initialism, not an acronym. :) But I'd just say "_FFI_ stands for..."

sjindel-google:
I've re-worked the wiki page into a "Walkthrough": https://github.com/flutter/flutter/wiki/Binding-to-native-code-via-FFI

Would you mind taking a look at that instead? I think it will be much more helpful to most users.

sfshaza2:
@sjindel-google, please take another look! It's staged on:

Staged: https://sz-flutter.firebaseapp.com/docs/development/platform-integration/c-interop

src/docs/development/platform-integration/c-interop.md
```diff
@@ -0,0 +1,342 @@
...
+systems about the native code so the code can be compiled
+and linked appropriately into the final application.
+
+You add the sources to the `ios` folder,
+because CocoaPods doesn't allow including sources
+above the podspec file, but Gradle allows you to point
+to the `ios` folder. It's not required to use the same
+sources for both iOS and Android;
+you may, of course, add Android-specific sources
+to the `android` folder and modify the `CMakeLists.txt`
```
sjindel-google: "modify `CMakeLists.txt` appropriately"

src/docs/development/platform-integration/c-interop.md
```diff
@@ -0,0 +1,342 @@
...
+file appropriately.
+
+The FFI library can only bind against C symbols,
+so in C++ these symbols must be marked "extern C".
```
sjindel-google: `extern C`

src/docs/development/platform-integration/c-interop.md
```diff
@@ -0,0 +1,342 @@
...
+You should also add attributes to indicate that the
+symbols are referenced from Dart,
+to prevent a linker from discarding the symbols
+during link-time optimization.
+
+For example:
+
+```
```
sjindel-google: This quote is Bash, a command-line which can be run from the root of the directly. Can we make it show up with bash highlighting in the website?

PR #2798, opened 2019-07-02
Create guide for adding a FlutterFragment to an existing Android app.

matthew-carroll:
Create guide for adding a FlutterFragment to an existing Android app.

Staged here:
https://flutter-website.firebaseapp.com/docs/development/platform-integration/add-to-app-android/add-flutter-fragment

PR #3000, opened 2019-09-14
Remove non-existent links

xster:
Took out 2 links to pngs that aren't checked-in. 

Otherwise, I can't build it on firebase.


xster:
@sfshaza2, did you mean to close this PR?

xster:
It's all in https://github.com/flutter/website/pull/3004 now. @matthew-carroll, let me know if it's ok to remove these non-existent links (or maybe I failed to recover these pngs correctly?)

matthew-carroll:
Do we know what happened to those PNGs?

xster:
I'm not sure. But it wasn't in https://github.com/flutter/website/pull/2798/files either. Was it from a different PR?

matthew-carroll:
It doesn't look like those other images are in there. But that's OK, I'll just add them back at some point.

sfshaza2:
I don't know what happened to those images. I don't *think* I was
responsible for those disappearing.

PR #3009, opened 2019-09-17
Adding a banner

sfshaza2:
Staged:  https://sz-flutter-2.firebaseapp.com/docs#latest-release

### DIFF
--- a/src/_includes/banner.html
+++ b/src/_includes/banner.html
@@ -5,7 +5,9 @@
     {% comment %} For headings use <h3> elements. {% endcomment -%}
       Flutter 1.9 is live! See what's new in
       <a href="https://developers.googleblog.com/2019/09/flutter-news-from-gdd-china-flutter1.9.html?m=1"
-         target="_blank">Flutter news from GDD China</a>.
+         target="_blank">Flutter news from GDD China</a>.<br>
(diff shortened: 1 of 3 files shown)


src/_includes/banner.html
```diff
@@ -5,7 +5,9 @@
     {% comment %} For headings use <h3> elements. {% endcomment -%}
       Flutter 1.9 is live! See what's new in
       <a href="https://developers.googleblog.com/2019/09/flutter-news-from-gdd-china-flutter1.9.html?m=1"
-         target="_blank">Flutter news from GDD China</a>.
+         target="_blank">Flutter news from GDD China</a>.<br>
+         See <a href="/docs#latest-release">What's new</a>
```
kwalrath: This seems long for a banner, but I can't see it on the staging server. Did you overwrite it with another PR?

kwalrath: The repetition of "See" and "docs" and "what's new" seems unfortunate. I'm having a hard time rewriting it, though. But here's something that might be better:

Also see: [What's new on the site]

I feel like if you want to call out the web docs, you should just link to them. But do we really want to do that?

kwalrath: Maybe "the site" -> "this site" (in both the banner and the section heading)?

PR #3013, opened 2019-09-18
Do not modify global defaultLocale

goderbauer:
Prior to this change we were telling people to modify the global `Intl.defaultLocale` when their custom `Localizations` for a new locale are loaded. That's a bad idea because Flutter allows that different parts of the same app use different localizations. For example, let's say your global locale is `de`. Now you have one screen in the app that uses a locale of `en` via the `Localizations.override` widget. When the first locale is loaded, `Intl.defaultLocale` is set to `de` and everything in your app appears in German as expected. When you navigate to the screen that has the `Localizations.override`, the `en` local is loaded and `Intl.defaultLocale` is set to `en` forcing the entire (!) app to English, even the parts outside of the `Localizations.override`.

This change no longer recommends the usage of `Intl.defaultLocale` and instead holds on to the locale applicable to the current Localizations instance in an instance variable.

Fixes https://github.com/flutter/flutter/issues/40353.

### DIFF
--- a/examples/internationalization/intl_example/lib/main.dart
+++ b/examples/internationalization/intl_example/lib/main.dart
@@ -44,13 +44,14 @@ import 'package:intl/intl.dart';
 import 'l10n/messages_all.dart';
 
 class DemoLocalizations {
+  DemoLocalizations(this.localeName);
+
   static Future<DemoLocalizations> load(Locale locale) {
     final String name = locale.countryCode.isEmpty ? locale.languageCode : locale.toString();
     final String localeName = Intl.canonicalizedLocale(name);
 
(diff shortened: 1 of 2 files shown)


examples/internationalization/intl_example/lib/main.dart
```diff
@@ -44,25 +44,29 @@ import 'package:intl/intl.dart';
...
   }
 
+  final String localeName;
+
   String get title {
     return Intl.message(
       'Hello World',
       name: 'title',
       desc: 'Title for the Demo application',
+      locale: localeName,
```
HansMuller: Good!

PR #3026, opened 2019-09-24
Update mocking.md isInstanceOf to TypeMatcher

CaseyHillers:
isInstanceOf was deprecated.

### DIFF
--- a/src/docs/cookbook/testing/unit/mocking.md
+++ b/src/docs/cookbook/testing/unit/mocking.md
@@ -141,7 +141,7 @@ main() {
       when(client.get('https://jsonplaceholder.typicode.com/posts/1'))
           .thenAnswer((_) async => http.Response('{"title": "Test"}', 200));
 
-      expect(await fetchPost(client), TypeMatcher<Post>());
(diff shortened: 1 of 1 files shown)


src/docs/cookbook/testing/unit/mocking.md
```diff
@@ -141,7 +141,7 @@ main() {
       when(client.get('https://jsonplaceholder.typicode.com/posts/1'))
           .thenAnswer((_) async => http.Response('{"title": "Test"}', 200));
 
-      expect(await fetchPost(client), isInstanceOf<Post>());
+      expect(await fetchPost(client), TypeMatcher<Post>());
```
a14n: nit: You can use `const TypeMatcher<Post>()`

CaseyHillers: Added

a14n: With matcher >= 0.12.5 you can also use [isA](https://pub.dev/documentation/matcher/latest/matcher/isA.html)

```dart
expect(await fetchPost(client), isA<Post>());
```

PR #3050, opened 2019-10-02
hide debug banner in front page samples

johnpryan:
this flag hides the 'debug mode' banner in the second two samples. (The first sample already sets this flag)


legalcodes:
LGTM, just waiting on checks to finish

Issue #3046, opened 2019-09-30
'Technical overview' page issue


kwalrath:
Page URL: https://flutter.dev/docs/resources/technical-overview#layer-cakes-are-delicious
Page source: https://github.com/flutter/website/tree/master/src/docs/resources/technical-overview.md

The SVG layer cake image uses title case; it should use sentence case. ~~But since it's SVG, it should be easy to fix!~~ I thought the text would be easily editable in the image source, but it isn't.

I noticed this because something else I've been editing reused this figure.

kwalrath:
Image source: https://github.com/flutter/website/tree/master/src//images/whatisflutter/diagram-layercake.svg

sfshaza2:
FYI:
- I downloaded Inkscape from the internal app place, and it won't launch on my Mac.
- Tried to fix this in the raw file - but no strings appear. They must use Unicode values or something.
- Tried to fix this in Omnigraffle - it wouldn't let me change the text.

sfshaza2:
OK, Ian took a look at the SVG file and says that the text is pre-vectorized, so it's not really text, it's shapes. This image needs to be recreated.

sfshaza2:
Fixed in PR https://github.com/flutter/website/pull/3067

PR #3067, opened 2019-10-07
Use sentence case for image

sfshaza2:
Replacing an SVG image with a PNG. The new image resides in _assets and uses sentence case.

Fixes https://github.com/flutter/website/issues/3046


kwalrath:
Should we put the Omnigraffle file somewhere in the repo, in case we ever need to change this again (e.g. if colors change)?

PR #3074, opened 2019-10-09
Allow check-code.sh to check code again

chalin:
Currently `check-code.sh` doesn't pretty much nothing. Move the `--no-check-code` option from the check-code.sh script into Travis config so that we can run `check-code.sh` manually over a local repo and have it actually check code once again. This PR has no net change relative to the Travis build.

This PR is a small step toward reinstating CI code checks.

cc @johnpryan


chalin:
Contributes to #3076

PR #3096, opened 2019-10-12
Added CD Option

nabilnalakath:
- Added Github Actions to CD Options
- Added example project link for CD using Git Hub actions


nabilnalakath:
@googlebot I signed it!

PR #3102, opened 2019-10-14
use site.custom instead of custom for dartpad URL

### LINKED ISSUE #3098: 'Basic Flutter layout concepts' page issue :p0-critical:
Page URL: https://flutter.dev/docs/codelabs/layout-basics
Page source: https://github.com/flutter/website/blob/master/src/docs/codelabs/layout-basics.md

Description of issue:
[Basic Flutter Layouts](https://flutter.dev/docs/codelabs/layout-basics) appears to be embedding a copy of itself inside the iframes where dartpad should be embedded.

It appears to be recursive and continues to embed copies of itself in nested iframes.  This causes massive memory usage and eventually, the page crashes or the entire OS locks-up.

Tested on Ubuntu 19.04 with current releases of Chrome, Chromium, and Firefox using no extensions.  Same issue across browsers.

johnpryan:
fixes #3098

### DIFF
--- a/src/docs/codelabs/layout-basics.md
+++ b/src/docs/codelabs/layout-basics.md
@@ -51,7 +51,7 @@ and `Column` lays out its widgets vertically.
 {% comment %}
   Gist: https://gist.github.com/4e11c4a7ec824685f963f25d7c30ba0b
 {% endcomment %}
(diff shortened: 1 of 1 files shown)


legalcodes:
Thanks @johnpryan for this fix.

PR #3105, opened 2019-10-15
Swift and Kotlin is the default for flutter create

digitaljoni:
Just an update to the website that Swift and Kotlin is the default option when creating a new flutter plugin package


digitaljoni:
@googlebot I signed it!

Issue #3107, opened 2019-10-15
Don't commit pubspec.lock files for tools and examples


chalin:
From: https://github.com/flutter/website/pull/2938/files#r335132003
> Hi @legalcodes @johnpryan et all. Why was it decided that we'd start committing the `pubspec.lock` files contained in the repo? 

I feel that having to maintain and commit the `pubspec.lock` files for this website repo introduces unnecessary churn. It usually makes sense to commit lock files for libraries. I'm not so sure that it is necessary for the tools and examples of this repo. Let me know what you think.

johnpryan:
Sounds good to me. When breakages are caused by a package upgrade it's better to pin the version in pubspec.yaml IMO (instead of updating the lockfile.)

chalin:
@legalcodes?

legalcodes:
Sure, I'm fine with it, thanks Patrice!
Best,
Jon

sfshaza2:
Can this be closed?

johnpryan:
Not yet, the fix will be to remove the lockfiles

PR #3108, opened 2019-10-15
Refresh pubspec.lock files

chalin:
I don't think that we should be committing the lock files (#3107), but since they are present, this updates them.


chalin:
Travis is green and this is a trivial infra-only PR. Merging (since I have followup PRs wanting to use the latest package version).

PR #3109, opened 2019-10-15
Check code-excerpt freshness

### LINKED ISSUE #3073: Revert or simplify recent change in Animation tutorial code
Hey @Piinks! This is a followup to #2842, in which you addressed a problem reported by https://github.com/flutter/flutter/issues/32772.

How were you able to reproduce the problem originally reported by https://github.com/flutter/flutter/issues/32772? I can't get the flutter analyzer to report an issue with the original code. I'm using the latest stable flutter (Flutter 1.9.1+hotfix.4 • channel stable).

The code updates of #2842, consisted in adding a cast:

```dart
final Animation<double> animation = listenable as Animation<double>;
```

The main point that I wanted to raise is that, IMHO, we don't need _both_ the cast and the type declaration, either should suffice by itself. So I was curious how you managed to get the analyzer to report an issue on the original code, which didn't have the cast.

The person who filed https://github.com/flutter/flutter/issues/32772 was using a beta version of flutter. Could the analyzer issue have only been transient (specific to that beta version), and so no longer be present in stable?

If we can't reproduce the problem with a stable version of flutter, then I suggest that we revert #2842. If we can reproduce the problem, then I suggest that we drop the LHS type declaration since type inference should kick in (given that the variable is `final`).

Thanks!

@legalcodes @johnpryan @sfshaza2: thoughts?

(Btw, I noticed this issue after manually running the code refresh script.)

### LINKED ISSUE #3075: CI: ensure that code-excerpt freshness is checked
- [ ] Re-adjust `tool/refresh-code-excerpts.sh` (following changes made in #2353) so that it exits with 1 when some excerpts are refreshed.
- [ ] Invoke `tool/refresh-code-excerpts.sh` from `tool/check-code.sh`

(More work needs to be done to `tool/refresh-code-excerpts.sh` so that it doesn't pick one `pub` version over another, but we'll address that in the scope of #3087.)

chalin:
Closes #3075

- tool/refresh-code-excerpts.sh: exit with 1 if code excerpts were refreshed
  (Readjustment following #2353.)
- tool/check-code.sh: also check code excerpt freshness

The code-excerpt freshness check will fail (it is a known failure), but the code check is currently allowed to fail so that's ok. (And, by having a known failure, we're testing that the script is working for that scenario.)

I'll refresh the code excerpts in a followup PR before or after we decide how to address #3073.

### DIFF
--- a/tool/check-code.sh
+++ b/tool/check-code.sh
@@ -2,6 +2,8 @@
 
 set -e # Fast fail the script on failures.
 
+readonly rootDir="$(cd "$(dirname "$0")/.." && pwd)"
+
 [[ -z "$DART_SITE_ENV_DEFS" ]] && . ./tool/env-set.sh
 
 if [[ $1 == --clean ]]; then
@@ -15,4 +17,15 @@ if [[ $1 == --help || $1 == -h ]]; then
   exit 0
 fi
 
-exec ./tool/build_check_deploy.sh --no-build --no-check-links $*
+errorMessage="
+Error: some code excerpts need to be refreshed.
+Rerun '$rootDir/tool/refresh-code-excerpts.sh' locally.
+"
+
+travis_fold start refresh_code_excerpts
+(set -x; $rootDir/tool/refresh-code-excerpts.sh) || (printf "$errorMessage" && exit 1)
+travis_fold end refresh_code_excerpts
+
+travis_fold start check_code
+./tool/build_check_deploy.sh --no-build --no-check-links $*
+travis_fold end check_code
--- a/tool/refresh-code-excerpts.sh
+++ b/tool/refresh-code-excerpts.sh
@@ -68,7 +68,7 @@ LOG_FILE="$TMP/refresh-code-excerpts-log.txt"
   "$SRC" 2>&1 | tee $LOG_FILE
 LOG=$(cat $LOG_FILE)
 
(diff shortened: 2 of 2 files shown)


chalin:
As expected, the [check-code job](https://travis-ci.org/flutter/website/jobs/598325993) fails because code excerpts aren't fresh:

```
Source:     /home/travis/build/flutter/website/src
Fragments:  /home/travis/build/flutter/website/tmp/_fragments/examples
Other args: --yaml --no-escape-ng-interpolation --replace=/\/\/!<br>//g;/ellipsis(<\w+>)?(\(\))?;?/.../g;/\/\*(\s*\.\.\.\s*)\*\//$1/g;/\{\/\*-(\s*\.\.\.\s*)-\*\/\}/$1/g;/\/\/!(analysis-issue|runtime-error)[^\n]*//g;/\x20*\/\/\s+ignore_for_file:[^\n]+\n//g;/\x20*\/\/\s+ignore:[^\n]+//g;
Processed 239 Dart/Jade/Markdown files: 3 out of 83 fragments needed updating.
...
Error: some code excerpts need to be refreshed.
Rerun '/home/travis/build/flutter/website/tool/refresh-code-excerpts.sh' locally.
```

legalcodes:
LGTM

PR #3110, opened 2019-10-15
Animation tutorial: drop unnecessary local var type decl

### LINKED ISSUE #3073: Revert or simplify recent change in Animation tutorial code

chalin:
- Closes #3073
- Followup to #2842
- Contributes to #3075.
- Refresh code excerpts.

cc @Piinks

### DIFF
--- a/examples/animation/animate2/lib/main.dart
+++ b/examples/animation/animate2/lib/main.dart
@@ -9,7 +9,7 @@ class AnimatedLogo extends AnimatedWidget {
       : super(key: key, listenable: animation);
 
   Widget build(BuildContext context) {
-    final Animation<double> animation = listenable as Animation<double>;
+    final animation = listenable as Animation<double>;
     return Center(
       child: Container(
         margin: EdgeInsets.symmetric(vertical: 10),
--- a/examples/animation/animate3/lib/main.dart
+++ b/examples/animation/animate3/lib/main.dart
@@ -8,7 +8,7 @@ class AnimatedLogo extends AnimatedWidget {
       : super(key: key, listenable: animation);
 
   Widget build(BuildContext context) {
-    final Animation<double> animation = listenable as Animation<double>;
+    final animation = listenable as Animation<double>;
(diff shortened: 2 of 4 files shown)


chalin:
FYI, I ran check using this version of flutter:

```console
$ flutter --version
Flutter 1.10.15-pre.131 • channel master • https://github.com/flutter/flutter
Framework • revision 3f1f22c26e (38 minutes ago) • 2019-10-15 13:34:04 -0700
Engine • revision 540fc977bb
Tools • Dart 2.6.0 (build 2.6.0-dev.7.0 70a7ef3f58)
```

And the analyzer was happy, for example:

```console
$ flutter analyze
Analyzing animate2...                                                   
No issues found! (ran in 2.0s)
```

PR #3114, opened 2019-10-16
update fetch-data article

### LINKED ISSUE #3106: 'Fetch data from the internet' spawns the network request during build
Page URL: https://flutter.dev/docs/cookbook/networking/fetch-data.html
Page source: https://github.com/flutter/website/tree/master/src/docs/cookbook/networking/fetch-data.md

Step 4 does this:
```dart
FutureBuilder<Post>(
  future: fetchPost(),
  builder: (context, snapshot) {
    if (snapshot.hasData) {
      return Text(snapshot.data.title);
    } else if (snapshot.hasError) {
      return Text("${snapshot.error}");
    }

    // By default, show a loading spinner.
    return CircularProgressIndicator();
  },
);
```

This is bad, because it means that every time this widget rebuilds (e.g. when you rotate the phone, or when any nearby widget rebuilds), it'll trigger a new network request.

The network request should be done in `initState`.

johnpryan:
This removes code that fetches data in build methods. We've gotten some feedback indicating that this is confusing for readers.

fixes #3106

### DIFF
--- a/src/docs/cookbook/networking/fetch-data.md
+++ b/src/docs/cookbook/networking/fetch-data.md
@@ -126,68 +126,8 @@ Future<Post> fetchPost() async {
 Hooray! Now you've got a function that fetches a post from the
 internet.
 
-## 4. Fetch and display the data
-
-To fetch the data and display it on screen, use the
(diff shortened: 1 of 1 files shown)


sfshaza2:
Lgtm!

PR #3130, opened 2019-10-20
[package.json] Flutter not Dart for name, description, and url

SamuelMarks:


sfshaza2:
@chalin, this looks good to me, can you confirm?

PR #3131, opened 2019-10-20
Add step for importing packages after dependency installation

andreidobrinski:
I was going through the steps when I realized that I had to scroll down to the final example to find the import statements.

Hopefully this makes it easier to follow and code along with sequentially :)


andreidobrinski:
@googlebot I signed it!

PR #3132, opened 2019-10-20
Add instructions for updating the version number

rohan20:
A lot of Flutter users are publishing their apps to the Play store but finding a way to update the version number and name isn't straightforward. 

Hence, this answer is gaining more attention every week: https://stackoverflow.com/a/56970752/5066615

Even though the instructions are directly available in `pubspec.yaml`, most users like me couldn't find it easily initially.  

Hence, this PR uses the same words from the `pubspec.yaml` on updating the version number of the app.


rohan20:
@googlebot I signed it!

PR #3139, opened 2019-10-22
Make the spinning curve more natural

filiph:
`easeInOutCubic` looks better, especially at the end of the animation.

From the API doc of [easeInOutCubic](https://api.flutter.dev/flutter/animation/Curves/easeInOutCubic-constant.html):

> The result is a safe sweet spot when choosing a curve for a widget whose initial and final positions are both within the viewport.


legalcodes:
Awesome, many thanks for this Filip!

PR #3141, opened 2019-10-22
(completely optional) Use BigInt for fibonacci computation

filiph:
I'm not sure this is a good idea, from educational perspective. But it does show that we have `BigInt`. And it shows attention to detail — Dart does think of big integers.

This will continue computing fib until the number fits into memory. Nobody will scroll that far.

The current version shows "seams" when you scroll far enough -- the `int` becomes JavaScript's `num` in the background, and switches to floating point when the int overflows, so you get numbers like 1.234234+e113.

In contrast, the new version shows:

Screen Shot 2019-10-22 at 11 31 53 AM

And then later:

Screen Shot 2019-10-22 at 11 32 04 AM


sfshaza2:
Not @legalcodes, but it LGTM

filiph:
Oh sorry, I missed `compile.sh` completely. I ran it and I think it also compiled the previous change to the flutterpad samples (https://github.com/flutter/website/pull/3139).

PR #3150, opened 2019-10-24
Update fetch-data.md

mit-mit:
Reformat the import step like other Dart code, as it's very easy to miss right now (I just did myself :-)


mit-mit:
That was quick! Sleep well :-)

PR #3197, opened 2019-11-02
Update internationalization.md

jsroest:
The casing should be the other way around, language in lowercase, country in uppercase.
see: https://en.wikipedia.org/wiki/Language_localisation


jsroest:
@googlebot I signed it!

Op za 2 nov. 2019 om 16:16 schreef googlebot <notifications@github.com>:

PR #3230, opened 2019-11-11
Update json_serializable example

includecmath:
`toJson` and `fromJson` code will be generated not `fromMappedJson`


sfshaza2:
It looks good to me, but can you verify @johnpryan?

PR #3265, opened 2019-11-19
Add an add-to-app landing page

xster:


src/_data/sidenav.yml
```diff
@@ -145,6 +145,8 @@
     - title: Add Flutter to existing app
       permalink: /docs/development/add-to-app
       children:
+      - title: Introduction
+        permalink: /docs/development/add-to-app/index
```
sfshaza2: You shouldn't have to specify index. Just /docs/development/add-to-app should be sufficient.

xster: ah sg, thanks!

PR #3271, opened 2019-11-19
Document new breaking change policy

Hixie:
This is the developer-facing side of our documentation for our new breaking-change policy.

### DIFF
--- a/src/docs/resources/compatibility.md
+++ b/src/docs/resources/compatibility.md
@@ -1,0 +1,57 @@
+---
+title: Flutter compatibility policy
+description: How Flutter approaches the question of breaking changes.
+---
+
+The Flutter team tries to balance the need for API stability with the
+need to keep evolving APIs to fix bugs, improve API ergonomics, and
+provide new features in a coherent manner.
+
+To this end, we have created a test registry where you can provide
+unit tests for your own applications or libraries that we should run
+on every change to help us track down when changes would break
+existing applications. Our commitment is that we will not make any
+changes that break these tests without working with the developers of
+those tests to (a) determine if the change is sufficiently valuable
+and (b) provide fixes for the code so that the tests continue to pass.
+
+If you would like to provide tests as part of this program, please
+submit a PR to the [flutter/tests
+repository][https://github.com/flutter/tests]. The README.md file on
+that repository describes the process in detail.
+
+
+## Announcements and migration guides
+
+If we do make a breaking change (defined as a change that caused one
+or more of these submitted tests to require changes), we will announce
+the change on our
+[flutter-announce][https://groups.google.com/forum/#!forum/flutter-announce]
+mailing list as well as in our release notes.
+
+We provide a list of [guides for migrating
+code][https://flutter.dev/docs/release/breaking-changes/] affected by
+breaking changes.
+
+
+## Deprecation policy
+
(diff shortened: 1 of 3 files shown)


Hixie:
cc @tvolkert @timsneath

src/docs/resources/compatibility.md
```diff
@@ -0,0 +1,57 @@
...
+be (if it breaks any of the submitted tests).
+
+
+## Dart and other libraries used by Flutter
+
+The Dart language itself has a separate breaking-change policy,
+[documented on the Dart
+wiki][https://github.com/dart-lang/sdk/blob/master/docs/process/breaking-changes.md].
+
+In general, the Flutter team does currently have any commitment
```
sfshaza2: Did you mean "does" or "does not"?

timsneath: does not?

src/docs/resources/compatibility.md
```diff
@@ -0,0 +1,57 @@
...
+The Flutter team tries to balance the need for API stability with the
+need to keep evolving APIs to fix bugs, improve API ergonomics, and
+provide new features in a coherent manner.
+
+To this end, we have created a test registry where you can provide
+unit tests that we should run on every change to help us track down
+when changes would break existing applications. Our commitment is that
+we will not make any changes that break these tests without working
+with the developers of those tests to (a) determine if the change is
+sufficiently valuable and (b) provide fixes for the code so that the
```
timsneath: I'm a bit confused by this philosophy. It _sounds_ from a) as if the first person to contribute a specific test gets special treatment in determining the value of a breaking change: is that what you mean? I wonder if contributing a unit test automatically gives someone a good view of the impact of that break?

I think b) is the key here: we want to communicate that we'll never arbitrarily break a test without fixing the test (in the same way that we don't just break google3). So contributing a test both acts as (i) a canary for inadvertent breaks plus (ii) a path to a documented fix for a breaking change. Does that sound about right?

src/docs/resources/compatibility.md
Hixie: oops, good catch

src/docs/resources/compatibility.md
Hixie: not sure what you mean by "first person". Anyone can contribute their app's tests, and then we won't break those tests without talking to them first.

timsneath: Ah... it wasn't clear to me at first that you were asking folk to contribute a link to _their app test ***suite***_. Without reading through the actual contributor guide, it looked like you're asking for individual unit tests. I was reading it as implying that I might (for instance) contribute a test for `FloatingActionButton`, and based on that test, if we broke the FAB you'd ask me about whether the change was sufficiently valuable.

Hixie: Ah yeah that wasn't clear. I've clarified the text to say:

> To this end, we have created a test registry where you can provide unit tests for your own applications or libraries that we should run on every change to help us track down when changes would break existing applications.

src/docs/resources/compatibility.md
Hixie: fixed

src/docs/resources/compatibility.md
timsneath: Great -- that totally resolves my confusion!

PR #3287, opened 2019-11-22
Fix VS Code link in FAQs

philipproplesch:
Minor fix to address an invalid link


philipproplesch:
@googlebot I signed it!

PR #3291, opened 2019-11-24
Fix links

pocari:
I changed the footer link so that it is in follwing page
https://flutter.dev/docs/cookbook/design


pocari:
@googlebot I signed it!

PR #3294, opened 2019-11-24
Update json.md

includecmath:
In my opinion, it is more appropriate to use variables <latest_version>.

### DIFF
--- a/src/docs/development/data-and-backend/json.md
+++ b/src/docs/development/data-and-backend/json.md
@@ -236,12 +236,12 @@ following the [pubspec file][] in the JSON serializable example.
 ```yaml
 dependencies:
   # Your other regular dependencies here
-  json_annotation: ^2.0.0
+  json_annotation: <latest_version>
 
 dev_dependencies:
   # Your other dev_dependencies here
-  build_runner: ^1.0.0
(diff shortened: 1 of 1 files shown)


includecmath:
Hi @sfshaza2 this file seems no default reviewers.

legalcodes:
Thanks for raising this issue @includecmath we are looking into how best to handle this case, will try to update this in about a week.

sfshaza2:
Hi, @includecmath! Using the latest version is not always the best case. APIs can change between versions and break your code. You can read more about it on dart.dev and the site it links to that talks about semantic versioning:  https://dart.dev/tools/pub/pubspec#version
Also, this page that talks about dependency versioning:  https://dart.dev/tools/pub/dependencies#version-constraints

includecmath:
Hi, @sfshaza2 😂I mean use this way in the tutorial for readers' reference. Like this [Add the http package](https://flutter.dev/docs/cookbook/networking/fetch-data#1-add-the-http-package) <latest_version> is just a reference, and before it we give package link in the pub.dev.

sfshaza2:
OK, now I understand what you mean, @includecmath! Yeah, I'm ok with this.  :)

PR #3296, opened 2019-11-24
Add another way to passing-data.md

includecmath:
In addition to pass arguments directly to the widget, we can use RouteSettings.


sfshaza2:
This PR looks ok to me (though the line length should be less than 80 chars...) but, @johnpryan, can you take a look?

includecmath:
@googlebot need review

includecmath:
Hi @sfshaza2  can you help contact John Ryan @johnpryan .

PR #3309, opened 2019-11-27
Revise iOS add-to-app usage page

xster:
Fixes https://github.com/flutter/flutter/issues/32863

Staging on https://add-to-app-2019.firebaseapp.com/docs/development/add-to-app/ios/add-flutter-screen


src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
+See [Loading sequence and performance](/docs/development/add-to-app/performance)
+for more analysis on the latency and memory trade-offs of pre-warming an engine.
+
+### Create a FlutterEngine
+
+The proper place to do this is specific to your host app. As an example, we will
+demonstrate creating a `FlutterEngine`, exposed as a property, on app startup in
+the app delegate.
+
+<ul class="nav nav-tabs" id="engine-language" role="tablist">
```
xster: @matthew-carroll see staging at https://add-to-app-2019.firebaseapp.com/docs/development/add-to-app/ios/add-flutter-screen. 

This seems like a reasonable way of toggling languages for instructions in a way consistent with the aesthetics of the rest of the website.

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
-In `AppDelegate.h`:
+The `FlutterEngine` may have the same lifespan as your `FlutterViewController`
+or outlive your `FlutterViewController`.
 
-```objective-c
+{{site.alert.tip}}
+It's generally recommended pre-warm a long-lived `FlutterEngine` for your
+application. This way,
+
+- Flutter screens will render their first frames faster.
```
dnfield: nit: `Flutter will start the work to get to its first frame earlier`.  

Not sure how much the precision matters here, but having no VC doesn't actually speed up frame time.

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
+- Your Flutter and Dart state could outlive one `FlutterViewController`.
```
dnfield: I'm always a bit thrown by what `Flutter` means in this context.  Do we just mean `Flutter-specific application state`?

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
 
 @implementation AppDelegate
 
-// This override can be omitted if you do not have any Flutter Plugins.
 - (BOOL)application:(UIApplication *)application
     didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
-  self.flutterEngine = [[FlutterEngine alloc] initWithName:@"io.flutter" project:nil];
-  [self.flutterEngine runWithEntrypoint:nil];
+  self.flutterEngine = [[FlutterEngine alloc] initWithName:@"my flutter engine"];
+  // Runs the default Dart entrypoint with a default Flutter route.
```
dnfield: I think we should avoid mentioning the `default Flutter route` here. If anything, we should outline how to use platform channels to communicate with the Dart side to set up any non-default state.

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
-class AppDelegate: FlutterAppDelegate {
-  var flutterEngine : FlutterEngine?;
-  // Only if you have Flutter plugins.
+class AppDelegate: FlutterAppDelegate { // More on the FlutterAppDelegate below.
+  lazy var flutterEngine = FlutterEngine(name: "my flutter engine")
+
   override func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
-    self.flutterEngine = FlutterEngine(name: "io.flutter", project: nil);
-    self.flutterEngine?.run(withEntrypoint: nil);
+    // Runs the default Dart entrypoint with a default Flutter route.
```
dnfield: Ditto.

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
+```objectivec
+@import Flutter;
+#import "AppDelegate.h"
+#import "ViewController.h"
+
+@implementation ViewController
+- (void)viewDidLoad {
+    [super viewDidLoad];
+
+    // Make a button to call the showFlutter function below when pressed.
```
dnfield: nit: maybe `showFlutterViewController`?

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
+import UIKit
+import Flutter
+
+class ViewController: UIViewController {
+  override func viewDidLoad() {
+    super.viewDidLoad()
+
+    // Make a button to call the showFlutter function below when pressed.
+    let button = UIButton(type:UIButton.ButtonType.custom)
+    button.addTarget(self, action: #selector(showFlutter), for: .touchUpInside)
```
dnfield: Ditto.

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
+`FlutterViewController` implicitly create its own `FlutterEngine` without
+pre-warming one ahead of time.
 
-Make your app delegate implement the `FlutterAppLifeCycleProvider` protocol, for
-example:
+This is not recommended since creating a `FlutterEngine` on-demand could
+introduce a noticeable latency between when the `FlutterViewController` is
+presented and when it will render its first frame. This could, however, be
+sometimes useful if the Flutter screen is rarely shown and there are no good
+heuristics to determine when the Dart VM should be started.
```
dnfield: `if the Flutter screen is rarely shown and does not need to persist state between apperances`?

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
+is recommended but not required.
+
+The `FlutterAppDelegate` performs functions such as
+
+- Forwarding application callbacks such as [`openURL`](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623112-application)
+  to plugins such as [local_auth](https://pub.dev/packages/local_auth).
+- Forwarding status bar taps (which can only be detected in the AppDelegate) to
+  Flutter for scroll-to-top behavior.
+
+If your app delegate cannot directly subclass `FlutterAppDelegate`, you can
```
dnfield: s/can/must right?

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -214,129 +429,190 @@ performFetchWithCompletionHandler:(void (^)(UIBackgroundFetchResult result))comp
...
+will run `myOtherEntrypoint()` in `lib/other_file.dart` instead of `main()` in
+`lib/main.dart`.
 
-You can have the Flutter app dismiss itself by calling `SystemNavigator.pop()`
-in the Dart code.
+### Route
 
-## Building and running your app
+An initial route can be set for your Flutter [`WidgetsApp`]({{site.api}}/flutter/widgets/WidgetsApp-class.html)
+when constructing the engine.
```
dnfield: `If you choose to do this, you _must_ do so before running the engine, and any calls after running the engine will be ignored.`

IMO, we should avoid documenting this API, deprecate it, and remove it.  It's the source of a lot of confusion and it has no clear fix.

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -214,129 +429,190 @@ performFetchWithCompletionHandler:(void (^)(UIBackgroundFetchResult result))comp
...
+`"setInitialRoute"` on the `navigationChannel` must be called before running your
+`FlutterEngine` in order for Flutter's very first frame to use the desired
+route.
+{{site.alert.end}}
+
+{{site.alert.tip}}
+In order to change your Flutter route after the `FlutterEngine` is already
+running, use [pushRoute]({{site.api}}/objcdoc/Classes/FlutterViewController.html#/c:objc(cs)FlutterViewController(im)pushRoute:)
+or [popRoute]({{site.api}}/objcdoc/Classes/FlutterViewController.html#/c:objc(cs)FlutterViewController(im)popRoute)
+on the `FlutterViewController`.
```
dnfield: This is markedly different from setting an initial route, since it mutates the route stack, whereas `setInitialRoute` initializes the route stack to a specific spot.

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -214,129 +429,190 @@ performFetchWithCompletionHandler:(void (^)(UIBackgroundFetchResult result))comp
...
+{{site.alert.end}}
+
+See the [Navigation and routing](/docs/development/ui/navigation) section for more on Flutter's routes.
+
+### Other
+
+The above only illustrates a few examples of ways to customize how a Flutter
+instance is initiated. Using [platform channels](/docs/development/platform-integration/platform-channels),
+you're free to push data or prepare your Flutter environment in any way you'd
+like before presenting the Flutter UI via a `FlutterViewController`.
```
dnfield: IME, the routing logic for add-to-app is pretty limited and does unexpected things. I think it would be better to document how to use platform channels in this way here, and then we could mention that you can use any of the framework-provided channels as well (such as the navigation channel).

In other words, I think a user is far more likely to want to tell the running Dart program what state it should initialize to before presenting, and that it's very unlikely that the right way to do that will be setting an initial route (that only works at a specific point in execution) or pushing/popping routes blindly from the background.

dnfield: Document platform channels in this way here -> document how to use a custom platform channel to communicate with Dart

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
+  input events into Flutter and to display frames rendered by the `FlutterEngine`.
+{{site.alert.end}}
 
-In `AppDelegate.h`:
+The `FlutterEngine` may have the same lifespan as your `FlutterViewController`
+or outlive your `FlutterViewController`.
 
-```objective-c
+{{site.alert.tip}}
+It's generally recommended pre-warm a long-lived `FlutterEngine` for your
```
jmagman: missing "to": "recommended to pre-warm"

src/docs/development/add-to-app/ios/add-flutter-screen.md
jmagman: "could" or "will"?

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
+application. This way,
+
+- Flutter screens will render their first frames faster.
+- Your Flutter and Dart state could outlive one `FlutterViewController`.
+- You and your plugins can interact with Flutter and your Dart logic before
```
jmagman: "Your application and plugins"?

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
+```objectivec
+#import <FlutterPluginRegistrant/GeneratedPluginRegistrant.h> // Used to connect plugins.
 
 #import "AppDelegate.h"
 
 @implementation AppDelegate
 
-// This override can be omitted if you do not have any Flutter Plugins.
 - (BOOL)application:(UIApplication *)application
     didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
```
jmagman: I know you didn't edit this, but it should be `NSDictionary<UIApplicationLaunchOptionsKey, id> *)`

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
 import Flutter
-import FlutterPluginRegistrant // Only if you have Flutter Plugins.
+import FlutterPluginRegistrant // Used to connect plugins.
 
 @UIApplicationMain
-class AppDelegate: FlutterAppDelegate {
-  var flutterEngine : FlutterEngine?;
-  // Only if you have Flutter plugins.
+class AppDelegate: FlutterAppDelegate { // More on the FlutterAppDelegate below.
+  lazy var flutterEngine = FlutterEngine(name: "my flutter engine")
```
jmagman: Do we have engine naming conventions?

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
+
+@implementation ViewController
+- (void)viewDidLoad {
+    [super viewDidLoad];
+
+    // Make a button to call the showFlutter function below when pressed.
+    UIButton *button = [UIButton buttonWithType:UIButtonTypeCustom];
+    [button addTarget:self
+               action:@selector(showFlutter)
+     forControlEvents:UIControlEventTouchUpInside];
```
jmagman: I think this is missing a few tabs in the front.

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
+    [button setTitle:@"Show Flutter!" forState:UIControlStateNormal];
+    [button setBackgroundColor:[UIColor blueColor]];
```
jmagman: I would love to start "modernizing" (~2012) at least our sample code with dot notation.
```objective-c
button.backgroundColor = UIColor.blueColor;
```

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
+    button.frame = CGRectMake(80.0, 210.0, 160.0, 40.0);
+    [self.view addSubview:button];
+}
+
+- (void)showFlutter {
+    FlutterEngine *flutterEngine =
+        [(AppDelegate *)[[UIApplication sharedApplication] delegate] flutterEngine];
```
jmagman:
```objective-c
UIApplication.sharedApplication.delegate.flutterEngine
```

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -6,77 +6,291 @@ description: Learn how to add a single Flutter screen to your existing iOS app.
...
+    <a class="nav-link" id="vc-swift" href="#vc-swift-tab" role="tab" aria-controls="vc-swift" aria-selected="false">Swift</a>
+  </li>
+</ul>
+
+<div class="tab-content"> {% comment %} FlutterViewController language tab start {% endcomment -%}
+
+<div class="tab-pane active" id="vc-objc-tab" role="tabpanel" aria-labelledby="vc-objc-tab" markdown="1">
+
+<?code-excerpt "ViewController.m" title?>
+```objectivec
```
jmagman: I thought the markdown language marker was objective-c?

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -214,129 +429,190 @@ performFetchWithCompletionHandler:(void (^)(UIBackgroundFetchResult result))comp
...
 
-```objective-c
-[flutterViewController setInitialRoute:@"route1"];
+<?code-excerpt "Objective-C" title?>
+```objectivec
+[flutterEngine runWithEntrypoint:@"myOtherEntrypoint" libraryURI:@"other_file.dart"];
 ```
 
-- Swift:
+or
```
jmagman: Should the two languages be in tabs?

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: We do have a default flutter route on the flutter side https://api.flutter.dev/flutter/widgets/Navigator/defaultRouteName-constant.html

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: changed to `should`. Must is a bit strong since it's not like the app won't run anymore without it. We can point out the reasoning and let them decide.

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: Added a warning box.

xster: Actually there is already a warning box.

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: I wasn't suggesting that they are equivalent. It's in a tangential tip box.

xster: Added a few more words to clarify difference.

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: thanks!

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: sg, changed

src/docs/development/add-to-app/ios/add-flutter-screen.md
dnfield: It exists but it causes much confusion and frequently doesn't work as expected

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: I'm not sure which syntax highlighter library we're using for the website. `objective-c` and `objc` don't work.

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: <I didn't try it yet> don't we need to keep the cast to AppDelegate? </>

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -346,7 +622,7 @@ Connected views:
 
 Attach to specific isolates instead in two steps:
 
-1. Name the Flutter root isolate of interest in its Dart source.
+1- Name the Flutter root isolate of interest in its Dart source.
```
jmagman: Is `1-` the normal markdown way to number steps on the website?

src/docs/development/add-to-app/ios/add-flutter-screen.md
```diff
@@ -358,7 +637,7 @@ void main() {
 }
 ```
 
-2. Run `flutter attach` with the `--isolate-filter` option.
+2- Run `flutter attach` with the `--isolate-filter` option.
 
 ```bash
```
jmagman: I think this is supposed to be ```terminal

src/docs/development/add-to-app/ios/add-flutter-screen.md
jmagman: Oh yeah `((AppDelegate *)UIApplication.sharedApplication.delegate).flutterEngine`

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: > VC doesn't actually speed up frame time

Not sure I understood the comment about having no VCs. I rephrased it a little bit nevertheless.

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: @dnfield I'm actually just kinda thinking anything you can conceptually assign to the term 'Flutter' and that has a state. Can you think of counter examples? 

Replaced could with will

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: Which part specifically are you commenting on? 

The fact that there's a cascade going on in https://api.flutter.dev/flutter/material/MaterialApp-class.html and the top level navigator behavior? It's indeed confusing but it's just how Flutter works at the moment. I don't think it should ever not work as expected. This is a mostly orthogonal concept to the timing of calls to setInitialRoute.

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: I don't think it adds much. Which misunderstanding are you trying to avoid for the reader?

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: added into an enumeration

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: I generally agree with you and using a custom channel to push an application specific DI-like bundle is what I would do too. 

Though on the flip side, there's a bit of a personal taste involved in the trade off in the sense that writing platform channels is extremely boring and tedious (and error-prone in its own right as well) currently until we do https://github.com/flutter/flutter/issues/32930. 

Routes is also a non-trivially important concept in Flutter and we're not necessarily going to be avoiding pains for developers if we avoided this whole topic in documentations. 

TL;DR, I think all present solutions have annoyances and we should properly handle this very common initialization use case and present a proper best practice with infrastructure in the coming phases, but we should have some docs in the meantime.

xster: Added a bit more caveating for more transparency in the docs in the meantime.

xster: oh look there's https://github.com/flutter/flutter/issues/34170 already

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: ah thanks

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: looking into it a bit, this only adds a debug label for the thread in instruments. I want to get rid of it so I'll not over-emphasize it here for now

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: this is just lining up the semi-columns. Did I understand correctly?

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: square brackets forever!

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: lol you caught me. I got tired of writing all that html boilerplate

src/docs/development/add-to-app/ios/add-flutter-screen.md
dnfield: Is `Flutter` here the Flutter framework?  The Flutter engine excluding the embedding?  The embedding?  Plugin code (dart/platform)?

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: It's not, but the code highlighter makes the markdown parser crap out (it would have the wrong indent with the code block anyway)

src/docs/development/add-to-app/ios/add-flutter-screen.md
dnfield: It's confusing becuase you can only set it at one point in the lifecycle of starting the engine, and you lose access to that completely if you set it up in certain ways, and if you call it too late it just silently fails, and there's no good way to fix any of those problems with the API so if you open a bug about it you'll be told you should be doing it differently anyway.

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: oh nice, it looks way nicer, thanks

src/docs/development/add-to-app/ios/add-flutter-screen.md
dnfield: what `Flutter` is.

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: done

src/docs/development/add-to-app/ios/add-flutter-screen.md
dnfield: SG.  I can understand including this - my comment above sketches out a bit more of why (I think users will read this, get it wrong, think it's a bug, and be frustrated when we tell them it's by design).  But maybe that's what we have right now.

xster: ya, I hear you. There are tons of bugs around this already and I don't think routes are generally useful APIs, but I don't think we can move the weight of the community right now and we might get more confusion if the code is in the repo but we avoided talking about it.

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: Which user-self-assigned definition of the term Flutter would be problematic?

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: Ya, the mechanism for changing the default route is confusing as mentioned below. But it doesn't change the fact that there is a default route when you don't change it.

src/docs/development/add-to-app/ios/add-flutter-screen.md
xster: Doesn't the sentence apply the same way for all those definitions? 

Tangent: I don't think the word embedding means anything for people outside the team and I'm tempted to keep it so for as long as we can.

dnfield: I guess a way of looking at this is: here you use `Flutter` in a way that excludes`FlutterViewController`.  Later in the document you're using `Flutter` in a way that explicitly means showing the `FlutterViewController`.

xster: Isn't it a bit like having the sentences:

- Having a backup of your movie will make it outlive one VHS tape
- 'showMovie'

and then asking whether readers will be struggling understanding whether 'movie' here encompasses the mylar tape ribbons or not :)

It's likely immaterial whether they assume one way or another.

PR #3317, opened 2019-11-28
Update iOS add-to-app project setup instructions

jmagman:
First part of https://github.com/flutter/flutter/issues/45224.


jmagman:
I need to figure out how to deploy this to see how it looks.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,87 +4,95 @@ short-title: Integrate Flutter
...
- 1. [Add Module Dependency in Existing
-    App](#make-the-host-app-depend-on-the-flutter-module)
-    Sets up Xcode to include your Flutter project into the build of your
-    existing app.
+From the command line, run:
 
-## Create a Flutter module
+```terminal
+cd some/path/
+flutter create --template module my_flutter
```
xster: maybe just -t?

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,87 +4,95 @@ short-title: Integrate Flutter
...
 
-In order to embed Flutter as a library, use the Flutter module template.
-Executing `flutter create -t module xxx` produces a Flutter project with a
-CocoaPods pod designed for consumption by your existing host app.
+In that directory you can run the same `flutter` commands you would
+in any other Flutter project, like `flutter run --debug` or `flutter build ios`.
+You can also run in [Android Studio/IntelliJ][] or [VS Code][] with the Flutter and Dart plugins.
+This will run a single-view example version of your module without being
+embedded in your existing application. This can be useful for testing
+quick, incremental changes with hot reload.
```
xster: this makes it sound like you can't hot reload if you embedded into your existing app

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,87 +4,95 @@ short-title: Integrate Flutter
...
+This creates a Flutter module project at `some/path/my_flutter/`.
 
-In order to embed Flutter as a library, use the Flutter module template.
-Executing `flutter create -t module xxx` produces a Flutter project with a
-CocoaPods pod designed for consumption by your existing host app.
+In that directory you can run the same `flutter` commands you would
+in any other Flutter project, like `flutter run --debug` or `flutter build ios`.
+You can also run in [Android Studio/IntelliJ][] or [VS Code][] with the Flutter and Dart plugins.
+This will run a single-view example version of your module without being
+embedded in your existing application. This can be useful for testing
```
xster: testing flutter only parts

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,87 +4,95 @@ short-title: Integrate Flutter
...
-code to get you started, and a `.ios/` hidden subfolder. The `.ios/` folder wraps
-the module project that contains some CocoaPods and a helper Ruby script.
+Add your Dart code to the `lib/` directory.
+
+You can add Flutter dependencies to `my_flutter/pubspec.yaml`, including Flutter plugins 
+with iOS platform code.
+
+There is a `.ios/` hidden subfolder which contains a Xcode workspace where you can 
+run a stand-alone version of your module. It also contains helper scripts to embed
+the module into your existing application with [CocoaPods][].
```
xster: or build frameworks from

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,87 +4,95 @@ short-title: Integrate Flutter
...
 
-## Make the host app depend on the Flutter module
+{{site.alert.note}}
+Custom iOS code should be added to your existing application or a plugin, not
+the module in `.ios/`. Changes made in `.ios/` will not be embedded in your existing application.
+The directory can be regenerated by running `flutter clean` in the `my_flutter` directory.
```
xster: Doesn't everything get overwritten too each time flutter packages get is run too? https://github.com/flutter/flutter/issues/28135

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,87 +4,95 @@ short-title: Integrate Flutter
...
+with iOS platform code.
+
+There is a `.ios/` hidden subfolder which contains a Xcode workspace where you can 
+run a stand-alone version of your module. It also contains helper scripts to embed
+the module into your existing application with [CocoaPods][].
 
-## Make the host app depend on the Flutter module
+{{site.alert.note}}
+Custom iOS code should be added to your existing application or a plugin, not
+the module in `.ios/`. Changes made in `.ios/` will not be embedded in your existing application.
```
xster: not the module in `.ios/` which is just a wrapper project to bootstrap Flutter.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,87 +4,95 @@ short-title: Integrate Flutter
...
-The description below assumes that your existing iOS app has a structure similar
-to what you get by asking Xcode to generate a new "Single View App" project
-using Objective-C. If your existing app has a different folder structure
-and/or existing `.xcconfig` files, you can reuse those, but you'll probably need
-to adjust some of the relative paths mentioned below accordingly.
+## Embed the Flutter module in your existing application
 
-The assumed folder structure is as follows:
+There are two ways to embed Flutter in your existing application.
+1. Use the CocoaPods dependency manager and `flutter`. Recommended.
```
xster: Not clear when you just say `flutter`. I suspect you're trying to say that in this case, you need a locally installed Flutter SDK.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,87 +4,95 @@ short-title: Integrate Flutter
...
-[cocoapods.org](https://cocoapods.org/).
-
-If your host application (`MyApp`) is already using CocoaPods, you only have to do the
-following to integrate with your `my_flutter` app:
+If your existing application (`MyApp`) does not already have a Podfile,  follow the
+[CocoaPods getting started guide][] to add a `Podfile` to your project.
 
 1. Add the following lines to your `Podfile`:
 
 ```ruby
```
xster: See https://github.com/flutter/website/pull/3318/files. I think it's kinda nice to add a `<?code-excerpt "MyApp/Podfile" title?>` type thing that titles what file you're showing. But up to you.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -94,28 +102,112 @@ following to integrate with your `my_flutter` app:
...
-vary on the pod dependencies of each plugin.
-- A build script is added to the Podfile targets that call `install_all_flutter_pods` to ensure that the binaries you build stay up to date
-with the Dart code that's actually in the folder.  It also uses your Xcode build configuration (Debug, Profile, Release) to embed the matching
-release type of `Flutter.framework` and `App.framework`.
\ No newline at end of file
+- `Flutter.framework` (the Engine library) is being embedded into your application by
+CocoaPods as a vendored framework. This has to match up with the release type
+(debug/profile/release) as well as the architecture for your app (arm*, 
+x86_64, etc.).
+- `App.framework` (your Flutter application binary) is being embedded into your
```
xster: The Flutter framework is in there too

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -94,28 +102,112 @@ following to integrate with your `my_flutter` app:
...
-- `App.framework` (your Flutter application binary) is embedded into your app.  CocoaPods also pulls this in as a vendored framework and
-makes sure it gets embedded into your native app.
-- Any plugins are added as CocoaPod pods.  In theory, it should be possible to manually merge those in as well, but those instructions
-vary on the pod dependencies of each plugin.
-- A build script is added to the Podfile targets that call `install_all_flutter_pods` to ensure that the binaries you build stay up to date
-with the Dart code that's actually in the folder.  It also uses your Xcode build configuration (Debug, Profile, Release) to embed the matching
-release type of `Flutter.framework` and `App.framework`.
\ No newline at end of file
+- `Flutter.framework` (the Engine library) is being embedded into your application by
+CocoaPods as a vendored framework. This has to match up with the release type
```
xster: `This has to match` sounds like you're telling the user to do something, but not what or how

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -94,28 +102,112 @@ following to integrate with your `my_flutter` app:
...
+
+### Embed frameworks in Xcode
+
+Alternatively, you can generate the necessary frameworks and embed them in your application
+by manually editing your existing Xcode project.
+
+If you are using the above [Embed with CocoaPods and Flutter tools](#embed-with-CocoaPods-and-Flutter-tools)
+method, you can skip these instructions.
+
+Let's assume you want to generate the frameworks to `some/path/MyApp/Flutter/`.
```
xster: Had to double take to realize that this is a future path you want to put things into

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -94,28 +102,112 @@ following to integrate with your `my_flutter` app:
...
+      <figure class="figure">
+        {% asset development/add-to-app/ios/project-setup/embed-xcode.png %}
+        <figcaption class="figure-caption">
+          Embed frameworks in Xcode
+        </figcaption>
+      </figure>
+    </div>
+  </div>
+</div>
+
```
xster: while playing, I also noticed that it might not select 'embed & sign' for them automatically. Worth double checking.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -94,28 +102,112 @@ following to integrate with your `my_flutter` app:
 
 3. Run `pod install`.
 
+{{site.alert.note}}
 Whenever you change the Flutter plugin dependencies in `some/path/my_flutter/pubspec.yaml`,
```
sfshaza2: I find this to be a bit confusing. Is it correct to say that you run this from the directory where the pubspec is? Or to say that you run it from the top of the package?

src/docs/development/add-to-app/ios/project-setup.md
sfshaza2: Can you use "*" to indicate a bulleted list, rather than minus? (I know this was there before...)

Also, I am not sure what the point of this section is. It seems like it's there to just tell the user what's going on, but you've deleted the introductory sentence explaining that it's there just to let you know what's going on. Can you add that back (or something similar)?

src/docs/development/add-to-app/ios/project-setup.md
sfshaza2: Also, can you make this sentence more active? ->
"CocoaPods embeds the engine library (`Flutter.framework`) into your app as a vendored framework.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -94,28 +102,112 @@ following to integrate with your `my_flutter` app:
...
-- A build script is added to the Podfile targets that call `install_all_flutter_pods` to ensure that the binaries you build stay up to date
-with the Dart code that's actually in the folder.  It also uses your Xcode build configuration (Debug, Profile, Release) to embed the matching
-release type of `Flutter.framework` and `App.framework`.
\ No newline at end of file
+- `Flutter.framework` (the Engine library) is being embedded into your application by
+CocoaPods as a vendored framework. This has to match up with the release type
+(debug/profile/release) as well as the architecture for your app (arm*, 
+x86_64, etc.).
+- `App.framework` (your Flutter application binary) is being embedded into your
+application by CocoaPods as a vendored framework.
```
sfshaza2: Ditto (make the sentence more active): ->
Cocoapods embeds your Flutter application binary (`App.framework`) into your app as a vendored framework.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -94,28 +102,112 @@ following to integrate with your `my_flutter` app:
...
+- Flutter plugins with iOS platform code are added as CocoaPod pods.
```
sfshaza2: Is this accurate? (I'm trying to make it more active): ->
CocoaPods adds the Flutter plugins with iOS platform code.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -94,28 +102,112 @@ following to integrate with your `my_flutter` app:
...
+- A build script is added to the Podfile targets that call `install_all_flutter_pods`
+to ensure that the binaries you build stay up to date with the Dart code in
+the Flutter module.  It also uses your Xcode build configuration (Debug, Profile, Release)
+to embed the matching build mode type of `Flutter.framework` and `App.framework`.
```
sfshaza2: Accurate?
The framework adds a build script to each Podfile target that calls `install_all_flutter_pods`.
This ensures that the binary stays up to date with the Dart code in the Flutter module.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -94,28 +102,112 @@ following to integrate with your `my_flutter` app:
...
+        {% asset development/add-to-app/ios/project-setup/embed-xcode.png %}
+        <figcaption class="figure-caption">
+          Embed frameworks in Xcode
+        </figcaption>
+      </figure>
+    </div>
+  </div>
+</div>
+
+You should now be able to build the project using `⌘B`.
```
sfshaza2: Again, this is very Mac specific. What about Windows? (We have far more Flutter devs on Windows than Mac.)

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -94,28 +102,112 @@ following to integrate with your `my_flutter` app:
...
+
+### Under the hood
+- The frameworks are added to the target's Link Binary with Libraries and Embed Frameworks
```
sfshaza2: So much passive voice! And non-asterisk bullets.  It would be nice if all items in this list used active voice, as well.

src/docs/development/add-to-app/ios/project-setup.md
xster: iOS add-to-app is only to iOS apps, which can only be developed on macs.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,87 +4,95 @@ short-title: Integrate Flutter
 description: Learn how to integrate a Flutter module into your existing iOS project.
 ---
 
-TODO(jmagman): add the pre-built framework alternative and review.
+Flutter can be embedded into your existing iOS application in a piecemeal fashion as
```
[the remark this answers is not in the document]
jmagman: > does it really make sense to add a "Flutter framework" to an iOS app

Yup, that's the feature!  The user has an existing Flutter-free iOS app, and they want to add Flutter bits to it without migrating to Flutter wholesale or starting over as a Flutter app.  These bits are generated as "libraries" or "frameworks" (Apple's wording, not ours) and then embedded into the existing iOS app.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,87 +4,95 @@ short-title: Integrate Flutter
...
 
-This creates a `some/path/my_flutter/` Flutter module project with some Dart
-code to get you started, and a `.ios/` hidden subfolder. The `.ios/` folder wraps
-the module project that contains some CocoaPods and a helper Ruby script.
+Add your Dart code to the `lib/` directory.
+
+You can add Flutter dependencies to `my_flutter/pubspec.yaml`, including Flutter plugins 
+with iOS platform code.
+
+There is a `.ios/` hidden subfolder which contains a Xcode workspace where you can
```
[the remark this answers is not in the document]
jmagman: iOS apps can only be developed on a Mac.  I'll add that caveat.

src/docs/development/add-to-app/ios/project-setup.md
jmagman: That too https://github.com/flutter/flutter/issues/42956.

src/docs/development/add-to-app/ios/project-setup.md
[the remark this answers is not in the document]
jmagman: > I'd probably delete "in a piecemeal fashion", which might confuse some

Sounds good, though I copied this from https://github.com/flutter/website/blob/staging-add-to-app-do-not-delete-until-2020/src/docs/development/add-to-app/index.md

src/docs/development/add-to-app/ios/project-setup.md
jmagman: I was copying https://github.com/flutter/website/blob/master/src/docs/development/packages-and-plugins/developing-packages.md#developing-dart-packages-dart, I would expect verbose to be better for docs but I'll defer to @sfshaza2

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,87 +4,95 @@ short-title: Integrate Flutter
...
 
-If you're starting off with an *existing* host app for either platform, you'll
-likely want to include your Flutter project in that app as some form of library
-instead.
+This creates a Flutter module project at `some/path/my_flutter/`.
 
-In order to embed Flutter as a library, use the Flutter module template.
-Executing `flutter create -t module xxx` produces a Flutter project with a
-CocoaPods pod designed for consumption by your existing host app.
+In that directory you can run the same `flutter` commands you would
```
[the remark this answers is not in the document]
jmagman: `some/path/my_flutter/` from the previous sentence.  I'll combine.

src/docs/development/add-to-app/ios/project-setup.md
jmagman: I meant the `flutter` command.  I will change to Flutter SDK.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,87 +4,95 @@ short-title: Integrate Flutter
...
-using Objective-C. If your existing app has a different folder structure
-and/or existing `.xcconfig` files, you can reuse those, but you'll probably need
-to adjust some of the relative paths mentioned below accordingly.
+## Embed the Flutter module in your existing application
 
-The assumed folder structure is as follows:
+There are two ways to embed Flutter in your existing application.
+1. Use the CocoaPods dependency manager and `flutter`. Recommended.
+1. Create frameworks for the Flutter engine, your compiled Dart code, and all Flutter plugins.
+Manually embed the frameworks and update your existing application's build settings in Xcode.
```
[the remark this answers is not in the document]
jmagman: There will be 2+ frameworks to embed, each one in the list are separate frameworks.  So the engine is one, the compiled dart code is the second, each plugin is one, etc.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,87 +4,95 @@ short-title: Integrate Flutter
...
+
+### Embed with CocoaPods and Flutter tools
+
+This method requires every developer working on your project to have `flutter` and [CocoaPods][] installed.
```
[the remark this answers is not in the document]
jmagman: Yes, I am all ears if there's a better way to organize the decision tree... Tabs?  Different pages?

src/docs/development/add-to-app/ios/project-setup.md
jmagman: I don't know if it's useful, the contents are almost untouched from what was there before.  I'm happy to remove the entire section if you don't think it's useful.  I removed the intro because I didn't want to imply anything would go wrong following these steps.

src/docs/development/add-to-app/ios/project-setup.md
jmagman: I removed the "Under the hood" sections.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,87 +4,96 @@ short-title: Integrate Flutter
 description: Learn how to integrate a Flutter module into your existing iOS project.
 ---
 
-TODO(jmagman): add the pre-built framework alternative and review.
+Flutter can be incrementaly added into your existing iOS application as embedded
+frameworks.
```
xster: Do we need to mention bitcode or the min OS version?

src/docs/development/add-to-app/ios/project-setup.md
xster: SG

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -94,28 +103,92 @@ following to integrate with your `my_flutter` app:
...
+
+The `podhelper.rb` script embeds your plugins, `Flutter.framework`, and
+`App.framework` into your project.
+
+You should now be able to build the project using `⌘B`.
+
+### Embed frameworks in Xcode
+
+Alternatively, you can generate the necessary frameworks and embed them in your application
+by manually editing your existing Xcode project.
```
xster: Add a sentence for why you might want to do this

src/docs/development/add-to-app/ios/project-setup.md
jmagman: Fortunately the bitcode bugs seem fixed so they don't have to turn it off anymore.  
I will add a systems requirements section.

src/docs/development/add-to-app/ios/project-setup.md
jmagman: Good idea.

src/docs/development/add-to-app/ios/project-setup.md
jmagman: The top of the package should be the same location as the pubspec.
How about:
> When you change the Flutter plugin dependencies in `my_flutter/pubspec.yaml`,
run `flutter pub get` in your Flutter module directory to refresh the list
of plugins read by the `podhelper.rb` script. Then run `pod install` again from
in your application at`some/path/MyApp`.

src/docs/development/add-to-app/ios/project-setup.md
jmagman: I added a system requirements section.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,88 +4,103 @@ short-title: Integrate Flutter
...
+## Embed the Flutter module in your existing application
+
+There are two ways to embed Flutter in your existing application.
+1. Use the CocoaPods dependency manager and installed Flutter SDK. Recommended.
+1. Create frameworks for the Flutter engine, your compiled Dart code, and all Flutter plugins.
+Manually embed the frameworks and update your existing application's build settings in Xcode.
+
+### Embed with CocoaPods and the Flutter SDK
+
+This method requires every developer working on your project to have a locally installed
```
xster: Add some pros and cons. e.g. you can just click run in xcode and not have to build 2 things when you're rapidly iterating.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -94,28 +109,96 @@ following to integrate with your `my_flutter` app:
...
 
-Whenever you change the Flutter plugin dependencies in `some/path/my_flutter/pubspec.yaml`,
-you need to run `flutter pub get` from `some/path/my_flutter` to refresh the list
+{{site.alert.note}}
+When you change the Flutter plugin dependencies in `my_flutter/pubspec.yaml`,
+run `flutter pub get` in your Flutter module directory to refresh the list
 of plugins read by the `podhelper.rb` script. Then run `pod install` again from
-`some/path/MyApp`.
+in your application at`some/path/MyApp`.
+{{site.alert.end}}
```
xster: Do you have to do this if you `flutter upgrade` or `flutter channel` too?

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -94,28 +109,96 @@ following to integrate with your `my_flutter` app:
...
+`App.framework` into your project.
+
+You should now be able to build the project using `⌘B`.
+
+### Embed frameworks in Xcode
+
+Alternatively, you can generate the necessary frameworks and embed them in your application
+by manually editing your existing Xcode project. You may choose to do this if members of your
+team cannot locally install Flutter SDK and CocoaPods, or if you do not wish to use CocoaPods
+as a dependency manager in your existing applications.
```
xster: Mention that the downside is that you have to run `flutter build ios-framework` each time you change things in Flutter

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -94,28 +109,96 @@ following to integrate with your `my_flutter` app:
...
+```
 
-The `podhelper.rb` script ensures that your plugins, the Flutter.framework, and
-the App.framework are embedded in your project.
+{{site.alert.tip}}
+With Xcode 11 installed you can generate [XCFrameworks][] instead of universal frameworks by adding
+the flags `--xcframework --no-universal`.
+{{site.alert.end}}
+
+Open your existing project in Xcode and drag the frameworks from `some/path/MyApp/Flutter/Release/` in Finder
```
xster: Say that this is just one arbitrary way of integrating we're using as an example. You don't have to do this and you can integrate in whichever way you prefer.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -4,88 +4,103 @@ short-title: Integrate Flutter
...
-manager. This is because the Flutter framework needs to be available to any
-Flutter plugins that you might include in `my_flutter`.
-
-For information on how to install CocoaPods on your development machine, see
-[cocoapods.org](https://cocoapods.org/).
-
-If your host application (`MyApp`) is already using CocoaPods, you only have to do the
-following to integrate with your `my_flutter` app:
+If your existing application (`MyApp`) does not already have a Podfile,  follow the
+[CocoaPods getting started guide][] to add a `Podfile` to your project.
```
xster: Actually I suspect a non-trivial amount of people might fall into the i-opened-the-xcodeproj-file-instead-of-the-xcworkspace-file and will run into an error. Perhaps spell that out as well.

src/docs/development/add-to-app/ios/project-setup.md
jmagman: I don't think upgrade or channel would alter anything in .flutter-plugins, but I can add it if you know something I don't.

xster: I'm just guessing. Wouldn't this change the engine framework which gets copied over to your .ios folder if you ran flutter packages get? Don't know if you need a subsequent pod install to make things match.

jmagman: The right version of the Flutter.framework should get copied the next time they run their app via the `xcode_backend` script.

src/docs/development/add-to-app/ios/project-setup.md
jmagman: I mentioned it.  Fortunately `pod` on its own has a better log about using `xcworkspace` than `flutter` does, so hopefully they see it.

PR #3337, opened 2019-12-04
remove add-to-app sidenav in preparation for merging to master

xster:


xster:
<nit>I think we want the merge conflicts so we properly merge into master and for people to deal with this on master such that it's the simplest possible switch flick next week</>

jmagman:
> I think we want the merge conflicts so we properly merge into master and for people to deal with this on master such that it's the simplest possible switch flick next week</>

Wouldn't a bunch of commented out code make that harder? Are we rebasing it or keeping the merge commits?
I'm always against commenting out code since it tends to kick around after it's needed, and it can always be grabbed out of source control. But if this makes it easier for you with the workflow you're picturing, this is fine.

xster:
We're just merging the diff into master. If there are conflicts, we'll rebase and deal with the conflicts now.

PR #3347, opened 2019-12-05
Update FFI iOS linking documentation

dcharkes:
Following the documentation led to https://github.com/dart-lang/ffi/issues/28.

@sfshaza2 PTAL.

The cpp is not automatically added by the podspec. This PR updates the documentation to reflect that. Though, maybe there's a way that it could be automatic?

Note: I was doing all of this from command line / VSCode, until I realized I needed to open Xcode to do the linking step.

### DIFF
--- a/src/docs/development/platform-integration/c-interop.md
+++ b/src/docs/development/platform-integration/c-interop.md
@@ -98,8 +98,9 @@ during link-time optimization.
 
 For example,
 to create a C++ file named `ios/Classes/native_add.cpp`,
-use the following instructions. Start from the root
-directory of your project:
+use the following instructions. (Note that the template
+has already created this file for you.) Start from the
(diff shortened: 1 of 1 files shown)


src/docs/development/platform-integration/c-interop.md
```diff
@@ -112,9 +113,11 @@ int32_t native_add(int32_t x, int32_t y) {
 EOF
 ```
 
-The podspec fle (`ios/native_add.podspec`) automatically
-includes all C/C++ files in `ios/Classes`, and they are
-compiled within the Xcode build system.
+On iOS, you need to tell xcode to statically link the file:
```
sfshaza2: xcode => Xcode
However, I won't hold up the merge for that..  :)

PR #3350, opened 2019-12-05
Introduce a Liquid tag to display code samples with different languages.

matthew-carroll:
Introduce a Liquid tag to display code samples with different languages.

### DIFF
--- a/src/_assets/js/main.js
+++ b/src/_assets/js/main.js
@@ -22,6 +22,7 @@ $(function () {
   setupClipboardJS();
 
   // New (dash) tabs
+  setupTabs($('#editor-setup'), 'io.flutter.tool-id');
   setupTabs($('.sample-code-tabs'), 'io.flutter.tool-id');
   // Old tabs
   setupToolsTabs($('#tab-set-install'), 'tab-install-', 'io.flutter.tool-id');
--- a/src/_assets/js/main.js
+++ b/src/_assets/js/main.js
@@ -22,7 +22,7 @@ $(function () {
   setupClipboardJS();
 
   // New (dash) tabs
-  setupTabs($('.nav-tabs'), 'io.flutter.tool-id');
+  setupTabs($('.sample-code-tabs'), 'io.flutter.tool-id');
   // Old tabs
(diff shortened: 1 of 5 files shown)


src/_assets/js/main.js
```diff
@@ -22,7 +22,7 @@ $(function () {
   setupClipboardJS();
 
   // New (dash) tabs
-  setupTabs($('.nav-tabs'), 'io.flutter.tool-id');
```
xster: #editor-setup was here previously right? Is that using the new tag too? If not, are we breaking it?

matthew-carroll: Where is the #editor-setup code example?

xster: https://github.com/flutter/website/blob/master/src/_assets/js/main.js#L25

xster:
Can you push this to your staging branch as well?

xster:
Doh, I can't change the merge base from the UI here directly. Can you do an interactive rebase of 356c448 and cc36363 onto master directly? It should merge cleanly since there aren't too many files.

matthew-carroll:
@xster I added back the editor configuration in main.js. I also converted the iOS samples to use the tag. I adjusted the behavior of the tag after seeing some different practices in the iOS code.

Can you take another look and let me know what you think?

matthew-carroll:
The CLA warning was the result of the rebasing process. After changing the target branch, the only commits are my own, so I manually added the CLA tag.

src/_plugins/code_tabs.rb
```diff
@@ -0,0 +1,223 @@
...
+#   // Swift code here
+#   ```
+#   {% endsamplecode %}
+#
+# The "samplecode" tag accepts a single parameter that represents the title of this sample. The
+# title is not displayed on the page, but it is injected into a number of HTML attributes to
+# differentiate itself from other code samples. The "samplecode" tag is a block tag and therefore
+# must be closed with a "endsamplecode" tag.
+#
+# The "sample" tag accepts one parameter: the name of the language as it should appear to
```
xster: These languages appear visually as tabs

sfshaza2:
I see @xster has approved. Merging.

PR #3351, opened 2019-12-05
Review of add-to-app/ios/project-setup.md

gaaclarke:


src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -32,18 +33,19 @@ embedded in your existing application, which is useful for incrementally
...
 ```text
 my_flutter/
-  .ios/
-    Runner.xcworkspace
-    Flutter/podhelper.rb
-  lib/
-    main.dart
-  test/
-  pubspec.yaml
+├─.ios/
```
jmagman: Thanks, missed this one.

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -81,10 +85,10 @@ The following example assumes that your existing application and the Flutter mod
 If you have a different directory organization, you may need to adjust the relative paths.
 
 ```text
-some/path
-├── my_flutter
-│   └── .ios
-│       └── Flutter
+some/path/
```
jmagman: I had removed the trailing `/` to match https://github.com/flutter/website/blob/staging-add-to-app-do-not-delete-until-2020/src/docs/development/add-to-app/android/project-setup.md#option-a---depend-on-the-android-archive-aar, it should be consistent either way.

gaaclarke: There were other places where the trailing slashes still existed.  I think it is more clear with the slashes.  I think I switched them all over here.

PR #3353, opened 2019-12-05
Added a tooltip that describes the frameworks.

gaaclarke:


src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -123,6 +123,11 @@ in your application at`some/path/MyApp`.
 The `podhelper.rb` script embeds your plugins, `Flutter.framework`, and
 `App.framework` into your project.
 
+{{site.alert.tip}}
+`Flutter.framework` is the library for the Flutter engine and `App.framework` is
```
xster: This is true for the option b section below too right? Is there another place we can put this box?

gaaclarke: I put the tip under the first instance of talking about the frameworks.

You definitely don't want the tool tip twice, and you definitely want the tooltip the first time the terms are introduced.  The only solution that solves both of those requirements is artificially talking about the terms earlier in the document.  Otherwise, this option is as good as any other.

xster:
LGTM

src/docs/development/add-to-app/ios/project-setup.md
jmagman: Can you say "framework" instead of "library"? We use them too interchangeably.

gaaclarke: Since these are term definitions I didn't want to use the word in the definition.  "`Flutter.framework` is a framework" is less informative than "`Flutter.framework` is a library".

jmagman: It's not a library though, it's a framework.  The terms mean something different, but it's interchangeably used in various places.

gaaclarke: Apple's `.framework`'s are libraries.  In the same way that DLLs are libraries, but DLLs are not libraries.  It is a type of library.  The thing that is really confusing is that `.framework`'s are not [frameworks](https://en.wikipedia.org/wiki/Software_framework).

jmagman: I know what you mean, the terms are overloaded across platforms, and general Flutter developers aren't necessarily iOS/macOS developers.  But within darwin world, frameworks (the bundle ending in .framework containing libraries) and libraries (the binary) are different, and the add-to-app developers will be more iOS savvy than most Flutter developers.

darwin

One of the reasons Flutter Swift apps build plugins as frameworks and not libraries, for example, is this bug: https://forums.swift.org/t/swift-static-libraries-dont-copy-generated-objective-c-header/19816

xster: I don't feel strongly about it but I think framework (in the Apple sense, not in the english word sense) is cleaner. .frameworks has-a library rather than is-a library in the same sense that .a, .so, .dylib would be.

gaaclarke: I think context matters, we are defining something called "App.framework".  I'm all for calling them frameworks everywhere else.  I think calling them "libraries" here at best helps people understand a concept better, at worst incenses pedantic people.  It is no way confusing to anyone because ".framework" is in the name.

Digging deeper into the philosophical black hole: A ".dll" is a library, right?  It is in the name after all.  There is conceptually nothing that a ".framework" can do that a ".dll" can't do.  So, if ".dll" files are libraries are libraries, ".framework"s must be libraries.  lol

PR #3355, opened 2019-12-05
fix missing tab setup

xster:
We accidentally removed a tab setup from the merge.


xster:
actually we don't need this since we're reusing an existing bootstrap class. We'll need it if we use our own class in https://github.com/flutter/website/pull/3350. Closing this.

PR #3369, opened 2019-12-07
fix link

rashedmyt:


src/docs/release/breaking-changes/index.md
```diff
@@ -4,7 +4,7 @@ description: A list of migration guides for breaking changes in Flutter.
 ---
 
 As described in our [breaking change
-policy][../../resources/compatibility], on occasion we publish guides
+policy](https://flutter.dev/docs/resources/compatibility), on occasion we publish guides
```
dnfield: I think you just want `(../../resources/compatibility)` here. This site gets deployed to other domains as well for staging/testing/etc.

dnfield: (so really the change should just be to fix the invalid markdown - just leave the URI alone)

rashedmyt: I saw [here](https://flutter.dev/docs/resources/compatibility#announcements-and-migration-guides), it used complete URL, so I thought of changing it. Will revert to as it was

rashedmyt:
@dnfield Should I revert to complete URL? The build was passing with complete URL

rashedmyt:
fixed in #3378

PR #3451, opened 2019-12-18
Fix markdown formatting errors in plugin migration page

xster:


src/docs/development/packages-and-plugins/plugin-api-migration.md
```diff
@@ -86,19 +88,19 @@ The following instructions outline the steps for supporting the new API:
...
+        super.onCreate(savedInstanceState);
+        GeneratedPluginRegistrant.registerWith(this);
+      }
     }
     ```
 
 1. (Optional) If you created an `EmbeddingV1Activity` in the step above, add the
-   `EmbeddingV1Activity` to the
-   <plugin_name>/example/android/app/src/main/AndroidManifest.xml.
+   `EmbeddingV1Activity` to the `<plugin_name>/example/android/app/src/main/AndroidManifest.xml`.
```
xster: bare angle brackets here causes the rest of the markdown to fail to parse

PR #3452, opened 2019-12-18
AlertDialog breaking change migration guide reprise

shihaohong:
I made some small edits to fix some small formatting issues as I saw on the website, since I figure setting the example correctly from the get-go will help define how it looks from here on out:
- "Scrollable AlertDialog" title shows up twice
- The numbered list doesn't appear as intended (shows up as 1. twice)
- Removes "two-phase update" section, since it's not really part of the migration guide

I also added a code snippet for one common migration case that will be observed, since it was a pattern that Flutter recommended in the API docs.


src/docs/release/breaking-changes/scrollable_alert_dialog.md
```diff
@@ -113,27 +109,43 @@ flutter:          └─SemanticsNode#32 <-- contents
...
+  title: Text('Very, very large title', textScaleFactor: 5),
+  content: Text('Very, very large content', textScaleFactor: 5),
+  actions: <Widget>[
+    FlatButton(child: Text('Button 1'), onPressed: () {}),
+    FlatButton(child: Text('Button 2'), onPressed: () {}),
+  ],
+)
+```
 
 ## Timeline
```
shihaohong: I did still have a question about the `Timeline` section -- What should this really contain? Or is this what I have already intended?

sfshaza2: Did you see Ian's PR? He fleshed this out a bit.

shihaohong: I just saw it! In that sense, I'm not far along enough with my migration (have not implemented the hard breaking change) that it applies yet because I made a mistake with understanding the steps in the process. 

Could I add a placeholder in here for now and come back and add the correct version to the timeline once the hard breaking change has landed?

sfshaza2: yes

PR #3453, opened 2019-12-18
Added breaking change doc for imagecache algorithm.

gaaclarke:
related pr: https://github.com/flutter/flutter/pull/47387

### DIFF
--- a/src/docs/release/breaking-changes/12_18_19_ImageCache_Logic.md
+++ b/src/docs/release/breaking-changes/12_18_19_ImageCache_Logic.md
@@ -24,7 +24,7 @@ void onLoadImage(Image image) {
     _cache.maxByteSize = image.byteSize + 1000;
   }
   _cache.add(image);
-  while (_cache.count > _cache.maxCount 
+  while (_cache.count > _cache.maxCount
       || _cache.byteSize > _cache.maxByteSize) {
     _cache.discardOldestImage();
   }
@@ -36,7 +36,7 @@ void onLoadImage(Image image) {
 void onLoadImage(Image image) {
   if (image.byteSize < _cache.maxByteSize) {
     _cache.add(image);
-    while (_cache.count > _cache.maxCount 
+    while (_cache.count > _cache.maxCount
         || _cache.byteSize > cache.maxByteSize) {
       cache.discardOldestImage();
     }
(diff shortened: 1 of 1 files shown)


src/docs/release/breaking-changes/12_18_19_ImageCache_Logic.md
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+}
+```
+
+## Migration guide
+
+There may be situations where the ImageCache is thrashing with the new logic
+where it wasn't previously, specifically if you are loading images that are
+larger than your cache.maxByteSize.  These can be remedied by one of the
```
dnfield: Does this cause thrash, or just mean you won't cache an image that you used to cache?

src/docs/release/breaking-changes/12_18_19_ImageCache_Logic.md
```diff
@@ -0,0 +1,69 @@
...
+following:
+
+  1) Increase the ImageCache.maxByteSize to accommodate the large images.
+  1) Adjust your image loading logic to guarantee that the images will fit
+     nicely into the ImageCache.maxByteSize of your choosing.
+  1) Subclass ImageCache, implement your desired logic, create a new bindings
+     that will serve up your subclass of ImageCache.
```
dnfield: It would help to have an example of how to do this here - where in a Flutter program should you set the cache, etc.

gaaclarke: I have an outstanding issue to add that documentation.  I don't think there is a good place to point people yet today: https://github.com/flutter/flutter/issues/47388

src/docs/release/breaking-changes/12_18_19_ImageCache_Logic.md
gaaclarke: I think it is unlikely that someone will get thrash where they didn't previously.  You would have to be causing build to happen regularly and also have one large image and small images that fit in 1000 bytes methinks.

gaaclarke: But thrashing would be the case where they'd want to address the change.

src/docs/release/breaking-changes/12_18_19_ImageCache_Logic.md
dnfield: We make this harder than I realized.  Nevermind.

sfshaza2:
Waiting for @Hixie and/or @dnfield to approve.

src/docs/release/breaking-changes/12_18_19_ImageCache_Logic.md
dnfield: Can you add a link in here to the example you're creating at https://github.com/flutter/flutter/pull/47467?  It can probably just be a link to the `PictureCache` class doc.

gaaclarke: Done.

PR #3454, opened 2019-12-19
Add .android note to add-to-app project setup page

jmagman:
Add note that `.android/` should not be edited to match [`.ios/` note.](https://flutter.dev/docs/development/add-to-app/ios/project-setup#module-organization)

See https://github.com/flutter/flutter/issues/32883

Fixes https://github.com/flutter/flutter/issues/32992.


src/docs/development/add-to-app/android/project-setup.md
```diff
@@ -97,6 +97,13 @@ This creates a `some/path/my_flutter/` Flutter module project with some Dart
 code to get you started and a `.android/` hidden subfolder that wraps up the
```
jmagman: @xster Is this specific enough to satisfy https://github.com/flutter/flutter/issues/32992?

PR #3468, opened 2019-12-20
Breaking change: MouseTrackerAnnotation requires parameter "key"

dkwingsmt:
Breaking change documentation for PR https://github.com/flutter/flutter/pull/46034.
Design doc: https://flutter.dev/go/key-based-mouse-tracker-annotation

### DIFF
--- a/src/docs/release/breaking-changes/mousetrackerannotation-requires-parameter-key.md
+++ b/src/docs/release/breaking-changes/mousetrackerannotation-requires-parameter-key.md
@@ -11,23 +11,29 @@ description: [Brief description similar to the "context" section below.]
 Mouse events, such as when a mouse pointer has entered a region, exited, or
 is hovering a region, are detected with the help of
 `MouseTrackerAnnotation`s that are placed on interested regions during the
-render phase. Upon each new frame or new event, `MouseTracker` compares the
-annotations hovered by the mouse pointer before and after the change, then
-dispatches callbacks accordingly. `MouseTracker` also tracks which
-annotations are on the screen by requring the annotation's owners to attach
-or detach the annotation promptly.
+render phase. Upon each update (a new frame or a new event), `MouseTracker`
+compares the annotations hovered by the mouse pointer before and after the
+update, then dispatches callbacks accordingly. `MouseTracker` also tracks
+which annotations are on the screen by requring the annotation's owners to
+attach or detach the annotation promptly.
 
-However, since the `MouseTrackerAnnotation` is a constant class, its properties
-can't be changed without replacing the annotation as a whole, which triggers
-the exit and enter callback if it's hovered by a mouse pointer.
+`MouseTracker` used to compare annotations based on identity, i.e. by
+checking if the _exact_ annotation is hovered before and after the update.
+Since the `MouseTrackerAnnotation` is a constant class, this means you
+couldn't change an annotation's properties without replacing the annotation
+as a whole, which would trigger the exit and enter callback.
 
-This issue is a blocker to the plan of supporting mouse cursor, since it’s
-very likely that the cursor assigned to a region needs to be changed.
+This issue blocked the support for mouse cursor, since it’s very likely that
+the cursor assigned to a region needs to be changed.
+
+This change fixes this issue by adding a key to `MouseTrackerAnnotation`,
+and making the comparison algorithm of `MouseTracker` based on the key.
 
 
 ## Description of change
 
-Class `MouseTrackerAnnotation` now requires a new parameter `key`:
+The `MouseTrackerAnnotation` class now has a new property `LocalKey key`,
+which is a required parameter in the constructor:
 
 ```diff
  class MouseTrackerAnnotation extends Diagnosticable {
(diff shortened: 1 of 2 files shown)


src/docs/release/breaking-changes/mousetrackerannotation-requires-parameter-key.md
```diff
@@ -0,0 +1,177 @@
...
+dispatches callbacks accordingly. `MouseTracker` also tracks which
+annotations are on the screen by requring the annotation's owners to attach
+or detach the annotation promptly.
+
+However, since the `MouseTrackerAnnotation` is a constant class, its properties
+can't be changed without replacing the annotation as a whole, which triggers
+the exit and enter callback if it's hovered by a mouse pointer.
+
+This issue is a blocker to the plan of supporting mouse cursor, since it’s
+very likely that the cursor assigned to a region needs to be changed.
```
Hixie: We should write this from the point of view of having already made the change. There used to be a problem, the problem was bla; this change resolves that issue.

src/docs/release/breaking-changes/mousetrackerannotation-requires-parameter-key.md
```diff
@@ -0,0 +1,177 @@
...
+
+
+## Description of change
+
+Class `MouseTrackerAnnotation` now requires a new parameter `key`:
```
Hixie: The `MouseTrackerAnnotation` class...

src/docs/release/breaking-changes/mousetrackerannotation-requires-parameter-key.md
```diff
@@ -0,0 +1,177 @@
...
++  final LocalKey key;
+
+   ...
+ }
+```
+
+Three of class `MouseTracker`'s methods, `attachAnnotation`,
+`detachAnnotation`, and `isAnnotationAttached`, instead of taking a
+single parameter of `MouseTrackerAnnotation`, will now take a
+single parameter `LocalKey`:
```
Hixie: here it gives the type of the parameter but earlier it gave the name of the parameter. we should be consistent.

src/docs/release/breaking-changes/mousetrackerannotation-requires-parameter-key.md
```diff
@@ -0,0 +1,177 @@
...
+
+-  void detachAnnotation(MouseTrackerAnnotation annotation);
++  void detachAnnotation(LocalKey annotationKey);
+ }
+```
+
+## Migration guide
+
+There are 2 ways of migrating affected code. The first way requires little
+work, while the second is much simpler when the annotation's properties
```
Hixie: "requires little work" means it's easy
"requires a little work" means it's not easy
I think this was supposed to be the second?

src/docs/release/breaking-changes/mousetrackerannotation-requires-parameter-key.md
```diff
@@ -0,0 +1,177 @@
...
+     super.attach(owner);
+-    RendererBinding.instance.mouseTracker.attachAnnotation(_hoverAnnotation);
++    RendererBinding.instance.mouseTracker.attachAnnotation(_hoverAnnotation.key);
+   }
+
+   @override
+   void detach() {
+-    RendererBinding.instance.mouseTracker.detachAnnotation(_hoverAnnotation);
++    RendererBinding.instance.mouseTracker.detachAnnotation(_hoverAnnotation.key);
+     super.detach();
```
Hixie: why does the render object do this instead of the layer tree? seems weird that you have to independently tell the layer tree and the mouse tracker about the annotations.

src/docs/release/breaking-changes/mousetrackerannotation-requires-parameter-key.md
dkwingsmt: Fixed.

src/docs/release/breaking-changes/mousetrackerannotation-requires-parameter-key.md
dkwingsmt: Fixed.

src/docs/release/breaking-changes/mousetrackerannotation-requires-parameter-key.md
dkwingsmt: Fixed.

src/docs/release/breaking-changes/mousetrackerannotation-requires-parameter-key.md
dkwingsmt: Yes. I used "requires less work" and made the difference of the 2 methods clearer.

src/docs/release/breaking-changes/mousetrackerannotation-requires-parameter-key.md
dkwingsmt: There are multiple reasons:
- Annotation attachment is used to indicate widget presence. Generally, an annotation should be detached if the owner widget is unmounted, for the reason described in https://github.com/flutter/flutter/pull/44631 "The change to onExit". Therefore I think it's better to handle it in the render object (or should it be the element?).
- Historical reason. Attaching and detaching annotations used to have immediate effects (triggering callbacks) and we had to do it during Element.activate and deactivate. This was no longer needed, so now we only need to do it in attach and detach, but we haven't fully moved it to layers yet.
- Future plan. We're planning to experiment a change that split annotations from the layer tree to fix issues such as https://github.com/flutter/flutter/issues/41194. Therefore we'd like to keep it separated from the layer tree (at least for now).

Hixie: It'll still have to be in the layer tree though right? Just done differently than today (many nodes dealing with one layer rather than one layer per node).

This just seems like an unnecessarily awkward API right now.

Anyway, the point of my questions is that you should explain this in the breaking change document. Especially if we're planning on changing this again, we should warn people that we're not done changing this API.

dkwingsmt:
Closing this because I found out that it's probably neither necessary nor sufficient. I've also closed the corresponding Flutter PR in favor of https://github.com/flutter/flutter/pull/48453.

PR #3516, opened 2020-01-09
Adding a new breaking change and cleaning up other files.

sfshaza2:
Staged:  https://sz-flutter.firebaseapp.com/docs/release/breaking-changes

Hi huge list of reviewers: @dnfield @shihaohong @Hixie @gaaclarke @HansMuller @Piinks!

Kate, I've included you in this PR, because I wasn't sure about the language used in one of the breaking change pages about golden tests. I've clarified the text a bit but am still not 100% sure about the wording.

Last month, I was slammed when some of you were adding our first breaking change pages to the website. I approved some PRs I probably shouldn't have. After all, it's "only" docs. ;) And it was a new thing, so we were still figuring it out.

Hans asked me today to add a breaking change page, so I took the opportunity to clean *everything* up. I would like you to EACH approve this PR, since each of you has contributed to work that I've heavily edited. I've also modified the template, @Hixie, which is why you're here. Hopefully I haven't messed up the meaning of any of the content.

Changes in this PR include:
* Adding Hans' breaking change page, theme-data-accent-properties.
* Renaming several files to be more consistent with our naming (removing dates, replacing underbars with hyphens, etc).
* Retitling several pages to be more consistent.
* Adding a missing page, as well as Hans' page, to the index file.
* Rewriting portions of every page to change future tense to past tense.
* Rewriting text to be more active and consistently replacing "may" with "might" except for the one place where it was correct.  :)
* Rewriting portions of every page to make info more clear and to explain what some concepts are to folks who aren't familiar.
* Updated links on every page to follow our conventions, and to replace hardcoded URLs to our most commonly used websites with macros.
* Updated the template page to reflect these changes. 

Thanks for your review!! I won't approve any new edits to this directory until we can land this PR.

Don't forget to check out the staged pages at the link above!

Oh, and from now on I'll be more strict on accepting PRs for breaking changes.  ;)


src/docs/release/breaking-changes/test-text-input.md
```diff
@@ -3,67 +3,73 @@ title: TestTextInput state reset
...
 ```dart
     expect(tester.testTextInput.editingState, isNull);
 ```
 
-to assert that the state hasn't been modified yet.
-
 ## Timeline
 
-This change is being proposed for December of 2019.
+TBA: This change is being proposed for December of 2019.
```
dnfield: This change landed already (in December)

dnfield: Otherwise this file LGTM

src/docs/release/breaking-changes/template.md
```diff
@@ -19,49 +25,71 @@ a breaking change?"]
...
 
-[This text should be backwards looking. People are reading this months from now, not today.
-Don't say "in a month" or talk about your plan to do something next week. You've done it,
-they're looking back to figure out how to migrate their code.]
+{% comment %}
+This text should be backwards looking, so write in past tense,
+not future tense. People are reading this months from now, not today.
+Don't say "in a month" or talk about your plan to do something next week.
+You've done it, they're looking back to figure out how to migrate their code.
+{% endcomment %}
```
Hixie: one thing i noticed in all the guides so far is that they seem a bit anemic in this section, we should probably work to be more detailed here going forwards. Often people will be looking at these a year from now and they'll want to know "when did this change, what version can i go back to to prepare, is there a deprecation window", etc. I think in the future we shouldn't accept the guides until this section can be filled in; we shouldn't publish guides with TBA in their timeline.

sfshaza2: @Hixie, I noticed that too and I completely agree. I did what I could to improve those sections, but they are definitely not consistent. I will think about updating this section and will definitely be more strict in future reviews.

sfshaza2:
To:  @dnfield @shihaohong @Hixie @gaaclarke @HansMuller @Piinks

Thanks so much for the LGTMs I have received so far! But there is one other thing I need from those who have authored breaking changes. Every single breaking change page needs an amended "Timeline" section. The timeline section should contain a couple of important pieces of info:

1. The version number where the change has landed, so something along the lines of:
```
Flutter 1.13.7-pre.38 • channel master •
```
Just specifying a date isn't sufficient.

2. (optional)  If any classes/methods/etc are being deprecated, in what version of the SDK will this occur?

So, if you own a breaking change page, please provide that info to me and I'll get it in.

shihaohong:
@sfshaza2 My change's "Timeline" section is contains "TBA" since it hasn't fully been landed yet and I know there was a discussion about that earlier in this PR. My plan was to update it accordingly once the change has been fully landed. Is that okay?

src/docs/release/breaking-changes/test-text-input.md
dnfield: This should be `v1.13.5` as the first version where this showed up.

dnfield: (unless we truly need the `pre` version that it showed up in before that, which would be really hard for anyone to sync to anyway)

PR #3536, opened 2020-01-12
No clear information on how to add child widgets to form

EmmanuelAmodu:
Advance users might figure what this is about intuitively but not new user. This information is not sufficient for beginners and those that might be using flutter for the first time. 

I recommend this
```
...
  @override
    // Build a Form widget using the _formKey created above.
    return Form(
      key: _formKey,
      child: // Build this out in the next steps.
    );
  }
}
```

Be changed to this

```
...
  @override
    // Build a Form widget using the _formKey created above.
    return Form(
      key: _formKey,
      child: Column(
          mainAxisAlignment: MainAxisAlignment.center,
          children: <Widget>[
                // Build this out in the next steps.
          ]
       )
    );
  }
}
```

### DIFF
--- a/src/docs/cookbook/forms/validation.md
+++ b/src/docs/cookbook/forms/validation.md
@@ -63,7 +63,11 @@ class MyCustomFormState extends State<MyCustomForm> {
     // Build a Form widget using the _formKey created above.
     return Form(
       key: _formKey,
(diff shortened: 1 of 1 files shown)


EmmanuelAmodu:
@googlebot I signed it!

PR #3537, opened 2020-01-13
Remove errant backtick

miketheman:
This extra backtick was causing further render issues for the rest of the paragraph, making it confusing to read.


legalcodes:
Nice catch @miketheman LGTM. Thanks!

PR #3556, opened 2020-01-17
Moved splash screen instructions to a page under /development/ui/splash-screen.

matthew-carroll:
Moved splash screen instructions to a page under /development/ui/splash-screen.


matthew-carroll:
@sfshaza2 do you know where I can specify a redirect so that the old splash screen URL redirects to the new one?

legalcodes:
> @sfshaza2 do you know where I can specify a redirect so that the old splash screen URL redirects to the new one?

If I understand your use case @matthew-carroll you should be able to do that here:
https://github.com/flutter/website/blob/master/firebase.json

matthew-carroll:
Thanks @legalcodes I added a redirect for the URL that I removed.

matthew-carroll:
@sfshaza2 @legalcodes do these changes look good?

PR #3561, opened 2020-01-19
Explain that test files must end with _test

### LINKED ISSUE #3560: 'An introduction to unit testing' page issue
The lack of mention that a test class needs to end with _test is confusing. It would be helpful if it was mentioned somewhere that there is a convention around naming test files. 

Example of my confusion:
https://stackoverflow.com/questions/59812714/testing-all-unit-test-in-dart

stimms:
Otherwise, the test runner will miss them. 

Fixes #3560


stimms:
@googlebot I signed it!

PR #3570, opened 2020-01-22
updated committee members name in alphabetically

nitishk72:
Earlier it was written that names are in alphabetically but it wasn't !


sfshaza2:
Closing.

PR #3572, opened 2020-01-22
Update CD for Android

woprandi:
Update CD page to use app bundle instead of apk since it's the recommanded upload format for the Google Play


woprandi:
@googlebot I signed it!

PR #3576, opened 2020-01-22
Break out shared content into separate .md file, use jekyll includes.

legalcodes:
Staged: 
https://jt-flutter-website-1.firebaseapp.com/docs/development/add-to-app/android/add-flutter-fragment#initial-route-with-a-cached-engine

https://jt-flutter-website-1.firebaseapp.com/docs/development/add-to-app/android/add-flutter-screen#initial-route-with-a-cached-engine/docs/development/add-to-app/android/add-flutter-screen#initial-route-with-a-cached-engine


sfshaza2:
Since this text applies *only* to add-to-app, I'd create the include file right in the same directory, as we do for the install pages, and preface the file name with an underbar. The _includes directory is good for include files that can occur anywhere (in multiple places) in the tree.

matthew-carroll:
Thanks for making this change @legalcodes - I actually made the same change yesterday but you beat me to a PR :)

I would probably recommend removing the title from the include file because the 2 places where it's included use different header levels.

Also, @sfshaza2 I'll put up a PR after this one that adjusts the navigation as you recommended.

legalcodes:
Thanks @matthew-carroll and @Sfshaza2 for the awesome feedback, will make these changes!

PR #3619, opened 2020-01-29
Update plugin-api-migration.md

cyanglaz:
Whiling doing migration for google map, I found that some information on this doc is missing and might be useful for people who works on the migration.

### DIFF
--- a/src/docs/development/packages-and-plugins/plugin-api-migration.md
+++ b/src/docs/development/packages-and-plugins/plugin-api-migration.md
@@ -74,27 +74,52 @@ The following instructions outline the steps for supporting the new API:
       // now automatically registers plugins.
     }
     ```
+    
+1. (Optional) If you removed `MainActivity.java`, update the `<plugin_name>/example/android/app/src/main/AndroidManifest.xml` to use `io.flutter.embedding.android.FlutterActivity`. For example:
+
+    <?code-excerpt "AndroidManifest.xml" title?>
+    ```xml
+     <activity android:name="io.flutter.embedding.android.FlutterActivity"
+            android:theme="@style/LaunchTheme"
+   android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|layoutDirection|fontScale"
+            android:hardwareAccelerated="true"
+            android:windowSoftInputMode="adjustResize">
+            <meta-data
+                android:name="io.flutter.app.android.SplashScreenUntilFirstFrame"
+                android:value="true" />
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN"/>
+                <category android:name="android.intent.category.LAUNCHER"/>
+            </intent-filter>
+        </activity>
+    ```
 
 1. (Optional) Create an `EmbeddingV1Activity.java` file that uses the v1
    embedding for the example project in the same folder as `MainActivity` to
-   keep testing the v1 embedding's compatibility with your plugin. For example:
+   keep testing the v1 embedding's compatibility with your plugin. Note that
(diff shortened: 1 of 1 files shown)


cyanglaz:
@xster seems to be out. Do any of you mind to review this one? @mklim @amirh

src/docs/development/packages-and-plugins/plugin-api-migration.md
```diff
@@ -74,27 +74,52 @@ The following instructions outline the steps for supporting the new API:
       // now automatically registers plugins.
     }
     ```
+    
+1. (Optional) If you removed `MainActivity.java`, update the `<plugin_name>/example/android/app/src/main/AndroidManifest.xml` to use `io.flutter.embedding.android.FlutterActivity`. For example:
```
mklim: The step above this says to keep the `MainActivity`. I don't think it makes sense to suggest this unless we also suggest removing the MainActivity. We suggest removing it in other places and last we talked about it that is @matthew-carroll's recommendation, so I think it makes sense to keep this note but also change the above to say to delete the `MainActivity` entirely instead.

src/docs/development/packages-and-plugins/plugin-api-migration.md
```diff
@@ -74,27 +74,52 @@ The following instructions outline the steps for supporting the new API:
...
+    package io.flutter.plugins.batteryexample;
 
     import android.os.Bundle;
+    import dev.flutter.plugins.e2e.E2EPlugin;
     import io.flutter.app.FlutterActivity;
-    import io.flutter.plugins.GeneratedPluginRegistrant;
+    import io.flutter.plugins.battery.BatteryPlugin;
 
-    public class EmbeddingV1Activity extends FlutterActivity {
+    public class EmbedderV1Activity extends FlutterActivity {
```
mklim: pedantic nit: We generally use "embedder" to refer to `embedder.h` in the engine, and "embedding" to refer to the actual Android embedding.

```suggestion
    public class EmbeddingV1Activity extends FlutterActivity {
```

src/docs/development/packages-and-plugins/plugin-api-migration.md
```diff
@@ -74,27 +74,52 @@ The following instructions outline the steps for supporting the new API:
...
       protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        GeneratedPluginRegistrant.registerWith(this);
+        BatteryPlugin.registerWith(registrarFor("io.flutter.plugins.battery.BatteryPlugin"));
+        E2EPlugin.registerWith(registrarFor("dev.flutter.plugins.e2e.E2EPlugin"));
       }
     }
     ```
+    
+1.  Add `<meta-data android:name="flutterEmbedding" android:value="2"/>` to the      `<plugin_name>/example/android/app/src/main/AndroidManifest.xml`.
```
mklim: nit: I would add a quick explanation, something like "This sets the example app to use the v2 embedding."

src/docs/development/packages-and-plugins/plugin-api-migration.md
cyanglaz: I think it might be beneficial to keep the above step completely, meaning leave the `MainActivity`; for cases like add-to-app, or people who edited their MainActivity for other purpose. If we say remove the `MainActivity` completely, people might panic if they have their custom stuff in the `MainActivity`.
Also, this doc is only for migration, so I don't know if we should consider add-to-app users getting confused by it.

src/docs/development/packages-and-plugins/plugin-api-migration.md
cyanglaz: done

PR #3643, opened 2020-02-01
Fix: FlutterEngine cannot be converted to PluginRegistry

EmmanuelAmodu:
Fix issues with wrong argument in 'MainActivity.java'

> error: incompatible types: FlutterEngine cannot be converted to PluginRegistry

### DIFF
--- a/src/docs/development/platform-integration/platform-channels.md
+++ b/src/docs/development/platform-integration/platform-channels.md
@@ -230,9 +230,9 @@ Flutter client side.
 
 <?code-excerpt "MainActivity.java" title?>
 ```java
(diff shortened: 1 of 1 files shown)


sfshaza2:
I've asked an engineer to verify this code.

johnpryan:
@EmmanuelAmodu The `configureFlutterEngine` hook is the correct one to use to register plugins. I couldn't reproduce the `FlutterEngine cannot be converted to PluginRegistry` issue. using this codelab. I'd recommend checking that you're on the latest `stable` channel, and doing a `flutter run` with first version of the project before making changes to `MainActivity.java`.

PR #3645, opened 2020-02-03
update constructor for less confusion

ninoM:


ninoM:
@googlebot I signed it!

PR #3648, opened 2020-02-03
Add docs linking to design docs

dnfield:
Adds a page under /docs/resources/design-docs that lists the currently available design docs.

@Sfshaza2 - is there some way to preview this?  I think it should work but haven't tested it.


christopherfujino:
Not sure if this is a problem with my local setup, but I'm getting this error when running the site locally and navigating to `/docs/resources/design-docs`:
```
  Server address: http://127.0.0.1:4002
  Server running... press ctrl-c to stop.
[2020-02-03 12:19:55] ERROR `/assets/js/go_link_index.js' not found.
[2020-02-03 12:19:56] ERROR `/assets/js/go_link_index.js' not found.
[2020-02-03 12:20:02] ERROR `/favicon.ico' not found.
```

sfshaza2:
@dnfield, a MILLION thanks for this PR though, yes. it doesn't quite work yet. I've staged it here:
https://sz-flutter.firebaseapp.com/docs/resources/design-docs

dnfield:
@christopherfujino helped me out with this locally.  It seems to work with the changes in the latest commit - @sfshaza2  mind taking another look?

dnfield:
I've also updated the side nav to include a link to the page.

src/_data/sidenav.yml
```diff
@@ -314,6 +314,8 @@
     # - title: Flutter learning resources # TODO
     - title: FAQ
       permalink: /docs/resources/faq
+    - title: Design Documents
+      permalink: /docs/resources/design-docs
```
dnfield: @sfshaza2  - this is what you were asking for right?

It looks like it's working in your latest staging of this.

PR #3650, opened 2020-02-03
Reorder add-to-app iOS instructions

jmagman:
Reorder the instructions to make it clear that embedding and linking are part of Option B.  Update Option C to refer to embedding in linking already mentioned in option B.

### DIFF
--- a/src/docs/development/add-to-app/ios/project-setup.md
+++ b/src/docs/development/add-to-app/ios/project-setup.md
@@ -221,45 +221,6 @@ some/path/MyApp/
   the flags `--xcframework --no-universal`.
 {{site.alert.end}}
 
-### Option C - Embed application and plugin frameworks in Xcode and Flutter framework with CocoaPods
-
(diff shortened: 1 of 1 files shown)


src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -313,6 +274,49 @@ You should now be able to build the project in Xcode using `⌘B`.
...
+Host apps using CocoaPods can add Flutter to their Podfile:
+
+<?code-excerpt "MyApp/Podfile" title?>
+```ruby
+pod 'Flutter', :podspec => 'some/path/MyApp/Flutter/{build_mode}/Flutter.podspec'
+```
+
+Embed and link the generated App.framework, FlutterPluginRegistrant.framework,
+and any plugin frameworks into your existing application
+as described in Option B.
```
jmagman: This paragraph is new.  Everything else was moved.

PR #3653, opened 2020-02-04
First time users should git clone from the stable branch

timsneath:
The getting started directions lead a novice user to `git clone` from the `master` branch, as a result of a recent edit. Instead, we should be encouraging users to start with the `stable` branch. Other branches are advanced scenarios that shouldn't be more than touched upon in the opening flow.


timsneath:
@sfshaza2 can you confirm as part of your review that this is the best code fence for Windows terminal commands? The two commands had different code fences, so I standardized on ` ```batchfile` per https://help.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks

PR #3661, opened 2020-02-06
Supplement commonly used @JsonKey annotation

includecmath:
Supplement commonly used @JsonKey annotation, reference high frequency questions in Stack Overflow and encountered at work.
In addition giving general solution to existing 'registration_date_millis' example:  
API snake_case  
Dart lowerCamelCase

### DIFF
--- a/src/docs/development/data-and-backend/json.md
+++ b/src/docs/development/data-and-backend/json.md
@@ -322,7 +322,7 @@ you can use the `@JsonKey` annotation with a name parameter:
 final int registrationDateMillis;
 ```
 
-Both server or client are best to follow the same naming strategy, 
+It's best if both server and client follow the same naming strategy.  
 `@JsonSerializable()` provide `fieldRename` enum to totally converting dart 
 fields into JSON keys.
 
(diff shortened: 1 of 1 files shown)


src/docs/development/data-and-backend/json.md
```diff
@@ -322,6 +322,35 @@ you can use the `@JsonKey` annotation with a name parameter:
 final int registrationDateMillis;
 ```
 
+Both server or client are best to follow the same naming strategy,
```
sfshaza2: => It's best if both server and client follow the same naming strategy.

src/docs/development/data-and-backend/json.md
```diff
@@ -322,6 +322,35 @@ you can use the `@JsonKey` annotation with a name parameter:
...
+adding `@JsonKey(name: '<snake_case>')` to each field.
+
+Sometimes server data is uncertain, so it is necessary to verify and protect data
+ on client.  
+Other commonly used `@JsonKey` annotations include: 
+
+<!-- skip -->
+```dart
+/// Tell json_serializable that "defaultValue" to use if the JSON does not 
+/// contain this key or if the value is `null`.
```
sfshaza2: => /// Tell json_serializable to use "defaultValue"  if the JSON does not 
      /// contain this key, or if the value is `null`.

src/docs/development/data-and-backend/json.md
```diff
@@ -322,6 +322,35 @@ you can use the `@JsonKey` annotation with a name parameter:
...
+@JsonKey(defaultValue: false)
+final bool isAdult;
+
+/// When `true` tell json_serializable that JSON must contains the key, 
+/// If the key does not exist, an exception will thrown.
```
sfshaza2: => /// When `true`, tell json_serializable that JSON must contain the key. 
      /// If the key doesn't exist, an exception is thrown.

Issue #3663, opened 2020-02-06
I want to add my article to state management approaches segment


bardiarastin:
Hi, Recently I've published an article about how to manage Flutter applications state using Flutter Redux
https://medium.com/@bardiarastin/flutter-meets-redux-the-redux-way-of-managing-flutter-applications-state-f60ef693b509

I would like to add this to the "List of state management approaches" page, on the Redux segment, can I make a PR, please?!

![image](https://user-images.githubusercontent.com/20615964/73958789-b226fc00-491d-11ea-8bef-3a5762bd8eb0.png)

sfshaza2:
Absolutely! It would happen faster if you could file a PR.  :)

bardiarastin:
@Sfshaza Hold my beer and wait a few minutes... <3

sfshaza2:
:D. Will do! I hate beer, so it's safe with me.

bardiarastin:
@Sfshaza so much thanks, https://github.com/flutter/website/pull/3664 there you go!

PR #3664, opened 2020-02-06
Update options.md

bardiarastin:
#3663


bardiarastin:
@googlebot I signed it!

sfshaza2:
The build choked, so I had to kick it off again.

PR #3670, opened 2020-02-07
Add survey-vers.json

jayoung-lee:
This is a json file for future quarterly surveys.

PR #3672, opened 2020-02-07
Create survey-vers.json under src/f/

jayoung-lee:
...to be able to get a URL for this file

### DIFF
--- a/f/survey-vers.json
+++ b/f/survey-vers.json
@@ -1,0 +1,7 @@
+{
+  "uniqueId": "2020Q1",
+  "startMsEpoch": "2020/03/03 9:00",
+  "endMsEpoch": "2020/03/10 9:00",
+  "title": "Help improve Flutter! Take our Q1 survey.",
+  "urlPrefix": "https://google.qualtrics.com/jfe/form/SV_6YBXYxIR69pgpr7"
+}


f/survey-vers.json
```diff
@@ -0,0 +1,7 @@
+{
+  "uniqueId": "2020Q1",
+  "startMsEpoch": "2020/03/03 9:00",
+  "endMsEpoch": "2020/03/10 9:00",
```
DanTup: @jayoung-lee @pq Should we include timezone info here to ensure everyone parses this the same? Using something like [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) might be better, since it's unambigious and includes timezones.

```
> new Date().toISOString()
"2020-02-07T18:56:09.261Z"
```

pq: I'm up for whatever is easiest for the authors.  My thinking is that clients will implement this just once but survey authors will have to revisit every time they want to open a new window.

@jayoung-lee what do you think?

jayoung-lee: ISO 8601 is definitely human-readable, so I have no complaints. Will the format below work? Some comments from @devoncarew on https://github.com/flutter/website/pull/3670 are addressed too. 

Filename: `src/f/flutter-survey-metadata.json`
```
{
"uniqueId": "2020Q1",
"title": "Help improve Flutter! Take our Q1 survey.",
"url": "https://google.qualtrics.com/jfe/form/SV_6YBXYxIR69pgpr7"
"startDate": "2020-03-03T09:00:00-08:00",
"endDate": "2020-03-10T09:00:00-07:00",
"_comment": "UniqueID must be updated with each new survey so IDEs know to re-prompt users. Title should not exceed 45 characters. Dates should follow ISO 8601 standard."
}
```
FYI the current title is 41 chars long.

devoncarew: Looks great!

If you want to allow the comment to span multiple lines, you could do something like:

```json
{
	"_comments": [
		"uniqueId must be updated with each new survey so IDEs know to re - prompt users.",
		"title should not exceed 45 characters.",
		"Dates should follow ISO 8601 standard."
	],
	"uniqueId": "2020Q1",
	"title": "Help improve Flutter! Take our Q1 survey.",
	"url": "https://google.qualtrics.com/jfe/form/SV_6YBXYxIR69pgpr7",
	"startDate": "2020-03-03T09:00:00-08:00",
	"endDate": "2020-03-10T09:00:00-07:00"
}
```

pq: Can we simplify the format to `"2020-03-03T09:00:00"`?

jayoung-lee: I think @DanTup needs time zone information. What about using `"2020-03-03T09:00:00"` and assume UTC? I can update the comment to `"Dates should follow ISO 8601 standard and use UTC time zone."`

DanTup: > "Dates should follow ISO 8601 standard and use UTC time zone."

I think this may be a bit confusing, since ISO 8601 says that if there's no timezone then it's local time. It'd be easy to read the above and assume an ISO 8601 parse method would do the right thing because it doesn't sound like it contradicts the standard (and in the UK, it would be correct for half of the year).

It's true that there aren't many consumers of this, but if the timezone is included (eg. `-8:00`) I think it'd be clear when modifying it what the timezone was - and it's easier to review diffs/PRs that might collapse the comments above out (though I suspect in general only the dates will be modified and times generally left alone).

(I don't feel strongly though - the only important thing to me is just not doing my own date parsing :-) manipulations like appending a `Z` are ofc trivial)

pq: >  It'd be easy to read the above and assume an ISO 8601 parse method would do the right thing

Maybe we can specify in the comment string how to parse it in, say Java, TS,  and Dart?  Or link to a doc that specifies it?

FWIW, the format proposed doesn't work w/ the preferred Java ISO Formatter.

```java
Instant.parse("2020-03-03T09:00:00-08:00");
```
throws a parse exception.

Some docs on format expectations here:

https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_INSTANT

Maybe we could add a unit test to the web site tests that validates that specified dates parse in the expected ways?

DanTup: > Maybe we can specify in the comment string how to parse it in, say Java, TS, and Dart? Or link to a doc that specifies it?

Sure, but something that just worked if you parsed it the simplest way without reading any docs seems better :-)

> FWIW, the format proposed doesn't work w/ the preferred Java ISO Formatter.

It looks like `Instant` only allows UTC times. I guess it would work with the trailing `Z`? Though it be better to use something that supports any ISO-8601 dates (possibly [this one](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_DATE_TIME)) to avoid requiring the date to be UTC?

pq: Sorry, yeah.   A lot has happened with date support in Java since I last looked!  As. you suggested, the proposed format does play nice w/

```java
LocalDate.parse("2020-03-03T09:00:00-08:00", DateTimeFormatter.ISO_DATE_TIME);
```

which looks good to me.

Unless there are any other reservations and assuming parsing it is equally standard for you,  I say let's go for it.

Thanks for the back-and-forth @DanTup !

pq:
@jayoung-lee: do you have a follow-up PR planned with the updated format?

jayoung-lee:
@pq @DanTup It seems like the suggested format `2020-03-03T09:00:00-08:00` works well for both of you!

I will submit another PR revising this file.

pq:
Thanks!

PR #3673, opened 2020-02-07
Migration guide: Annotations return local position relative to object

dkwingsmt:


src/docs/release/breaking-changes/annotations-return-local-position-relative-to-object.md
```diff
@@ -0,0 +1,100 @@
+---
+title: [Annotations return local position relative to object]
+description: [Provide mouse events with reliable and meaningful local positions.]
```
sfshaza2: Please remove the square brackets on both of these lines.

sfshaza2:
Actually, I didn't mean to approve. Please first make the changes I mentioned.  :)

dkwingsmt:
@sfshaza2 Thanks! I've fixed the issues

src/docs/release/breaking-changes/annotations-return-local-position-relative-to-object.md
```diff
@@ -0,0 +1,100 @@
+---
+title: Annotations return local position relative to object
+description: Provide mouse events with reliable and meaningful local positions.
+---
+
+## Summary
+
+The local position returned in an annotation search is now relative to the
+region that contains the annotation object instead of the layer. This majorly
```
Hixie: s/majorly/significantly/

src/docs/release/breaking-changes/annotations-return-local-position-relative-to-object.md
```diff
@@ -0,0 +1,100 @@
...
+
+This change shouldn't affect most people, if any. The `MouseRegion` couldn't
+output a valid `localPosition` at the time of the change, while only other
+use case in the Flutter framework was by `SystemUiOverlayStyle`, which didn't
+depend on the local position.
+
+If any one happens to have relied on this property, they should have found
+this to be unreliable, in which case this change is a fix. In the case that
+someone actually wants the relative location to the layer, a case-by-case
+analysis is needed to figure out a migration plan.
```
Hixie: i don't really understand how this is a breaking change if we're telling people they can't be broken?

src/docs/release/breaking-changes/annotations-return-local-position-relative-to-object.md
```diff
@@ -0,0 +1,100 @@
...
+design mistake. The offset from the layer is meaningless and unreliable. For
+example, a `Transform` widget will draw on the same layer with an offset if
+the transform matrix is a simple translation, or push a dedicated
+`TransformLayer` if the matrix is non-trivial. The former case will keep the
+previous coordinate origin (for example, the top left corner of the app),
+while the latter case will move the position origin since it's on a new
+layer, and the difference of them might just be an unnoticeable scale of
+99%.
+
+Moreover, the major (if not only) users of `AnnotatedRegionLayer` are render
```
Hixie: This is written in a very defensive tone. The only people who will be reading this are people who have been affected. So we should start from the assumption that this had an effect, and work from there. Assume someone was broken by this. Why would they be broken? What can they do about it? etc.

src/docs/release/breaking-changes/annotations-return-local-position-relative-to-object.md
dkwingsmt: If some developers use `AnnotatedRegionLayer` instead of the widgets or objects, and they rely on the `localPosition` of the result, they will be affected. However, I don't know what they expected from `localPosition` if so, since the old behavior is unreliable, hence I can't give more information about how to migrate until more discussion. If they expected local position relative to the object, this change is a fix. If they expected local position relative to the layer, the feature is gone, and it'll be a feature request to add it back (if the use case is deemed necessary).

It's a migration guide because it broke a test. When I added this feature, a test was added, although I failed to think twice about the use case due to my poor understanding about painting at the time.

PR #3678, opened 2020-02-08
Emphasize that fetchPost should not return null.

petermichaux:
A bit more clarity for someone new to `FutureBuilder` so that the don't misinterpret the value of `snapshot.hasData`. When Dart has non-null by default types, this will be less critical. But for now, danger lurks in the future resolving to `null`.


petermichaux:
I signed the Google CLA.

petermichaux:
@googlebot I signed it!

PR #3682, opened 2020-02-10
Update and rename survey-vers.json

jayoung-lee:
Updated some details in the file as per https://github.com/flutter/website/pull/3672

CC: @pq @DanTup @devoncarew

### DIFF
--- a/f/flutter-survey-metadata.json
+++ b/f/flutter-survey-metadata.json
@@ -1,0 +1,12 @@
+{
+	"_comments": [
+		"uniqueId must be updated with each new survey so IDEs know to re-prompt users.",
+		"title should not exceed 45 characters.",
+		"startDate and endDate should follow ISO 8601 standard with a timezone offset."
+	],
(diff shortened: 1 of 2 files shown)


f/flutter-survey-metadata.json
```diff
@@ -0,0 +1,12 @@
+{
+	"_comments": [
```
legalcodes: Nit: I haven't found any of our `.json` files that use tabs (they use spaces). Is there a need to use tabs for `f/*.json` files?

sfshaza2: Interesting. I googled tabs and JSON, and it seems like it can be problematic. But I have no vote in this race.  :)

jayoung-lee: I did not know that. Let me fix it!

jayoung-lee: Done! I just removed the tabs just in case.

PR #3685, opened 2020-02-11
Add Floating SnackBar Offset migration guide

shihaohong:
cc/ @filaps


HansMuller:
Will land this change after checking the 3rd party test suite (assuming that the tests pass).

PR #3687, opened 2020-02-12
Update inject_embed URL

legalcodes:


johnpryan:
was the update to the flutter submodule intentional?

Issue #2836, opened 2019-07-17
'Fetch data from the internet' page issue, Post is a bad example class name for an HTTP example


dejaime:
Page URL: https://flutter.dev/docs/cookbook/networking/fetch-data.html
Page source: https://github.com/flutter/website/tree/master/src/docs/cookbook/networking/fetch-data.md

Hi,
I got pretty confused while reading the docs, with it doing post and get requests, until I realized post was actually  referring to a blog post or similar, not an HTTP post request. It was partially my fault for skipping a small portion of the page, but also due to the confusing name of the "Post" class.

Take the following code from the http example: 
```
FutureBuilder<Post>(
  future: fetchPost(),
  builder: (context, snapshot) {
    if (snapshot.hasData) {
      return Text(snapshot.data.title);
    } else if (snapshot.hasError) {
      return Text("${snapshot.error}");
    }

    // By default, show a loading spinner.
    return CircularProgressIndicator();
  },
);
```

Reading this, a beginner like me would assume this is doing a **post** request, due to `FutureBuilder<Post>` and `fetchPost()` (although fetch+Post did sound kind of weird). Still, `fetchPost` actually executes a **get** request.

I'd propose renaming Post to something like BlogPost, to make it a tad more clear. The example would then read something like this:
```
FutureBuilder<BlogPost>(
  future: fetchBlogPost(),
  builder: (context, snapshot) {
    [...]
  },
);
```

Thank you for your time!

petermichaux:
I agree this is confusing. It should be called "Article".

Unfortunately, the API uses "posts".

https://jsonplaceholder.typicode.com/posts/1

Also I think that inside `_MyAppState` the member named `post` should be called `_futurePost`. It references a `Future<Post>` object. It does not reference a `Post` object. (If the change to "Article" is possible then it should be `_futureArticle`.

sfshaza2:
Ugh, yes, this is unfortunate. @petermichaux, are you suggesting that, instead of fetching blog posts that we fetch photo albums? I am definitely ok with that. PR me!

petermichaux:
@sfshaza2 yes that's what I'm suggesting. It will help with the POST/PUT/DELETE recipes I've proposed.

sfshaza2:
I like it.

dejaime:
For some reason, I thought my July 2019 commit dc3f567 had already gone through.

sfshaza2:
Was that commit attached to a PR?

petermichaux:
[...]

PR #3693, opened 2020-02-13
Replace Post with Album

hemanthrajv:
Changing the cookbook `fetch-data` & `authenticated-requests` examples to use `albums` instead if `posts` as someone new to the example would misinterpret the model `Post` with the http request type `POST`.

Closes: #2836


hemanthrajv:
@googlebot I signed it!

PR #3694, opened 2020-02-13
Moving survey metadat to src/f/

jayoung-lee:
Moving the metadata file to the correct directory. Sorry! 😅


jayoung-lee:
Thanks @legalcodes  :)

PR #3704, opened 2020-02-15
Fix typos in references in internationalization.md

orestesgaolin:
There were missing colons in references section that caused page to render incorrectly.


orestesgaolin:
@googlebot I signed it!

PR #3749, opened 2020-02-28
Adds 'zip' to linux prereqs

zanderso:
unzip does not necessarily imply zip, so it needs to be listed separately.


zanderso:
Thanks!

PR #3768, opened 2020-03-05
Update Maven URL to https

jmagman:
Match URL in tool: https://github.com/flutter/flutter/pull/52042

### DIFF
--- a/src/docs/development/add-to-app/android/project-setup.md
+++ b/src/docs/development/add-to-app/android/project-setup.md
@@ -202,7 +202,7 @@ repositories {
     // if using a relative path.
   }
   maven {
-    url 'http://download.flutter.io'
+    url 'https://storage.googleapis.com/download.flutter.io'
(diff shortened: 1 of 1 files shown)


src/docs/development/add-to-app/android/project-setup.md
```diff
@@ -202,7 +202,7 @@ repositories {
     // if using a relative path.
   }
   maven {
-    url 'http://download.flutter.io'
+    url 'https://storage.googleapis.com/download.flutter.io'
```
xster: sad loss of vanity url, but seems sensible. I have a feeling this string exists in more places such as flutter/samples etc.

jmagman: yup https://github.com/flutter/samples/blob/master/add_to_app/android_using_prebuilt_module/app/build.gradle#L36

PR #3776, opened 2020-03-08
Added update and delete recipe to the cookbook

### LINKED ISSUE #3679: Cookbook recipes: POST/PUT/DELETE data to the internet.
The cookbook has the recipe [Fetch data from the internet](https://flutter.dev/docs/cookbook/networking/fetch-data). Recipes also showing how to POST/PUT/DELETE data to the internet would be good to make a complete set of recipes for all four of the fundamental CRUD operations using `FutureBuilder`.

For example, the main point in the POST and PUT recipes would be to show that the future would be `null` to begin with and there would be no `initState`. When the future is `null`, the widget will show the save button. When a user presses the save button, then `setState` to change the future to non-`null`. When the future is non-`null` then the UI will show spinner, error message, or something indicating success. Anyway, it would be a very helpful companion to the "Fetch data from the internet" recipe.

With DELETE, there is an issue with what the future will return. It can't return `null` for success as then `snapshot.hasData` becomes complicated. It could return boolean value `true`.

Making this all easy for someone new would help get them up to speed with Flutter with less rethinking about things others have already had to think through.

Perhaps POST could be covered in one recipe called "Create data on the internet". PUT in a recipe called "Update data on the internet". DELETE in a recipe called "Delete data from the internet".

hemanthrajv:
Closes #3679

### DIFF
--- a/src/docs/cookbook/networking/update-data.md
+++ b/src/docs/cookbook/networking/update-data.md
@@ -40,7 +40,7 @@ import 'package:http/http.dart' as http;
 
 ## 2. Updating data over the internet
 
-In this example, let's update an album title to the
+In this example, you'll see how to update an album title to the
 [JSONPlaceholder][] using the [http.put()][] method
 
 <!-- skip -->
@@ -75,9 +75,9 @@ convert the `http.Response` into a Dart object.
 
 ### Create an Album class
 
-First, create a `Album` class that contains the data from the
+First, create an `Album` class that contains the data from the
 network request. It includes a factory constructor that
-creates a `Album` from JSON.
+creates an `Album` from JSON.
 
 Converting JSON by hand is only one option.
 For more information, see the full article on
@@ -100,7 +100,7 @@ class Album {
 }
 ```
 
-### Convert the `http.Response` to a `Album`
+### Convert the `http.Response` to an `Album`
 
 Now, use the following steps to update the `updateAlbum()`
 function to return a `Future<Album>`:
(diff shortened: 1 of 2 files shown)


src/docs/cookbook/networking/update-data.md
```diff
@@ -0,0 +1,377 @@
...
+
+While it's easy to make a network request, working with a raw
+`Future<http.Response>` isn't very convenient. To make your life easier,
+convert the `http.Response` into a Dart object.
+
+### Create an Album class
+
+First, create a `Album` class that contains the data from the
+network request. It includes a factory constructor that
+creates a `Album` from JSON.
```
sfshaza2: a `Album` => an `Album` (in lines 78 and 80)

src/docs/cookbook/networking/update-data.md
```diff
@@ -0,0 +1,377 @@
...
+  factory Album.fromJson(Map<String, dynamic> json) {
+    return Album(
+      id: json['id'],
+      title: json['title'],
+    );
+  }
+}
+```
+
+### Convert the `http.Response` to a `Album`
```
sfshaza2: a -> an

src/docs/cookbook/networking/update-data.md
```diff
@@ -0,0 +1,377 @@
...
+
+Now, use the following steps to update the `updateAlbum()`
+function to return a `Future<Album>`:
+
+  1. Convert the response body into a JSON `Map` with the `dart:convert`
+     package.
+  2. If the server returns a `UPDATED` response with a status code of 200, then convert
+     the JSON `Map` into a `Album` using the `fromJson()` factory method.
+  3. If the server doesn't return a `UPDATED` response with a status code of 200,
```
sfshaza2: a `Album` -> an `Album`
a `UPDATED` => an `UPDATED` (in both steps 2 and 3)

hemanthrajv:
Hi @sfshaza2 , sorry about the errors, I fixed them as per your request, do let me know if any other changes needed. Thanks!

hemanthrajv:
@sfshaza2 done! Hope this is good.

hemanthrajv:
Thank for accepting. I also made changes of the send-data recipe. 
PTAL: #3780

sfshaza2:
I've already merged with my PR.  :D. Thanks, again!

PR #3788, opened 2020-03-10
Fix add-to-app iOS project setup page

mbaleczny:
Error in markdown caused incorrectly generated page where **Option B - Embed frameworks in Xcode** was hidden in a code block. Also, did resolve some other small markdown warnings.


mbaleczny:
@googlebot I signed it!

PR #3796, opened 2020-03-12
Update platform-channels.md

ivanryndyuk:
Binary messenger `controller.binaryMessenger` should be passed to FlutterMethodChannel initializer in Objective-C code exmaple.


ivanryndyuk:
@googlebot I signed it!

PR #3805, opened 2020-03-16
Adding a note about a current bug.

sfshaza2:
Thanks for this info, @gspencergoog!

### DIFF
--- a/src/docs/cookbook/forms/focus.md
+++ b/src/docs/cookbook/forms/focus.md
@@ -25,9 +25,10 @@ you can set the focus to the text field for the search term.
 This allows the user to start typing as soon as the screen
 is visible, without needing to manually tap the text field.
 
-In this recipe, learn how to give the focus to a text field as soon
-as it's visible,
-as well as how to give focus to a text field when a button is tapped.
+In this recipe, learn how to give the focus
+to a text field as soon as it's visible,
+as well as how to give focus to a text field
+when a button is tapped.
 
 ## Focus a text field as soon as it's visible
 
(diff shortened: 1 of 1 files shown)


src/docs/cookbook/forms/focus.md
```diff
@@ -142,6 +143,14 @@ FloatingActionButton(
 
 ## Interactive example
 
+{{site.alert.note}}
+  While the following code runs in DartPad, it fails
+  due to [Issue 52221][] when run against the 
+  stable (1.12) or beta (1.14) versions of the
+  Flutter SDK. Until this issue has landed,
```
gspencergoog:
```suggestion
  Flutter SDK. Until the fix for this issue has landed in the channel you are using,
```

I should probably track down what actually fixed this problem, so we can link to the PR/hash that fixes it.

sfshaza2: If you do, let me know, but I'm going to push it, with your edit. Thx!

gspencergoog:
FYI, this issue was fixed in https://github.com/flutter/flutter/pull/50372, which is hash bf551a31fe7ef45c854a219686b6837400bfd94c, so once that makes it into the channel you're interested in, the example should be fine.

src/docs/cookbook/forms/focus.md
sfshaza2: Thanks for the LGTM, @gspencergoog!

gspencergoog:
What happened to the "fix for" in my edit?  And also, now that I found where it was fixed, we should add something that describes when it will land (give the hash where it was fixed).

PR #3809, opened 2020-03-17
Fix up code analysis/testing

dnfield:
Pretty sure this passed locally but I want to check a run through on travis to be sure.

### DIFF
--- a/src/docs/development/ui/interactive.md
+++ b/src/docs/development/ui/interactive.md
@@ -256,9 +256,12 @@ it. In the same location, create the stateful widget:
          ],
        ),
      );
-@@ -117,2 +108,2 @@
-     );
-   }
+@@ -114,3 +105,5 @@
+           ),
+         ),
+       ],
++    );
++  }
 ```
 
 That's it! When you hot reload the app, the star icon should now respond to taps.
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,10 +28,6 @@ env:
     - TASK="./tool/check-code.sh"
     - TASK="bundle exec jekyll build"
 
-matrix:
-  allow_failures:
-    - env: TASK="./tool/check-code.sh"
-
 before_install:
   - source ./tool/env-set.sh
   - ./tool/before-install.sh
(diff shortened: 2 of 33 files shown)


dnfield:
Draft prevents it from being merged until it's marked ready.  CI is still failing for some reason not clear to me yet.  I'll get back to this soon and get it running.

dnfield:
List of changes/fixes:

- Remove the .dart_tool directories before running analysis, due to a bug in the Dart SDK on Flutter stable related to this. It is fixed on master and we should remove this change (TODO in the script about this) once we roll the next stable.
- Add some more missing skips to incomplete code examples.
- Revert some previous changes that were fixes for rolling to master, not to stable (mostly around theme breaking changes).
- Fix some broken tests that were setting `FlutterError.onError` without resetting it - replaced with `tester.takeException`.
- Rework code extraction logic to preserve the path, to avoid confusion when similarly named markdown files exist in different directories (e.g. `foo/bar/introduction.md` and `/baz/qux/introduction.md`).
- Remove optional `new` keywords from one document that was missing them.

dnfield:
Going to land this since CI is now passing.  We'll keep an eye on it if the tasks start failing again.

src/docs/development/ui/interactive.md
```diff
@@ -256,9 +256,12 @@ it. In the same location, create the stateful widget:
...
      );
-@@ -117,2 +108,2 @@
-     );
-   }
+@@ -114,3 +105,5 @@
+           ),
+         ),
+       ],
++    );
++  }
```
chalin: Hey @dnfield, just checking: you ran the refresh script under Linux to get this result, right? (I've resumed investigation of this variance in behavior in macOS vs Linux.)

Good idea to commit the Linux refresh results so that you could get tests to pass.

dnfield: Correct.  I didn't get to dig into the platform differences much - I was a bit surprised by that, although I wonder if it's differences between `diff` on Linux and macOS at this point.

chalin: Yeah, it is due to `diff` version difference, see #3916

Issue #3811, opened 2020-03-17
add shorter devtools urls, and have them re-direct to the devtools pages


devoncarew:
We'd like to have slightly shorter urls for the devtools documentation (in order to have terse urls to display in the app).

Instead of the current urls:

```
https://flutter.dev/docs/development/tools/devtools/inspector
```

it would be great to have a url like:

```
https://flutter.dev/devtools/inspector
```

I think this would be easiest if flutter.dev had some re-directs - something like `https://flutter.dev/devtools/`* ==> `https://flutter.dev/docs/development/tools/devtools/*`.

cc @sfshaza2

sfshaza2:
Yes, this is achievable by editing the firebase.json file. PRs accepted!  :)

PR #3815, opened 2020-03-18
add redirects for devtools pages

devoncarew:
- this PR adds 301 redirects for devtools - it allows us to use short urls - `https://flutter.dev/devtools/memory` - and re-direct to the devtools content
- fix https://github.com/flutter/website/issues/3811

cc @sfshaza2


sfshaza2:
Thanks for this PR! Let me stage to confirm that it works.

sfshaza2:
It works!

PR #3818, opened 2020-03-19
Fix package name and fix example code

westy92:


sfshaza2:
Hi, @westy92! Thanks for this PR! Is this change intended to update the example to the 2.0 plugin API? I ask, because the advice is to keep this call for apps that don't use v2 embedding. [This wiki page](https://github.com/flutter/flutter/wiki/Experimental:-Create-Flutter-Plugin) has advice on how to support both.

@ditman

westy92:
As far as I can tell, the code example I touched was already using the 2.0 plugin API and it did not compile until I removed those lines.

westy92:
`GeneratedPluginRegistrant.registerWith(flutterEngine);` doesn't take a `FlutterEngine` as a parameter.

westy92:
I followed [this guide](https://github.com/flutter/flutter/wiki/Upgrading-pre-1.12-Android-projects) last night, to give context of my current understanding.

sfshaza2:
Thanks, @westy92. I've been working on some updated plugin docs in PR https://github.com/flutter/website/pull/3816, so it's fresh on my mind. I'll go ahead and merge this, but we may want to add a note later. I'm not sure how prevalent it is for folks to support both.

THANKS for this PR! Both I and my reviewers missed this change.

Issue #3800, opened 2020-03-13
'Build and release an iOS app' page issue


syonip:
Page URL: https://flutter.dev/docs/deployment/ios.html
Page source: https://github.com/flutter/website/tree/master/src/docs/deployment/ios.md

Description of issue:

The part about verifying the `Display Name` setting in xcode is misleading.
Changing the Display Name to anything else than `Runner` will cause the app to not run with flutter tools, as described here https://github.com/flutter/flutter/issues/41793

micpap25:
Looks like this is fixed, can be closed.

PR #3838, opened 2020-03-24
Adding a note about the Runner display name.

sfshaza2:
Addresses issue https://github.com/flutter/website/issues/3800.

Please review, @jmagman


sfshaza2:
Thanks for reviewing, @jmagman!

Issue #3784, opened 2020-03-10
'/docs/release/breaking-changes/container-color.html' page issue


kaboc:
Page URL: https://flutter.dev/docs/release/breaking-changes/container-color.html
Page source: https://github.com/flutter/website/tree/master/src/docs/release/breaking-changes/container-color.md

Pages linked from `ColoredBox` and `Colored box and container optimization #50979` in the `References` section at the bottom are not found.

kaboc:
As for the latter, I noticed the link was just broken, and made a PR (#3785).

sfshaza2:
Thanks for fixing the link to the PR! I've fixed the other links, which should link to the *master* API docs, not the stable API docs. (Unless the fix lands in stable..)

micpap25:
Fixed, looks like this should be closed

PR #3839, opened 2020-03-24
Correcting the API links to point at master.

sfshaza2:
The API links need to point to master.

Addresses issue https://github.com/flutter/website/issues/3784

### DIFF
--- a/src/docs/release/breaking-changes/container-color.md
+++ b/src/docs/release/breaking-changes/container-color.md
@@ -89,9 +89,9 @@ Relevant issues:
 Relevant PRs:
 * [Colored box and container optimization #50979][]
 
-[`Container`]: {{site.api}}/flutter/widgets/Container-class.html
-[`ColoredBox`]: {{site.api}}/flutter/widgets/ColoredBox-class.html
-[`BoxDecoration`]: {{site.api}}/flutter/painting/BoxDecoration-class.html
+[`Container`]: https://master-api.flutter.dev/flutter/widgets/Container-class.html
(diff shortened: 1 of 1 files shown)


src/docs/release/breaking-changes/container-color.md
```diff
@@ -89,9 +89,9 @@ Relevant issues:
 Relevant PRs:
 * [Colored box and container optimization #50979][]
 
-[`Container`]: {{site.api}}/flutter/widgets/Container-class.html
-[`ColoredBox`]: {{site.api}}/flutter/widgets/ColoredBox-class.html
-[`BoxDecoration`]: {{site.api}}/flutter/painting/BoxDecoration-class.html
+[`Container`]: https://master-api.flutter.dev/flutter/widgets/Container-class.html
+[`ColoredBox`]: https://master-api.flutter.dev/flutter/widgets/ColoredBox-class.html
+[`BoxDecoration`]: https://master-api.flutter.dev/flutter/widgets/BoxDecoration-class.html
```
legalcodes: Do we have a way to indicate to users when we are linking to non-stable versions of the api docs?

sfshaza2: Nope. It's never been an issue before but, now with breaking changes living on our site, it has become an issue. Feel free to file a bug against the infrastructure to identify this in some way, but please don't hold up this PR.

legalcodes: I wasn't talking about infrastructure -- I meant something like:

## References

{{site.alert.note}}
The following references link to api docs for non-stable versions of Flutter.
{{site.alert.

PR #3849, opened 2020-03-25
More updating the obfuscation and split-debug-info information.

sfshaza2:
Staged: https://sz-flutter.firebaseapp.com

I did a sweep through the docs, looking for more places to talk about obfuscating and split-debug-info. Also looked for links to the wiki page.


src/docs/perf/app-size.md
```diff
@@ -107,7 +109,13 @@ section in the Flutter [FAQ][].
 
 ## Reducing app size
 
-Some of the obvious things you can do to make your app smaller
+When building a release version of your app,
+consider using the `--split-debug-info` tag.
+This tag can dramatically reduce code size.
+For an example of using this tag, see
+[Obfuscating Dart code][].
```
jonahwilliams: Might be worth noting that `--split-debug-info` can be used without obfuscate.

src/docs/deployment/obfuscate.md
```diff
@@ -83,9 +83,24 @@ flutter symbolize -i <stack trace file> -d /out/android/app.android-arm64.symbol
...
+Be aware of the following when coding an app that will
+eventually be an obfuscated binary.
+
+* Code that relies on matching specific class, function,
+  or library names will fail.
+  For example, the following call to `expect()` will not
+  work in an obfuscated binary:
+
+```dart
+expect(Foo.runtimeType.toString(), equals('Foo'))
```
jonahwilliams: I would make it more obvious this is an instance instead of the class, since Foo.runtimeType should be `Type`

```suggestion
expect(foo.runtimeType.toString(), equals('Foo'))
```

src/docs/deployment/android.md
```diff
@@ -273,6 +276,9 @@ the Play Store.
 This section describes how to build a release app bundle.
 If you completed the signing steps,
 the app bundle will be signed.
+At this point, you might consider [obfuscating your Dart code][]
+to make it more difficult to reverse engineer. Obfuscating
+your code involves adding a couple flags to your build command.
```
jonahwilliams: Perhaps: "..and maintaining additional files to de-obfuscate stack traces"? or something to that nature.

sfshaza2:
Thanks for the LGTM, @jonahwilliams!

PR #3861, opened 2020-03-27
Clarify no mixing frameworks from different directories

xster:


src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -233,7 +239,7 @@ Embed and link the generated frameworks into your existing
 application in Xcode.  There are multiple ways to do
 this&mdash;use the method that is best for your project.
 
-### Link on the frameworks
+#### Link on the frameworks
```
xster: This makes the table of content indent to the right level

src/docs/development/add-to-app/ios/project-setup.md
```diff
@@ -223,6 +223,12 @@ some/path/MyApp/
         └── example_plugin.framework
 ```
 
+{{site.alert.warning}}
+  You mustn't mix `.framework` imports from different directories, such as
```
jmagman: "mustn't " is an unusual contraction, and @sfshaza2 always advises against future tense.
How about this, though she probably has a better solution for the parenthetical example:
```
Mixing `.framework` imports from different directories (Profile/Flutter.framework` and `Debug/App.framework`) causes runtime crashes.
```

dnfield: Maybe we could also just positively state what you're supposed to do first?

> Always use the same runtime mode for `Flutter.framework` and `App.framework.`  Mixing `.framework` imports from different directories (`Profile/Flutter.framework` and `Debug/App.framework`) will result in runtime crashes.

jmagman: I like the positive statement better, though "results in runtime crashes" or "causes runtime crashes" instead of "will result"

sfshaza2: I like the proposed two sentences! And I want to give @jmagman a badge for being a "Friend of the Google Developer Documentation Style Guidelines".   :D  If we had such a thing, that is.

xster: Thanks for the suggestions all
🏅

PR #3867, opened 2020-03-31
Add specific Xcode error text to promote search discoverability

jmagman:
@kf6gpe hit this error, but couldn't find the migration guide by Googling around.
Add the Xcode error text to increase discoverability

### DIFF
--- a/src/docs/development/ios-project-migration.md
+++ b/src/docs/development/ios-project-migration.md
@@ -47,4 +47,4 @@ add `$(inherited) -framework Flutter`.
 </li>
 </ol>
 
-[experience issues switching between iOS devices and simulators]: https://github.com/flutter/flutter/issues/50568
+[see the following errors when switching between iOS devices and simulators:]: https://github.com/flutter/flutter/issues/50568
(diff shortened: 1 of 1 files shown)


src/docs/development/ios-project-migration.md
```diff
@@ -10,7 +10,15 @@ description: How to migrate existing Flutter iOS projects to Xcode 11.4.
 
 To develop Flutter apps for iOS, you need a Mac with Xcode installed.
 Xcode 11.4 changed the way frameworks are linked and embedded, and
-you may [experience issues switching between iOS devices and simulators][].
+you may [see the following errors when switching between iOS devices and simulators:][]
```
christopherfujino: doesn't this text have to match the link at the bottom (line 50)?

kf6gpe: I think @christopherfujino is correct. Otherwise, looks good to me! Thanks for the speedy response to my lack of search-fu!

jmagman: sigh. yes. thanks for noticing.

jmagman: (the sigh comment was directed at @christopherfujino for noticing the link issue, not at @kf6gpe 😃 )

sfshaza2: @jmagman, another trick you might like better. You can do this:

```
you may [see the following errors when switching between iOS devices and simulators:][errors]

[errors]: URL
```

Putting a short-cut in the second pair of otherwise empty brackets, then put the shortcut in the list of URL references.

jmagman: ooo nice, done.

PR #3872, opened 2020-04-01
Document APK shared object compression

dcharkes:
See discussions on https://github.com/flutter/flutter/issues/53738 and https://github.com/objectbox/objectbox-dart/issues/97.

### DIFF
--- a/src/docs/development/platform-integration/c-interop.md
+++ b/src/docs/development/platform-integration/c-interop.md
@@ -359,7 +359,7 @@ This feature is not yet supported for web plugins.
 
 Flutter APKs by default don't follow these guidelines and compress `libflutter.so` and `libapp.so` - this leads to smaller APK size but larger on device size.
 
-Using `dart:ffi` inverts this default setting, and stops the compression of `libflutter.so`, `libapp.so`, and any user-added shared objects. To re-enable compression change `android/app/src/main/AndroidManifest.xml` in the following way.
+Shared objects from third parties can change this default setting with `android:extractNativeLibs="true"` in their `AndroidManifest.xml` and stop the compression of `libflutter.so`, `libapp.so`, and any user-added shared objects. To re-enable compression, override the setting in `your_app_name/android/app/src/main/AndroidManifest.xml` in the following way.
 
 ```diff
 @@ -1,5 +1,6 @@
--- a/src/docs/development/platform-integration/c-interop.md
+++ b/src/docs/development/platform-integration/c-interop.md
@@ -355,9 +355,9 @@ This feature is not yet supported for web plugins.
 
 ### Android APK size (shared object compression)
 
-Android guidelines in general recommend distributing native shared objects uncompressed because that actually saves on device space. Shared objects can be directly loaded from the APK instead of unpacking them on device into a temporary location and then loading. APKs are additionally packed in transit - that is why you should be looking at download size. 
+[Android guidelines][] in general recommend distributing native shared objects uncompressed because that actually saves on device space. Shared objects can be directly loaded from the APK instead of unpacking them on device into a temporary location and then loading. APKs are additionally packed in transit - that is why you should be looking at download size. 
 
-Flutter APKs by default don't follow these guidelines (we are in discussions about that) and compress `libflutter.so` and `libapp.so` - this leads to smaller APK size but larger on device size.
+Flutter APKs by default don't follow these guidelines and compress `libflutter.so` and `libapp.so` - this leads to smaller APK size but larger on device size.
 
 Using `dart:ffi` inverts this default setting, and stops the compression of `libflutter.so`, `libapp.so`, and any user-added shared objects. To re-enable compression change `android/app/src/main/AndroidManifest.xml` in the following way.
 
@@ -391,4 +391,5 @@ Using `dart:ffi` inverts this default setting, and stops the compression of `lib
 [FFI]: https://en.wikipedia.org/wiki/Foreign_function_interface
 [ffi issue]: {{site.github}}/dart-lang/sdk/issues/34452
 [Upgrading Flutter]: /docs/development/tools/sdk/upgrading
+[Android guidelines]: https://developer.android.com/topic/performance/reduce-apk-size#extract-false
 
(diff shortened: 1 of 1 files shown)


dcharkes:
@ditman @sfshaza2 I did not add a code snippet with 6 lines in this file in this PR. Can you please check why Travis is failing?

src/docs/development/platform-integration/c-interop.md
```diff
@@ -351,6 +351,35 @@ in binary form, use the following instructions:
 
 This feature is not yet supported for web plugins.
 
+## FAQ
+
+### Android APK size (shared object compression)
+
+Android guidelines in general recommend distributing native shared objects uncompressed because that actually saves on device space. Shared objects can be directly loaded from the APK instead of unpacking them on device into a temporary location and then loading. APKs are additionally packed in transit - that is why you should be looking at download size.
```
mraleph, line 358: I think you should link to https://developer.android.com/topic/performance/reduce-apk-size#extract-false in this sentence somewhere.

src/docs/development/platform-integration/c-interop.md
```diff
@@ -351,6 +351,35 @@ in binary form, use the following instructions:
 
 This feature is not yet supported for web plugins.
 
+## FAQ
+
+### Android APK size (shared object compression)
+
+Android guidelines in general recommend distributing native shared objects uncompressed because that actually saves on device space. Shared objects can be directly loaded from the APK instead of unpacking them on device into a temporary location and then loading. APKs are additionally packed in transit - that is why you should be looking at download size. 
+
+Flutter APKs by default don't follow these guidelines (we are in discussions about that) and compress `libflutter.so` and `libapp.so` - this leads to smaller APK size but larger on device size.
```
mraleph, line 360: Maybe don't include `we are in discussion about that` here.

src/docs/development/platform-integration/c-interop.md
```diff
@@ -351,6 +351,35 @@ in binary form, use the following instructions:
...
+
+Using `dart:ffi` inverts this default setting, and stops the compression of `libflutter.so`, `libapp.so`, and any user-added shared objects. To re-enable compression change `android/app/src/main/AndroidManifest.xml` in the following way.
```
mraleph, line 362: I would like to understand how `dart:ffi` impact this setting. Right now it is completely magical that this happens.

greenrobot: As a library provider, I'd like to add that people get confused when they use 3rd party libs and suddenly their APK size explodes. Ofc, they blame the library first.

If this is this a case of "works as designed" and you want to stick with, how about something like a build log to inform users more prominently about this behavior change when it kicks in.

dcharkes: Without `tools:replace="android:extractNativeLibs"` we get a source:

> is also present at [io.objectbox:objectbox-android:2.5.0] AndroidManifest.xml:22:9-42 value=(false).

@greenrobot what is `objectbox-android`? And where do I find its source code?

greenrobot: `objectbox-android` is an AAR (aka Android library) also containing the native ObjectBox .so files. You can download it from the repository, if you want to have a closer look: https://bintray.com/beta/#/objectbox/objectbox/io.objectbox:objectbox-android/2.5.1?tab=files

dcharkes: > This is because ObjectBox adds `extractNativeLibs="false"` to your `AndroidManifest.xml` as recommended by Google.

https://docs.objectbox.io/faq#how-much-does-objectbox-add-to-my-apk-size

This might very well be the cause.

mraleph: Based on this understanding I think we should rewrite this text and remove paragraphs that blame `dart:ffi`. Instead we can say that some libraries do this and it can be counteracted by doing `extractNativeLibs="true"`

dcharkes:
@ditman @sfshaza2 How do I run the website Travis scripts locally? I tried the following, but that didn't work.

```
$ FLUTTER_ROOT=~/flt/engine/flutter ./tool/build_check_deploy.sh --no-build --no-check-links
Using FLUTTER_ROOT '/Users/dacoharkes/flt/engine/flutter'
Using Dart SDK: /Users/dacoharkes/flt/engine/flutter/bin/cache/dart-sdk/bin/dart
Dart VM version: 2.8.0-dev.18.0.flutter-0f89df0c00 (be) (Wed Apr 1 00:35:37 2020 +0000) on "macos_x64"
+ rm -rf example.g
+ mkdir -pv example.g
+ cp example/README.md example/analysis_options.yaml example/pubspec.yaml example.g/
Running "flutter pub get" in example.g...                           2.6s
EXTRACTING code snippets from the markdown:
Error: Could not resolve the package 'path' in 'package:path/path.dart'.
tool/extract.dart:4:8: Error: Not found: 'package:path/path.dart'
import 'package:path/path.dart' as p;
       ^
tool/extract.dart:99:27: Error: Method not found: 'basename'.
  final String filename = p.basename(sourceFile.path);
                          ^^^^^^^^
tool/extract.dart:100:31: Error: Method not found: 'join'.
  final String dartFileName = p.join(
                              ^^^^
tool/extract.dart:105:13: Error: Method not found: 'join'.
  Directory(p.join(
            ^^^^
```

sfshaza2:
Hi, @dcharkes! Nice to meet you!

The problem is due to the fact that we recently re-enabled code testing on the website's code snippets. There is a 6-line code snippet in the `c-interop.md` file, at line 167 that is failing analysis, because it's only a partial example.

Just add the following line immediately before the code to tell the analyzer to skip this snippet:

```
<!-- skip -->
```

I'm not sure why this didn't trigger an error before.  @dnfield, any idea?

Also, please note that we break lines at 80 chars or less in our markdown.

sfshaza2:
Oh, and as far as running Travis locally. That is difficult for the SWEs, apparently because we use different versions of the tooling. The instructions are in the README file for the repo, but I'm happy to run it for you. I always have the server running and with `git pr <PR#>` it's super easy.

PR #3892, opened 2020-04-03
Update Basic Widgets example to avoid deprecated field.

Roomka:
Update Theme.of(context).primaryTextTheme.title to Theme.of(context).primaryTextTheme.headline6 as the previous one was deprecated after version v1.13.8

### DIFF
--- a/src/docs/development/ui/widgets-intro.md
+++ b/src/docs/development/ui/widgets-intro.md
@@ -143,7 +143,7 @@ class MyScaffold extends StatelessWidget {
           MyAppBar(
             title: Text(
               'Example title',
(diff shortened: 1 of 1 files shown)


Roomka:
@googlebot I signed it!

sfshaza2:
Not merging for now.

sfshaza2:
OK, I'm going to close this for now. The reason being that the Travis tests the website code to the current stable channel, which is 1.12. These Material Design fields were changed in 1.13. So, if we change them now, the build fails. I think a better approach is to leave them at 1.12, but add notes, saying what they are changed to in 1.13.

PR #3896, opened 2020-04-03
Optional banner change

timsneath:
I wondered what you thought of this alternative? The current banner text felt a little lacking in context to this casual reader.


sfshaza2:
A definite improvement! For context, Martin asked me to remove the Clock notice, so I disabled the banner, but this caused tabs to break! That meant that people were unable to download Flutter. So, I added back the banner and just threw up some alternate text. I figured we could fix it later, and you have!  :D

timsneath:
Yeah, I figured from the log!

I have a more substantive PR for the /web page pending your review; feel
free to ping me if there's anything there you want to chat about.

PR #3900, opened 2020-04-05
Add <code> tag

lightkeima:
I think there is a code tag for 'flutter create' command in this paragraph:
             <p>The ZIP file may contain other files and directories that are
               created using the **\<code\> flutter create \</code\>** command, such as
               <code>pubspec.yaml</code>; it may also include asset files
               (e.g. images, fonts) or data. We recommend running
               <code>flutter clean</code> before creating the ZIP file to
               strip out unnecessary binaries.</p>
I just added it.


lightkeima:
@googlebot I signed it!

PR #3906, opened 2020-04-06
Update assets-and-images.md

includecmath:
assets files located in directory or subdirectories are very common,  Note and example should help users more clearly. I find other page use {{site.alert.note}}.

### DIFF
--- a/src/docs/development/ui/assets-and-images.md
+++ b/src/docs/development/ui/assets-and-images.md
@@ -31,12 +31,15 @@ specify the directory name with the `/` character at the end:
 ```yaml
 flutter:
   assets:
-    - assets/
+    - directory/
+    - directory/subdirectory/
 ```
 
-Note that only files located directly in the directory are
-included. To add files located in subdirectories,
-create an entry per directory.
+{{site.alert.note}}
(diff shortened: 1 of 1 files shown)


src/docs/development/ui/assets-and-images.md
```diff
@@ -31,20 +31,23 @@ specify the directory name with the `/` character at the end:
...
-    - assets/
+    - directory/
+    - directory/subdirectory/
 ```
 
-Note that only files located directly in the directory are
-included. To add files located in subdirectories,
-create an entry per directory.
+{{site.alert.note}}
+  That only files located directly in the directory are
```
sfshaza2, line 39: Delete "That", so:  "Only files..."

includecmath: Thank you sfshaza2 for always helping me correct my English.
Where are you from? It's 1am in my country, but I have a little insomnia and learn flutter for a while.

sfshaza2: Yeah, I often suffer from insomnia, too.  Though I seem to be sleeping better with the social isolation!

PR #3908, opened 2020-04-06
Add showAutocorrectionPromptRect

LongCatIsLooong:
in order to the autocorrection highlight on iOS.

### DIFF
--- a/src/docs/release/breaking-changes/add-showAutocorrectionPromptRect.md
+++ b/src/docs/release/breaking-changes/add-showAutocorrectionPromptRect.md
@@ -86,7 +86,7 @@ API documentation:
 * [`TextInputClient`]: https://master-api.flutter.dev/flutter/services/TextInputClient-class.html
 
 Relevant issues:
-* [Issue 12920][{{site.github}}/flutter/flutter/issues/12920]
+* [Issue 12920]: {{site.github}}/flutter/flutter/issues/12920
 
 Relevant PRs:
 * [iOS UITextInput autocorrection prompt]: {{site.github}}/flutter/flutter/pull/54119/
--- a/src/docs/release/breaking-changes/add-showAutocorrectionPromptRect.md
+++ b/src/docs/release/breaking-changes/add-showAutocorrectionPromptRect.md
@@ -27,8 +27,7 @@ support autocorrect, you only need to add an empty implementation for the new me
 <!-- skip -->
 ```dart
 class CustomTextInputClient implements TextInputClient {
-  void showAutocorrectionPromptRect(int start, int end) {
-  }
+  void showAutocorrectionPromptRect(int start, int end) {}
 }
 ```
 
@@ -39,8 +38,8 @@ Code after migration:
 
 <!-- skip -->
 ```dart
-// Assume your `TextInputClient` is a `State` subclass, and it has a variable `_currentPromptRectRange` that controls 
-// the autocorrection highlight.
+// Assume your `TextInputClient` is a `State` subclass, and it has a variable 
+// `_currentPromptRectRange` that controls the autocorrection highlight.
 class CustomTextInputClient extends State<...> implements TextInputClient {
   @override
   void updateEditingValue(TextEditingValue value) {
@@ -48,7 +47,7 @@ class CustomTextInputClient extends State<...> implements TextInputClient {
     if (value.text != _value.text) {
       setState(() {
         _currentPromptRectRange = null;
-      });  
+      });
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/docs/release/breaking-changes/add-showAutocorrectionPromptRect.md
```diff
@@ -0,0 +1,90 @@
+---
+title: New method `showAutocorrectionPromptRect` added to `TextInputClient` 
+description: a new method `void showAutocorrectionPromptRect(int start, int end)` was added to the `TextInputClient` interface
```
Hixie, line 3: capitalization?

Hixie: (question because i don't know if this should be capitalized or not, we should check the other docs to be consistent)

src/docs/release/breaking-changes/add-showAutocorrectionPromptRect.md
```diff
@@ -0,0 +1,90 @@
...
+  }
+}
+```
+
+Otherwise, if your app targets iOS and supports autocorrect on iOS, it is recommended to add a sensible implementation of `void showAutocorrectionPromptRect(int start, int end)` 
+to your `TextInputClient` subclass. 
+
+Code after migration:
+
+<!-- skip -->
```
Hixie, line 39: why skip?

Hixie: we should make this valid code that we can analyze, so that we continue to update it

src/docs/release/breaking-changes/add-showAutocorrectionPromptRect.md
```diff
@@ -0,0 +1,90 @@
...
+        _currentPromptRectRange = null;
+      });  
+    }
+  }
+
+  void _handleFocusChanged() {
+    // When this text input loses focus, the autocorrection highlight needs
+    // to be dismissed.
+    if (!_hasFocus)
+      setState(() { _currentPromptRectRange = null; });
```
Hixie, line 58: we should have consistent style in this snippet. probably easiest way is to dartfmt it.

src/docs/release/breaking-changes/add-showAutocorrectionPromptRect.md
```diff
@@ -0,0 +1,90 @@
...
+    // When this text input loses focus, the autocorrection highlight needs
+    // to be dismissed.
+    if (!_hasFocus)
+      setState(() { _currentPromptRectRange = null; });  
+  }
+  
+  @override
+  void showAutocorrectionPromptRect(int start, int end) {
+    // Updates the range of the highlight, as iOS requested.
+    // Note this method will not be called when iOS decides to
```
Hixie, line 64: remote "Note this", it adds nothing (see style guide under "empty prose")

src/docs/release/breaking-changes/add-showAutocorrectionPromptRect.md
```diff
@@ -0,0 +1,90 @@
...
+}
+```
+
+## Timeline
+
+{% comment %}
+The version # of the SDK where this change was
+introduced.  If there is a deprecation window,
+the version # to which we guarantee to maintain the old API.
+{% endcomment %}
```
Hixie, line 79: update before landing

src/docs/release/breaking-changes/add-showAutocorrectionPromptRect.md
sfshaza2: Yes, the description should start with "A", rather than "a". Also, we don't codefont a title. These aren't huge issues. The issue blocking me from merging is the release info.

Thanks, @LongCatIsLooong!

src/docs/release/breaking-changes/add-showAutocorrectionPromptRect.md
sfshaza2: Yes, please do use dartfmt. Also, if the code, by itself, can successfully run through the analyzer, then don't skip. Most snippets aren't complete enough to run through the analyzer without errors, which is what skip is for.

src/docs/release/breaking-changes/add-showAutocorrectionPromptRect.md
LongCatIsLooong: The code doesn't compile: `class CustomTextInputClient extends State<...> implements TextInputClient`

LongCatIsLooong:
@sfshaza2 it seems the breaking change is tagged now: https://github.com/flutter/flutter/commit/4af5fe2c370df36e5640c79d4080981879d0de69. Is there anything else left before we can land this?

sfshaza2:
@LongCatIsLooong, from that commit, it appears that this change is merged in 1.18.0-dev.5.0. Can you add that info to this PR? Also, can you make sure that the code is formatted according to dartfmt? Otherwise, it looks good to me.

PR #3933, opened 2020-04-09
Update flutter-survey-metadata.json for Q2

jayoung-lee:
Q2 survey schedule is updated. 

cc @pq @DanTup


jayoung-lee:
Thank you!

PR #3959, opened 2020-04-14
add migration guide for render editable

chunhtai:


src/docs/release/breaking-changes/rendereditable-layout-before-hit-test.md
```diff
@@ -0,0 +1,84 @@
...
+title: The RenderEditable needs to be laid out before hit testing
+description: The hit testing of RenderEditable requires additional information that is only available after the layout.
+---
+
+## Summary
+
+The `RenderEditable` needs to be laid out in order to process hit
+testing; Otherwise, the framework throws an assertion error indicates
+the `RenderEditable` has not been laid out.
+`Failed assertion: line xxxx pos xx: '!debugNeedsLayout': is not true.`
```
Hixie, line 11:
```suggestion
Instances of `RenderEditable` must be laid out before processing hit
testing; trying to hit-test a `RenderEditable` object before layout will 
result in an assertion such as the following:

`Failed assertion: line 123 pos 45: '!debugNeedsLayout': is not true.`
```

src/docs/release/breaking-changes/rendereditable-layout-before-hit-test.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Context
+
+To support gesture recognizers in selectable text, the
+`RenderEditable` requires the layout information about the
+text spans to determine which text span receive the
+pointer event. Therefore, we made the layout to be a hard
+requirement for performing the hit testing to the
+`RenderEditable`.
```
Hixie, line 20:
```suggestion
To support gesture recognizers in selectable text, the
`RenderEditable` requires the layout information for its
text spans to determine which text span receive the
pointer event. (Before this change, `RenderEditable` objects 
did not take their text into account when evaluating hit tests.)
To implement this, layout was made a prerequisite for performing
hit testing on a `RenderEditable` object.
```

src/docs/release/breaking-changes/rendereditable-layout-before-hit-test.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Description of change
+
+The `RenderEditable` needs to be laid out before hit testing.
```
Hixie, line 26: Maybe just remove this section, it doesn't add anything and this change is simple enough that the context is sufficient.

src/docs/release/breaking-changes/rendereditable-layout-before-hit-test.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Migration guide
+
+If you are seeing `'!debugNeedsLayout': is not true.` assertion error while hit testing the
+`RenderEditable`, make sure to layout the `RenderEditable` before doing so.
```
Hixie, line 31:
```suggestion
`RenderEditable`, lay out the `RenderEditable` before doing so.
```

src/docs/release/breaking-changes/rendereditable-layout-before-hit-test.md
```diff
@@ -0,0 +1,84 @@
...
+  editable.attach(owner);
+  // This should throw an assertion error because the render editable has not been laid out.
+  editable.handleEvent(const PointerDownEvent(), BoxHitTestEntry(editable, const Offset(10,10)));
+  editable.detach();
+});
+```
+
+Code after migration:
+
+<!-- skip -->
```
Hixie, line 51: see if you can remove the skip somehow. it'd be good to syntax check the after code

src/docs/release/breaking-changes/rendereditable-layout-before-hit-test.md
```diff
@@ -0,0 +1,84 @@
...
+```dart
+test('attach and detach correctly handle gesture', () {
+  final RenderEditable editable = RenderEditable(
+    //...
+  );
+  final PipelineOwner owner = PipelineOwner(onNeedVisualUpdate: () { });
+  editable.attach(owner);
+  // This should throw an assertion error because the render editable has not been laid out.
```
Hixie, line 43:
```suggestion
  // This will throw an assertion error because the render editable has not been laid out.
```

src/docs/release/breaking-changes/rendereditable-layout-before-hit-test.md
```diff
@@ -0,0 +1,84 @@
...
+
+Code after migration:
+
+<!-- skip -->
+```dart
+test('attach and detach correctly handle gesture', () {
+  final RenderEditable editable = RenderEditable(
+    //...
+  );
+  // Layouts the render editable first.
```
Hixie, line 57:
```suggestion
  // Lay out the RenderEditable first.
```

src/docs/release/breaking-changes/rendereditable-layout-before-hit-test.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Context
+
+To support gesture recognizers in selectable text, the
+`RenderEditable` requires the layout information about the
+text spans to determine which text span receive the
+pointer event. Therefore, we made the layout to be a hard
+requirement for performing the hit testing to the
+`RenderEditable`.
+
```
Hixie, line 23: also add something like "In practice, this is rarely an issue. The widget library ensures that layout is performed before any hit test on all render objects. This problem is only likely to be seen in code that directly interacts with render objects, for example in tests of custom render objects."

src/docs/release/breaking-changes/rendereditable-layout-before-hit-test.md
sfshaza2: @Hixie, it occurs to me that removing the skip tag on new API will fail, because Travis checks the code against the stable release, or 1.12. Also, the latest commit DID cause Travis to fail, presumably from the new code.

sfshaza2:
@chunhtai, Travis is failing, as of your last commit, when it checked the code snippet. Travis tests the code against the stable release (1.12), so if your new code relies on 1.18, that won't work. You'll need to put the skip tag back. I can't merge it until Travis is happy.

chunhtai:
@sfshaza2 I fixed travis.

Issue #3795, opened 2020-03-11
'Build and release an Android app' page issue


r-silver1:
Page URL: https://flutter.dev/docs/deployment/android.html
Page source: https://github.com/flutter/website/tree/master/src/docs/deployment/android.md

Description of issue:

Hi,

Referencing the "signing the app" part of the tutorial, it seems that the keygen is not correct. It would not let me run commands in the way that you described. I think that if you are using the same system I am (flutter on windows 10, git bash installed) it doesn't work to put quotes around space separated folder names like "Program Files" as you all suggested. This didn't work in flutter terminal or bash for me. On bash the command you gave worked fine if i space separated the folders like /Program\ Files/

Issue #2066, opened 2018-12-17
Documentation for keytool


zoechi:
@wmleler commented on [Thu May 18 2017](https://github.com/flutter/flutter/issues/10159)

On https://flutter.io/android-release/ we should add some documentation that explains what keytool does and especially say where it leaves the key.jks file. It wasn't in the current directory, nor in the app directory. I had to search and finally found it in my home directory (ick!).

---

@jakobr-google commented on [Thu May 18 2017](https://github.com/flutter/flutter/issues/10159#issuecomment-302375805)

Keytool leaves the keystore in your home directory because you told it to (the `-keystore ~/key.jks` part of the command line), and we point you to the directory in the next section, 'Reference the keystore from the app', but of course we can improve the text.

You don't want to keep the keystore with the app, since you'll want to use the same key for other apps as well. There's a ton of information on this in the Studio documentation that we could link to, like https://developer.android.com/studio/publish/app-signing.html.

---

@mit-mit commented on [Thu May 18 2017](https://github.com/flutter/flutter/issues/10159#issuecomment-302423640)

@wmleler if you have concrete suggestions for edits, you are very much invited to send a PR! Page source is here: https://github.com/flutter/website/blob/master/android-release-build.md

---

@drexel-ue commented on [Tue Dec 04 2018](https://github.com/flutter/flutter/issues/10159#issuecomment-392914778)

I am getting this error:
```
Generating 2,048 bit RSA key pair and self-signed certificate (SHA256withRSA) with a validity of 10,000 days
        for: CN=Ikkesh Pack, OU=Development, O=PackApps, L=Columbus, ST=OH, C=US
[Storing ~/key.jks]
keytool error: java.io.FileNotFoundException: ~\key.jks (The system cannot find the path specified)
java.io.FileNotFoundException: ~\key.jks (The system cannot find the path specified)
        at java.base/java.io.FileOutputStream.open0(Native Method)
        at java.base/java.io.FileOutputStream.open(FileOutputStream.java:299)
        at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:238)
        at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:127)
        at java.base/sun.security.tools.keytool.Main.doCommands(Main.java:1308)
        at java.base/sun.security.tools.keytool.Main.run(Main.java:397)
        at java.base/sun.security.tools.keytool.Main.main(Main.java:390)
```

could you advise?

---

[...]

PR #3962, opened 2020-04-15
Fixing a couple of keytool-related doc bugs.

sfshaza2:
Fixes https://github.com/flutter/website/issues/3795
Fixes https://github.com/flutter/website/issues/2066

### DIFF
--- a/src/docs/deployment/android.md
+++ b/src/docs/deployment/android.md
@@ -78,7 +78,7 @@ This command stores the `key.jks` file in your home
 directory. If you want to store it elsewhere, change
 the argument you pass to the `-keystore` parameter.
 **However, keep the `keystore` file private;
-do not check it into public source control!**
+don't check it into public source control!**
 
 {{site.alert.note}}
 * The `keytool` command might not be in your path&mdash;it's
@@ -88,9 +88,10 @@ do not check it into public source control!**
   'Java binary at:'. Then use that fully qualified path
   replacing `java` (at the end) with `keytool`.
   If your path includes space-separated names,
-  such as `Program Files`, use a delimeter appropriate
-  to your platform. For example, use `/"Program Files"/`
-  on Mac/Linux, or `/Program\ Files/` on Windows.
+  such as `Program Files`, use platform-appropriate
+  notation for the names. For example, on Mac/Linux
+  use `/"Program Files"/`, and on Windows use
+  `/Program\ Files/`.
 
 * The `-storetype JKS` tag is only required for Java 9
   or newer. As of the Java 9 release,
@@ -111,7 +112,7 @@ storeFile=<location of the key store file, such as /Users/<user name>/key.jks>
 
 {{site.alert.warning}}
   Keep the `key.properties` file private;
-  do not check it into public source control.
+  don't check it into public source control.
 {{site.alert.end}}
 
 ### Configure signing in gradle
--- a/src/docs/deployment/android.md
+++ b/src/docs/deployment/android.md
@@ -18,7 +18,7 @@ This page covers the following topics:
 
 * [Adding a launcher icon](#adding-a-launcher-icon)
 * [Signing the app](#signing-the-app)
-* [R8](#r8)
+* [Shrinking your code with R8](#shrinking-your-code-with-r8)
 * [Reviewing the app manifest](#reviewing-the-app-manifest)
 * [Reviewing the build configuration](#reviewing-the-build-configuration)
 * [Building the app for release](#building-the-app-for-release)
@@ -193,7 +193,7 @@ Release builds of your app will now be signed automatically.
 For more information on signing your app, see
 [Sign your app][] on developer.android.com.
 
-## R8
+## Shrinking your code with R8
 
 [R8][] is the new code shrinker from Google, and it's enabled by default
 when you build a release APK or AAB. To disable R8, pass the `--no-shrink`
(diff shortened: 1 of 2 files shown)


src/docs/deployment/android.md
```diff
@@ -74,10 +74,11 @@ On Windows, use the following command:
...
 
-{{site.alert.warning}}
-  Keep the `keystore` file private;
-  do not check it into public source control.
-{{site.alert.end}}
+This command stores the `key.jks` file in your home
+directory. If you want to store it elsewhere, change
+the argument you pass to the `-keystore` parameter.
+**However, keep the `keystore` file private;
+do not check it into public source control!**
```
kwalrath, line 81: do not -> don't

(per Google style)

src/docs/deployment/android.md
```diff
@@ -87,8 +88,9 @@ keytool -genkey -v -keystore c:/Users/USER_NAME/key.jks -storetype JKS -keyalg R
   'Java binary at:'. Then use that fully qualified path
   replacing `java` (at the end) with `keytool`.
   If your path includes space-separated names,
-  such as `Program Files`, place quotes around the
-  space-separated names. For example: `/"Program Files"/`
+  such as `Program Files`, use a delimeter appropriate
```
kwalrath, line 91: delimeter -> delimiter... but I'm not sure _delimiter_ is the word we want here. Looking up the definition of delimiter, it's a character that separates parts of a string. Maybe:

...use platform-appropriate notation for the names. For example, on Mac/Linux use ..., and on Windows use ....

src/docs/deployment/android.md
```diff
@@ -88,9 +88,10 @@ do not check it into public source control!**
   'Java binary at:'. Then use that fully qualified path
   replacing `java` (at the end) with `keytool`.
   If your path includes space-separated names,
-  such as `Program Files`, use a delimeter appropriate
-  to your platform. For example, use `/"Program Files"/`
-  on Mac/Linux, or `/Program\ Files/` on Windows.
+  such as `Program Files`, use platform-appropriate
+  notation for the names. For example, on Mac/Linux
+  use `/"Program Files"/`, and on Windows use
```
kwalrath, line 93: Text looks good, but I just noticed that this doesn't render correctly (at least in https://flutter.dev/docs/deployment/android). Should it be these instead?

`"Program Files"`
`Program\ Files`

(And actually, I think the first one might work for both. Maybe we should get a DPE to advise?)

sfshaza2:
I saw your note. I will play with it.  :)

src/docs/deployment/android.md
sfshaza2: @kwalrath, I've staged the site and it looks fine to me. Please confirm:

https://sz-flutter-2.firebaseapp.com/docs/deployment/android

sfshaza2:
@kwalrath, I think the formatting is ok.  I've included a staging link this time. Please confirm.

src/docs/deployment/android.md
kwalrath: The formatting does look better than before (not cut off any more), but are the `\` characters really supposed to be in the strings? And why would there be a `\` _after_ the `"`? This is just a command-line string, not a regex.

image

sfshaza2:
@johnpryan, can you weigh in? The staged version seems good to me.

src/docs/deployment/android.md
johnpryan: I think shells on macOS and Linux support escaping using the space character (`cd Program\ Files/`) and the Windows command prompt needs the path is wrapped quotation marks (`cd "C:\Program Files"`)

src/docs/deployment/android.md
```diff
@@ -438,21 +444,20 @@ The resulting app bundle or APK files are located in
...
 [appid]: {{site.android-dev}}/studio/build/application-id
 [applicationtag]: {{site.android-dev}}/guide/topics/manifest/application-element
 [arm64-v8a]: {{site.android-dev}}/ndk/guides/abis#arm64-v8a
 [armeabi-v7a]: {{site.android-dev}}/ndk/guides/abis#v7a
-[crash-issue]: https://issuetracker.google.com/issues/147096055
-[x86-64]: {{site.android-dev}}/ndk/guides/abis#86-64
 [bundle]: {{site.android-dev}}/platform/technology/app-bundle
 [bundle2]: {{site.android-dev}}/guide/app-bundle
 [configuration qualifiers]: {{site.android-dev}}/guide/topics/resources/providing-resources#AlternativeResources
+[crash-issue]: https://issuetracker.google.com/issues/147096055
```
johnpryan, line 457: Is this link publicly accessible?

src/docs/deployment/android.md
kwalrath: The `\` characters shouldn't be in the examples, right?

kwalrath: Also, iirc quotation marks work for Mac/Linux as well as Windows. (Though we might or might not want to say that.)

src/docs/deployment/android.md
kwalrath: You have to sign in, but it accepted my gmail account... so probably?

johnpryan: 👍

src/docs/deployment/android.md
johnpryan: Yeah I would change this to "on Mac/Linux use `Program\ Files` and on Windows use `"Program Files"`

PR #3976, opened 2020-04-20
Changing punctuation in banner

sfshaza2:


sfshaza2:
Thanks for you LGTM, @maguinis!

PR #4006, opened 2020-04-27
Added a tip with iOS camera permission

edvinasgestautas:
it's not always obvious that you need it in order to have this example working.


edvinasgestautas:
@googlebot I signed it!

PR #4010, opened 2020-04-28
Add a link for Flutter Engine CPU Profiling

iskakaushik:
This is a design document that talks about changes to Flutter engine CPU profiling.


iskakaushik:
cc: @liyuqian

PR #4017, opened 2020-04-30
Add subscription form

johnpryan:
staged: 

https://flutter-website-jr-staging.firebaseapp.com/subscribe
https://flutter-website-jr-staging.firebaseapp.com/subscribe/preferences.html


johnpryan:
That's a good point - I'll get an answer on that.

sfshaza2:
Actually, I do have another suggestion. You can no longer submit until you check the box. Can you disable the button until they check the box?

johnpryan:
Good suggestion - I added that, and some more context on the preferences page.

PR #4019, opened 2020-04-30
Add tip to remove disable_input_output_paths from Podfile

jmagman:
This migration means users can remove workaround introduced in https://github.com/flutter/flutter/issues/20685#issuecomment-542322890, which will improve compilation time.


jmagman:
@sfshaza2 Should I mention https://github.com/flutter/flutter/issues/20685 anywhere?  That's the issue where that workaround was introduced, but now we don't need that workaround with these documented migration steps.  But I don't want to confuse anyone.

PR #4020, opened 2020-04-30
Explicit animations codelab

legalcodes:


sfshaza2:
Hey, Jon, can you maybe put those code snippets back and submit again? The integration.md error is fixed. (But I did have to update diffutils, and it sounds like you do, too.)

sfshaza2:
@legalcodes, I've examined the Travis log file, and the breakages are specific to the code in the codelab. Does this code expect a specific release? Can you take a look at this?

PR #4025, opened 2020-05-02
add note for china in maven instructions

xster:
Part of https://github.com/flutter/flutter/issues/55134

### DIFF
--- a/src/docs/development/add-to-app/android/project-setup.md
+++ b/src/docs/development/add-to-app/android/project-setup.md
@@ -214,6 +214,13 @@ dependencies {
 }
 ```
 
+{{site.alert.important}}
+If you're located in China, use a mirror site such as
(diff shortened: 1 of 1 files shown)


src/docs/development/add-to-app/android/project-setup.md
```diff
@@ -214,6 +214,13 @@ dependencies {
 }
 ```
 
+{{site.alert.important}}
```
sfshaza2, line 217: This is a minor nit, and don't fix it unless you have to make other changes, but notes should be indented on each line by 2 spaces. We also no longer use that format for URLs, but I can fix that on a future pass. 

Otherwise, LGTM, but will wait for @ blasten to review.

sfshaza2:
Thanks for the LGTM, @blasten!

PR #4028, opened 2020-05-02
Explicit commands for "flutter create"

chibuezeukachi:
This is clearer for beginners and ensures that the directory structure is created correctly


chibuezeukachi:
@googlebot I signed it!

PR #4037, opened 2020-05-04
Adding info on when not to use Flutter for web

sfshaza2:


src/docs/development/platform-integration/web.md
```diff
@@ -9,32 +9,36 @@ As of the 1.12 release, web support for Flutter
 is available on the beta channel.
 **We don't recommend deploying a web app to production.**
 
+### What sort of apps are not ideal for Flutter on the web?
```
mariamhas, line 12: It may be better to have the question be more positive : What scenarios are ideal for Flutter on the web?

src/web.md
```diff
@@ -83,6 +83,15 @@ The following resources can help you get started:
...
 
+---
+
+<sup>*</sup> We don’t recommend using Flutter to build static
+  content websites with a lot of text and images such as blogs,
+  event pages, and news sites. For these types of sites,
+  it's best to use traditional web approaches,
+  but you _can_ use Flutter to embed interactive experiences
+  into these websites.
+
```
timsneath, line 94: This feels duplicative of the text below, which falls just after the list of scenarios?

> Not every HTML scenario is ideally suited for Flutter at this time. For example, text-rich flow-based content such as blog articles benefit from the document-centric model that the web is built around, rather than the app-centric services that a UI framework like Flutter can deliver.

src/docs/development/platform-integration/web.md
```diff
@@ -9,32 +9,36 @@ As of the 1.12 release, web support for Flutter
...
+
```
mariamhas, line 13: To answer the question of what scenarios are ideal, we can either summarize what we have listed on flutter.dev/web here or just link them to the page and make the scenarios more explicit on that page.

"While traditional web approaches remain a great choice, we envision the web support available in Flutter being valuable in many scenarios. For example:

Rich interactive single page application
Connected Progressive Web Application 
Embedding dynamic/interactive content in an existing website 


^^ these need to be updated in flutter.dev/web... ill submit a PR soon for updates there.

src/docs/development/platform-integration/web.md
timsneath, line 12: Can we use the language below for consistency with the /web page? 

> Not every HTML scenario is ideally suited for Flutter at this time. For example, text-rich flow-based content such as blog articles benefit from the document-centric model that the web is built around, rather than the app-centric services that a UI framework like Flutter can deliver.

src/docs/development/platform-integration/web.md
```diff
@@ -9,32 +9,36 @@ As of the 1.12 release, web support for Flutter
...
+While **we don’t recommend using Flutter to build static content
```
mariamhas, line 14: This could go after we list the scenarios that it is good for within the same question

src/docs/development/platform-integration/web.md
```diff
@@ -9,32 +9,36 @@ As of the 1.12 release, web support for Flutter
 is available on the beta channel.
 **We don't recommend deploying a web app to production.**
```
timsneath, line 10: Can we update this question while we're tackling this page? We've made a lot of progress since 1.12, and we have real customers in production (including our own product), so we should start to turn the dial a bit. 

I'd say: 
> Flutter web support is now available in the `beta` channel. As is typical for beta software, we're still making changes to improve performance and quality, and you may discover rendering bugs compared to the stable mobile platforms. While the web support is in beta, we recommend careful testing on a variety of platforms before deploying to production, and that you track and apply Flutter updates to your deployed site on a regular basis.

src/docs/development/platform-integration/web.md
mariamhas: +1 this as an answer sounds good to me

sfshaza2:
@mariamhas @timsneath 

I have incorporated your feedback. Tim, that paragraph *was* duplicative, but that's because I managed to MISS it on my previous read-through! Anyway, I mostly kept the original text, but added a little tweak.

sfshaza2:
@mariamhas, oops, I had missed that feedback before. I've now incorporated it.

Thanks for your review!

PR #4038, opened 2020-05-05
Refresh code excerpt

legalcodes:


johnpryan:
I think it's OK to merge this, something is wrong with the check-code.sh script.

sfshaza2:
I still don't get why SOME PRs were fine, though, and others not so much...

PR #4040, opened 2020-05-05
temporarily disable check-code.sh

johnpryan:


sfshaza2:
This info could come in VERY handy for me in future! I'll make sure to save this info in the breakages doc.  :D. Thanks, @johnpryan!

PR #4055, opened 2020-05-06
Note Flutter version where Runner display name issue is fixed

jmagman:
Note the version in which https://github.com/flutter/flutter/issues/41793 is fixed.

### DIFF
--- a/src/docs/deployment/ios.md
+++ b/src/docs/deployment/ios.md
@@ -89,10 +89,10 @@ In the **Identity** section:
 : Runner
 
 {{site.note.alert}}
-  Changing the display name to anything other
-  than `Runner` causes the app not to run with
-  the Flutter tools. For more information,
-  see [Issue 41793][].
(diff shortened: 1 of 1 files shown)


jmagman:
@sfshaza2 Not sure why that isn't showing up as an alert:
Screen Shot 2020-05-06 at 2 37 27 PM

src/docs/deployment/ios.md
```diff
@@ -89,10 +89,10 @@ In the **Identity** section:
 : Runner
 
 {{site.note.alert}}
```
sfshaza2, line 91: The macro is:

```
{{site.alert.note}}
   xxxx
{{site.alert.end}}
```

jmagman: Well that would do it.
https://github.com/flutter/website/pull/3838/files#diff-e8a9c09b65f16491acbbde77bcbf4883R94

Issue #4061, opened 2020-05-08
'Flutter SDK releases' page issue


tibmartSynergy:
Page URL: https://flutter.dev/docs/development/tools/sdk/releases.html
Page source: https://github.com/flutter/website/tree/master/src/docs/development/tools/sdk/releases.md

Description of issue:

The link for the new MacOS version has a typo in the URL. (1.17* instead of v.17*)

Expected:
https://storage.googleapis.com/flutter_infra/releases/stable/macos/flutter_macos_v.17.0-stable.zip

Existing:
https://storage.googleapis.com/flutter_infra/releases/stable/macos/flutter_macos_1.17.0-stable.zip

sfshaza2:
Hi, @tibmartSynergy! I was able to successfully download the macOS stable 1.17 release just now. Is this still happening for you?

pcsosinski:
we removed the "v" from our versioning scheme starting with this release.

sfshaza2:
Thanks, @pcsosinski, for weighing in!  @tibmartSynergy, we announced this change in our Medium publication. See the [Spring 2020 Updates](https://medium.com/flutter/flutter-spring-2020-update-f723d898d7af) post.

Closing.

PR #4064, opened 2020-05-08
Adding a note about the updated release process

sfshaza2:
Inspired by https://github.com/flutter/website/issues/4061

### DIFF
--- a/src/docs/development/tools/sdk/releases.md
+++ b/src/docs/development/tools/sdk/releases.md
@@ -16,9 +16,9 @@ The {{site.sdk.channel | capitalize }} channel contains the
 most stable Flutter builds. See [Flutter’s channels][] for details.
 
 {{site.alert.note}}
-  As of release 1.17 (stable), we have updated our release process.
-  This includes how we name the releases and their corresponding files.
-  For more information, see the "Updating our release process" section
+  As of release 1.17 (stable), we've updated our release process,
+  including changes to the names of releases and their download files.
+  For details, see the "Updating our release process" section
   in the [Flutter Spring 2020 Update][] blog post.
 {{site.alert.end}}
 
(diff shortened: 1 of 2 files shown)


kwalrath:
I think I need to add the same note to dart.dev...

src/docs/development/tools/sdk/releases.md
```diff
@@ -15,6 +15,13 @@ toc: false
 The {{site.sdk.channel | capitalize }} channel contains the
 most stable Flutter builds. See [Flutter’s channels][] for details.
 
+{{site.alert.note}}
+  As of release 1.17 (stable), we have updated our release process.
```
kwalrath, line 19:
```suggestion
  As of release 1.17 (stable), we've updated our release process,
```

src/docs/development/tools/sdk/releases.md
```diff
@@ -15,6 +15,13 @@ toc: false
...
+  This includes how we name the releases and their corresponding files.
```
kwalrath, line 20:
```suggestion
  including changes to the names of releases and their download files.
```

src/docs/development/tools/sdk/releases.md
```diff
@@ -15,6 +15,13 @@ toc: false
...
+  For more information, see the "Updating our release process" section
```
kwalrath, line 21:
```suggestion
  For details, see the "Updating our release process" section
```

src/docs/development/tools/sdk/releases.md
pcsosinski: "how we name" makes more sense in this context to me.  if we want to be more specific we can say "we've changed our versioning schema," but I think "how we name" makes sense.

kwalrath: Maybe:

...changes to how we name releases and their download files.

kwalrath: I was trying to avoid too much "we" and unadorned "This". Although given the name of the link I didn't avoid "we" entirely.

Issue #4039, opened 2020-05-05
Update FAQ on x86


mit-mit:
From https://flutter.dev/docs/resources/faq#what-devices-and-os-versions-does-flutter-run-on:

> Note Flutter currently does not support building for x86 Android (issue #9253) directly, however apps built for ARMv7 or ARM64 run fine (via ARM emulation) on many x86 Android devices.

However, https://github.com/flutter/flutter/issues/9253 is resolved, so this likely needs an update

sfshaza2:
Fixed.

kw2019ltd:
still not supported
flutter build apk --target-platform android-x86 is failing

FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':app:compileReleaseJavaWithJavac'.
> Could not resolve all task dependencies for configuration ':app:releaseCompileClasspath'.
   > Could not find io.flutter:x86_release:1.0.0-540786dd51f112885a89792d678296b95e6622e5.
     Required by:
         project :app

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1m 5s
Running Gradle task 'assembleRelease'...
Running Gradle task 'assembleRelease'... Done                      66.6s
Gradle task assembleRelease failed with exit code 1

PR #4065, opened 2020-05-08
Updating the faq

sfshaza2:
Staged: https://sz-flutter.firebaseapp.com/docs/resources/faq

I started this PR to specifically address https://github.com/flutter/website/issues/4039.

However, as I scanned through it, it is in serious need of updating. I updated some of it, but did not tackle everything.

@jonahwilliams, please look at the "What devices does Flutter run on?" section, which is the reason that the doc bug was filed.

@timsneath, as I said, I've tackled *some* of the rest of this doc, but maybe a PM could do a more thorough pass on it? In a new PR, perhaps?  :)

### DIFF
--- a/src/docs/resources/faq.md
+++ b/src/docs/resources/faq.md
@@ -18,8 +18,8 @@ and is free and open source.
 For users, Flutter makes beautiful app UIs come to life.
 
 For developers, Flutter lowers the bar to entry
-for building mobile apps. It speeds up development of
-mobile apps and reduces the cost and complexity
+for building apps. It speeds up development of
+apps and reduces the cost and complexity
 of app production across platforms.
 
 For designers, Flutter helps deliver the original
@@ -29,20 +29,20 @@ It also acts as a productive prototyping tool.
 ### Who is Flutter for?
 
 Flutter is for _developers_ that want a faster
-way to build beautiful mobile apps, or
+way to build beautiful apps, or
 a way to reach more users with a single investment.
 
-Flutter is also for _engineering managers_ that need to lead
-mobile development teams. Flutter allows eng managers
-to create a single _mobile app dev team_, unifying their
-development investments to ship more features faster,
-ship the same feature set to iOS and Android at the same
+Flutter is also for _engineering managers_ that lead
+development teams. Flutter allows eng managers
+to create a single _mobile, web, and desktop app dev team_,
+unifying their development investments to ship more features faster,
+ship the same feature set to multiple platform at the same
 time, and lower maintenance costs.
 
 While not the initial target audience,
 Flutter is also for _designers_
 that want their original design visions delivered
-consistently, with high fidelity, to all users on mobile.
+consistently, with high fidelity, to all users.
 
 Fundamentally, Flutter is for users that want beautiful apps,
 with delightful motion and animation, and UIs with character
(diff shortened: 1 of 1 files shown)


src/docs/resources/faq.md
```diff
@@ -416,59 +415,62 @@ Hot reload works by injecting updated source code files
...
 
-Mobile hardware: iOS devices (iPhone 4S or newer) and ARM Android devices.
+* Mobile hardware: iOS devices (iPhone 4S or newer)
+  and ARM Android devices.
 
-Note Flutter currently does not support building for x86 Android
-([issue #9253][]) directly, however apps built for ARMv7 or
-ARM64 run fine (via ARM emulation) on many x86 Android devices.
+* Flutter supports building ahead-of-time (AOT) compiled libraries
+  for `x86_64`, `armeabi-v7a`, and `arm64-v8a`.
```
jonahwilliams, line 440: The note about ARM emulation for x86 devices is still important

src/docs/resources/faq.md
```diff
@@ -416,59 +415,65 @@ Hot reload works by injecting updated source code files
...
-We test on a variety of low-end to high-end phones
-but we don't yet have an official device compatibility guarantee.
+* We test on a variety of low-end to high-end phones
+  but we don't yet have an official device compatibility guarantee.
 
-We believe Flutter works well on tablets. We do not currently implement
-all of the tablet-specific adaptations recommended by Material Design,
-though we are planning further investment in this area.
+* We believe Flutter works well on tablets. We do not
+  currently implement all of the tablet-specific adaptations
```
timsneath, line 452: I'm not sure if this is still true. We should check with @willlarche whether this call-out is necessary. I suspect it dates back from a much earlier time in Flutter's Material development.

src/docs/resources/faq.md
```diff
@@ -416,59 +415,65 @@ Hot reload works by injecting updated source code files
...
 
-We test on a variety of low-end to high-end phones
-but we don't yet have an official device compatibility guarantee.
+* We test on a variety of low-end to high-end phones
+  but we don't yet have an official device compatibility guarantee.
 
-We believe Flutter works well on tablets. We do not currently implement
-all of the tablet-specific adaptations recommended by Material Design,
-though we are planning further investment in this area.
+* We believe Flutter works well on tablets. We do not
```
timsneath, line 451: We can be stronger than "We _believe_ Flutter works well on tablets". There's nothing fundamentally different about iPadOS or Android on tablets in terms of its compilation / execution, and we test on those devices just like we test on phones.

src/docs/resources/faq.md
```diff
@@ -570,19 +575,20 @@ Learn more about a11y in the
...
-are able to run in parallel (usually implemented as separate threads). Isolates
-communicate by sending and receiving asynchronous messages. Flutter does not
-currently have a shared-memory parallelism solution, although we are evaluating
-solutions for this.
+Flutter supports isolates. Isolates are separate heaps in
+Flutter's VM, and they are able to run in parallel
+(usually implemented as separate threads). Isolates
+communicate by sending and receiving asynchronous messages.
+Flutter does not currently have a shared-memory parallelism
+solution, although we are evaluating solutions for this.
```
timsneath, line 583: Let's remove this last sentence. It seems unnecessary to make a promise here.

src/docs/resources/faq.md
willlarche: I would say it's not necessary anymore. Happy to take community feedback on features they'd like tho!

PR #4070, opened 2020-05-09
JavaScript sample code fixed.

yuta-ike:
`const data = json.origin` is not asynchronous process, so that `await` keywords is not needed.


yuta-ike:
@googlebot I signed it!

PR #4071, opened 2020-05-09
Back slash for windows keytool command

ajmaln:
Change path in keytool command for windows from forward slash to back slash.


ajmaln:
@googlebot I signed it!

PR #4075, opened 2020-05-11
Make the 'update' docs current

mit-mit:
* Remove setup steps (they are already documented as part of Getting Started, and in our templates)
* Remove text that suggests that `flutter upgrade` updates package dependencies (did we ever do that?)
* Remove `pub get` content; I don't think that belongs in an "upgrading" topic
* Add `pub outdated` mentioning
* Don't encourage package pinning


mit-mit:
Also, @jonahwilliams, we do we still want to mention `flutter version`? I thought I saw a comment somewhere that this command has some known issues, and might get deprecated?

PR #4080, opened 2020-05-11
Document creating an Xcode App Size Report

jmagman:
https://flutter.dev/docs/perf/app-size#ios says:

> To get a rough idea of how large your release IPA is, run the following:
> 
> ```
> flutter build ios && tar -zcf build/app.ipa build/ios/iphoneos/Runner.app && ls -lh build/app.ipa
> ```
> The resulting IPA file for the example/helloworld app (as of this writing) is 8.3 MB.
> 
> ```
> -rw-r--r--  1 userName  primarygroup   8.3M Oct 25 13:47 build/app.ipa
> ```

On my machine I get 179M.
```
-rw-r--r--  1 userName  primarygroup   179M May 11 15:05 build/app.ipa
```
[Bitcode and Swift libraries](https://github.com/flutter/flutter/issues/38016#issuecomment-584498505) really blow up this number, and don't reflect final IPA sizes.

Remove suggestion to approximate size with a tarball, and instead document how to produce a Xcode App Size Report.

Fixes https://github.com/flutter/flutter/issues/38016.

### DIFF
--- a/src/docs/perf/app-size.md
+++ b/src/docs/perf/app-size.md
@@ -97,7 +97,7 @@ Then:
 `Strip Swift symbols`
 : Selected
 
-Sign and export the IPA. The exported directory will contain
+Sign and export the IPA. The exported directory contains
 `App Thinning Size Report.txt` with details about your projected
 application size on different devices and versions of iOS.
 
@@ -111,7 +111,7 @@ App size: 5.3 MB compressed, 12.9 MB uncompressed
 On Demand Resources size: Zero KB compressed, Zero KB uncompressed
 ```
 
-In this example, the app will have an approximate
+In this example, the app has an approximate
 download size of 5.3 MB and an approximate
 install size of 12.9 MB on an iPhone12,1 ([Model ID / Hardware
 number][] for iPhone 11)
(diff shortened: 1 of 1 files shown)


jmagman:
\cc @xster

jmagman:
> A great improvement, @jmagman! LGTM
> 
> (There was one "will" that I'd remove, but it's not worth holding up the merge.)

Done. Don't know why "will" is so tempting to add.

src/docs/perf/app-size.md
```diff
@@ -74,30 +74,48 @@ Built build/app/outputs/apk/release/app-arm64-v8a-release.apk (4.6MB).
...
+
+`Strip Swift symbols`
+: Selected
+
+Sign and export the IPA. The exported directory contains
+`App Thinning Size Report.txt` with details about your projected
+application size on different devices and versions of iOS.
+
+The App Size Report for the default demo app
+(as of this writing) shows:
```
xster, line 105: nit, maybe just say as of May 2020

jmagman:
> You had some trick for doing this from the command line right? Is that worth including here?

It's documented at https://developer.apple.com/documentation/xcode/reducing_your_app_s_size#3458578.  I started thinking about it generically in https://github.com/flutter/flutter/issues/40471 but it will be hard to document the right `ExportOptions.plist` for everyone.

src/docs/perf/app-size.md
jmagman: I'll check out stable, regenerate, and say "The App Size Report for the default demo app in Flutter 1.17 shows:"

PR #4085, opened 2020-05-12
Fix starter DartPad code to match instructions

carlospero:
The initial DartPad has outdated names using "Login" instead of "SignUp" which don't match the instructions. It was also missing a step to add the /welcome route at the proper time.


carlospero:
@googlebot I signed it!

PR #4093, opened 2020-05-14
Performing some tweaks to the codelab

sfshaza2:


src/docs/codelabs/explicit-animations.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Explicit animations codelab"
+title: "Explicit animations"
```
kwalrath, line 2: No need for quotation marks, I think. (Not that you added them. More an FYI.)

sfshaza2: You are quite right! And yes, 'twasn't me. I'll fix in a subsequent PR.

PR #4099, opened 2020-05-16
Avoid error caused by flutter upgrade.

basashi-G:
Error occurs like this when I run example code. 
```
FlutterError (ServicesBinding.defaultBinaryMessenger was accessed before the binding was initialized.
If you're running an application and need to access the binary messenger before `runApp()` has been called (for example, during plugin initialization), then you need to explicitly call the `WidgetsFlutterBinding.ensureInitialized()` first.
If you're running a test, you can call the `TestWidgetsFlutterBinding.ensureInitialized()` as the first line in your test's `main()` method to initialize the binding.)
```
I dealt with this problem by reference to 
https://stackoverflow.com/questions/57689492/flutter-unhandled-exception-servicesbinding-defaultbinarymessenger-was-accesse

### DIFF
--- a/src/docs/cookbook/persistence/sqlite.md
+++ b/src/docs/cookbook/persistence/sqlite.md
@@ -312,10 +312,15 @@ To run the example:
 ```dart
 import 'dart:async';
 
+import 'package:flutter/widgets.dart';
+
 import 'package:path/path.dart';
 import 'package:sqflite/sqflite.dart';
 
 void main() async {
+  // Avoid errors coused by flutter upgrade. 
+  // Importing 'package:flutter/widgets.dart' is required.
+  WidgetsFlutterBinding.ensureInitialized();
   final database = openDatabase(
     // Set the path to the database. Note: Using the `join` function from the
     // `path` package is best practice to ensure the path is correctly


basashi-G:
@googlebot I signed it!

sfshaza2:
@johnpryan, here's an interesting PR. The build is breaking on the example code in [flutter.dev/docs/cookbook/persistence/sqlite](https://flutter.dev/docs/cookbook/persistence/sqlite). All @basashi-G did was to add a call, as the first line in `main()`, to `WidgetsFlutterBinding.ensureInitialized();`, and to import the corresponding package. @basashi-G links to a stackoverflow issue, which I read through, as well as some related bugs on flutter/flutter that are all closed. This problem occurred to numerous people when upgrading their release of Flutter. Solutions included wiping the emulator, restarting the app, even rebooting.

I don't get why the build fails on this. I restarted the build, and it failed again. [Here's the logfile](https://travis-ci.org/github/flutter/website/jobs/687679753). If you look at the diff, it only lists a comment. Thanks for looking into this!

sfshaza2:
@basashi-G, can you update this PR? It's spelled "caused", not "coused."

thx

src/docs/cookbook/persistence/sqlite.md
```diff
@@ -312,10 +312,15 @@ To run the example:
 ```dart
 import 'dart:async';
 
+import 'package:flutter/widgets.dart';
+
 import 'package:path/path.dart';
 import 'package:sqflite/sqflite.dart';
 
 void main() async {
+  // Avoid errors coused by flutter upgrade.
```
sfshaza2, line 321: @basashi-G, can you correct this line?

  // Avoid errors coused by flutter upgrade.  =>   // Avoid errors caused by flutter upgrade.

johnpryan:
I ran `dartfmt` on this snippet to see if that would fix the build: https://github.com/flutter/website/pull/4103

johnpryan:
@sfshaza2 it looks like https://github.com/flutter/website/pull/4103 fixes it

sfshaza2:
WOOT, thanks John!

sfshaza2:
@johnpryan, I merged the other one too fast. Do you mind fixing the conflict?

PR #4105, opened 2020-05-20
Add information about --track-widget-creation

Hixie:


src/docs/testing/debugging.md
```diff
@@ -222,6 +222,49 @@ If you use Travis or Cirrus for testing, increase the number of
...
+To disable this behavior, pass `--no-track-widget-creation` to the `flutter run` command.
+With that flag set, the code above prints "1" in debug and release builds, and error messages
+will bemoan that they cannot provide all the information that they would otherwise be able
+to provide if widget creation tracking was enabled.
+
+See also:
+
+ * [WidgetInspectorService.isWidgetCreationTracked](https://api.flutter.dev/flutter/widgets/WidgetInspectorService/isWidgetCreationTracked.html).
+ * The `_Location` class in [widget_inspector.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/widgets/widget_inspector.dart).
+ * The kernel transform that implements this feature: XXX
```
Hixie, line 266: cc @jacob314 where is the kernel transform implemented?

src/docs/testing/debugging.md
```diff
@@ -222,6 +222,50 @@ If you use Travis or Cirrus for testing, increase the number of
...
 
+### Widgets marked const that should be equal to each other, aren't
+
+In debug mode, you may find that two `const` widgets that should to all
+appearances be equal (because of Dart's constant deduplication) are not.
+
+For example, this code should print 1:
+
+```dart
+print(<Widget>{ // this is the syntax for a Set<Widget> literal
```
Hixie, line 233: @Sfshaza2 any idea why this isn't passing the syntax checks? do i have to declare that i'm importing something or...?

src/docs/testing/debugging.md
Hixie: nevermind, jonah found it for me

src/docs/testing/debugging.md
sfshaza2: Ian, I'm not sure how code checks work on flutter/flutter, but on flutter/website, the code must be a full example that can pass analysis without any other code. This is why we use the skip tag so liberally.

src/docs/testing/debugging.md
```diff
@@ -222,6 +222,50 @@ If you use Travis or Cirrus for testing, increase the number of
...
+  const SizedBox(),
+  const SizedBox(),
+}.length);
+```
+
+...because the two constants should be considered equal (and indeed the analyzer will
```
sfshaza2, line 239: Could you make this a standalone sentence? As simple as:

"The two constants should be considered equal..."

"the analyzer will complain" => "the analyzer complaints"

src/docs/testing/debugging.md
```diff
@@ -222,6 +222,50 @@ If you use Travis or Cirrus for testing, increase the number of
...
+complain that "Two elements in a set literal shouldn't be equal"). As expected, in release
+builds, it does print 1. However, in debug builds it will print 2. This is because the
```
sfshaza2, line 241: "it will print 2" => "it prints 2"

src/docs/testing/debugging.md
```diff
@@ -222,6 +222,50 @@ If you use Travis or Cirrus for testing, increase the number of
...
+time, so the code is effectively:
+
+```dart
+print(<Widget>{
+  const SizedBox(location: Location(file: 'foo.dart', line: 12)),
+  const SizedBox(location: Location(file: 'foo.dart', line: 13)),
+}.length);
+```
+
+...and thus the instances are different. We use this to make the error messages clearer when
```
sfshaza2, line 252: "....and thus the instances are different" => "Therefore, the instances are different."

src/docs/testing/debugging.md
```diff
@@ -222,6 +222,50 @@ If you use Travis or Cirrus for testing, increase the number of
...
+a widget is involved in an exception, by reporting where the relevant widget was created.
+Unfortunately, it has the visible side-effect of making otherwise-identical constants be
+different at compile time.
+
+To disable this behavior, pass `--no-track-widget-creation` to the `flutter run` command.
+With that flag set, the code above prints "1" in debug and release builds, and error messages
+will bemoan that they cannot provide all the information that they would otherwise be able
```
sfshaza2, line 259: will bemoan => bemoan

(Do you think most folks, including non-native-English speakers, understand this word? Maybe use "complain" instead?)

src/docs/testing/debugging.md
```diff
@@ -222,6 +222,50 @@ If you use Travis or Cirrus for testing, increase the number of
...
+to provide if widget creation tracking was enabled.
+
+See also:
+
+ * Our documentation on [how the Widget Inspector uses widget creation tracking][]
```
sfshaza2, line 264: For parallelism, either add a period here, or remove it from all of the lines.

src/docs/testing/debugging.md
```diff
@@ -269,6 +313,11 @@ You might find the following docs useful:
 [debug mode]: /docs/testing/build-modes#debug
 [profile mode]: /docs/testing/build-modes#profile
 [release mode]: /docs/testing/build-modes#release
+[how the Widget Inspector uses widget creation tracking]: /docs/development/tools/devtools/inspector#track-widget-creation
 
 [`Assert`]: {{site.dart-site}}/guides/language/language-tour#assert
 [Dart language tour]: {{site.dart-site}}/guides/language/language-tour
+
+[WidgetInspectorService.isWidgetCreationTracked]: https://api.flutter.dev/flutter/widgets/WidgetInspectorService/isWidgetCreationTracked.html
```
sfshaza2, line 321: Use the macro:
```
[WidgetInspectorService.isWidgetCreationTracked]: {{site.api}}/flutter/widgts/WidgetInspectorService/isWidgetCreationTracked.html
```

src/docs/testing/debugging.md
Hixie: ah ok, will skip then. thanks.

Hixie:
I addressed the review comments. Thanks!

Hixie:
Yup, this should be ready to land.

PR #4114, opened 2020-05-20
add mgm card

chunhtai:
This is how it looks
Screen Shot 2020-05-20 at 4 33 25 PM

It looks a bit out of place because the backgound color is purple while all other cards are gray. Should we ask them to provide a pic with similar background?
Screen Shot 2020-05-20 at 4 41 32 PM

Also, do we want it to be in the feature section?


sfshaza2:
(I'll merge after @maguinis approves...)

maguinis:
LGTM

PR #4116, opened 2020-05-21
Update python server command

kasp1:
The SimpleHTTPServer module has been merged into http.server in Python 3.0.

https://docs.python.org/2/library/simplehttpserver.html


kasp1:
@googlebot I signed it!

PR #4125, opened 2020-05-25
Main git clone command retrieves all branches

wayne-li2:
I started using Flutter a few days ago and was following the Get Started documentation.

Without too much thought, I used the git clone command, then forgot about it.

Today, as I was tinkering, I realized I needed to be on the beta channel, not stable. But ```flutter channel``` was only giving me the stable channel and trying to switch caused an error.

It took me a while before I realized that it was because of this earlier command.

In my opinion, to reduce confusion for new users, the main command should be to simply clone the entire repository. Open to feedback though, so I figure I make this PR and have maintainers decide.


wayne-li2:
@googlebot I signed it!

PR #4132, opened 2020-05-26
Fix function name ref

zentby:
The function mentioned in the description is different than the one referred in the example.
```dart
// Open the database and store the reference.
final Future<Database> database = openDatabase(
  // Set the path to the database. Note: Using the `join` function from the
  // `path` package is best practice to ensure the path is correctly
  // constructed for each platform.
  join(await getDatabasesPath(), 'doggie_database.db'),
);
```


zentby:
@googlebot I signed it!

Issue #3952, opened 2020-04-13
Update code example on 'Mock dependencies using Mockito' page


ajonno:
Page URL: https://docs.flutter.dev/cookbook/testing/unit/mocking
Page source: https://github.com/flutter/website/tree/master/src/docs/cookbook/testing/unit/mocking.md

Description of issue:
this line of code no longer works (test always fails) in the test example as ```TypeMatcher``` is deprecated. please fix, thanks.

```expect(await fetchPost(client), const TypeMatcher<Post>());```

sfshaza2:
@ajonno, can you tell me which version of the Flutter sdk you are using?

ajonno:
Flutter 1.12.13+hotfix.8

sfshaza2:
Thanks, @ajonno! @filiph, as you used this very recently, can you confirm? He is using the current stable version.

samisnotinsane:
What should be passed into `expect` instead of `TypeMatcher<Post>()`? Please update as I'm using the cookbook to learn Flutter and deprecated code in official documentation is just poor.

samisnotinsane:
Found the solution using official API docs for class [isA](https://api.flutter.dev/flutter/package-matcher_matcher/isA.html).

So change the line **from**:
`expect(await fetchPost(client), const TypeMatcher<Post>());`
**to**:
`expect(await fetchPost(client), isA<Post>());`

micpap25:
@samisnotinsane Hi, it seems like you pushed these commits but never actually made a PR for them. Do you plan to do that, and if not, can I copy your commits on to my branch and make a PR? Thanks.

samisnotinsane:
@micpap25 Hey, thanks for your comment. I believe I did raise PR #4155 which solves this issue and it was subsequently merged. Let me know if it isn't checking out. ![image](https://user-images.githubusercontent.com/11885489/94531361-60c4d780-0234-11eb-9146-a6272a2b45bc.png)

micpap25:
Ah, so in that case it looks like this issue can be closed.

atsansone:
This issue no longer exists on the page. Closing this issue.

PR #4155, opened 2020-05-30
Fixing issue #3952

samisnotinsane:
`const TypeMatcher` can no longer be passed into `expect` as it is deprecated and causes unit test to fail. API docs indicate `isA` should be used instead. Therefore this PR updates the code example to reflect this change.


samisnotinsane:
@googlebot I signed it!

PR #4165, opened 2020-06-02
Code fixes of "FlutterFragment"

chenglu:
Hi team:

There's a feedback from the CN community here https://github.com/cfug/flutter.cn/issues/657, the `flutterFragment` here at `flutterFragment.withCachedEngine("my_engine_id").build(); ` should be upper case to `FlutterFragment`.

cc/ @xster, could you please valid this?

Thanks to @SevenWalnut for having this feedback.


xster:
LGTM. Thanks!

PR #4173, opened 2020-06-03
Add link for iOS pull-to-refresh improvement

edwardaux:


edwardaux:
@sfshaza2 I have resolved the conflict and saved. 

I note that you mentioned the "editing" tag was removed and I should update my URL to not contain the trailing `/edit`. If I look at the changes it looks like it was the "sharing" tag that was removed and all of the other docs still have the trailing `/edit` (Google Docs redirects you to a URL with that trailing path in any case).  I've left mine matching the existing patterns, but please let me know if I need further changes. Ta

sfshaza2:
You are right, @edwardaux! Thanks for fixing!

Issue #4168, opened 2020-06-02
Update info about formatting code in Android Studio and IntelliJ


alexander-doroshko:
Page: https://flutter.dev/docs/development/tools/formatting#automatically-formatting-code-in-android-studio-and-intellij

Source: https://github.com/flutter/website/blob/master/src/docs/development/tools/formatting.md#automatically-formatting-code-in-android-studio-and-intellij

'Reformat Code with dartfmt' action is still there but there's no more need in it. Standard 'Reformat Code' action now invokes dartfmt. Standard shortcut is `Cmd+Alt+L` on Mac or `Ctrl+Alt+L` on Windows and Linux.

Also, there's a check box `Format code on save` on the Flutter page in Preferences (Mac) / Settings (Windows, Linux).

sfshaza2:
Nice additions!

EmperorYP7:
Linked a PR to resolve this issue. I'd be glad if someone reviews it!

sfshaza2:
Fixed! Thanks, @alexander-doroshko!

alexander-doroshko:
Thanks! Sorry for being a bit late. Added some comments to the PR.

PR #4178, opened 2020-06-03
Updating formatting code section

EmperorYP7:
This is my implementation on #4168 
Updating info about formatting code in Android Studio and IntelliJ.

### DIFF
--- a/src/docs/development/tools/formatting.md
+++ b/src/docs/development/tools/formatting.md
@@ -17,11 +17,11 @@ where time might be better spent on code behavior rather than code style.
 
 Install the `Dart` plugin (see [Editor setup](/docs/get-started/editor))
 to get automatic formatting of code in Android Studio and IntelliJ.
-
-To automatically format the code in the current source code window,
-right-click in the code window and select `Reformat Code with dartfmt`.
-You can add a keyboard shortcut to this in Keymap section of IntelliJ
-Preferences.
+To automatically format your code in the current source code window,
+use `Cmd+Alt+L` (on Mac) or `Ctrl+Alt+L` (on Windows and Linux).
+Android Studio and IntelliJ also provides a check box named `Format code on save` on
+the Flutter page in Preferences (on Mac) or Settings (on Windows and Linux)
+which will format the current file automatically when you save it.
 
 ## Automatically formatting code in VS Code
 


EmperorYP7:
@googlebot I signed it!

src/docs/development/tools/formatting.md
```diff
@@ -17,11 +17,11 @@ where time might be better spent on code behavior rather than code style.
...
-
-To automatically format the code in the current source code window,
-right-click in the code window and select `Reformat Code with dartfmt`.
-You can add a keyboard shortcut to this in Keymap section of IntelliJ
-Preferences.
+To automatically format your code in the current source code window,
+use `Cmd+Alt+L` (on Mac) or `Ctrl+Alt+L` (on Windows and Linux).
+Android Studio and IntelliJ also provides a check box named `Format code on save` on
+the Flutter page in Preferences (on Mac) or Settings (on Windows and Linux)
+which will format the current file automatically when you save it.
```
alexander-doroshko, line 24: There's no Save File action in Android Studio/IntelliJ (well, strictly speaking it's there but its a kind of hidden action and has a different name). Typical action (Ctrl+S / Cmd+S) is Save All. So it would be more correct to write smth like `which will ensure that the IDE formats edited files when it saves them`. There might be better words, I'm not a native speaker.

src/docs/development/tools/formatting.md
```diff
@@ -17,11 +17,11 @@ where time might be better spent on code behavior rather than code style.
 
 Install the `Dart` plugin (see [Editor setup](/docs/get-started/editor))
```
alexander-doroshko, line 18: To be precise, we may add Flutter plugin:
```
Install the `Dart` and the `Flutter` plugins ...
```

src/docs/development/tools/formatting.md
```diff
@@ -17,11 +17,11 @@ where time might be better spent on code behavior rather than code style.
...
 to get automatic formatting of code in Android Studio and IntelliJ.
-
-To automatically format the code in the current source code window,
-right-click in the code window and select `Reformat Code with dartfmt`.
-You can add a keyboard shortcut to this in Keymap section of IntelliJ
-Preferences.
+To automatically format your code in the current source code window,
+use `Cmd+Alt+L` (on Mac) or `Ctrl+Alt+L` (on Windows and Linux).
```
alexander-doroshko, line 21: maybe add an action name? 
```
use the standard `Reformat Code` action (`Cmd+Alt+L` on Mac or `Ctrl+Alt+L` on Windows and Linux).
```

src/docs/development/tools/formatting.md
```diff
@@ -17,11 +17,11 @@ where time might be better spent on code behavior rather than code style.
 
 Install the `Dart` plugin (see [Editor setup](/docs/get-started/editor))
 to get automatic formatting of code in Android Studio and IntelliJ.
-
-To automatically format the code in the current source code window,
-right-click in the code window and select `Reformat Code with dartfmt`.
-You can add a keyboard shortcut to this in Keymap section of IntelliJ
-Preferences.
+To automatically format your code in the current source code window,
```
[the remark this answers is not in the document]
EmperorYP7: You might want to consider the similarity between this and the 'Using trailing commas' section on the same page. Yet if the paragraph spacing is too hard on the eyes, I'll get it done nicely.

src/docs/development/tools/formatting.md
EmperorYP7: The option is defined as 'Format code on save' and there is no context of me referring to any action to save a file. Rather, it just conveys to the client that whenever they intend to save their files, it will be formatted.

src/docs/development/tools/formatting.md
```diff
@@ -17,11 +17,11 @@ where time might be better spent on code behavior rather than code style.
...
+use `Cmd+Alt+L` (on Mac) or `Ctrl+Alt+L` (on Windows and Linux).
+Android Studio and IntelliJ also provides a check box named `Format code on save` on
```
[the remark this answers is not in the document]
EmperorYP7: I think 'provides' is more apt for the place it is used in.

src/docs/development/tools/formatting.md
alexander-doroshko: I mean words "will format the current file". There's no *current* file when we talk about saving files in the IDE. It formats *unsaved*, i.e. *edited* files, not *current*.

src/docs/development/tools/formatting.md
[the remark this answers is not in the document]
alexander-doroshko: Is this an exception I'm not aware of?
What I've found is that [as a general rule](https://www.grammarbook.com/grammar/subjectVerbAgree.asp), we use a plural verb with two or more subjects when they are connected by *and*.
image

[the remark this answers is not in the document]
EmperorYP7: > Is this an exception I'm not aware of?
> What I've found is that [as a general rule](https://www.grammarbook.com/grammar/subjectVerbAgree.asp), we use a plural verb with two or more subjects when they are connected by _and_.
> image

Well, that proves my point.

src/docs/development/tools/formatting.md
EmperorYP7: > I mean words "will format the current file". There's no _current_ file when we talk about saving files in the IDE. It saves and formats _unsaved_, i.e. _edited_ files, not _current_.

Oh yes, you're right. Thanks for pointing that out!

src/docs/development/tools/formatting.md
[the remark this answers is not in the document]
alexander-doroshko: Hm, the rule says that 'Android Studio and IntelliJ' should be used as plural (they are 2 independent IDEs, it's not a compound noun), so the verb should be 'provide'.

PR #4194, opened 2020-06-06
Fixing method call

diegokeller:


src/docs/get-started/codelab-web.md
```diff
@@ -440,7 +440,7 @@ Add the code below marked as NEW:
 ```dart
 ...
 return Form(
-  onChanged: () => _updateFormProgress(), // NEW
+  onChanged: () => _updateFormProgress, // NEW
```
EmperorYP7, line 443: There is no need of arrow function here. A function pointer will do the work!

src/docs/get-started/codelab-web.md
```diff
@@ -852,7 +852,7 @@ class _SignUpFormState extends State<SignUpForm> {
   @override
   Widget build(BuildContext context) {
     return Form(
-      onChanged: () => _updateFormProgress(),
+      onChanged: () => _updateFormProgress,
```
EmperorYP7, line 855: I think the implementation should be:
`onChanged: _updateFormProgress,`
rather than using arrow function.

diegokeller:
I hope it is correct now. Glad to help!

PR #4212, opened 2020-06-11
Removing word duplication

BrenoBaronte:
The word 'Open' is appearing duplicated in the cookbook/testing/integration/profiling page. I just removed it.


BrenoBaronte:
@googlebot I signed it!

PR #4111, opened 2020-05-20
Updating flavors page

sfshaza2:
Thanks to @AngeloAvv for raising this issue.

PR #4217, opened 2020-06-11
Add flavor help articles back to flavor page

jmagman:
Re-add some helpful flavor articles that were removed in https://github.com/flutter/website/pull/4111.

Also add https://medium.com/@animeshjain/build-flavors-in-flutter-android-and-ios-with-different-firebase-projects-per-flavor-27c5c5dac10b which is a good Firebase flavor reference.

### DIFF
--- a/src/docs/deployment/flavors.md
+++ b/src/docs/deployment/flavors.md
@@ -6,8 +6,13 @@ description: How to create build flavors specific to different release types or 
 
 Do you need to set up product flavors for different development
 environments or release types?
-The community has created some packages that you might find useful.
+The community has written some articles and packages you might find useful.
+These articles address flavors for both iOS and Android.
 
+* [Creating flavors of a Flutter app][]
+* [Flavoring Flutter][]
+* [Flutter Ready to Go][]
+* [Build flavors in Flutter (Android and iOS) with different Firebase projects per flavor][]
 
 The following packages are listed alphabetically:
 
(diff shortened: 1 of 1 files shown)


jmagman:
\cc @matthew-carroll

matthew-carroll:
Thanks @jmagman. I'll try to work through one of those references today or tomorrow for my own flavor issue. I glanced at the one that looks most relevant for me, which is Firebase, but I didn't notice any discussion of cocoapods. Are cocoapod plugins supposed to work automatically when introducing flavors, or are there manual modifications that one should expect to make?

The first issue I run into is the `podfile` and its hard-coded expectation of a `Runner` target. 

If I hardcode my own targets instead of `Runner`, then I run into an issue where some kind of subdirectory is expected with a given name, like `Pod-Production` or `Pod-Dev`, but I have no idea what leads to that assumption or how to comply.

Do you know if those details are covered anywhere?

jmagman:
If you follow the instructions for build configurations from https://medium.com/@animeshjain/build-flavors-in-flutter-android-and-ios-with-different-firebase-projects-per-flavor-27c5c5dac10b it should mostly work. 

But I'm not sure what state you've gotten into with your build configurations.  CocoaPods will generate different config files (xcconfigs) for each build configuration.  Looking at the screenshot you emailed, that's Dev, Staging, Production. You're using the Flutter app generated Debug.xcconfig and Release.xcconfig for your "Based on Configuration File".  You probably need to create ones that point to the right generated CocoaPods configs.  Something like:

Staging.xcconfig
```
#include "Pods/Target Support Files/Pods-Runner/Pods-Runner.staging.xcconfig" // or whatever it is
#include "Generated.xcconfig" 
```
Dev.xcconfig
```
#include "Pods/Target Support Files/Pods-Runner/Pods-Runner.dev.xcconfig"
#include "Generated.xcconfig"
```

> If I hardcode my own targets instead of `Runner`, then I run into an issue where some kind of subdirectory is expected with a given name, like `Pod-Production` or `Pod-Dev`, but I have no idea what leads to that assumption or how to comply.

Runner target was required until 1.19.0-2.0.pre (currently master, dev, beta channels), see https://github.com/flutter/flutter/pull/56620 and https://github.com/flutter/flutter/issues/14648#issuecomment-627079766

> Do you know if those details are covered anywhere?

Probably not.

PR #4219, opened 2020-06-12
Add flavor help article

mono0926:
In addition to #4217, I think this article is very useful.


sfshaza2:
The note above lies, this PR DID pass Travis.

PR #4258, opened 2020-06-23
update DevTools docs for Flutter Web version - all except memory docs

kenzieschmoll:
This should be landed after we release DevTools 0.8.0 tomorrow.


kenzieschmoll:
Added Network page docs - PTAL @sfshaza2

PR #4261, opened 2020-06-23
Add performance basics video

filiph:
This adds an iframe that plays https://www.youtube.com/watch?v=PKGguGUwSYE at the top of the performance overview page. I'm not 100% we want it like this -- feel free to push back.


filiph:
Thanks for merging! I'm relatively happy with this: https://flutter.dev/docs/perf. Any suggestions for improvement?

PR #4270, opened 2020-06-26
Just some light cleanup.

sfshaza2:


kwalrath:
I think all this does is update the codelabs submodule. If so, lgtm, but I'm adding Brett just to make sure I haven't missed something.

PR #4299, opened 2020-07-02
Add data table theme document link.

perclasson:
Add data table theme document link.


kwalrath:
Sorry for the delay, @perclasson, is this link no longer needed?

perclasson:
@kwalrath Yes it was decided that no link / external doc was needed. Thanks!

PR #4310, opened 2020-07-07
Update "Get Started" instructions for launching app using VS Code

jeremy-w:
Before, the text directed the reader to press a settings button.
Until the Run sidebar is opened, no such settings button exists.
(The text surrounding the button has also changed since initial writing.)

Now, the text includes the essential step of opening the sidebar,
and it points the reader towards the VS Code User Guide if they
need further background on launch configurations.

This confused me for a good while yesterday. I've mostly run tools (e.g. test runners and watchers) using the VS Code embedded Terminal, so I had the Run sidebar icon hidden in my local config, which made the UI previously mentioned in the walkthrough doubly undiscoverable.

### DIFF
--- a/src/docs/get-started/test-drive/_vscode.md
+++ b/src/docs/get-started/test-drive/_vscode.md
@@ -28,12 +28,19 @@ contains a simple demo app that uses [Material Components][].
       click **No Devices** and launch a simulator.
     - To setup a real device, follow the device-specific instructions on the
       [Install][] page for your OS.
- 1. Press the **Settings button**&mdash;a cog icon gear
-    on the top right (now marked with a red or orange indicator)
-    next to the DEBUG text box that reads **No Configuration**.
-    Select flutter. And choose the debug configuration:
-    To create your emulator if it is closed or to run the
-    emulator or device that is now connected.
+ 1. Open the **Run** sidebar by invoking **View > Run**.
+ 1. Activate a **Flutter** launch configuration. If needed, add one like:
+
+    ```json
+    {
+      "name": "Flutter",
+      "type": "dart",
+      "request": "launch"
+    }
+    ```
+    
+    For more information on adding a launch configuration, see
+    ["Launch configurations" in the Visual Studio Code User Guide][launch config].
  1. Invoke **Run > Start Debugging** or press <kbd>F5</kbd>.
  1. Wait for the app to launch &mdash; progress is printed
     in the **Debug Console** view.
@@ -58,6 +65,7 @@ contains a simple demo app that uses [Material Components][].
 [Install]: /docs/get-started/install
 [Material Components]: {{site.material}}/guidelines
 [Quickly switching between Flutter devices]: https://dartcode.org/docs/quickly-switching-between-flutter-devices
+[launch config]: https://code.visualstudio.com/Docs/editor/debugging#_launch-configurations
 [status bar]: {% asset tools/vs-code/device_status_bar.png @path %}
 [trusted your computer]: /docs/get-started/install/macos#trust
 


src/docs/get-started/test-drive/_vscode.md
```diff
@@ -28,12 +28,19 @@ contains a simple demo app that uses [Material Components][].
...
     - To setup a real device, follow the device-specific instructions on the
       [Install][] page for your OS.
- 1. Press the **Settings button**&mdash;a cog icon gear
-    on the top right (now marked with a red or orange indicator)
-    next to the DEBUG text box that reads **No Configuration**.
-    Select flutter. And choose the debug configuration:
-    To create your emulator if it is closed or to run the
-    emulator or device that is now connected.
+ 1. Open the **Run** sidebar by invoking **View > Run**.
+ 1. Activate a **Flutter** launch configuration. If needed, add one like:
```
chalin, line 32: AFAICT (after testing using the stable and master branch releases), these two steps are optional. Can we simplify the test drive instructions and just drop these two new steps?

(I've tested by running on an emulated iPhone and an Android Pixel 2.)

chalin: @DanTup @dnfield ☝️ thoughts?

jeremy-w: I just tried creating a new Flutter app and launching it in VS Code by skipping those steps and going straight to the Run > Start Debugging step, and it did not go well for me. I wound up flailing around for about a few minutes - not a smooth getting started experience.

[Here is a screen recording of duration 4 min 23 sec.](https://drive.google.com/file/d/1w2W_oZX-cDtXiAb_zERvUzL8DGa61H37/view?usp=sharing) It goes from `flutter create test123` at 0:22, through attempting to run the app a few times in VS Code, to finally succeeding in creating a runnable Flutter launch config at 2:58.

Key points in my flailing are:

- Around 1:07, I try just doing **Run > Start Debugging.** Rather than running anything, I instead get an **error sheet**: "Could not find the task 'build'.", with default option "Configure Task", a "Cancel" option, and an alternative option of "Open launch.json".
- I select the default "Configure Task" option. This opens **tasks.json** (not launch.json!), which is only useful if you want to automate running Flutter yourself using shell commands, not if you want to use the launch configuration supported by the Dart extension. This is a dead end.
- Around 2:18, after hoping maybe selecting lib/main.dart and running would make a difference (it doesn't), I'm back at the dialog I originally saw when I attempted Start Debugging. This time, I try selecting the **"Open launch.json"** option. VS Code opens up **settings.json** - this is still not launch.json.
- At last, around 2:53, I notice a "play button" symbol in the status bar. (This is analogous to the button in the Run sidebar.) I click the widget containing it, which says "Jest Tests (user settings)". A few clicks later, I at last have a Flutter launch config, and soon, the Flutter app is running in an iPhone Simulator.

The current steps, by funneling you into the Run sidebar, avoid the confusion I experienced in the recording. The approach I wound up using in the screen recording relies on two bits of UI that can be hidden: the status bar itself, and the Debug widget in the status bar. Because View > Run always adds and focuses the sidebar containing the required Debug widget, the existing steps seems the approach most likely to succeed in the face of any possible user configuration.

(So, why did "Open launch.json" instead open settings.json for me at 2:18? It turns out that another extension had added a launch config to settings.json. Since the opened settings file doesn't place the cursor at the launch config therein, that is not immediately obvious. I found this files as an existing VS Code issue https://github.com/microsoft/vscode/issues/90216, but it is marked as out of scope, so it is not expected to be fixed.)

DanTup: @chalin is correct - those two steps are not only not required, I would recommend *not* creating a `launch.json` unless you need one (eg. to customise launch args, or if you want multiple launch configs). I would recommend removing those steps (eg., after selecting device, you just press `F5`).

@jeremy-w it sounds like your issue is related to a launch config in your user settings - I don't think the instructions here should try to account for that as it seems like a bad configuration. VS Code deliberately supports launching apps without a `launch.json` (and for Flutter I consider it the preferred way), so having a config in user settings that would override that seems undesirable (and would impact other languages too).

DanTup: @jeremy-w btw in your video you ran `flutter create` from the terminal, but the instructions say to use **flutter: New Project** from within VS Code. It's a minor difference, however using the built-in command will open `lib/main.dart` automatically - and due t how VS Code picks a debugger, this can prevent a slightly confusing step you would've hit where VS Code asks you which debugger to use (if the active file is Dart, it would automatically use the Dart debugger).

chalin: >  I would recommend removing those steps

Done, see #4334

chalin: Thanks for the feedback @DanTup!

jeremy-w: thanks @DanTup. i'm glad to hear there's an easier on-ramp. :) i'm deleting the troublesome settings.json launch config now.

PR #4316, opened 2020-07-08
Updated linux snap instructions to match snapcraft release

kenvandine:
There were a couple of changes to the syntax in what was released in snapcraft, this updates to match.


kwalrath:
Thanks for the PR, @kenvandine!

PR #4325, opened 2020-07-09
Update license to notices

jonahwilliams:
As part of https://github.com/flutter/flutter/issues/16723 , the aggregate file is now called NOTICES

Update the documentation to clarify.


kwalrath:
@jonahwilliams is this change already visible in a stable release? If not, will it be in the next stable Flutter release?

Thanks!

jonahwilliams:
This will be in the next release candidate, but for web is already available on dev

PR #4354, opened 2020-07-15
Update _get-sdk.md

csells:
tweaking the Flutter snap-related links

PR #4355, opened 2020-07-15
Updating the snap store links.

sfshaza2:
Changing the links as per PR https://github.com/flutter/website/pull/4354, which changed the wrong file.

### DIFF
--- a/src/docs/get-started/install/_get-sdk-linux.md
+++ b/src/docs/get-started/install/_get-sdk-linux.md
@@ -135,7 +135,7 @@ command again to verify that you’ve set everything up correctly.
 {% include_relative _analytics.md %}
 
 [Flutter repo]: {{site.github}}/flutter/flutter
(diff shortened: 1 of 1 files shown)


src/docs/get-started/install/_get-sdk-linux.md
```diff
@@ -134,9 +135,8 @@ command again to verify that you’ve set everything up correctly.
 {% include_relative _analytics.md %}
 
 [Flutter repo]: {{site.github}}/flutter/flutter
+[install Flutter using the Snap Store]: https://snapcraft.io/store
```
kenvandine, line 138: It would be better to link directly to the store page for the flutter snap, which would provide an install button.

https://snapcraft.io/flutter

sfshaza2: Done

sfshaza2:
Approved offline.

PR #4364, opened 2020-07-18
Get started > Install: list all OSs as cards

chalin:
#### Before:

> Screen Shot 2020-07-18 at 16 01 33

#### After:

> Screen Shot 2020-07-18 at 16 00 25

On mobile:

> Screen Shot 2020-07-18 at 15 59 44

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -101,8 +101,6 @@ toc:
 
 ## Site-wide shorthands
 
-os-list: [Windows, macOS, Linux]
-
 alert:
   bonus: >-
     <aside class="alert alert-info" role="alert" markdown="1">
--- a/src/docs/get-started/install/index.md
+++ b/src/docs/get-started/install/index.md
@@ -4,22 +4,17 @@ description: How to set up your code editor.
 next:
   title: Set up an editor
   path: /docs/get-started/editor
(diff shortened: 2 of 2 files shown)


chalin:
@sfshaza2 - hopefully, you'll find this uniformity better (and that Chrome OS wasn't excluded for a reason).

Note: When resizing the page, just before it collapses into the narrow "mobile-view layout", the "OS" of "Chrome OS" can overflow the card boundary a bit. It isn't that visually distracting, so it didn't seem worth the extra effort to rework the design to fix that very special case. If you'd like that addressed, I'd suggest that we do it in a followup PR.

chalin:
Actually, I think that it would be better to give prominence to the OS icon, something like this:

> Screen Shot 2020-07-19 at 09 57 33

If you also prefer this design, let me know and I'll commit that change too. /cc @kwalrath 

Here's what it looks like on mobile:

> Screen Shot 2020-07-19 at 09 57 23

chalin:
@sfshaza2 - Glad that you like it!

So, do you want the (big) icons-first version too? See https://github.com/flutter/website/pull/4364#issuecomment-660647922 for screenshots.

sfshaza2:
I think I would! :)

chalin:
Done, see #4370

PR #4385, opened 2020-07-23
Adding pointers to the web performance articles

sfshaza2:
Also doing some general cleanup.

John/Brett, I don't need both of you to review.  :)

### DIFF
--- a/src/docs/perf/rendering/best-practices.md
+++ b/src/docs/perf/rendering/best-practices.md
@@ -10,6 +10,23 @@ instead of needing to micro-optimize with complicated profiling tools.
 These best recommendations will help you write the most performant
 Flutter app possible.
 
+If you are writing web apps in Flutter, you might be interested
+in a series of articles, written by the Flutter Material team,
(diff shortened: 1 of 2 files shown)


src/docs/perf/rendering/best-practices.md
```diff
@@ -24,7 +41,7 @@ when designing your app:
 * Avoid overly large single Widgets with a large `build()` function.
   Split them into different Widgets based on encapsulation but also on how
   they change:
-  * When `setState()` is called on a State, all descendent widgets will
```
lucalves, line 27: I believe that in this case "will" is correct.

sfshaza2:
@johnpryan?

johnpryan:
Sorry, I didn't see this, LGTM

src/docs/perf/rendering/best-practices.md
johnpryan: I don't think it's adding any clarity here. The sentence is still correct without it.

sfshaza2:
Thanks, @johnpryan!

PR #4398, opened 2020-07-28
Add note about potential Dex merging issues

arnemolland:
After encountering [issues](https://github.com/flutter/flutter/issues/62388) related to the module package name after manually adding a Flutter module to an existing Android project, I thought it'd be worth mentioning in the docs that identical package names in the host app and module can (and most likely will) lead to Dex merging issues.


arnemolland:
@googlebot I fixed it.

PR #4402, opened 2020-07-29
Update the survey metadata for the Q3 survey

jayoung-lee:


jayoung-lee:
Survey ID, URL, and the dates are updated. 

FYI @pq @DanTup @jackkim9

PR #4407, opened 2020-07-30
Update constraints.md to fix grammar typo

sponrad:
Not sure how the bad grammar crept in, it was correct in the original article.


sponrad:
@googlebot I signed it!

PR #4430, opened 2020-08-04
Updating performance docs

sfshaza2:
Here is the PR that moves the contents of the wiki page to flutter.dev. Once I land the PR, I will update the wiki to redirect to the new page.

Staged:  https://sz-flutter.firebaseapp.com/docs/perf/rendering/shader

Thanks for your review!


src/docs/perf/rendering/shader.md
```diff
@@ -0,0 +1,161 @@
+---
+title: Reduce shader compilation jank on mobile
+short-title: Shader jank
+description: What is shader jank and how to minimize it.
+---
+
+If the animations on your mobile app appear to be janky,
+but only on the first run, you can _warm up_ the
```
filiph, line 8: It would be great to specify here or elsewhere what we mean by "first run". Is it the first time the user opens the app, and then the second open is fine? Is it every time the user opens the app from scratch. If I remember correctly, it's the former for Android and the latter for iOS.

You probably want to go into this detail somewhere else than in the first paragraph.

src/docs/perf/rendering/shader.md
```diff
@@ -0,0 +1,161 @@
...
+
+## What is shader compilation jank?
+
+If an app has janky animations during the first run,
+and later becomes smooth for the same animation,
+then it's very likely due to shader compilation jank.
+
+More technically, a shader is a piece of program that runs on
+a GPU (graphics processing unit).
+When a shader is first used, it needs to be compiled.
```
filiph, line 22:
```suggestion
When a shader is first used, it needs to be compiled on the device.
```

src/docs/perf/rendering/shader.md
```diff
@@ -0,0 +1,161 @@
...
+The compilation could cost up to a few hundred milliseconds
+whereas a smooth frame needs to be drawn within 16 milliseconds
+for a 60 fps (frame-per-second) display.
+Therefore, a compilation could cause tens of frames to be missed,
+and drop the fps from 60 to 6. This is _compilation jank_.
```
filiph, line 27: > drop the fps from 60 to 6

Well, doesn't this make the problem look worse than it is? I mean, compilation jank is bad, but it's not like we're compiling on every frame. We typically have several missed frames, but after that, the app is smooth again.

This almost looks like the app starts on a consistent 6 FPS framerate for some time, which, I believe, is not the case.

src/docs/perf/rendering/shader.md
```diff
@@ -0,0 +1,161 @@
...
+1. **Why not just compile or warm up _all_ possible shaders?**<br>
+   If there are only a limited number of possible shaders,
+   then Flutter could do a warm-up and compile all of them
+   before-hand to avoid such jank.
+   However, for the best overall performance,
+   the Skia GPU backend used by Flutter can dynamically generate
+   shaders based on a lot of parameters at runtime
+   (for example, draws, device models, and driver versions).
+   Due to all possible combinations of those parameters,
+   the number of possible shaders is unfortunately exponential.
```
filiph, line 120:
```suggestion
   the number of possible shaders is exponential, unfortunately.
```

(Or just remove "unfortunately" altogether.)

src/docs/perf/rendering/shader.md
sfshaza2: @liyuqian, can you clarify?

src/docs/perf/rendering/shader.md
sfshaza2: Done

src/docs/perf/rendering/shader.md
sfshaza2: Done

src/docs/perf/rendering/shader.md
liyuqian: Filip's understanding is correct: Android is the former and iOS is the latter.

src/docs/perf/rendering/shader.md
```diff
@@ -0,0 +1,171 @@
+---
+title: Reduce shader compilation jank on mobile
+short-title: Shader jank
+description: What is shader jank and how to minimize it.
+---
+
+If the animations on your mobile app appear to be janky,
+but only on the first run, you can _warm up_ the
+shader for the Skia Shader Language (SkSL) for a
```
liyuqian, line 9: `shader for the Skia Shader Language (SkSL)` -> `shader captured in the Skia Shader Language (SkSL)`

src/docs/perf/rendering/shader.md
```diff
@@ -0,0 +1,171 @@
...
+Definitive evidence for the presence of shader compilation jank
+is to see `GrGLProgramBuilder::finalize` in the tracing.
+See the following screenshot for an example timeline tracing.
+
+![A tracing screenshot verifying jank]({% asset perf/render/tracing.png @path %}){:width="100%"}
+
+## What do we mean by "first run"?
+
+On Android, "first run" means that the user might see
+jank the first time opening the app. Subsequent opens
```
liyuqian, line 39: `first time opening the app` -> `first time opening the app after a fresh installation`

src/docs/perf/rendering/shader.md
liyuqian: Shall we  change `drop the fps from 60 to 6` to `drop the fps from 60 to 6 for a short duration in the first run`?

src/docs/perf/rendering/shader.md
```diff
@@ -0,0 +1,171 @@
...
+should be fine.
+
+On iOS, "first run" means that the user might see
+jank the first time the user opens the app from scratch.
```
liyuqian, line 43: `the first time the user opens the app from scratch` -> `when an animation first occurs every time the user opens the app from scratch`?

src/docs/perf/rendering/shader.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+With such [integration tests][], you can easily and reliably get the
+new SkSLs when the app code changes, or when Flutter upgrades.
+Such tests can also be used to verify the performance change
+before and after the SkSL warm-up. Even better, you can put
+those tests into a CI (continuous integration) system so the
+SkSLs are generated and tested automatically over the lifetime of an app.
+
+Take [Flutter Gallery][] as an example. The CI system is set up to
```
liyuqian, line 106: Shall we specify this `Flutter Gallery` to be the old `Flutter Gallery` that's not in the Google play store?

src/docs/perf/rendering/shader.md
```diff
@@ -0,0 +1,171 @@
...
+   In short, Flutter uses programs (app, Flutter, and Skia code)
+   to generate some other programs (shaders). The number of possible
+   shader programs that Flutter can generate is simply too large.
+
+2. **Can SkSLs captured from one device help shader compilation jank
+   on another device?**<br>
+   Theoretically, there's no guarantee that the SkSLs from one device
+   would help on another device (but they also won't cause any troubles
+   if SkSLs aren't compatible across devices).
+   Practically, as shown in the table on this [SkSL-based warmup PR][],
```
liyuqian, line 140: `[SkSL-based warmup PR][]` -> `[SkSL-based warmup issue][]`
(And update the link name below accordingly.)

src/docs/perf/rendering/shader.md
```diff
@@ -0,0 +1,171 @@
...
+   SkSLs work surprisingly well
+   even if 1) SkSLs are captured from iOS and then applied to Android devices,
+   or 2) SkSLs are captured from emulators and then applied to real mobile
+   devices. As the Flutter team has only a limited number of devices in the lab,
+   we currently don't have enough data to provide a big picture of cross-device
+   helpfulness. We'd love you to provide us more data points to see how it
+   works in the wild. ([`FrameTiming`][] can be used to compute the worst frame
+   rasterization time in release mode.)
```
liyuqian, line 148: Maybe add `And the worst frame rasterization time is a good indicator on how severe the shader compilation jank is.`

src/docs/perf/rendering/shader.md
```diff
@@ -0,0 +1,171 @@
...
+   [GitHub issue][]. In the longer term, we have a plan to use test-based
+   shader warm-up to mitigate this. If there's an urgent need for fixing
+   shader compilation jank on newer iPhones, please leave feedback on
+   [Issue 61045][], and we can help you turn on OpenGL for your app.
+
+[GitHub issue]: {{site.github}}/flutter/flutter/issues
+[Issue 61045]: {{site.github}}/flutter/flutter/issues/61045
+
+If you have any questions or suggestions on warm-up in general,
+please comment on [Issue 60313][] and [Issue 53607][].
```
liyuqian, line 168: Issue 60313 and 53607 are specific to SkSL shader warm-up. For general shader compilation jank (warm-up) discussions, we should direct them to https://github.com/flutter/flutter/issues/32170

src/docs/perf/rendering/shader.md
sfshaza2: Yes, it's staged here:  https://sz-flutter.firebaseapp.com/docs/perf/rendering/shader

I'd like to go live tonight, not tomorrow.

src/docs/perf/rendering/shader.md
sfshaza2: I didn't realize that!!!

src/docs/perf/rendering/shader.md
sfshaza2: I added a follow on sentence saying it's smooth after.

src/docs/perf/rendering/shader.md
sfshaza2: DOne

src/docs/perf/rendering/shader.md
sfshaza2: Done

src/docs/perf/rendering/shader.md
sfshaza2: I didn't realize this, but I added a note.

src/docs/perf/rendering/shader.md
```diff
@@ -0,0 +1,171 @@
...
+short-title: Shader jank
+description: What is shader jank and how to minimize it.
+---
+
+If the animations on your mobile app appear to be janky,
+but only on the first run, you can _warm up_ the
+shader for the Skia Shader Language (SkSL) for a
+significant improvement.
+
+![Side-by-side screenshots of janky mobile app next to non-janky app]({% asset perf/render/shader-jank.gif @path %}){:width="100%"}
```
liyuqian, line 12: The `{:width="100%"}` looks bad in my browser as the GIF is upscaled to full width which looks very blurry on my high-resolution screen. Shall we keep the GIF to be rendered at the original resolution?

src/docs/perf/rendering/shader.md
sfshaza2: Done.

src/docs/perf/rendering/shader.md
sfshaza2: Done

src/docs/perf/rendering/shader.md
liyuqian: Oh, my bad. Somehow I thought that I saw "Wednesday" in some emails, maybe on another topic. CC @chinmaygarde @zanderso for the updated time window.

src/docs/perf/rendering/shader.md
sfshaza2: Done

src/docs/perf/rendering/shader.md
sfshaza2: I removed it for the first image. The tracing image was HUGE and extended beyond the window, so I sized it down.

src/docs/perf/rendering/shader.md
sfshaza2: The release goes live tomorrow at 8am, but this needs to be in place first.

sfshaza2:
I can't show you the staged version at the moment, because I'm using that for another PR. But both images look better now..

sfshaza2:
@liyuqian, I can land this once you LGTM it. If @chinmaygarde or @zanderso have changes, I can do those in a followup PR. This is essentially the same thing on the wiki.

src/docs/perf/rendering/index.md
```diff
@@ -33,7 +35,29 @@ see the following docs:
 * [Performance best practices][]
 * [Flutter performance profiling][]
 
+## Mobile-only advice
+
+Do you see noticeable jank on your mobile app, but only on
+the first run of an animation? If so, see
+[Reduce shader animation jank on mobile][].
+
+[Reduce shader animation jank on mobile]: /docs/perf/rendering/shader
```
liyuqian, line 44: For future PRs: maybe it's nice to include some of Filip's articles here so the mobile section won't look very empty compared to the web section.

src/docs/perf/rendering/shader.md
liyuqian: `questions on shader warm-up` -> `questions on SkSL shader warm-up`

src/docs/perf/rendering/shader.md
```diff
@@ -0,0 +1,177 @@
...
+   shader warm-up to mitigate this. If there's an urgent need for fixing
+   shader compilation jank on newer iPhones, please leave feedback on
+   [Issue 61045][], and we can help you turn on OpenGL for your app.
+
+[GitHub issue]: {{site.github}}/flutter/flutter/issues
+[Issue 61045]: {{site.github}}/flutter/flutter/issues/61045
+
+If you have questions on shader warm-up,
+please comment on [Issue 60313][] and [Issue 53607][].
+If you have general warm-up questions, please refer to
```
liyuqian, line 171: `warm-up` -> `shader warm-up`

sfshaza2:
Thanks for your review, @filiph and @liyuqian!

PR #4431, opened 2020-08-04
A few tweaks on the deployment/web file

sfshaza2:
Thanks for your review!


sfshaza2:
Yeah, either link works, but I updated it.

PR #4449, opened 2020-08-05
Clarifying and reordering

mariamhas:
Clarifying Edge is supported for Windows devices as a default option as well as Chrome.
Reordering a few questions so they flow better

### DIFF
--- a/src/docs/development/platform-integration/web.md
+++ b/src/docs/development/platform-integration/web.md
@@ -67,7 +67,7 @@ Safari (mobile & desktop)
 Edge (mobile & desktop)
 Firefox (desktop)
 
-During development, Chrome and Edge(only on Windows devices) are supported as the default browsers for debugging your app. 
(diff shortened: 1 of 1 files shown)


src/docs/development/platform-integration/web.md
```diff
@@ -68,19 +67,13 @@ Safari (mobile & desktop)
 Edge (mobile & desktop)
 Firefox (desktop)
 
-During development, Chrome is the default browser for running your app. 
+During development, Chrome and Edge(only on Windows devices) are supported as the default browsers for debugging your app.
```
timsneath, line 70: Suggest: "Chrome (on macOS, Windows and Linux) and Edge (on Windows) are supported as..." 

to avoid any confusion as to whether the parentheses apply to one or both browsers.

timsneath:
Yup LGTM!

PR #4457, opened 2020-08-07
Move the link to avoid interrupt markdown ordered Lists

CareF:
The current version is causing https://flutter.dev/docs/perf/rendering/shader#limitations-and-considerations showing 1, 2, 1 as the enumeration label:

image


CareF:
Probably we should also consider it a bug of the website builder.

PR #4464, opened 2020-08-10
Document Xcode symbol stripping

dcharkes:
Add solution to https://github.com/flutter/flutter/issues/62666 to documentation.


kwalrath:
@Sfshaza I'm not sure where this info should go, but I figure we can tweak that later if necessary.

PR #4536, opened 2020-08-27
fix #4505 : Refactoring final example code.

### LINKED ISSUE #4505: 'Persist data with SQLite' page issue
Page URL: https://flutter.dev/docs/cookbook/persistence/sqlite.html
Page source: https://github.com/flutter/website/tree/master/src/docs/cookbook/persistence/sqlite.md

Description of issue: 
In step 3 we create a database reference through `final Future<Database> database`, in the full code example this gets changed to `final database`. I had wrote the code step by step and it would not print out the correct output until I made that change.

aromalanil:
Since the openDatabase() function returns a Future object, I changed the type of the variable.

### DIFF
--- a/src/docs/cookbook/persistence/sqlite.md
+++ b/src/docs/cookbook/persistence/sqlite.md
@@ -324,7 +324,7 @@ void main() async {
   // Avoid errors caused by flutter upgrade.
   // Importing 'package:flutter/widgets.dart' is required.
   WidgetsFlutterBinding.ensureInitialized();
-  final database = openDatabase(
+  final Future<Database> database = openDatabase(
     // Set the path to the database. Note: Using the `join` function from the
     // `path` package is best practice to ensure the path is correctly
     // constructed for each platform.


sfshaza2:
@aromalanil, I tested the example against the latest stable release (1.20) and it passed analysis without problem. What makes you think that you need to explicitly type the `database` field?

@domesticmouse, if you have a minute, can you look at the last example on this sqflite persistence page in the cookbook? I build the example, ran the analyzer over it (it was happy), and ran the test, as described, using the iOS simulator. I saw the following in the console:
```
Xcode build done.                                           43.7s
flutter: [Dog{id: 0, name: Fido, age: 35}]                              
flutter: [Dog{id: 0, name: Fido, age: 42}]                              
flutter: []                                                             
Waiting for iPhone 11 Pro Max to report its views...                13ms
```

But nothing appeared on-screen. I tried a hot restart and that did not work (the screen just turned to black).
Can you take a look at this? I don't *think* it's related to the issue in this PR, but it is weird.

aromalanil:
Since the `database` variable is declared here using the `final` keyword, it will automatically assign the data type to the variable based on the return type of the `openDatabase( )` function. 

But in this case, we clearly know the type which will be assigned to the `database` variable,( Refer : [this](https://github.com/tekartik/sqflite/blob/master/sqflite/lib/sqflite.dart#L127-L146) ) and I think it is a good practice to specify type annotation to a variable when possible. 

Also in the _[Step 3](https://flutter.dev/docs/cookbook/persistence/sqlite#3-open-the-database)_ of this page, the type annotation is given to the `database` variable, and that leads to my second point which is consistency. We need to maintain consistency over different steps of the documentation and the final example code.

Issue #4537, opened 2020-08-27
Add symlink for code_excerpt_framer.rb


johnpryan:
https://github.com/dart-lang/site-shared/pull/121 has been merged into site-shared - code_excerpt_framer.rb should be replaced with the a symlink to the site-shared version.

micpap25:
If a PR has been merged for this I think it can be closed.

johnpryan:
@micpap25 good catch, thanks!

PR #4541, opened 2020-08-27
Move _plugins/code_excerpt_framer.rb into site_shared

limexp:
Replacing code_excerpt_framer.rb with a symlink to the site-shared version
Solving https://github.com/flutter/website/issues/4537


sfshaza2:
Can you weigh in, @kwalrath?

PR #4545, opened 2020-08-28
Update UI for FavoriteWidget

gaetansenn:
I just did the tutorial and I think we should add this to match with the initial UI


gaetansenn:
@googlebot I signed it!

sfshaza2:
@johnpryan, Travis is failing here because it wants the code snippets to be refreshed. But, given that this PR is externally contributed, how do we handle this?

johnpryan:
@sfshaza2 Here's a PR for that: https://github.com/flutter/website/pull/4601. The commit authorship will be included when it's merged.

PR #4546, opened 2020-08-28
remove a quick link

xster:
For various reasons, we need to remove that last link


sfshaza2:
I resolved the conflict. Waiting for Travis.

Issue #4520, opened 2020-08-21
'Building a web application with Flutter' page issue


satriahrh:
Page URL: https://flutter.dev/docs/get-started/web.html
Page source: https://github.com/flutter/website/tree/master/src/docs/get-started/web.md

Description of issue: 

Hi, I am new to flutter. So I try to build a mini project.  It was great when I can test on iOS simulator, very neat.

Then I was curious that I can run on flutter app from web browser, and got to this page.

Asking user to `flutter channel beta` without caution is a big mistake, especially for some of us that have problem with Internet connection issue. I was stumbled upon this kind of condition when knowing this answer on [so](https://stackoverflow.com/questions/61556268/flutter-how-to-switch-flutter-channel-without-downloading-flutter-dart-sdk-e), `flutter channel <something>` will remove previous selected channel and force download selected channel, every time.

Hoping that we can do something to notice the user of the side effect on switching flutter channel.

aromalanil:
I also think that flutter team should add a way to cache the SDK from the previous channel during a  channel switching. This will save the user time and internet data.

sfshaza2:
I think you have a very good point. Please file a request on the flutter/flutter repo, as SWEs don't generally read this one. (But first search the issues - it wouldn't surprise me if someone has already requested this. Then you can upvote and add your voice.)

aromalanil:
@sfshaza2 Sure, I'll do that. As of now I have added a pull request which adds a warning about switching the channel. I request you to take a look into it #4547

aromalanil:
@sfshaza2 I have created an [issue](https://github.com/flutter/flutter/issues/64822) in `flutter/flutter` as I couldn't find any similar issues.

aromalanil:
@sfshaza2 This issue is solved, so I think this can be closed.

parlough:
The SDK issue is being tracked there and there is now a warning on the site as of https://github.com/flutter/website/pull/4547. Thanks everyone :)

PR #4547, opened 2020-08-28
Added a warning on changing the flutter channel

aromalanil:
I have added a warning on changing the flutter channel to beta
This will solve issue #4520

### DIFF
--- a/src/docs/get-started/web.md
+++ b/src/docs/get-started/web.md
@@ -60,9 +60,11 @@ $ flutter config --enable-web
 ```
 
 {{site.alert.warning}}
-By running `flutter channel beta` you will be downloading a less stable 
-version of the Flutter SDK each time you run `flutter upgrade`, and the channel 
-will remain as beta until you change it.
+  Running `flutter channel beta` replaces your current version of Flutter
+  with the beta version and can take time if your connection is slow.
+  After this, running `flutter upgrade` upgrades your install to the latest
+ `beta`.  Returning to the stable channel (or any other) requires calling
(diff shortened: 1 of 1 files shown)


src/docs/get-started/web.md
```diff
@@ -59,6 +59,12 @@ $ flutter upgrade
 $ flutter config --enable-web
 ```
 
+{{site.alert.warning}}
```
sfshaza2, line 62: I thought that one of your primary concerns was that changing channels is slow because it doesn't retain both the original channel and the new channel? It's a full replace? (Also, indent the note by 2 spaces per line.)

How about:

```
{{site.alert.warning}}
  Running `flutter channel beta` replaces your current version of Flutter
  with the beta version and can take time if your connection is slow.
  After this, running `flutter upgrade` upgrades your install to the latest
 `beta`.  Returning to the stable channel (or any other) requires calling
 `flutter channel <channel>` explicitly.
{{site.alert.end}}
```

@jonahwilliams, can you vet this phrasing please?

jonahwilliams: That seems reasonable to me (though it is more evidence that we need non-destructive channel updates...)

sfshaza2: YES YES YES! I have to often change channels and I try to avoid it.  :D

aromalanil: @sfshaza2 @jonahwilliams I'm Making the changes right away

sfshaza2 (approved):
LGTMM

PR #4539, opened 2020-08-27
Updated broken links to pub.dev closes #4538

zvikarp:
I updated all the links, (searched for `site.pub-pkg` in the whole site to find all links to pub.dart. Also because anchors are not used, I Removed pub.dev's condition from the `tool/config/linkcheck-skip-list.txt`.

full issue description is at #4538

PR #4557, opened 2020-08-31
Run travis 4539

johnpryan:
Merges master into https://github.com/flutter/website/pull/4539. Creating this PR so that I can re-run travis


sfshaza2:
@johnpryan, can you just restart the build? If not, we need to fix that. For awhile I couldn't restart the build either, but logging out of Travis and back in again fixed it for both @kwalrath and I. Let's get that working for you.

johnpryan:
It looks like travis is passing

PR #4576, opened 2020-09-04
Revise documentation for breaking socket policy

mehmetf:
This breaking change was reverted at https://dart-review.googlesource.com/c/sdk/+/161581. This is now making its way to flutter/flutter likely landing Sept 6-7.

We are pushing a different version of the breaking change with https://dart-review.googlesource.com/c/sdk/+/161801. This is likely to land to 1.23 instead. Hide the breaking change doc until the change is in master.


sfshaza2 (approved):
LGTM

PR #4585, opened 2020-09-07
Remove reference to Flutter Gallery and fastlane

guidezpl:
The gallery doesn't use fastlane (but we're looking at options)


sfshaza2 (approved):
LGTM

PR #4588, opened 2020-09-07
Add details to MainAxisSize behavior

### LINKED ISSUE #2988: Layout codelab not accurately demonstrating mainAxisSize behavior
(This issue was pointed out to us by someone commenting on a flutter/flutter issue: https://github.com/flutter/flutter/issues/4566)

The main axis size section of the layout basics codelab includes a code exercise that's not explaining things correctly:

https://flutter.dev/docs/codelabs/layout-basics#example-modifying-axis-size

When the main axis size is set to max, the Row fills the entire width and its children are left-aligned. No problem there. However, when it's set to min, because of the way the test code is structured, the row becomes centered within its parent as it shrinks. This could confuse some codelab users into thinking that the Row itself is centering its children, which it's not.

The test result message unfortunately reinforces this message:

```
Row lays out the BlueBox widgets without extra space, and the BlueBox widgets are positioned at the middle of Row's main axis.
```

While technically true (they occupy the entire main axis space of the Row and are therefore centered in it by default), someone unfamiliar with Flutter is likely to miss the subtle moment when the Row's empty space (marked with a dark grey color) disappears. As a result, they could process the new UI as a Row that takes up the same amount of width and has simply centered its children.

The test code and the test result for this exercise should be updated to make the true effect of MainAxisSize.min clear.

micpap25:
Fixes #2988
Clarifies how `MainAxisSize.min` works with a tip after a code section and draws attention to the disappearance of the dark grey space in said coding section.

### DIFF
--- a/src/docs/codelabs/layout-basics.md
+++ b/src/docs/codelabs/layout-basics.md
@@ -105,6 +105,13 @@ The `mainAxisSize` property has two possible values:
 {% endcomment %}
 <iframe src="{{site.custom.dartpad.embed-flutter-prefix}}?id=928d699ea0869e75d072e6e9c4e63397&amp;theme=dark&amp;split=60&amp;ga_id=axis_size" width="100%" height="400px"></iframe>
 
+{{site.alert.tip}}
+Note how the dark grey section of the Row disappears 
+when you change `MainAxisSize.max` to `MainAxisSize.min`.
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
I like it! LGTM

(Though in future, the lines inside an alert should be indented by 2 spaces.)

PR #4607, opened 2020-09-14
Update obfuscate.md

amanv8060:
```terminal
flutter build apk --obfuscate --split-debug-info=../<project-name>/<directory>
```
otherwise it would cause error

Adding `..` no error shows up , wihout that it shows https://pastebin.com/ga9D6uZB

A little example why this happens

```
aman@amanverma:~/Desktop$ mkdir /am
mkdir: cannot create directory ‘/am’: Permission denied
``` 

while this :
```
aman@amanverma:~/Desktop$ mkdir am
```
 works so in the case mentioned above i guess the same problem is there
other way around

```
flutter build apk --obfuscate --split-debug-info=<directory>
``` 
would mean same and would rather be more convenient

### DIFF
--- a/src/docs/deployment/obfuscate.md
+++ b/src/docs/deployment/obfuscate.md
@@ -51,8 +51,9 @@ channels.)
 For example:
 
 ```terminal
-flutter build apk --obfuscate --split-debug-info=../<project-name>/<directory>
+flutter build apk --obfuscate --split-debug-info=<directory>
 ```
+<directory> would be created inside your project folder
 
 Once you've obfuscated your binary, save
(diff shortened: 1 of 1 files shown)


sfshaza2 (changes requested):
I do thinks this is an improvement, @amanv8060, but could you also update the `flutter symbolize` command?

amanv8060:
@sfshaza2 i already changed it that day , is there something else that i need to change?

sfshaza2:
It looks like we've lost @amanv8060. I'm closing this for now. We can always revisit later.

PR #4622, opened 2020-09-16
Add iOS 14 Settings > Local Network instructions

jmagman:
If they miss the local network permission dialog, users can also enable permissions in `Settings > Privacy > Local Network > Your App` on the device.

### DIFF
--- a/src/docs/development/ios-14.md
+++ b/src/docs/development/ios-14.md
@@ -40,7 +40,7 @@ functionalities such as hot-reload and DevTools.
 
 This affects debug and profile builds only and won't
 appear in release builds. The permission can also be allowed
-by enabling `Settings > Privacy > Local Network > Your App`.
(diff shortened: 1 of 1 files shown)


src/docs/development/ios-14.md
```diff
@@ -39,7 +39,8 @@ functionalities such as hot-reload and DevTools.
 ![Screenshot of "allow network connections" dialog]({% asset development/device-connect.png @path %})
 
 This affects debug and profile builds only and won't
-appear in release builds. 
+appear in release builds. The permission can also be allowed
+by enabling `Settings > Privacy > Local Network > Your App`.
```
xster, line 43: I think Shams wants these UI references to be bolded rather than code formatted

xster (approved):

src/docs/development/ios-14.md
jmagman: You're right, I forgot.

Issue #3758, opened 2020-03-03
Update 'Persist data with SQLite' page with fix for code insertion


bergamin:
Page URL: https://docs.flutter.dev/cookbook/persistence/sqlite
Page source: https://github.com/flutter/website/tree/master/src/docs/cookbook/persistence/sqlite.md

This page is a little confusing. It doesn't explain where to add the following code:

```dart
// Open the database and store the reference.
final Future<Database> database = openDatabase(
  // Set the path to the database. Note: Using the `join` function from the
  // `path` package is best practice to ensure the path is correctly
  // constructed for each platform.
  join(await getDatabasesPath(), 'doggie_database.db'),
);
```

Also, wherever I put this block of code, Android Studio doesn't like the word `await`.  I have all three imports in the file:

```dart
import 'package:path/path.dart';
import 'package:sqflite/sqflite.dart';
import 'dart:async';
```

Where I tried to add the code:

* As an attribute of my "Dog" (not dog in my case, but keeping the example)
* As an attribute of my "HomeWidgetState", which is the state of the home widget called from the MaterialApp in main.dart and where I expect to use add/remove/update the data from the database

Should I create a new file/class for the database?

Edit: looking forward in the document, there are other blocks of code that aren't clear where they should be placed

micpap25:
At the end of the file, there's a large finished version of the cookbook. That should resolve any of the issues that you are having. I think this can be closed.

atsansone:
Issue resolved. Closing.

PR #4626, opened 2020-09-17
Clean Up the "Persist data with SQLite" Page

micpap25:
Makes a bunch of fixes to the page:
* Removes typos and grammar mistakes in the first paragraph. 
* Breaks up Dart sections that misleadingly look like they are next to each other when they aren't. (From Issue #3758)
* Also adds a note to make the final example easier to arrive at by following the tutorial.
* Basically makes the final example consistent with the rest of the cookbook.

### DIFF
--- a/src/docs/cookbook/persistence/sqlite.md
+++ b/src/docs/cookbook/persistence/sqlite.md
@@ -179,6 +179,7 @@ class Dog {
 }
 ```
 
+<!-- skip -->
 ```dart
 // Define a function that inserts dogs into the database
 Future<void> insertDog(Dog dog) async {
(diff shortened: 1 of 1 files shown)


sfshaza2:
Thanks for this PR, @micpap25! Overall, it looks good, however, Travis is failing and it appears to be because some of the code snippets aren't tagged with the "skip" tag. The skip tag is required for incomplete code examples or code that uses APIs that aren't yet in our latest stable channel. It tells the analyzer not to analyze that code.

sfshaza2 (changes requested):
It looks good overall, but is breaking the build.

micpap25:
Fixed! I didn't know about the skip tags, so thank you for pointing that out.

sfshaza2 (approved):
LGTM

PR #4637, opened 2020-09-21
Add libblkid in the dependencies list

dedesite:
On my system (Xubuntu 18.04) at least it is required

### DIFF
--- a/src/desktop.md
+++ b/src/desktop.md
@@ -62,6 +62,7 @@ you need the following in addition to the Flutter SDK:
 * [GTK development headers][]
 * [Ninja build][]
 * [pkg-config][]
+* [libblkid][]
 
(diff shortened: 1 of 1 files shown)


dedesite:
@googlebot I signed it!

sfshaza2:
@stuartmorgan, can you verify this?

stuartmorgan:
That's correct:
https://github.com/flutter/flutter/blob/3a5e45404d2a79501c84334bcebd5d65df28280d/packages/flutter_tools/lib/src/base/user_messages.dart#L238
It was made an explicit dependency post-alpha and I forgot we'd need to update these instructions.

sfshaza2 (approved):
LGTM! Thanks!

PR #4639, opened 2020-09-21
Rename all uses of "code size" to "app size" in DevTools docs

kenzieschmoll:


sfshaza2 (approved):
Thanks for doing this, Kenzie!!

PR #4641, opened 2020-09-22
Fix types in "Staggered Animations"

### LINKED ISSUE #3865: Incorrect type for `controller` on 'Staggered Animations' page issue
Page URL: https://flutter.dev/docs/development/ui/animations/staggered-animations.html
Page source: https://github.com/flutter/website/tree/master/src/docs/development/ui/animations/staggered-animations.md

Description of issue:

`controller` should be of type `AnimationController` instead of `Animation<double>` (although technically speaking the former is a subclass of a latter class so it should be OK...), which might be less confusing?

micpap25:
Fixes #3865
Changes one type on the page so that the main controller is an `AnimationController` rather than an `Animation<double`. 
@xcoderreal , who opened this issue, made this commit on their branch of the repo but hasn't made the PR since it came up in March/April. Hopefully it's OK that I'm making a PR now for this.

### DIFF
--- a/src/docs/development/ui/animations/staggered-animations.md
+++ b/src/docs/development/ui/animations/staggered-animations.md
@@ -20,7 +20,7 @@ short-title: Staggered
   If the concept of tweens or tweening is new to you, see the
   [Animations in Flutter tutorial][].
 {{site.alert.end}}
-
+<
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
Thanks! LGTM

PR #4648, opened 2020-09-23
Desktop updates

sfshaza2:


src/desktop.md
```diff
@@ -97,29 +134,41 @@ $ flutter upgrade
...
+$ flutter config --enable-windows-desktop
 $ flutter config --enable-macos-desktop
 $ flutter config --enable-linux-desktop
 ```
 
 To ensure that desktop _is_ installed,
 list the devices available.
 You should see something like the following
-(you'll see either macOS _or_ Linux, not both):
+(you'll see windows, macOS, or Linux,
```
stuartmorgan, line 148: nit: Windows, not windows

stuartmorgan (approved):
LGTM with one typo fix.

src/desktop.md
```diff
@@ -5,32 +5,31 @@ toc: true
...
-  [Desktop shells][] page in the [Flutter wiki][]
-  to understand known limitations and ongoing work.
 {{site.alert.end}}
 
 {{site.alert.note}}
-  To compile a macOS desktop app, you must build the app on a Mac.
-  To compile a Linux desktop app, you must build the app on Linux.
+  To compile a desktop app, you must build the app **on**
+  the targeted platform: build a Windows app on Windows,
+  a macOS app on macOS, and so on.
```
csells, line 25: there's only one more platform left, so perhaps "and so on" becomes "and a Linux app on Linux"

csells (approved):

Issue #4661, opened 2020-09-25
Define the differences between "packages" and "plugins"


jmagman:
Plugins and packages are used interchangeably in the "Using packages" guide, but I don't see a definition between the two terms.
https://flutter.dev/docs/development/packages-and-plugins/using-packages

sfshaza2:
Yes, this page should be clarified, though the info is on the [developing packages](https://flutter.dev/docs/development/packages-and-plugins/developing-packages#types) page. Though, again, it could be made more clear.

jmagman:
That's exactly what I was looking for!  Maybe https://flutter.dev/docs/development/packages-and-plugins/using-packages just needs a link to that.

jmagman:
I was trying to define it for using plugins in an app, not developing them, so I didn't look on the developing packages page.

Roshan13046:
> That's exactly what I was looking for! Maybe https://flutter.dev/docs/development/packages-and-plugins/using-packages just needs a link to that.

I would like to work on this issue.
so, should I add a link or define difference between packages and plugins?

sfshaza2:
Go for it! I would like to see a section added to the beginning of the page defining the difference. Then we can link to that text from other pages, as needed.

parlough:
This has been expanded on in https://github.com/flutter/website/pull/5287 :)

PR #4662, opened 2020-09-25
Remove pod setup instructions, note the sudoless CocoaPods install

jmagman:
1. `pod setup` is no longer required as of CocoaPods 1.8. Remove it.
2. Fix Xcode casing.
3. Add link to CocoaPods installation docs.
4. Add note about Ruby version managers not requiring `sudo`.
5. Add caveat that CocoaPods is only needed when using Flutter plugins.

https://github.com/flutter/flutter/issues/10122#issuecomment-698499467
Fixes https://github.com/flutter/flutter/issues/65131.

### DIFF
--- a/src/docs/get-started/install/_ios-setup.md
+++ b/src/docs/get-started/install/_ios-setup.md
@@ -90,69 +90,83 @@ you'll need to set up physical device deployment in Xcode
 and an Apple Developer account. If your app is using Flutter plugins,
 you will also need the third-party CocoaPods dependency manager.
 
- 1. You can skip this step if your apps do not depend on
- [Flutter plugins][] with native iOS code.
- [Install and set up CocoaPods][] by running the following commands:
+<ol markdown="1">
+<li markdown="1">
 
-    ```terminal
-    $ sudo gem install cocoapods
-    ```
+You can skip this step if your apps do not depend on
+[Flutter plugins][] with native iOS code.
+[Install and set up CocoaPods][] by running the following commands:
+
+```terminal
+$ sudo gem install cocoapods
+```
 {{site.alert.note}}
   The default version of Ruby requires `sudo` to install the CocoaPods gem.
   If you are using a Ruby Version manager, you may need to run without `sudo`.
 {{site.alert.end}}
 
- 1. Follow the Xcode signing flow to provision your project:
+</li>
 
-     {: type="a"}
-     1. Open the default Xcode workspace in your project by
-        running `open ios/Runner.xcworkspace` in a terminal
-        window from your Flutter project directory.
-     1. Select the device you intend to deploy to in the device
-        drop-down menu next to the run button.
-     1. Select the `Runner` project in the left navigation panel.
-     1. In the `Runner` target settings page,
-        make sure your Development Team is selected.
-        The UI varies depending on your version of Xcode.
-        * For Xcode 10, look under **General > Signing > Team**.
-        * For Xcode 11 and newer, look under
-        **Signing & Capabilities > Team**.
+<li markdown="1">
 
-        When you select a team,
-        Xcode creates and downloads a Development Certificate,
-        registers your device with your account,
-        and creates and downloads a provisioning profile (if needed).
+Follow the Xcode signing flow to provision your project:
 
-        * To start your first iOS development project,
-          you might need to sign into
-          Xcode with your Apple ID. ![Xcode account add][]{:.mw-100}
-          Development and testing is supported for any Apple ID.
-          Enrolling in the Apple Developer Program is required to
-          distribute your app to the App Store.
-          For details about membership types,
-          see [Choosing a Membership][].
+   {: type="a"}
+   1. Open the default Xcode workspace in your project by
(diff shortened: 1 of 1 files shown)


src/docs/get-started/install/_ios-setup.md
```diff
@@ -152,6 +157,8 @@ to set up physical device deployment in Xcode.
 [Check the app's Bundle ID]: /images/setup/xcode-unique-bundle-id.png
 [Choosing a Membership]: https://developer.apple.com/support/compare-memberships
 [Mac App Store]: https://itunes.apple.com/us/app/xcode/id497799835
+[Flutter plugins]: /docs/development/packages-and-plugins/using-packages
```
jmagman, line 160: I wanted to link somewhere more specific that explained what plugins were (with native code), but I couldn't really find anything.  Filed https://github.com/flutter/website/issues/4661.

src/docs/get-started/install/_ios-setup.md
```diff
@@ -86,16 +86,21 @@ follow these steps:
...
- 1. Install and set up CocoaPods by running the following commands:
+ 1. You may skip this step if your apps do not depend on
+ [Flutter plugins][] with native iOS code.
+ [Install and set up CocoaPods][] by running the following commands:
 
     ```terminal
     $ sudo gem install cocoapods
-    $ pod setup
     ```
+{{site.alert.note}}
```
jmagman, line 100: @sfshaza2 Would you mind deploying this so we can check if this note messed up the numbering?  I know we've needed to add `ol`s in a few places.

src/docs/get-started/install/_ios-setup.md
```diff
@@ -86,16 +86,21 @@ follow these steps:
...
+ The default version of Ruby requires `sudo` to install the CocoaPods gem.
+ If you are using a Ruby Version manager, you may need to run without `sudo`.
```
xster, line 102: I'm under the impression you need 2 indents inside these alerts

src/docs/get-started/install/_ios-setup.md
xster: random tangent: you don't need to deploy it to check formatting. You can just follow https://github.com/flutter/website#developing locally.

xster (commented):
LGTM

src/docs/get-started/install/_ios-setup.md
sfshaza2: @xster is correct! Just run the script in tool/serve.sh and you can view it on localhost. You just can't share that link with others.  :)

src/docs/get-started/install/_ios-setup.md
```diff
@@ -86,16 +86,21 @@ follow these steps:
...
 To deploy your Flutter app to a physical iOS device
-you need the third-party CocoaPods dependency manager
-and an Apple Developer account. You'll also need
-to set up physical device deployment in Xcode.
+you'll need to set up physical device deployment in Xcode
+and an Apple Developer account. If your app is using Flutter plugins,
+you will also need the third-party CocoaPods dependency manager.
 
- 1. Install and set up CocoaPods by running the following commands:
+ 1. You may skip this step if your apps do not depend on
```
sfshaza2, line 93: We avoid "may".
=? You can skip this...

src/docs/get-started/install/_ios-setup.md
sfshaza2: Thanks for filing that bug, @jmagman. The info is a bit buried on the [developing packages](https://flutter.dev/docs/development/packages-and-plugins/developing-packages#types) page, but both could use some clarification.

sfshaza2 (changes requested):
A couple of minor comments.

src/docs/get-started/install/_ios-setup.md
jmagman: Ah ha, there it is!  Will update the link to that anchor.

src/docs/get-started/install/_ios-setup.md
jmagman: I'm not willing to install nvm and rvm on my development machine.  See: this very diff for all the versioning trouble it causes.

xster: ya, it's definitely a hassle. Though on the flip side, fwiw, I never managed to get ruby or python with a reasonably high version to install nicely on mac either without their own ruby or python specific version manager.

jmagman: You're right, I didn't need `rvm`, I was just missing `jekyll` and `npx`.

jmagman: This had two #1s, I changed it to `ol` and `li`s, now it increments to 1 and 2, and the alphabetized `{: type="a"}` also looks right.
Screen Shot 2020-09-25 at 6 40 08 PM

jmagman:
Ready for re-review, @sfshaza2!

sfshaza2 (approved):
LGTM

PR #4663, opened 2020-09-25
Add Section on Internet Permission to "Fetch Data from the Internet" Page

### LINKED ISSUE #4008: 'Fetch data from the internet' page issue
Page URL: https://flutter.dev/docs/cookbook/networking/fetch-data.html
Page source: https://github.com/flutter/website/tree/master/src/docs/cookbook/networking/fetch-data.md

Description of issue:
Nowhere on this page or in the http package documentation does it say that we should add the INTERNET permission in the AndroidManifest.xml. As a result, things work fine on the emulator, and even on my device in debug mode, but fail when installed from the Play Store.

I had to find the solution somewhere on a GitHub issue (https://github.com/flutter/flutter/issues/27883#issuecomment-486942699).

Can you add this note somewhere, that we should add this snippet to the manifest:
```xml
    <!-- Required to fetch data from the internet. -->
    <uses-permission android:name="android.permission.INTERNET" />
```

micpap25:
Fixes #4008
Adds a short segment on setting permissions in the AndroidManifest.xml file.

### DIFF
--- a/src/docs/cookbook/networking/fetch-data.md
+++ b/src/docs/cookbook/networking/fetch-data.md
@@ -42,6 +42,15 @@ Import the http package.
 import 'package:http/http.dart' as http;
 ```
 
+Additionally, in your AndroidManifest.xml file, 
+add the Internet permission.
+
+<!-- skip -->
+```xml
+<!-- Required to fetch data from the internet. -->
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
LGTM

PR #4672, opened 2020-09-26
minor clarifications and typo fix for App Clip page

xster:


src/docs/development/platform-integration/ios-app-clip.md
```diff
@@ -302,9 +303,11 @@ flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
 
 target 'Runner'
 target '<name of your App Clip target>'
-end
 ```
 
+At the top of the file, also uncomment `platform :ios, '9.0'` and set the
```
jmagman, line 308: Why do this?

xster: Maybe there's another way to handle this. But if this isn't added, then there'll be 2 divergent deployment targets. The default 8 or 9 from the main app and the App Clip which defaults to 14. When the podfile "target"s the 2 of them, it'll create 2 sets of configs to target them individually but if the plugins don't use any specific versioned API, we should just let it create 1 set of config. 

In your original guide, you just set the whole project's deployment target to 13. But I might be nicer to let people target different versions per target.

jmagman: They should fix that in the respective Xcode targets, you don't need CocoaPods to use App Clips.

jmagman: Oh this is in the `Integrate plugins` section.  You're saying we're trying to keep CocoaPods from making multiple xcconfigs?  Does that hurt anything?

sfshaza2 (approved):
LGTM

PR #4675, opened 2020-09-26
Update index.md

pdhankhar61:
now it is correct.


pdhankhar61:
> @googlebot I signed it!

PR #4683, opened 2020-09-28
:memo: fixed a typo

pranjaljain0:
the original line uses "may" when it should use "might". (Search for "may" in the wordlist: https://developers.google.com/style/word-list).
Pointed out by @Sfshaza


sfshaza2 (approved):
LGTM. Thanks!

PR #4686, opened 2020-09-29
Fixed incorrect wording in 'Using the Performance view' page.

### LINKED ISSUE #4633: Unclear/incorrect wording in 'Using the Performance view', profile-granularity
Page URL: https://flutter.dev/docs/development/tools/devtools/performance#profile-granularity
Page source: https://github.com/flutter/website/tree/master/src/docs/development/tools/devtools/performance.md

Description of issue:
> The sampling rates for low, medium, and high granularity are 1 / 50 μs, 1 / 250 μs, and 1 / 1000 μs, respectively.

I think this should be the opposite way around (low is a sample every 1000 μs, high is one every 50 µs) and could be clearer (like saying _every_ x μs).

aromalanil:
fix #4633


sfshaza2 (approved):
LGTM

PR #4710, opened 2020-09-30
navigation bar issue

Niranjan215:
Navigation bar issue for small screen size.responsiveness is to be corrected by Using CSS properties.


kwalrath:
It looks like you're trying to change Hixie's branch to match the master branch; that should be up to Hixie. We'd be very interested in your navbar improvements in a new PR, if you can do that (and sign the CLA, of course).

PR #4711, opened 2020-09-30
improved docs

IshaPreethvi:


kwalrath:
The README isn't really the place for this information, although it could point out to an external description.

PR #4713, opened 2020-09-30
Update README.md

PriyanshuRajSingh:


sfshaza2:
Thanks for this PR, but there are a few problems.

1. You've misspelled "through".
2. We avoid using "below" in our official docs.
3. Converting to a string for the date makes sense when you have a date like 2/5, where it's confusing, but 10/26 can only be converted in one way.

PR #4693, opened 2020-09-30
Includes the download link for the Dream11 iOS app

lucalves:
Hello everyone, everything good?

This PR includes the download link for the iOS Dream11 application.

:)

PR #4714, opened 2020-09-30
:memo: Added URL to an App

pranjaljain0:
Added a URL to the Dream 11 IOS application.


kwalrath:
Duplicate of #4693 (which I didn't merge because someone more familiar with the showcase should review it).

pranjaljain0:
OKay

Issue #4586, opened 2020-09-07
'Windows install' page issue - Use where command instead of which


pavi2410:
Page URL: https://flutter.dev/docs/get-started/install/windows.html
Page source: https://github.com/flutter/website/tree/master/src/docs/get-started/install/windows.md

Description of issue:
Under the ["Update your path"](https://flutter.dev/docs/get-started/install/windows#update-your-path) header, there is a note which tells us to run this command `which flutter dart` which is not available on Windows. I request you to change that to use the `where` command built into Windows. This will avoid confusion among Windows users.

TheBiirb:
It is pretty clear in the documentation:

> The following command (### on macOS, linux, and chrome OS), tells you whether the flutter and dart commands originate from the same bin directory and are therefore compatible. (**Some versions of Windows support a similar where command**.)

sfshaza2:
Thanks for that info, @pavi2410! PRs welcome, by the way.  :D

micpap25:
Just going to add that the section @pavi2410 is referring to is part of a file in _includes that is shared between the Install page for Windows, Mac, and Linux.
I also agree that it is pretty clear unless we were to make a completely separate version of this section for Windows. Is that a good idea? If so I think the title of this issue should be changed to reflect that, and also the time to resolve might need to be increased.

sfshaza2:
I'm closing this issue. The doc does mention using "where" on Windows.

PR #4719, opened 2020-09-30
:memo: Issue #4586

pranjaljain0:
Page URL: https://flutter.dev/docs/get-started/install/windows.html
Page source: https://github.com/flutter/website/tree/master/src/docs/get-started/install/windows.md

Description of issue:
Under the "Update your path" header, there is a note which tells us to run this command which flutter dart which is not available on Windows. I request you to change that to use the where command built into Windows. This will avoid confusion among Windows users.

### DIFF
--- a/src/_includes/dart-tool.md
+++ b/src/_includes/dart-tool.md
@@ -14,7 +14,7 @@
   support a similar `where` command.)
 
   ```terminal
-  $ which flutter dart
+  $ where flutter dart
   /path-to-flutter-sdk/bin/flutter
   /usr/local/bin/dart
   ```
@@ -29,7 +29,7 @@
   now come from the same directory.
 
   ```terminal
-  $ which flutter dart
+  $ where flutter dart
   /path-to-flutter-sdk/bin/flutter
   /path-to-flutter-sdk/bin/dart
   ```


kwalrath:
Googlers don't generally have access to Windows, so our docs tend to be UNIX-centric. It's a good idea to cover Windows, but I'm not sure how we should do it, in general.

pranjaljain0:
@kwalrath You're right, But it was an issue raised in which @Sfshaza said that PRs are welcome so I tried to fix the issue :)

pranjaljain0:
@kwalrath on the thing you said, It got me thinking that indeed there is quality documentation needed for other OSs as well.

kwalrath:
For the record, the issue is #4586. (We don't get the nice cross-link if the issue number is only in the title.)

pranjaljain0:
@kwalrath how do I refer to the issue in the right way?

pranjaljain0:
It's my second day as an open-source contributor 😂

kwalrath:
> It's my second day as an open-source contributor 😂

Wow, congrats and welcome! GitHub has some [help on how to mark that a PR will fix an issue](https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

If the PR is related to an issue (or other PR) but won't necessarily close it, I'll just mention the issue/PR # in the description (first comment). Sometimes I'll say something like `Contributes to #4586.`

Also, since we're here... It helps to have PR titles that have a very brief description of the change. For example, "Improve path example".

pranjaljain0:
Okay @kwalrath That was actually a huge help
I was just doing it in a haphazard way right now.
But this helps.
Thanks a lot.

sfshaza2:
Hi, @pranjaljain0! I've closed the issue, as the page does clearly mention that the `where` command is available on some Windows releases. I don't want to use your PR, as it would confuse UNIX users (Linux and macOS). 

Thanks for the PR otherwise, and I appreciate that you looked for the "PRs welcome" label.

PR #4733, opened 2020-10-01
Capitalized Important words.

nirupam090:


sfshaza2 (commented):
We use sentence case, not title case. Not an improvement.

kwalrath:
Our style is to use sentence case for headings.

PR #4752, opened 2020-10-01
A amazingwebsiteRAEADME.md

nil901:


kwalrath:
Not an improvement.

GermaVinsmoke:
These people doesn't even try to check what they're changing. Atleast go and check how .md files are written.

PR #4782, opened 2020-10-01
Adds the need for "--platforms" section to plugin creation

### LINKED ISSUE #4551: 'Binding to native code using dart:ffi' page issue
https://flutter.dev/docs/development/platform-integration/c-interop

Step two fails because ios and android directories do not exist.
It appears to be missing:
```bash
flutter create -t plugin --platforms android .
flutter create -t plugin --platforms ios .
```
after step one and before step two.

micpap25:
Fixes #4551
Shows that the `--platforms` option is necessary in part 1 to avoid failure in part 2. Also adds a note that you can exclude platforms you aren't building to.

### DIFF
--- a/src/docs/development/platform-integration/c-interop.md
+++ b/src/docs/development/platform-integration/c-interop.md
@@ -59,9 +59,15 @@ To create a plugin called "native_add",
 do the following:
 
 ```terminal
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
LGTM

PR #4793, opened 2020-10-02
fixed a typo in src/docs/codelabs/explicit-animations.md

Vishal19111999:
"the" was used twice in a scentence


sfshaza2 (commented):
This page is not yet complete nor public.

PR #4812, opened 2020-10-03
Update desktop.md

### LINKED ISSUE #4805: 'Desktop support for Flutter' page issue
Page URL: https://flutter.dev/desktop.html
Page source: https://github.com/flutter/website/tree/master/src/desktop.md

Description of issue: Add the command for dnf package manager

In the session [#additional-linux-requirements](https://flutter.dev/desktop#additional-linux-requirements), there is a command for installing it manually for Linux. The command is using APT, that is only for Debian-based distros. It should change from:

> If snapd is unavailable on the Linux distro you’re using, you might use the following command:

for something like:

> If snapd is unavailable on the Linux distro you’re using, you might use the following command for distros that use APT (Debian, Ubuntu, Mint):

Also, you could add the Fedora version of the command, that is the main point of this issue, that is this one:

`dnf install clang cmake ninja-build pkg-config gtk3-devel libblkid-devel`

therealansh:
NOTE: We love contributions, but we hate spam. If your PR doesn't improve flutter.dev, we'll close it and label it as `invalid`.

Fixes #4805 

*  Added commands to fulfil Linux requirement for the Linux distros that do not use apt as there package manager.

* Commands added for pacman, dnf, yum package managers

### DIFF
--- a/src/desktop.md
+++ b/src/desktop.md
@@ -106,19 +106,19 @@ you might use the following command for the distros that use apt (Debian, Ubuntu
 $ sudo apt-get install clang cmake ninja-build pkg-config libgtk-3-dev libblkid-dev
 ```
 
-However, for distros like Fedora, which use DNF package manager, you might use the following command:
+However, for distros like Fedora, which use <em>`DNF`</em> package manager, you might use the following command:
 
 ```terminal
 $ sudo dnf install clang cmake ninja-build pkg-config gtk3-devel libblkid-devel
 ```
 
-For Arch Linux, which uses pacman package manager, the following command should be used:
+For Arch Linux, which uses <em>`pacman`</em> package manager, the following command should be used:
 
 ```terminal
 $ sudo pacman -S clang cmake ninja-build pkg-config gtk3-devel libblkid-devel
 ```
 
-For users using YUM package manager, the following command should be used:
+For users using <em>`YUM`</em> package manager, the following command should be used:
 
 ```terminal
 $ sudo yum install clang cmake ninja-build pkg-config gtk3-devel libblkid-devel
--- a/src/desktop.md
+++ b/src/desktop.md
@@ -106,7 +106,7 @@ you might use the following command for the distros that use apt (Debian, Ubuntu
 $ sudo apt-get install clang cmake ninja-build pkg-config libgtk-3-dev libblkid-dev
 ```
 
-However, for distros like Fedora, which uses dnf package manager, you might use the following command:
+However, for distros like Fedora, which use DNF package manager, you might use the following command:
 
 ```terminal
 $ sudo dnf install clang cmake ninja-build pkg-config gtk3-devel libblkid-devel
--- a/src/desktop.md
+++ b/src/desktop.md
@@ -100,10 +100,28 @@ $ sudo snap install flutter --classic
 ```
 
 If `snapd` is unavailable on the Linux distro you're using,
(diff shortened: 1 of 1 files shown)


therealansh:
@googlebot I signed it!

sfshaza2:
@stuartmorgan, this LGTM, but can you weigh in?  Thx!

src/desktop.md
```diff
@@ -100,12 +100,30 @@ $ sudo snap install flutter --classic
...
 
 If `snapd` is unavailable on the Linux distro you're using,
-you might use the following command:
+you might use the following command for the distros that use apt (Debian, Ubuntu, Mint):
 
 ```terminal
 $ sudo apt-get install clang cmake ninja-build pkg-config libgtk-3-dev libblkid-dev
 ```
 
+However, for distros like Fedora, which uses dnf package manager, you might use the following command:
```
micpap25, line 109: "...for distros like Fedora, which *use DNF* package manager..."

micpap25 (commented):
I think this is great, but it does need one grammatical change.

stuartmorgan:
I'm very skeptical that we actually want to maintain four versions of these instructions. And this isn't exhaustive either; if we setting a precedent that it's intended to be, then this will only grow over time as people submit instructions for increasingly obscure package managers.

The `apt` instructions weren't meant to be exhaustive, but representative, with people who use other package managers adjusting as necessary. See https://flutter.dev/docs/get-started/install/linux#update-your-path for example; we provide instructions for the common case, and a general description of what needs to be done otherwise. We don't provide instructions for every possible shell, or even the top four.

If we were to do this, at the very least we should:
- get stats on percentage of users using different package managers and set a hard threshold on what the criteria for inclusion are, and
- use an inline tab view for the different variants so that we not using so much of the visual weight of the page on infrequently-used variant instructions.

sfshaza2:
That makes sense, @stuartmorgan, thanks for weighing in!

AllanDaemon:
> The `apt` instructions weren't meant to be exhaustive, but representative, with people who use other package managers adjusting as necessary. See https://flutter.dev/docs/get-started/install/linux#update-your-path for example; we provide instructions for the common case, and a general description of what needs to be done otherwise. We don't provide instructions for every possible shell, or even the top four.
> 
> If we were to do this, at the very least we should:
> 
> * get stats on percentage of users using different package managers and set a hard threshold on what the criteria for inclusion are

I agree with that. I opened the issue because it wasn't to find the rpm equivalents of the apt version, so I just want to share it. 

If you would maintain, just .deb and .rpm packages would be enough, as many places just support these ones. If you put archlinux packages, that would more than 90% of the Linux market for sure (ditro share can confirm this). But for Arch, there is the AUR that probably someone would do the job and share it there, without the need of sharing this here.

stuartmorgan:
A tab-based version with `deb` and `rpm` variants sounds reasonable to me. I don't think we should break it down by individual installer tools there, just pick one for the example and mention that the specific command may be different depending on distro—I expect that people will know how to install packages, it's just the names of the packages that are non-obvious.

AllanDaemon:
> A tab-based version with `deb` and `rpm` variants sounds reasonable to me. I don't think we should break it down by individual installer tools there, just pick one for the example and mention that the specific command may be different depending on distro—I expect that people will know how to install packages, it's just the names of the packages that are non-obvious.

I agree 100%.

PR #4826, opened 2020-10-06
remove zip as a system requirement

christopherfujino:
We haven't actually required the `zip` binary in a long time, and yet the website still lists it as a macos dependency. This change removes it. This will likely not affect users, as `zip` comes installed on Mac OS X (this is why we use zip archives, rather than tarballs).


sfshaza2 (approved):
Thanks for this update!

PR #4839, opened 2020-10-06
Add `build appbundle`

liyuqian:
This should reduce confusions like
https://github.com/flutter/flutter/issues/32170#issuecomment-703320921


sfshaza2 (approved):
LGTM

PR #4840, opened 2020-10-07
removed obsolete hint

diskostu:

*  updated the docs

### DIFF
--- a/src/docs/get-started/codelab.md
+++ b/src/docs/get-started/codelab.md
@@ -399,11 +399,6 @@ class _RandomWordsState extends State<RandomWords> {
     [!}!]
   }
   ```
-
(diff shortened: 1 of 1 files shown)


sfshaza2:
How is this obsolete, @diskostu?

diskostu:
Because the info 

> After adding the state class, the IDE complains that the class is missing a build method.

is not valid in this place. Just before the info, in the code snippet, we already added the `build`-method.

> Next, you’ll add a basic build method that generates the word pairs by moving the word generation code from MyApp to _RandomWordsState.

is also not valid, because that's just what we added before.

sfshaza2 (approved):
LGTM

PR #4853, opened 2020-10-09
Remove iOS app display name 1.18 caveat

jmagman:
https://github.com/flutter/flutter/issues/41793 was fixed in 1.18, which was released in May.  We're 2 stable versions past it, so let's remove the warning to only allow the Display Name field in the Xcode UI to be `Runner`.

See also #3828 and #3838 and https://github.com/flutter/website/pull/4055


sfshaza2 (approved):
LGTM!

PR #4866, opened 2020-10-13
Update sample code for CustomPainter in ios-devs

turara:

* Use same sample code for CustomPainter in iOS page as those in Android and Xamarin pages

I think sample code in Android/Xamarin pages is easier to run and see how it works.


sfshaza2 (approved):
LGTM, though I am taking it a bit on faith.

PR #4868, opened 2020-10-13
Subscribe page: enable double opt for certain country codes

johnpryan:
Austria, Germany, Greece, Luxembourg, and Norway now require double-opt-in consent.


sfshaza2 (approved):
LGTM

PR #4883, opened 2020-10-15
Add breaking change "Use maxLengthEnforcement instead of maxLengthEnforced"

AlexV525:
Migration guide for  https://github.com/flutter/flutter/pull/63754 and https://github.com/flutter/flutter/pull/68086 .

Changes proposed in these pull requests:

*  Add the migration guide for `MaxLengthEnforcement` about the composing editing value truncate behavior.

This may also require a public announcement.

Design doc: https://docs.google.com/document/d/1S900lQaTPOsePnIeE7o3-Ia8f39VZaSv0JnjNyBfCvU


src/docs/release/breaking-changes/composing-value-wont-be-truncated-by-default.md
```diff
@@ -0,0 +1,68 @@
...
+<!-- skip -->
+```dart
+TextField(
+  inputFormatters: <TextInputFormatter>[
+    LengthLimitingTextInputFormatter(
+      6,
+      composingMaxLengthEnforced: true, // <-- Enforce truncate
+    ),
+  ],
+)
```
pedromassango, line 45: Why add a formatter if we can just specify the enforcement type? I think it is a good time to promote the newly added `MaxLengthEnforcement` API.

```dart
TextField(
    maxLength: 6,
    maxLengthEnforcement = MaxLengthEnforcement.truncateComposing,
)
```

AlexV525: Document is not following up the progress of the PR (It's a draft). 😄

AlexV525: (Only three hours since the new API has commited)

pedromassango: Haa, OK. 
Will take a nother look once you're done.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Summary
+
+_If you didn't use `maxLengthEnforced` in `TextField`/`TextFormField`/`CupertinoTextField`, or you don't care about the truncate behavior for the composing text editing value, you can stop reading._
+
+To control the behavior of `maxLength` in the `LengthLimitingTextInputFormatter`, use `maxLengthEnforcement` instead of `maxLengthEnforced`, and `maxLengthEnforced` is deprecated.
+
+## Context
+
+Before [the first PR][Fix TextField crashed with composing and maxLength set], `LengthLimitingTextInputFormatter` will always truncate the editing value, which may leads to engine crashes when the composing range is overflowed sometimes. With this changes, a composing `TextEditingValue` will not be truncated if only `maxLength` set, and still be truncated after the composing ends (See [`maxLength` documentation][] for more specific behavior). But it's not optional.
```
sfshaza2, line 14: I don't think the square brackets work the way you think. The text in the second set of square brackets is not displayed. It serves only as a shortcut. So:

```
Before [PR 63754][]: Fix TextField crashed with composing and maxLength set,
the `LengthLimitingTextInputFormatter` always truncated the editing value,
which resulted in engine crashes when the composing range sometimes overflowed.
With this change, an editable `TextEditingValue` is not truncated if only `maxLength` is set,
but _is_ truncated after editing is complete. (For more specific behavior, see the
[`maxLength` documentation][]. 

[PR 63754]:  {{site.github.com}}//flutter/flutter/pull/63754
```

I deleted the "But it's not optional" line because I don't understand. What isn't optional?

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
...
+
+With [the second PR][Introduce `MaxLengthEnforcement`], all behaviors of max length limiting will be controlled by passing `maxLenghLimitingEnforcement` to the `LengthLimitingTextInputFormatter`. It's also exposed to `TextField`/`TextFormField`/`CupertinoTextField`. Since then, `maxLengthEnforced` is deprecated. (See also: [`MaxLengthEnforcement` documentation][`MaxLengthEnforcement`] .)
```
sfshaza2, line 16: =>
```
With [PR 68086][]: Introduce `MaxLengthEnforcement`,
the truncation behavior is now controlled by passing
`maxLengthLimitingEnforcement` to the `LengthLimitingTextInputFormatter`,
which is then exposed to the `TextField`, `TextFormField`, and `CupertinoTextField` widgets.
The `maxLengthEnforced` property is now deprecated.
(For more information, see the docs for[`MaxLengthEnforcement`] .)

[PR 68086]: {{site.github}}/flutter/flutter/pull/68086
```

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
...
+)
+```
+
+Code after migration:
+
+<!-- skip -->
+```dart
+TextField(
+  maxLength: 6,
+  maxLengthEnforcement: MaxLengthEnforcement.truncateComposing, // <-- Enforce truncate
```
sfshaza2, line 70: => Enforces truncation

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
...
+
+## References
+
+API documentation:
+* [`MaxLengthEnforcement`][]
+* [`LengthLimitingTextInputFormatter`][]
+
+Relevant PR:
+* [Fix TextField crashed with composing and maxLength set][]
+* [Introduce `MaxLengthEnforcement`][]
```
sfshaza2, line 87: Replace both of these lines:

```
* [PR 63754][]: Fix TextField crashed with composing and maxLength set
* [PR 68086][]: Introduce `MaxLengthEnforcement`
```

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
...
+
+[`MaxLengthEnforcement`]: {{site.api}}/flutter/services/MaxLengthEnforcement-class.html
+
+[`LengthLimitingTextInputFormatter`]: {{site.api}}/flutter/services/LengthLimitingTextInputFormatter-class.html
+
+[`maxLength` documentation]: {{site.api}}/flutter/services/LengthLimitingTextInputFormatter/maxLength.html
+
+[Fix TextField crashed with composing and maxLength set]: {{site.github}}/flutter/flutter/pull/63754
+
+[Introduce `MaxLengthEnforcement`]: {{site.github}}/flutter/flutter/pull/68086
```
sfshaza2, line 97: Delete this line.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
...
+* [Fix TextField crashed with composing and maxLength set][]
+* [Introduce `MaxLengthEnforcement`][]
+
+[`MaxLengthEnforcement`]: {{site.api}}/flutter/services/MaxLengthEnforcement-class.html
+
+[`LengthLimitingTextInputFormatter`]: {{site.api}}/flutter/services/LengthLimitingTextInputFormatter-class.html
+
+[`maxLength` documentation]: {{site.api}}/flutter/services/LengthLimitingTextInputFormatter/maxLength.html
+
+[Fix TextField crashed with composing and maxLength set]: {{site.github}}/flutter/flutter/pull/63754
```
sfshaza2, line 95: Delete this line.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Description of change
+
+* Add `MaxLenghLimitingEnforcement` enum, which controlls all max length limiting behavious.
+* `maxLengthEnforced` in text fields is deprecated.
+* A composing `TextEditingValue` will skip the `maxLength` limit by default.
+* A named optional parameter `maxLenghLimitingEnforcement` has been added to the `TextField`/`TextFormField`/`CupertinoTextField`/`LengthLimitingTextInputFormatter`, defaults to `MaxLengthLimitingEnforcement.regular`.
+
+## Migration guide
+
```
sfshaza2, line 26: Add a line of text here. For example:

```
The following tips describe how to emit a warning, but not truncate,
on entering text, and how to continue with the former behavior
of truncating as text is entered.
```

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
...
+### Not enforces limitation
+
+If you want a max length warning, but not truncate when it's exceed the limitation, use `MaxLengthEnforcement.warningOnly` instead of `maxLengthEnforced: false`.
```
sfshaza2, line 29:
```
To print a max length warning, but _not_ truncate when the limit is exceeded,
use `MaxLengthEnforcement.warningOnly`, instead of `maxLengthEnforced: false`.

```

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
...
+```dart
+TextField(
+  maxLength: 6,
+  maxLengthEnforcement: MaxLengthEnforcement.warningOnly,
+)
+```
+
+### Limit composing
+
+If you still want to limit all kinds of inputting value (such as "verification code"), you should use `MaxLengthEnforcement.truncateComposing` from now.
```
sfshaza2, line 53:
```
To truncate text as it's entered (for example, "verification code"),
use `MaxLengthEnforcement.truncateComposing`:
```

To be honest, I don't get why verification code is truncated...

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Context
+
+Before [the first PR][Fix TextField crashed with composing and maxLength set], `LengthLimitingTextInputFormatter` will always truncate the editing value, which may leads to engine crashes when the composing range is overflowed sometimes. With this changes, a composing `TextEditingValue` will not be truncated if only `maxLength` set, and still be truncated after the composing ends (See [`maxLength` documentation][] for more specific behavior). But it's not optional.
+
+With [the second PR][Introduce `MaxLengthEnforcement`], all behaviors of max length limiting will be controlled by passing `maxLenghLimitingEnforcement` to the `LengthLimitingTextInputFormatter`. It's also exposed to `TextField`/`TextFormField`/`CupertinoTextField`. Since then, `maxLengthEnforced` is deprecated. (See also: [`MaxLengthEnforcement` documentation][`MaxLengthEnforcement`] .)
+
+## Description of change
+
+* Add `MaxLenghLimitingEnforcement` enum, which controlls all max length limiting behavious.
```
sfshaza2, line 20:
```
* Added a `MaxLengthLimitingEnforcement` enum,
  which controls all max length limiting behavior.
```

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
...
+* `maxLengthEnforced` in text fields is deprecated.
```
sfshaza2, line 21:
```
* The `maxLengthEnforced` property in editable text field classes is deprecated.

```

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
...
+* A composing `TextEditingValue` will skip the `maxLength` limit by default.
```
sfshaza2, line 22: ?? Not quite sure what this means.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
...
+* A named optional parameter `maxLenghLimitingEnforcement` has been added to the `TextField`/`TextFormField`/`CupertinoTextField`/`LengthLimitingTextInputFormatter`, defaults to `MaxLengthLimitingEnforcement.regular`.
```
sfshaza2, line 23:
```
* A named optional parameter, `maxLengthLimitingEnforcement`,
  is added to the `TextField`, `TextFormField`, `CupertinoTextField`,
  and `LengthLimitingTextInputFormatter` classes,
  which defaults to `MaxLengthLimitingEnforcement.regular`.
```

sfshaza2 (changes requested):
Some feedback for you.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
+---
+title: Use maxLengthEnforcement instead of maxLengthEnforced
+description: Introduing MaxLengthEnforcement enum.
+---
+
+## Summary
+
+_If you didn't use `maxLengthEnforced` in `TextField`/`TextFormField`/`CupertinoTextField`, or you don't care about the truncate behavior for the composing text editing value, you can stop reading._
```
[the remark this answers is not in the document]
pedromassango: I think it is important to mention that the default value of `maxLengthEnforced` is true so a lot of people uses it implicitly by just setting the `maxLength` property. 

May be it is worth mentioning the `maxLenght` parameter as well (or even mention it in place of `maxLengthEnforced` since it is implicitly used by default).

Cc @justinmc @LongCatIsLooong @chunhtai (you might know better about what I'm trying to say).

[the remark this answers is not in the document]
justinmc: Yeah I think it's worth at least saying "the `maxLength` or `maxLengthEnforced` properties".

Users of `maxLength` will not be broken if they don't also use `maxLengthEnforced`, but it's good to know anyways.  They may want to change to using `MaxLengthEnforcement.allowComposingTextToFinish`.

AlexV525:
@sfshaza2 The PR is still changing and not stabled yet so I draft this. Sorry for the early review request 😢

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,97 @@
...
+## Timeline
+
+The first change was made in August of 2020 after the 1.22.0-1.0.pre release.
+The second change was made in October of 2020 after the v1.23.0-18.0.pre release.
+
+## References
+
+API documentation:
+* [`MaxLengthEnforcement`][]
+* [`LengthLimitingTextInputFormatter`][]
```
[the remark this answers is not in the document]
AlexV525: https://api.flutter.dev/flutter/services/LengthLimitingTextInputFormatter-class.html

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
AlexV525: > I deleted the "But it's not optional" line because I don't understand. What isn't optional?

@sfshaza2 The behavior, allow composing exempt, is not optional. It provides as `allowComposingTextToFinish` enforcement now.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
AlexV525: 81 for this...

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
AlexV525: Removed

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,139 @@
...
+  limiting behavior.
+* The `maxLengthEnforced` property in editable text field classes is deprecated.
+* A named optional parameter, `maxLenghLimitingEnforcement` is added to
+  the `TextField`, `TextFormField`, `CupertinoTextField`,
+  and `LengthLimitingTextInputFormatter` classes, which defaults to
+  `MaxLengthLimitingEnforcement.enforced`.
+
+## Migration guide
+
+The following tips describe how to emit a warning, but not truncate,
```
AlexV525, line 49: @sfshaza2 Actually warning can only show with a `TextField`, should I add `(TextField only)` after "how to emit..." ?

> The following tips describe how to emit a warning (`TextField` only), but not truncate,

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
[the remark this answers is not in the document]
sfshaza2: @AlexV525, It should be reflected on the master branch.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
justinmc: Maybe you could just say "The following tips describe how to produce common behaviors." and let the sections below elaborate.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,140 @@
...
+on entering text, and how to truncate as text is entered.
+
+### Not enforces limitation
```
justinmc, line 52: Maybe: "To not enforce the limitation"

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,140 @@
...
+
+To print a max length warning in `TextField`, but _not_ truncate when the limit
```
justinmc, line 54: Saying "print" and "warning" makes me think this is something that will be logged on the console.  Maybe say "To show a max length error message"?

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,140 @@
...
+
+<!-- skip -->
+```dart
+TextField(
+  maxLength: 6,
+  maxLengthEnforcement: MaxLengthEnforcement.none,
+)
+```
+
+For `CupertinoTextField`, just don't set the `maxLength` value.
```
justinmc, line 78: Maybe explain that CupertinoTextField doesn't show error messages?

"For `CupertinoTextField`, which isn't able to show error messages, just don't set the `maxLength` value."

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,140 @@
...
+Code after migration:
+
+<!-- skip -->
+```dart
+CupertinoTextField(
+  // Leave it not set.
+)
+```
+
+### Unenforce composing
```
justinmc, line 99: "To enforce the limit, but not for composing text"

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,140 @@
...
+Code for the implementation:
+
+<!-- skip -->
+```dart
+TextField(
+  maxLength: 6,
+  maxLengthEnforcement: MaxLengthEnforcement.allowComposingTextToFinish, // <-- Temporarily lift the limit
+)
+```
+
```
justinmc, line 115: Should there be a section for MaxLengthBehavior.enforced, too?

justinmc (commented):
A few more small improvements I saw.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
AlexV525: Hmm, from your point I think "message" is from console too, maybe To show a max length error in TextField?

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
AlexV525: Done

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
justinmc: True, that sounds good to me.

justinmc (approved):
LGTM from a technical point of view.  The migration guide looks accurate and up-to-date with the new plan for the PR.

AlexV525:
[`MaxLengthEnforcement` doc](https://master-api.flutter.dev/flutter/services/MaxLengthEnforcement-class.html) is exist on master-api now.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,192 @@
...
+* The `maxLengthEnforced` property in editable text field classes is deprecated.
+* A named optional parameter, `maxLengthEnforcement` is added to
+  the `TextField`, `TextFormField`, `CupertinoTextField`,
+  and `LengthLimitingTextInputFormatter` classes.
+* The default behavior for different platforms is updated, including Android,
+  iOS, Web, Windows, MacOS, Linux, and Fuchsia.
+
+## Migration guide
+
+The following tips describe how to produce common behaviors, and which behavior
```
LongCatIsLooong, line 50: Maybe replace this paragraph with one that describes how every value in the deprecated parameter map to the new parameter (maybe with a table?), and when should the new `truncateAfterCompositionEnds` value be used, for people that just want to move away from the deprecated parameter.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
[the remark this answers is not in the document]
sfshaza2: It won't show up in the api.flutter.dev docs until our next stable release.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,192 @@
...
+The `maxLengthEnforced` property is now deprecated.
+(For more information, see the docs for [`maxLength`][] and
+[`MaxLengthEnforcement`][].)
+
+The default max length truncation behavior is based on the `TargetPlatform`
+that you're using or running on.
+
+## Description of change
+
+* Added a `MaxLengthEnforcement` enum, which controls all max length
```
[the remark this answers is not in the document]
AlexV525: Like this?

```
* Added a `MaxLengthEnforcement` enum as a replacement for the now-deprecated
  boolean `maxLengthEnforced` parameter on `TextField`, `TextFormField`,
  and `CupertinoTextField`.
```

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,192 @@
...
+  limiting behavior.
+* The `maxLengthEnforced` property in editable text field classes is deprecated.
+* A named optional parameter, `maxLengthEnforcement` is added to
+  the `TextField`, `TextFormField`, `CupertinoTextField`,
+  and `LengthLimitingTextInputFormatter` classes.
+* The default behavior for different platforms is updated, including Android,
```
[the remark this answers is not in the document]
AlexV525: Should we raise the `Platforms behavior` part to here, or just add a short description like you suggested?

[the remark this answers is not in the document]
LongCatIsLooong: Just thought it would be useful to provide the defaults since it's brought up here, otherwise this line will probably leave people wondering what the defaults are.

AlexV525:
Can I add some screenshots to explain more details about text fields behavior?

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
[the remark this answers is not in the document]
AlexV525: I'm experiencing watching these long parameters at everywhere 🤣

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
AlexV525: There might be too many variable for "when should use which parameter", such as "CJK or not", "Gboard or not", "Which platform is"...

AlexV525: But I think we should recommend users to use platforms default behavior in order to avoid breaking someone that is not propose to input like the behavior that the developer has used.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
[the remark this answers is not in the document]
AlexV525: All API url references are now master-api.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,208 @@
...
+
+_Use the default behavior for the current platform is recommended, since it
+won't break someone that is not surposed to input like the behavior that
+developers has defined._
+
+### Platforms behavior
+
+* Android, Windows: `MaxLengthEnforcement.enforced`. The native behavior of
+  these platforms is enforced. The composing will be handled by the IME while
+  users are entering CJK characters.
```
LongCatIsLooong, line 75: "The composing will be handled by the IME while users are entering CJK characters." doesn't seem true for windows? I tried the default pinyin IME on windows 10 and it put the composing text into the text field.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,208 @@
...
+* iOS: `MaxLengthEnforcement.truncateAfterCompositionEnds`. iOS has no default
+  behavior and it requires users implement the behavior themselves. Allow the
+  composition to exceed to avoid breaking CJK input.
+* Web, MacOS, Linux, Fuchsia: `MaxLengthEnforcement.truncateAfterCompositionEnds`.
+  These platforms allow the composition to exceed by default.
```
LongCatIsLooong, line 80: I'm a bit curious about macOS and Fuchsia. I'd expect macOS text fields to not have a `maxLength` attribute so it falls into the same category as iOS (but I'm not sure).

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,208 @@
...
+The following tips describe how to produce common behaviors, and which behavior
+is using on different platforms by default. Notice that the enforcement API is
+exposed to `TextField`, `TextFormField`, `CupertinoTextField` and
+`LengthLimitingTextInputFormatter`.
+
+_Use the default behavior for the current platform is recommended, since it
+won't break someone that is not surposed to input like the behavior that
+developers has defined._
+
+### Platforms behavior
```
LongCatIsLooong, line 71: maybe: ### Default values of `maxLengthEnforcement`

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,208 @@
...
+  behavior and it requires users implement the behavior themselves. Allow the
+  composition to exceed to avoid breaking CJK input.
+* Web, MacOS, Linux, Fuchsia: `MaxLengthEnforcement.truncateAfterCompositionEnds`.
+  These platforms allow the composition to exceed by default.
+
+### To enforce the limit all the time
+
+To enforce the limit that always truncate the value when it reaches the limit,
+use `MaxLengthEnforcement.enforced` in editable text fields.
+
```
LongCatIsLooong, line 86: nit: as you previously mentioned, maybe add that this does not work well if you expect general user input in the text field and the user tries to enter CJK characters with a phonetic keyboard, to raise awareness.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,208 @@
...
+TextField(
+  maxLength: 6,
+  maxLengthEnforcement: MaxLengthEnforcement.enforced,
+)
+```
+
+### To not enforce the limitation
+
+To show a max length error in `TextField`, but _not_ truncate when the limit
+is exceed, use `MaxLengthEnforcement.none` instead of
```
LongCatIsLooong, line 117: is exceeded

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,208 @@
...
+)
+```
+
+### To enforce the limit, but not for composing text
+
+To avoid truncating text while the user is editing, specify
+`MaxLengthEnforcement.truncateAfterCompositionEnds`. This behavior allows some
+Chinese, Japanese, and Korean (CJK) characters to temporarily ignore the limit
+until editing is complete.
+
```
LongCatIsLooong, line 168: I think the Gboard Android case is worth mentioning, It is going to be a common pitfall.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,208 @@
...
+<!-- skip -->
+```dart
+CupertinoTextField(
+  // Leave it not set.
+)
+```
+
+### To enforce the limit, but not for composing text
+
+To avoid truncating text while the user is editing, specify
```
LongCatIsLooong, line 164: editing -> inputting text via composition

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,208 @@
...
+`maxLengthEnforced` used to decide whether text fields should truncate
+the inputting value when it reached the `maxLength` limit. Additionally,
+`TextField` and `TextFormField` will display a warning message in the
+characters counter.
+
+Although `maxLengthEnforced` allows users to control the behavior of the
+truncation, so does the `LengthLimitingTextInputFormatter`, they didn't handle
+the truncate behavior for CJK (Chinese, Japanese, and Korean) characters
+properly, while these characters require users to type a sequence of Latin
+characters to enter one such character, which will cause composing CJK
```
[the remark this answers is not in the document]
LongCatIsLooong: also mention that this is called composing text/text composition so we can use these terms later.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
AlexV525: Got it

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,208 @@
...
+
+To control the behavior of `maxLength`
+in the `LengthLimitingTextInputFormatter`,
+use `maxLengthEnforcement` instead of `maxLengthEnforced`, which was deprecated.
+
+## Context
+
+`maxLengthEnforced` used to decide whether text fields should truncate
+the inputting value when it reached the `maxLength` limit. Additionally,
+`TextField` and `TextFormField` will display a warning message in the
```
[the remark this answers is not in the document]
AlexV525: I think most of people don't know anything about `maxLengthEnforced` before, so there would be a short description about it, so does `TextField`'s warning.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
AlexV525: > Makes sense to me. But I feel we shouldn't "recommend" the default behavior since we want developers to pick the right behavior for their use cases.

We make different platforms have their default behavior because we want the behavior is the same as the original native platform, aren't we? So why we shouldn't recommend it? IMO the goal is developers don't need to choose which behavior they should use, but satisfied with the default behavior.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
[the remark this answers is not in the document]
AlexV525: I think I'll add a short description like `The text composition usually displays with an underline decoration, or with a colored background`

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
AlexV525: Although this is the migration guide, which means people who's using the `maxLengthEnforced` might read this guide, there will be more people that may found different behaviors and come here to see the guide.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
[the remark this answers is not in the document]
AlexV525: > (on a phonetic keyboard).

> 这句是什么意思？

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
AlexV525: That is same as the Gboard on Android, like we discussed before? https://github.com/flutter/flutter/pull/68086#discussion_r538741280

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
AlexV525: * MacOS: Yes it's the same as iOS because maybe they both require developers handle it manually.
* Fuchsia: Because it's flutter native, I simply make the definition to make it as default to avoid break CJKs.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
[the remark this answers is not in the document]
LongCatIsLooong: https://en.wikipedia.org/wiki/Phonetic_keyboard_layout. Basically it's not going to be an issue if the user uses handwriting input.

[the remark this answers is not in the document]
AlexV525: Oh that's interesting. Actually the handwriting keyboard can produce composing text too.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
AlexV525: Same as above, I think the handwriting keyboard might be broken too.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,211 @@
...
+To control the behavior of `maxLength`
+in the `LengthLimitingTextInputFormatter`,
+use `maxLengthEnforcement` instead of `maxLengthEnforced`, which was deprecated.
+
+## Context
+
+The `maxLengthEnforced` parameter was used to decide whether text fields should
+truncate the input value when it reaches the `maxLength` limit. Additionally,
+`TextField` and `TextFormField` will display a warning message in the
+characters counter.
```
LongCatIsLooong, line 22: ... when this parameter is set to false and the length of the user input exceeds `maxLength`.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
```diff
@@ -0,0 +1,211 @@
...
+
+### To enforce the limit, but not for composing text
+
+To avoid truncating text while the user is inputting text via composition,
+specify `MaxLengthEnforcement.truncateAfterCompositionEnds`. This behavior
+allows some Chinese, Japanese, and Korean (CJK) characters to temporarily
+ignore the limit until editing is complete.
+
+_Gboard on Android produces the composing text when the user is entering Latin
+character/words by default._
```
LongCatIsLooong, line 170: "Gboard's English keyboard on Android (and many other Android input methods) creates a composing region
for the current English word being entered. When used in a `truncateAfterCompositionEnds` text field, the
user may not be stopped right away at the `maxLength` limit. Consider the `enforced` option if the text field 
does not expect characters that may require multistage text input on a phonetic keyboard (e.g., CJK characters)."

LongCatIsLooong: I'm not sure CJK scripts are the only ones that require multiple key presses to enter one character. I'd like to avoid using "multistage text input" since it's not very clear but can't find a better way to put it.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
AlexV525: Might be better:
> The inputting value will be truncated whether the user is entering with composition or not.

src/docs/release/breaking-changes/use-maxLengthEnforcement-instead-of-maxLengthEnforced.md
AlexV525: Haven't found other inputs that related to multistage text input besides CJKs. I think we can use "CJK characters" directly because we use this sentence through the whole guide?

AlexV525: > Consider the `enforced` option if the text field does not expect CJK characters on a phonetic keyboard.

AlexV525:
@sfshaza2 Could you take another look?

LongCatIsLooong (approved):
LGTM. Sorry for the delay.

PR #4888, opened 2020-10-16
Updating travis, submodules

RedBrogdon:
* Shards the travis configuration into eight jobs. The jekyll build and code check are each run on these four Flutter versions:
  - Whatever commit the submodule is checked in with
  - dev
  - beta
  - stable
* Stable is not allowed to fail, but the other versions are.
* Updates the Adopt-a-Widget page to remove some whitespace that was causing tests to fail.
* Updates the Flutter submodule to the latest stable commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,33 +32,39 @@ jobs:
 
     - name: "Dev channel: code check"
       env:
-        - TASK="flutter/bin/flutter channel dev && ./tool/check-code.sh"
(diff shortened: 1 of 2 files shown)


sfshaza2 (approved):
I'm taking this on faith and because the build seems to be happy. LGTM

sfshaza2:
I take it back. Build is not happy and it's a WIP anyway.  :)

PR #4905, opened 2020-10-21
Fix image size in narrow width window

haruelico:
These images protruded out of the area in a narrow width window.

This is just one example.

| before | after |
| --- | --- |
| Screenshot from 2020-10-22 01-58-35 | Screenshot from 2020-10-22 01-58-45 |


haruelico:
@googlebot I signed it!

sfshaza2 (approved):
I staged the site and checked all oof the affected images. A definite improvement! Thanks, @haruelico!  LGTM

PR #4922, opened 2020-10-27
Update memory.md

omatt:
Added missing `<dd>` closing tag. Removed `{:width="100%"}` on overly-stretched images.

NOTE: We love contributions, but we hate spam. If your PR doesn't improve flutter.dev, we'll close it and label it as `invalid`.

PR #4599, opened 2020-09-10
Updated Memory Docs

terrylucas:
fixes https://github.com/flutter/devtools/issues/1563

PR #4925, opened 2020-10-28
Fixed closing tags

omatt:
I've missed some missing closing tags on my previous PR (#4922 ). I've only noticed the issue after being published. I wish there's a better way to check the doc's preview and see if there's any breaking changes.

The missing closing tags seems to originate from #4599


sfshaza2 (approved):
Thanks again, @omatt! Actually, there is a way to stage it. The instructions are in the README. It's pretty easy to stage it on localhost on your machine, it can get trickier if you want to create a stageable version that others can see (then you need a firebase instance.)
LGTM

omatt:
Thanks! I'll do that in the future.

PR #4927, opened 2020-10-28
Includes learn more link to eBay motors app

lucalves:
Hello everyone, everything good?

I thought it would be interesting to include the "learn more" link for the eBay motors app, given that they have a page just to talk about the app.

The content about the app can be seen at this [link](https://pages.ebay.com/motors/motorsapp)! 😉


sfshaza2 (approved):
LGTM! Thanks for that, @lucalves! (And everything is ok here. I hope it's the same for you!)

PR #4932, opened 2020-10-29
Updated names from Xcode settings

ksilz:
*  It's "Build Settings, not "Deployment Info".
* It's "iOS Deployment Target", not just "Deployment Target".

Here's a screenshot from Xcode 12.1:

image


sfshaza2 (approved):
Thanks, again, for this update to Xcode 12! LGTM

PR #4936, opened 2020-10-29
Specify how to check if an iOS device uses Metal

liyuqian:
Helps situations like https://github.com/flutter/flutter/issues/60313#issuecomment-718327995


sfshaza2 (approved):
LGTM

Issue #4920, opened 2020-10-26
'Display a snackbar' page requires an update


ramanathanp:
[Display a snackbar](https://flutter.dev/docs/cookbook/design/snackbars) page needs to be updated because `Scaffold.of(context).showSnackBar()` method is deprecated and instead `ScaffoldMessenger.of(context).showSnackBar()` must be used.

P.S. - Thanks to Flutter team for maintaining a wonderful UI framework.

piedcipher:
I think this one is fixed. Good to close?

parlough:
Fixed by https://github.com/flutter/website/pull/5386. Thanks again for reporting this!

PR #4937, opened 2020-10-30
Scaffold changed to ScaffoldMessenger to show snackbar

yash1200:
Fixes https://github.com/flutter/website/issues/4920

*  Changed Scaffold to ScaffoldMessenger to show snackbar

### DIFF
--- a/src/docs/cookbook/design/snackbars.md
+++ b/src/docs/cookbook/design/snackbars.md
@@ -56,7 +56,7 @@ First, create a `SnackBar`, then display it using the `Scaffold`.
 final snackBar = SnackBar(content: Text('Yay! A SnackBar!'));
 
 // Find the Scaffold in the widget tree and use it to show a SnackBar.
-Scaffold.of(context).showSnackBar(snackBar);
+ScaffoldMessenger.of(context).showSnackBar(snackBar);
 ```
 
 ## 3. Provide an optional action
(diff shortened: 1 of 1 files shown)


sfshaza2:
Hi, @yash1200! Thanks for your update, but ScaffoldMessenger is not yet available in the stable channel (we test the code in flutter.dev against stable). I do see it on the master channel.

So, we either wait to update this page OR, we offer both solutions, mentioning which channel supports it *and* putting a `<!--skip-->1` on the code snippet that shows the new code. Usually, we wait until the change hits stable....

sfshaza2:
(And, by the way, that is why the Travis build is failing, because it can't find the ScaffoldMesenger class in stable.)

yash1200:
Then we should wait for this feature to land in stable version and then update it.

src/docs/cookbook/design/snackbars.md
```diff
@@ -56,7 +56,7 @@ First, create a `SnackBar`, then display it using the `Scaffold`.
 final snackBar = SnackBar(content: Text('Yay! A SnackBar!'));
 
 // Find the Scaffold in the widget tree and use it to show a SnackBar.
```
Piinks, line 58: For example here and elsewhere this reference needs to be updated.
```suggestion
// Find the ScaffoldMessenger in the widget tree and use it to show a SnackBar.
```

It would also be a good idea to mention the ScaffoldMessenger, what it is, and that the one we are calling on here is the default ScaffoldMessenger provided by the MaterialApp.

Piinks (commented):
Thank you @yash1200 for updating this, and welcome! 🎉 

When ScaffoldMessenger does land on stable and it's time to update this, we should also update the documentation on this page, not just the invocations.

sfshaza2 (commented):
Thanks for your PR, @yash1200, but we fixed this when preparing for the Flutter 2 release.  Closing.

PR #4943, opened 2020-11-02
Fixed typos and repeated words. 

jimmycasey:

* Fixed typos and repeated words. 
* Added spellcheck-cli and created a dictionary of whitelisted words. Could remove these updates and just submit the fixes on there own if people would rather not add this to repo.


jimmycasey:
@googlebot I fixed it.

sfshaza2 (commented):
Hi, @jimmycasey! Well, this is mighty tempting.... But is this a filtered set of edits? I ask because much of markdown or DartPad/image/video embeds/etc would trigger errors in a conventional spell check.

jimmycasey:
Hi @sfshaza2 ,
Ya, It was easier to pipe all the errors to the dictionary.txt and then review that file for valid mistakes and remove them and then run it again and actually fix the relevant file. I'm sure there is still valid typos that should be removed from dictionary.txt.

src/docs/get-started/install/_android-setup.md
```diff
@@ -52,7 +52,7 @@ follow these steps:
  1. Choose a device definition and select **Next**.
  1. Select one or more system images for the Android versions you want
     to emulate, and select **Next**.
-    An _x86_ or _x86\_64_ image is recommended.
+    A _x86_ or _x86\_64_ image is recommended.
```
micpap25, line 55: This is nit-picky, I know, but I think it would make more sense to keep this as "An _x86...". Just because of how it's actually pronounced, saying "A ex-eighty-six" doesn't make sense while "An ex-eighty-six" does.

sfshaza2: @micpap25 is quite right. It should be "An", not "A".

@ianloic, could you remove the spellcheck-cli from this PR? I would very much like to merge the fixed errors.  :)

sfshaza2 (changes requested):
See inline request.

src/docs/get-started/install/_android-setup.md
jimmycasey: Will sort this within the week of that's ok. Just moved job and waiting for new laptop to be posted out 😃

sfshaza2: That sounds good, @jimmycasey. And congrats on the new job!

sfshaza2:
Fixed the typos and misspellings in PR https://github.com/flutter/website/pull/5273

PR #4949, opened 2020-11-03
Updated Flutter platform view demo for iOS Swift

omatt:
**FLNativeViewFactory.swift**
Calling `super.init()` before `_view = UIView()` causes errors, swapped.
Added demo on how [UIView](https://developer.apple.com/documentation/uikit/uiview) can be modified inside `view()` function.

**AppDelegate.swift**
Added unwrap on `registrar! `, and changed function call from `messenger` to `messenger()`
`GeneratedPluginRegistrant.register(with: self)` expected argument is `with` instead of `withRegistry`

**FLPlugin.swift**
Changed function call from `registrar.messenger` to `registrar.messenger()`

Expected output when demo is run on iOS:

Screen Shot 2020-11-05 at 12 20 55 AM

Fixes [#67393](https://github.com/flutter/flutter/issues/67393).

Proposed changes mentioned above. Copy-pasting this iOS Swift sample should now run without issues on Xcode.

### DIFF
--- a/src/docs/development/platform-integration/platform-views.md
+++ b/src/docs/development/platform-integration/platform-views.md
@@ -445,12 +445,12 @@ import UIKit
 
 class FLNativeViewFactory: NSObject, FlutterPlatformViewFactory {
     private var messenger: FlutterBinaryMessenger
-    
+
     init(messenger: FlutterBinaryMessenger) {
         self.messenger = messenger
         super.init()
     }
-    
+
     func create(
         withFrame frame: CGRect,
         viewIdentifier viewId: Int64,
@@ -466,7 +466,7 @@ class FLNativeViewFactory: NSObject, FlutterPlatformViewFactory {
 
 class FLNativeView: NSObject, FlutterPlatformView {
     private var _view: UIView
-    
+
     init(
         frame: CGRect,
         viewIdentifier viewId: Int64,
@@ -476,18 +476,18 @@ class FLNativeView: NSObject, FlutterPlatformView {
         _view = UIView()
         super.init()
     }
-    
+
     func view() -> UIView {
         // This function returns a UIView and you can add native iOS UI here.
         _view.backgroundColor = UIColor.blue
-        
+
         let nativeLabel = UILabel()
         nativeLabel.text = "Native text from iOS"
         nativeLabel.textColor = UIColor.white
         nativeLabel.textAlignment = .center
         nativeLabel.frame = _view.frame
         _view.addSubview(nativeLabel)
-        
+
         return _view
     }
 }
--- a/src/docs/development/platform-integration/platform-views.md
+++ b/src/docs/development/platform-integration/platform-views.md
@@ -475,20 +475,22 @@ class FLNativeView: NSObject, FlutterPlatformView {
     ) {
         _view = UIView()
         super.init()
+        // iOS views can be created here
+        createNativeView(view: _view)
     }
 
     func view() -> UIView {
-        // This function returns a UIView and you can add native iOS UI here.
-        _view.backgroundColor = UIColor.blue
+        return _view
+    }
 
+    func createNativeView(view _view: UIView){
+        _view.backgroundColor = UIColor.blue
         let nativeLabel = UILabel()
         nativeLabel.text = "Native text from iOS"
         nativeLabel.textColor = UIColor.white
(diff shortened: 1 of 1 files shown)


src/docs/development/platform-integration/platform-views.md
```diff
@@ -444,13 +444,13 @@ import Flutter
 import UIKit
 
 class FLNativeViewFactory: NSObject, FlutterPlatformViewFactory {
-    private weak var messenger: FlutterBinaryMessenger
-
+    private var messenger: FlutterBinaryMessenger
+
```
blasten, line 448: nit: remove this and the trailing whitespaces below.

src/docs/development/platform-integration/platform-views.md
```diff
@@ -466,18 +466,28 @@ class FLNativeViewFactory: NSObject, FlutterPlatformViewFactory {
...
     func view() -> UIView {
+        // This function returns a UIView and you can add native iOS UI here.
+        _view.backgroundColor = UIColor.blue
+        
+        let nativeLabel = UILabel()
+        nativeLabel.text = "Native text from iOS"
+        nativeLabel.textColor = UIColor.white
+        nativeLabel.textAlignment = .center
+        nativeLabel.frame = _view.frame
+        _view.addSubview(nativeLabel)
```
blasten, line 489: we actually recommend to do this work in the constructor. is there a reason why the getter needs to do this work? cc @bparrishMines

src/docs/development/platform-integration/platform-views.md
omatt: Trailing whitespaces removed

src/docs/development/platform-integration/platform-views.md
omatt: I've tried creating views in the constructor, but for some reason I'm unable to fetch the `UIView.frame` where `UILabel` in the sample will be placed. The output here is that the `UILabel` won't be displayed, but the background color was able to change.

```swift
init(
    frame: CGRect,
    viewIdentifier viewId: Int64,
    arguments args: Any?,
    binaryMessenger messenger: FlutterBinaryMessenger?
) {
    _view = UIView()
    super.init()
    createNativeView(view: _view)
}
```

Moved creating views into a function

```swift
func createNativeView(view _view: UIView){
    _view.backgroundColor = UIColor.blue

    let nativeLabel = UILabel()
    nativeLabel.text = "Native text from iOS"
    nativeLabel.textColor = UIColor.white
    nativeLabel.textAlignment = .center
    nativeLabel.frame = _view.frame
    _view.addSubview(nativeLabel)
}
```

Any idea what might have been the issue? And any suggestions for best practices that we can add in the docs?

omatt: Thanks! I've moved the creation of views in `init()`

sfshaza2:
Can you review, @johnpryan?

johnpryan:
I'm not sure I'm the right person to review. Maybe @blasten could take a look?

blasten:
LGTM

@jmagman do you mind also reviewing the Swift code?

blasten (approved):

src/docs/development/platform-integration/platform-views.md
```diff
@@ -473,13 +473,25 @@ class FLNativeView: NSObject, FlutterPlatformView {
         arguments args: Any?,
         binaryMessenger messenger: FlutterBinaryMessenger?
     ) {
-        super.init()
         _view = UIView()
+        super.init()
```
blasten, line 477: doesn't this need to be called first?

jmagman:
> @jmagman do you mind also reviewing the Swift code?

I've never written any Swift code.  Maybe @xster?  
There's an internal Swift readability group, maybe we can ask someone there?

src/docs/development/platform-integration/platform-views.md
xster: either is fine but I'd just leave this first. We don't expect the super to access any overridden properties.

xster (approved):

src/docs/development/platform-integration/platform-views.md
omatt: Calling `super.init()` first throws this error in Xcode.

Screen Shot 2020-11-21 at 4 42 37 PM

PR #4954, opened 2020-11-04
Remove reference to flutter version command

### LINKED ISSUE #4935: 'Upgrading Flutter' page issue
Page URL: https://flutter.dev/docs/development/tools/sdk/upgrading.html
Page source: https://github.com/flutter/website/tree/master/src/docs/development/tools/sdk/upgrading.md

Description of issue: Tried the `flutter version` command and did not work. Plus, `flutter -h` seems to do not list it anymore.

micpap25:
Fixes #4935

*  Remove the section in the "Upgrading Flutter" page that discusses `flutter version`. The page already has a note that tells users how to properly switch to a specific version. Thus, there's no reason to continue presenting this outdated information.

### DIFF
--- a/src/docs/development/tools/sdk/upgrading.md
+++ b/src/docs/development/tools/sdk/upgrading.md
@@ -84,17 +84,6 @@ You can also ask questions on the [Flutter dev mailing list][flutter-dev].
 Aside from subscribing to receive announcements,
 we'd love to hear from you!
 
-
-## Selecting a specific version
-
-If you have a specific version of Flutter that you'd like to switch to,
-you can use the `flutter version` command:
-
-```terminal
-$ flutter version v1.9.1+hotfix.3
-```
-
-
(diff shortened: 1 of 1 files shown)


sfshaza2 (commented):
`flutter upgrade` upgrades to the latest, and `flutter channel` changes channels. But sometimes folks need a specific release, especially if they have a bug report for a specific release. Also, these instructions are at the end. So I would like to leave them as-is.

micpap25:
In that case, I think it makes sense to at least expand the section so it mentions that `flutter version` is deprecated and maybe explain cases where it is better to directly download the SDK for a different version rather than use `flutter version`. If those exist, of course.

sfshaza2:
What makes you think that `flutter --version` is deprecated?

sfshaza2:
D'OH. `flutter version` , not `flutter --version`. I have, quite literally, NEVER used `flutter version`.

sfshaza2 (approved):
LGTM

PR #4966, opened 2020-11-06
added flutter_command and GetIt

escamoteur:
Hi, 

I realized that my packages for state management weren't listed despite being quite popular. @filiph told me I should make a PR so that they get included in the list.

Cheers Thomas

### DIFF
--- a/src/docs/development/data-and-backend/state-mgmt/options.md
+++ b/src/docs/development/data-and-backend/state-mgmt/options.md
@@ -98,6 +98,14 @@ A family of stream/observable based patterns.
 * [Reactive Programming - Streams - BLoC - Practical Use Cases][],
   by Didier Boelens
 
+## GetIt
+
+A service locator based state management approach that doesn't need a `BuildContext`.
+
+* [get_it package][], The service locator. It can also be used together with BloCs.
+* [get_it_mixin package][], Mixin that completes `get_it` to a full state management solution.
+* [get_it_hooks package][], Same as the mixin in case you already use `flutter_hooks`.
+
 ## MobX
 
 A popular library based on observables and reactions.
@@ -105,6 +113,13 @@ A popular library based on observables and reactions.
 * [MobX.dart, Hassle free state-management for your Dart and Flutter apps][]
 * [Getting started with MobX.dart][]
 * [Flutter: State Management with Mobx][], a video by Paul Halliday
+
+## Flutter Commands
+
+Reactive state management that uses the Command Pattern and is based on `ValueNotifiers`. Best in combination with [get_it](#getit), but can be used with `provider` or other locators too.
+
+* [flutter_command package][] 
+* [rx_command package][], `Stream` based implementation.
 
 ## GetX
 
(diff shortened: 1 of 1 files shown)


sfshaza2 (commented):
Yes, @filiph is quite right! This LGTM, but I'll let Filip review it.

src/docs/development/data-and-backend/state-mgmt/options.md
```diff
@@ -106,6 +114,13 @@ A popular library based on observables and reactions.
 * [Getting started with MobX.dart][]
 * [Flutter: State Management with Mobx][], a video by Paul Halliday
 
+## Flutter Commands
+
+Reactive state management that uses the Command Pattern and is based on `ValueNotifiers`. Best in combination with [get_it](#getit), but can be used with `provider` or other locators too.
+
+* [flutter_command package][] 
+* [rx_command package][], `Stream` based implementation.
+
```
filiph, line 123: Since you're saying "Best in combination" anyway, I suggest collapsing this into GetIt, as one of the `* ...` list items. This way, we keep this list nice and short, and it makes the GetIt option a little less sparse.

I know the `flutter_command` package is also useful without GetIt, but that doesn't mean it automatically gets its own section. If you feel strongly about this, let me know — I'm open to discussion.

src/docs/development/data-and-backend/state-mgmt/options.md
```diff
@@ -98,6 +98,14 @@ A family of stream/observable based patterns.
 * [Reactive Programming - Streams - BLoC - Practical Use Cases][],
   by Didier Boelens
 
+## GetIt
+
+A service locator based state management approach that doesn't need a `BuildContext`.
+
+* [get_it package][], The service locator. It can also be used together with BloCs.
```
filiph, line 105: [nit] Consistent capitalization would help here. If the package's name is "GetIt", I'd go with "GetIt package" (similar to "Provider package" and "Flutter Redux package" elsewhere on this page). Conversely, if the preferred stylization is `get_it`, then please update the heading of this section.

This applies to the other bullet points as well.

filiph, line 105:
```suggestion
* [get_it package][], the service locator. It can also be used together with BloCs.
```

src/docs/development/data-and-backend/state-mgmt/options.md
```diff
@@ -98,6 +98,14 @@ A family of stream/observable based patterns.
...
+* [get_it_mixin package][], Mixin that completes `get_it` to a full state management solution.
```
filiph, line 106:
```suggestion
* [get_it_mixin package][], a mixin that completes `get_it` to a full state management solution
```

src/docs/development/data-and-backend/state-mgmt/options.md
```diff
@@ -98,6 +98,14 @@ A family of stream/observable based patterns.
...
+* [get_it_hooks package][], Same as the mixin in case you already use `flutter_hooks`.
```
filiph, line 107:
```suggestion
* [get_it_hooks package][], same as the mixin in case you already use `flutter_hooks`
```

filiph (changes requested):
Thanks! I have some stylistic nits and one conceptual suggestion.

src/docs/development/data-and-backend/state-mgmt/options.md
escamoteur: I think the Command pattern represents a very different approach than the others, therefore I think it's own section is justified.

src/docs/development/data-and-backend/state-mgmt/options.md
escamoteur: ok

sfshaza2:
Note to self. Leave this review to @filiph.

filiph (approved):
LGTM

PR #4973, opened 2020-11-09
Update errors.md

polina-c:
Fixes #70059.

*  Add example for customizing error widget.

### DIFF
--- a/src/docs/testing/errors.md
+++ b/src/docs/testing/errors.md
@@ -16,7 +16,8 @@ objects mentioned in the message.
 
 When an error occurs during the build phase, the [`ErrorWidget.builder`][]
 callback is invoked to build the widget that is used instead of the
-one that failed. By default, in debug mode this shows an error message
+one that failed.
+By default, in debug mode this shows an error message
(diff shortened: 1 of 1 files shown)


src/docs/testing/errors.md
```diff
@@ -52,6 +52,25 @@ This handler can also be used to report errors to a logging service.
 For more details, see our cookbook chapter for 
 [reporting errors to a service][].
 
+To customize error widget at root, set [`MaterialApp.builder`]:
```
sfshaza2, line 55: Hi @polina-c, nice to "meet" you and thanks for this PR! But I have some questions. When you say "to customize error widget at root", do you mean to set customize the error widget at the root of the widget tree so it is available to all widgets? Is something else going on? I need a bit more context.

Also, I'm not sure this bit belongs at the end of the page. If you want to talk offline, just drop me a message or an email..

sfshaza2 (changes requested):
See inline comment.

sfshaza2 (approved):
I will make a few tweaks in a followup PR. Thanks!

PR #5000, opened 2020-11-19
Update null-safety.md

kwalrath:
Staged: https://kw-flutter-1.firebaseapp.com/docs/null-safety


mit-mit:
Should we start with a line that says something like this?

Null safety is available in the Flutter beta channel, starting with build `1.24.0-10.2.pre`.

mit-mit (approved):

PR #5003, opened 2020-11-19
Add links to all the null safety docs

kwalrath:
Staged: https://kw-flutter-1.firebaseapp.com/docs/null-safety

/cc @timsneath @kevmoo @mit-mit

### DIFF
--- a/src/docs/null-safety.md
+++ b/src/docs/null-safety.md
@@ -24,11 +24,11 @@ check out these pages:
 * [Unsound null safety][]:
   Conceptual and practical information about mixed-mode programs,
   in which only some libraries are null safe.
-* [Null safety FAQ][]:
+* [FAQ][]:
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
LGTM, except, do we want to reinforce here that, at this time, only migrate packages?

kwalrath:
> LGTM, except, do we want to reinforce here that, at this time, only migrate packages?

Good point. I've tweaked the text to mention packages, but not specifically apps. Should we mention apps specifically, @timsneath / @kevmoo / @mit-mit?

timsneath:
Packages, not apps, for now. Follow the language in the beta blog post, if
in any doubt :)

Thanks!

kwalrath:
I think we're good for now. We can always tweak.

PR #5004, opened 2020-11-19
A few minor tweaks

sfshaza2:
@shihaohong, your last PR was a significant improvement. Can you remind me what you're adding next?

### DIFF
--- a/src/docs/development/accessibility-and-localization/internationalization.md
+++ b/src/docs/development/accessibility-and-localization/internationalization.md
@@ -13,7 +13,7 @@ description: How to internationalize your Flutter app.
 {{site.alert.end}}
 
 If your app might be deployed to users who speak another language then
-you'll need to internationalize it. That means you'll need to write
+you'll need to internationalize it. That means you need to write
 the app in a way that makes it possible to localize values like text
 and layouts for each language or locale that the app
 supports. Flutter provides widgets and classes that help with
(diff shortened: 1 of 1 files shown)


src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -46,12 +46,12 @@ This section provides a tutorial on how to internationalize
 a Flutter application, along with any additional setup that a
 target platform might require.
 
-### Setting up an internation&shy;alized app: the flutter<wbr>_localizations package {#setting-up}
+### Setting up an internation&shy;alized app: the Flutter<wbr>_localizations package {#setting-up}
```
shihaohong, line 49: `flutter_localizations` here is correct, since that's the package name: https://github.com/flutter/flutter/tree/master/packages/flutter_localizations

shihaohong (commented):
Changes look good! Just one nit.

As for further work, I've currently linked the Google doc that is the "more technical guide" at the bottom of the new "Adding your own localized messages" section. If that doc itself is sufficient and its format is fine, then there's nothing more to do. I originally considered adding a codelab for the more nitty gritty details like handling plurals or introducing parameters (basically, content between page 11-17 on [the user guide](https://docs.google.com/document/d/10e0saTfAv32OZLRmONy866vnaw0I2jwL8zukykpgWBc/edit#). However, that may not be necessary at this point and we could wait for more user feedback about the tool before we add additional content?

src/docs/development/accessibility-and-localization/internationalization.md
sfshaza2: Ah, then we code font it.

PR #5009, opened 2020-11-22
Update architectural-overview.md

MeandNi:
Fix typo in docs: bypassing -> by passing


sfshaza2 (commented):
Hi, @MeandMi! Thanks for your PR. Wow, one space can really change the meaning here, but actually I believe the original is correct. Would you like to confirm and close this PR, @timsneath?

timsneath (commented):
Bypassing is the correct word here. We don't transmit the UI to the system widgets, we replace the system widgets.

MeandNi:
Wow, I was wrong, thank you for your reply. @timsneath @sfshaza2

PR #5038, opened 2020-12-02
Desktop - specify support for adding to an existing Flutter project, not a native project

xster:
I got confused reading this and thought it was supporting adding to an existing desktop app. But it was referring to adding to an existing Flutter mobile app.


sfshaza2 (approved):
LGTM

PR #5039, opened 2020-12-03
Add sysctl to macOS system requirements

jmagman:
`sysctl` is required as of https://github.com/flutter/flutter/pull/67970

See also:
https://github.com/flutter/flutter/issues/71516

### DIFF
--- a/src/docs/get-started/install/macos.md
+++ b/src/docs/get-started/install/macos.md
@@ -23,6 +23,7 @@ your development environment must meet these minimum requirements:
   - `git` 2.x
   - `mkdir`
   - `rm`
+  - `sysctl`
   - `unzip`
   - `which`
 


src/docs/get-started/install/macos.md
```diff
@@ -23,6 +23,7 @@ your development environment must meet these minimum requirements:
   - `git` 2.x
   - `mkdir`
   - `rm`
```
xster, line 25: I wonder how useful this information is especially so early in the onboarding journey. For things that, unless something is wrong, should be on the default macOS 10, should we just show a longer message when they're missing rather than stating these needs ahead of time? i.e. what are we asking the user to do specifically? That they `which bash` etc for all of these? 

Maybe a check for all these should be just executed in flutter doctor "safe mode" that still runs even if the rest of the checks failed.

sfshaza2: I agree with this. This could easily stress folks out needlessly.

jmagman: On the tools team we find these lists useful to close bugs.  "unzip isn't on your PATH, here's the link to system requirements, closing" type thing.

I guess we only need commands to be in this list that are used by the bootstrapping scripts, not the flutter tool.
Like if `unzip` fails you can't even download the dart SDK, the tool can't compile and give a nicer error.

PR #4887, opened 2020-10-16
Add note for xcarchive command available in dev channel

jmagman:
https://github.com/flutter/flutter/pull/67598 has reached the dev channel.
Add a note to the [iOS deployment docs](https://flutter.dev/docs/deployment/ios#create-a-build-archive).

Part of https://github.com/flutter/website/issues/4852.

The inlined `site.alert.note` required `<ol>` to keep the numbering.

Issue #4852, opened 2020-10-09
Update iOS deployment docs with `flutter build ipa` command


jmagman:
https://github.com/flutter/flutter/pull/67598 and https://github.com/flutter/flutter/pull/67781 introduces a new `flutter build ipa` command.
Reference in the [iOS deployment docs](https://flutter.dev/docs/deployment/ios#create-a-build-archive).

1. Remove the `flutter build ios` instructions.
1. Replace 
> 1. Select Product > Archive to produce a build archive
> 1. In the sidebar of the Xcode Organizer window, select your iOS app, then select the build archive you just produced.

with something like:
> 1. Run `flutter build ipa`
> 2. Open the archive in`build/ios/archive` in Xcode by running `open -a Xcode build/ios/archive/YourApp.xcarchive` or locating it in Finder and opening it.

Keep the "Click the Validate App button." next step.

Add something about `flutter build ipa` with hints for how to populate the optional `--export-options-plist` file.

NTA-trongpq:
@jmagman Sorry if I disturb you. Does this command `flutter build ipa` run a clean build iOS (In Xcode, Product -> Clean build folder)?

jmagman:
> @jmagman Sorry if I disturb you. Does this command `flutter build ipa` run a clean build iOS (In Xcode, Product -> Clean build folder)?

No, neither does `flutter run` or `build`.  You can run `flutter clean`.

r-dev-limited:
how can get non signed archive....someone mentioned that build ipa does not sign...but thats not true.

jmagman:
> how can get non signed archive....someone mentioned that build ipa does not sign...but thats not true.

See https://github.com/flutter/flutter/issues/101765

PR #5040, opened 2020-12-03
Add note for flutter build ipa command

jmagman:
https://github.com/flutter/flutter/pull/67781 has reached the beta channel.
Add notes to the [iOS deployment docs](https://flutter.dev/docs/deployment/ios#create-a-build-archive).

Part of https://github.com/flutter/website/issues/4852, will follow up with more detailed docs once the command reaches stable and it doesn't have to be a caveat `note`.

The inlined `site.alert.note` required `<ol>` to keep the numbering.

Follow up to https://github.com/flutter/website/pull/4887 but `flutter build ipa` instead of `flutter build xcarchive`.

## Screenshot
Screen Shot 2020-12-02 at 6 25 09 PM

### DIFF
--- a/src/docs/deployment/ios.md
+++ b/src/docs/deployment/ios.md
@@ -193,16 +193,53 @@ In Xcode, configure the app version and build:
 
 Finally, create a build archive and upload it to App Store Connect:
 
-1. Select **Product > Archive** to produce a build archive.
-1. In the sidebar of the Xcode Organizer window, select your iOS app,
-   then select the build archive you just produced.
-1. Click the **Validate App** button. If any issues are reported,
-   address them and produce another build. You can reuse the same
-   build ID until you upload an archive.
-1. After the archive has been successfully validated, click
-   **Distribute App**. You can follow the status of your build in the
-   Activities tab of your app's details page on
-   [App Store Connect][appstoreconnect_login].
+<ol markdown="1">
+<li markdown="1">
+
+Select **Product > Archive** to produce a build archive.
+
+{{site.alert.note}}
+  On Flutter version 1.24.0-6.0 and later you can create an
+  archive by instead running `flutter build ipa`.
+  Then open `build/ios/archive/MyApp.xcarchive` in Xcode to
+  validate and distribute your app.
+  See `flutter build ipa -h` for available flags.
+{{site.alert.end}}
+
+</li>
+<li markdown="1">
+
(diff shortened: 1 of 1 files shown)


src/docs/deployment/ios.md
```diff
@@ -193,16 +193,53 @@ In Xcode, configure the app version and build:
...
-   Activities tab of your app's details page on
-   [App Store Connect][appstoreconnect_login].
+<ol markdown="1">
+<li markdown="1">
+
+Select **Product > Archive** to produce a build archive.
+
+{{site.alert.note}}
+  On Flutter version 1.24.0-6.0 and later you can create an
+  archive by instead running `flutter build ipa`.
```
jmagman, line 203: @xster Is this confusing since it produces an `xcarchvie`, not an `ipa`?  `flutter build xcarchive` is an alias of `flutter build ipa` so I can say either.

src/docs/deployment/ios.md
```diff
@@ -193,16 +193,53 @@ In Xcode, configure the app version and build:
...
+
+Click the **Validate App** button. If any issues are reported,
+address them and produce another build. You can reuse the same
+build ID until you upload an archive.
+
+</li>
+<li markdown="1">
+
+After the archive has been successfully validated, click
+**Distribute App**. You can follow the status of your build in the
```
xster, line 226: Add a note about making various appropriate selections in the subsequent wizard? (so users don't get confused that it doesn't just do it but presents them with choices)

xster (approved):

src/docs/deployment/ios.md
xster: I think ipa is fine since that's the intent. It just happens to have not built one since it doesn't have all the information to take it all the way to the end. 

I forgot the verbiage we put. As long as we print a thing at the end that says "I know you wanted to make an IPA but I can't make one since I don't know how you're gonna distribute it and I can't really ask those choices to you here in the CLI, so why don't you continue the last bit in the GUI by typing open ...", I think it's ok.

sfshaza2 (approved):
LGTM. I'll let you merge when @xster is satisfied.

sfshaza2:
It looks like @xster has approved. Merging.

src/docs/deployment/ios.md
jmagman: Sorry, lost track of this.  I'll follow up in another PR.

PR #5044, opened 2020-12-04
Correct Web FAQ

creativecreatorormaybenot:

*  Add Firefox mobile to supported web platforms.

The mobile Firefox browser runs Flutter web apps (at least on Android).

For example: https://app.simpleclub.com

### DIFF
--- a/src/docs/development/platform-integration/web.md
+++ b/src/docs/development/platform-integration/web.md
@@ -66,7 +66,7 @@ Flutter web apps can run on the following browsers:
 * Chrome (mobile & desktop)
 * Safari (mobile & desktop)
 * Edge (mobile & desktop)
-* Firefox (desktop)
+* Firefox (mobile & desktop)
 
 During development, Chrome (on macOS, Windows, and Linux) and Edge (on Windows)
 are supported as the default browsers for debugging your app. In order to test


sfshaza2:
@johnpryan, can you confirm?

johnpryan:
It may run, but that doesn't mean we intend to support it. cc: @mariamhas

creativecreatorormaybenot:
@johnpryan Then we should maybe update the wording. At the moment, it literally says:

>Flutter web apps can run on the following browsers:

johnpryan:
Yeah, that's a little confusing. It currently says:

```
### Which web browsers are supported by Flutter?
Flutter web apps can run on the following browsers:
```

But probably should say:

```
### Which web browsers are supported by Flutter?
Flutter web apps support the following browsers:
```

sfshaza2 (changes requested):
@creativecreatorormaybenot, would you mind making the change that @johnpryan suggests? Otherwise, I'll have to close this PR.  Thx!

creativecreatorormaybenot:
@sfshaza2 But is Firefox mobile supported or not?

I see no reason why it would not be 🤔 Especially on iOS, it just uses WebKit, like all other browsers.

sfshaza2:
@creativecreatorormaybenot, my understanding from the discussion is that it is not officially supported, though it may work.

creativecreatorormaybenot:
@sfshaza2 

From:

> It may run, but that doesn't mean we intend to support it. cc: @mariamhas

I concluded that we are still waiting for an update on that matter (from @mariamhas).

sfshaza2:
She doesn't monitor website issues, but @johnpryan speaks to her regularly. I'm assuming he's checked with her. @johnpryan, can you confirm?

creativecreatorormaybenot:
@johnpryan @sfshaza2 Any update on this? I think we have concluded that the way it is currently is slightly misleading.

I understand that Firefox mobile might not have active support (even though we do not know why that would be, especially on iOS). Then again, I do not want to undersell Flutter because of that 😃

johnpryan:
I'll see if I can get a concrete answer.

sfshaza2:
Where are we on this PR, @johnpryan?

johnpryan:
Right now we don't have automated testing set up for Firefox on mobile devices, so we can't guarantee support. It's likely that most of the features will work, but your milage may vary.

PR #5052, opened 2020-12-05
Documentation: update "Build and release a Linux app"

Arenukvern:

I would like to improve documentation as during my experience for Flutter snap publishing I stack for some time with problems and therefore would like to share these workarounds.

1. Replace push to upload to snapstore
```terminal
   $ snapcraft upload --release=<channel> <file>.snap
```
The "push" cause the following error:
```
DEPRECATED: The 'push' set of commands have been replaced with 'upload'.
See http://snapcraft.io/docs/deprecation-notices/dn11 for more information.
Usage: snapcraft push [OPTIONS] <snap-file>
Try 'snapcraft push -h' for help.
````

2. To run correctly (at least with Ubuntu) there is also a need to setup DBus interface within yaml file
See more at https://snapcraft.io/docs/dbus-interface
Without it, it was not possible to run snap from snapstore - by getting the following error:
`
Failed to register: GDBus.Error:org.freedesktop.DBus.Error.AccessDenied: Connection ":1.1881" is not allowed to own the service "org.foo.my_super_app" due to AppArmor policy
`

3. Added basic section for icon and desktop file - without it the app will not shown in menu. 

Sorry for my english.
Hope it helps

### DIFF
--- a/src/docs/deployment/linux.md
+++ b/src/docs/deployment/linux.md
@@ -143,13 +143,20 @@ confinement: strict
 base: core18
 grade: stable
 
+slots:
+  dbus-super-cool-app: # adjust accordingly to your app name
+    interface: dbus
+    bus: session
+    name: org.bar.super_cool_app # adjust accordingly to your app name and
+    
 apps:
   super-cool-app:
     command: super_cool_app
     extensions: [flutter-master] # Where "master" defines which Flutter channel to use for the build
     plugs:
     - network
-
+    slots:
+      - dbus-super-cool-app
 parts:
   super-cool-app:
     source: .
@@ -235,6 +242,37 @@ apps:
   when snaps are strictly confined. This Flutter snap needs
   access to the network.
 
+**DBus interface**
+
+: The dbus interface provides a way for snaps to 
+  communicate over DBus. The snap providing the DBus 
+  service declares a slot with the well-known DBus name 
+  and which bus it uses. Snaps wanting to communicate 
+  with the providing snap’s service declare a plug for 
+  the providing snap. Note that a snap declaration is 
+  needed for your snap to be delivered via the snap store 
+  and claim this well-known DBus name (simply upload the 
+  snap to the store and request a manual review and 
+  a reviewer will take a look).
+
+  When a providing snap is installed, snapd will 
+  generate security policy that will allow it to 
+  listen on the well-known DBus name on the specified 
+  bus. If the system bus is specified, snapd will also 
+  generate DBus bus policy that allows ‘root’ to own 
+  the name and any user to communicate with the 
+  service. Non-snap processes are allowed to 
+  communicate with the providing snap following 
+  traditional permissions checks. Other (consuming) 
+  snaps may only communicate with the providing 
+  snap by connecting the snaps’ interface.
+```
+dbus-super-cool-app: # adjust accordingly to your app name
+  interface: dbus
+  bus: session
+  name: dev.site.super_cool_app 
+```
+
 ### Parts
 
 This section defines the sources required to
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Arenukvern** [comment]
@googlebot I signed it!

File: src/docs/deployment/linux.md [hunk 1]
```diff
@@ -235,6 +242,37 @@ apps:
...
+**DBus interface**
+
+: The dbus interface provides a way for snaps to 
+  communicate over DBus. The snap providing the DBus 
+  service declares a slot with the well-known DBus name 
+  and which bus it uses. Snaps wanting to communicate 
+  with the providing snap’s service declare a plug for 
+  the providing snap. Note that a snap declaration is 
+  needed for your snap to be delivered via the snap store 
+  and claim this well-known DBus name (simply upload the
```
sfshaza2, line 254: Is this "well known" name actually DBus or something else?

src/docs/deployment/linux.md
```diff
@@ -262,6 +300,36 @@ parts:
     flutter-target: lib/main.dart # The main entry-point file of the application
 ```
 
+
+## Desktop file and icon
+: Desktop entry files are used to add an application
```
sfshaza2, line 305: After a header, the following text should be a paragraph, not a definition list.

src/docs/deployment/linux.md
```diff
@@ -262,6 +300,36 @@ parts:
...
+  Specification version 1.1.
+  
+## Flutter super-cool-app.desktop example
+
+: Place the .desktop file in your Flutter
+  project under `<project root>/snap/gui/super-cool-app.desktop`.
+  **Notice**: name for icon and .desktop file must be the same as you name your app in yaml file
+For example:
+
+```desktop
```
sfshaza2, line 319: I'm not sure that desktop is a valid type for a code block. Perhaps it should be `terminal`?

sfshaza2 (commented):
Thanks so much for this PR, @Arenukvern! Unfortunately, I don't know snap or Linux development, so it's hard for me to evaluate. @stuartmorgan, can you take a look?

src/docs/deployment/linux.md
Arenukvern: @sfshaza2 I can only assume from my very limited experience:
First about name:

It seems that it not necessary should be "well-known" from the beginning. In reality: 
- every "new" DBus name will be reviewed manually after first snap will be loaded to Snapstore.
- after DBus name have been at least once manually reviewed and approved - it going to be "well-known" and not requiring manual review when you uploading new snap to Snapstore and will be approved automatically.

About your question and  maybe I give you a wrong answer:

I'm not sure that the name is actually the DBus itself, it seems like it is a name of DBus service for snap app.
For me it looks very similar to AndroidManifest.xml beginning
```
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
    package="org.foo.my_super_cool_app">
```
But again, I maybe wrong, because I have just user look for what's going on under the hood.

src/docs/deployment/linux.md
Arenukvern: Thank you, changes commited

stuartmorgan:
I'm not familiar with snapcraft. @kenvandine can you route this to a good reviewer?

sfshaza2:
@kenvandine, gentle ping?  Thx!

stuartmorgan:
@robert-ancell 2 is what you recently fixed in the template, right?

robert-ancell:
> @robert-ancell 2 is what you recently fixed in the template, right?

Yes (not yet merged, awaiting tree to reopen).

sfshaza2:
I just kicked off the build again - Travis had failed but the problem may have been fixed.

sfshaza2:
@stuartmorgan, @kenvandine, Travis is happy now. Can I merge this? (Ack, it is over a month old.)

kenvandine:
Actually, @robert-ancell recently updated the template so the DBus portion of this will no longer be required.  I guess we should land this as is and drop the DBus portion of it once the template changes make it's way to beta.

stuartmorgan:
> Actually, @robert-ancell recently updated the template so the DBus portion of this will no longer be required. I guess we should land this as is and drop the DBus portion of it once the template changes make it's way to beta.

Desktop is only supported on master and dev, so we don't need to wait for any propagation.

PR #5079, opened 2020-12-12
Add --web-renderer documentation

johnpryan:
The `flutter` command  supports the `--web-renderer` option on the beta, dev, and master channels. This adds a new page,  `docs/deployment/web-renderers.md` and links to it from [flutter.dev/web](flutter.dev/web).

Staged: https://flutter-website-jr-stagi-988d4.firebaseapp.com/docs/development/tools/web-renderers

cc: @yjbanov @ferhatb @mariamhas


src/docs/deployment/web-renderers.md
```diff
@@ -0,0 +1,55 @@
...
+The `--web-renderer` command line option takes one of three values, `html`,
+`canvaskit`, or `auto`.
+
+Choose the **html** renderer when optimizing for code size over performance on
+desktop and mobile devices.
+
+Choose the **canvaskit** renderer for applications that prioritize performance
+and pixel-perfect consistency
+
+Choose the **auto** option (default) for applications that are accessed from
```
mariamhas, line 26: the wording of this confusing to me with how many times the word "accessed" is used and the "but" doesn't make sense.

it doesn't make it clear to me when I read it that auto will cause the app to use the HTML renderer on mobile devices, and CanvasKit on the desktop

johnpryan: Maybe it's clearer to say "Choose the **auto** option (default) to switch between html on mobile devices and canvaskit on desktop devices." WDYT?

src/docs/deployment/web-renderers.md
```diff
@@ -0,0 +1,55 @@
...
+  performance, and is less likely to have differences across browsers, but adds
+  about 2MB in download size.
+
+## Command line flags
+
+The `--web-renderer` command line option takes one of three values, `html`,
+`canvaskit`, or `auto`.
+
+Choose the **html** renderer when optimizing for code size over performance on
+desktop and mobile devices.
```
sfshaza2, line 21: I find this confusing. Maybe we can chat?

src/docs/deployment/web-renderers.md
sfshaza2: I'm also confused.

sfshaza2 (changes requested):
I need more explanation.  :)

src/docs/deployment/web-renderers.md
mariamhas: agreed a chat would be good to go over this

johnpryan: There's more information in [this design doc](https://docs.google.com/document/d/1aY0iU16wf_sdT7nwfpjgT-IatHNfF3slTiYHKmxcIog) that perhaps we can integrate here.

src/docs/deployment/web-renderers.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Web renderers
+description: How to choose a web renderer for running and building a web app.
+---
+
```
sfshaza2, line 5: Add a note stating that this is available on which channels. (NOT STABLE yet)

src/docs/deployment/web-renderers.md
sfshaza2: What do. you mean "on desktop and mobile devices"? Choose this for an Android/iOS app? How does it even affect that? Choose when running either a desktop app or a WEB app on mobile, like Chrome? This is unclear.

src/docs/deployment/web-renderers.md
```diff
@@ -0,0 +1,59 @@
...
+## Command line flags
+
+The `--web-renderer` command line option takes one of three values, `html`,
+`canvaskit`, or `auto`.
+
+Choose the **html** renderer when optimizing for code size over performance on
+desktop and mobile devices.
+
+Choose the **canvaskit** renderer for applications that prioritize performance
+and pixel-perfect consistency.
```
sfshaza2, line 29: What does this really mean? Is it web only? What is an example of when you might choose this?

src/docs/deployment/web-renderers.md
```diff
@@ -0,0 +1,59 @@
...
+
+Choose the **auto** option (default) to switch between `html` on mobile browsers
+and `canvaskit` on desktop browsers.
```
sfshaza2, line 32: Ahhhh. OK, this is a bit more clear. HOWEVER, are you calling desktop browsers "desktop"? I'm thinking of the previous definitions. Desktop and desktop browsers are very different. We need to be really clear in the terminology.

src/docs/deployment/web-renderers.md
```diff
@@ -0,0 +1,59 @@
...
+
+This flag can be used with the `run` or `build` subcommands. For example:
+
+```
+flutter run -d chrome --web-renderer html
+flutter build web --web-renderer canvaskit
+```
```
sfshaza2, line 39: So, this page is in the deployment section of the site, but it rather glosses over building an app for deployment. 

- What if you are building a single app for all platforms? Can you drop "web" and leave the "web-renderer" tag?
- Any other considerations to take into account?

src/docs/deployment/web-renderers.md
```diff
@@ -0,0 +1,59 @@
...
+Run in Chrome using the default renderer option (`auto`):
+
+```
+flutter run -d chrome
+```
+
+Build your app in release mode, using just the CanvasKit renderer:
+
+```
+flutter build web --web-renderer canvaskit --release
```
sfshaza2, line 52: Again, is the assumption that you will only use this parameter when building a web-only app?

sfshaza2 (changes requested):
I've left the questions that I would otherwise ask in a GVC chat. I'm happy to publish by Thursday if we can:
- Add a bit more clarity
- Add a note stating that this is not yet on the stable channel
- Add the page to the sidenav in src/_data/sidenav.

src/docs/deployment/web-renderers.md
johnpryan: I see, I'm trying to differentiate from `auto` here. You should choose the `html` renderer if you care about code size **regardless of whether or not the app is accessed from a mobile browser or desktop browser**. 

The `auto` option should be chosen if you care more about code size when the app is accessed from a mobile browser, and if you care about fast performance when the app is accessed from a desktop browser. 

The `auto` option is provided to handle what we think is the most common use case; that most apps will choose to use HTML for their users who are using a mobile browser (e.g. desktop Safari) and would prefer that their users see the CanvasKit version for their users who are using a desktop browser.

src/docs/deployment/web-renderers.md
johnpryan: yes, good point.

src/docs/deployment/web-renderers.md
johnpryan: yes, I believe it will either throw an error or be ignored when a non-web platform is specified.

johnpryan: It looks like this option is ignored on other platforms.

johnpryan:
I also moved the doc from `docs/deployment/` to  `docs/development/tools/` - it's probably more tooling-related than deployment related.

src/docs/deployment/web-renderers.md
johnpryan: I moved this to `docs/development/tools/` since it's not only related to deployment. 

You can't build for all platforms using a single command (you need to choose a subcommand like `web` or `ios`. I added a clarification that this flag is ignored for non-web device targets.

src/docs/deployment/web-renderers.md
johnpryan: Updated - this page now says "browser" whenever we specify a device like desktop or mobile.

src/docs/deployment/web-renderers.md
johnpryan: clarified how this works in the "choosing which option to use" section.

src/docs/deployment/web-renderers.md
johnpryan: added the note from flutter.dev/web

johnpryan:
@sfshaza2 this is ready for another look, I added a "Choosing which option to use" and tried to clarify things.

https://flutter-website-jr-stagi-988d4.firebaseapp.com/docs/development/tools/web-renderers

sfshaza2 (approved):
Much better!!! LGTM

PR #5082, opened 2020-12-12
Fix a broken link on the community page.

micpap25:
Just a quick fix to a link that seems to be broken now; I made it specifically re-direct to the flutter..dev community page.


sfshaza2 (approved):
LGTM

PR #5117, opened 2020-12-20
difference of plugins and packages added.

### LINKED ISSUE #4661: Define the differences between "packages" and "plugins"

Roshan13046:
Fixes #4661.

*  Added the definitions of flutter plugins and packages.
*  Added differences between plugins and packages.

### DIFF
--- a/src/docs/development/packages-and-plugins/using-packages.md
+++ b/src/docs/development/packages-and-plugins/using-packages.md
@@ -3,6 +3,22 @@ title: Using packages
 description: How to use packages in your Flutter app.
 ---
 
+**Plugin**
+A plugin is ready to use software component that can be added to existing 
+software for more features. It is about making native functionality 
+available to Flutter.
+
+**Package**
+A package is about integrating Flutter with an existing native application.
+
+**Difference** between packages and plugins:
+
+1. Packages are usually new components or code written purely in Dart
+whereas plugins work to allow more functionality on the device side 
+using native code. 
+2. Plugins are downloaded and integrated into the existing software without
+the need to compile/build. 
+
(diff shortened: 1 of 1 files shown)


src/docs/development/packages-and-plugins/using-packages.md
```diff
@@ -3,6 +3,22 @@ title: Using packages
 description: How to use packages in your Flutter app.
 ---
 
+**Plugin**
+A plugin is ready to use software component that can be added to existing 
+software for more features. It is about making native functionality 
+available to Flutter.
+
+**Package**
+A package is about integrating Flutter with an existing native application.
```
sfshaza2, line 12: I think this is misleading.  How about:
```
***Packages***
A package contains shared and reusable code and includes a list of
dependencies (specified in the `pubspec.yml` file), and (optionally)
Dart libraries, apps, resources, tests, images, and examples.
The pub.dev site lists many packages developed by Google engineers
and other members of the Flutter and Dart community that you can use in your app.
```

Also, I'd place the definition for packages BEFORE plugins, as plugins are a TYPE of package.

```
***Plugins***
A plugin package is a special kind of package that makes platform functionality
available to the app. Plugin packages can be written for Android (using Kotlin or Java),
iOS (using Swift or Objective-C), web, macOS, Windows, or Linux,
or any combination thereof. An example of a plugin would provide the Flutter app
the ability to access a device's camera.
```

sfshaza2 (changes requested):
I've wanted to add this sort of text here for ages.  :)

Issue #5062, opened 2020-12-08
[i18n] Documentation issues with "adding your own localizations" section


drekka:
Page URL: https://flutter.dev/docs/development/accessibility-and-localization/internationalization.html
Page source: https://github.com/flutter/website/tree/master/src/docs/development/accessibility-and-localization/internationalization.md

Description of issue:

Hi guys, I've been trying to follow this page for several days and it appears that the Intl package no longer works as described in this page. Specifically, it no longer generates any files to support internationalisation as described.

sfshaza2:
@shihaohong, have these changes reached stable?

drekka:
Actually can close this. It turns out that the Flutter docs are out of date and the files are actually generated into the `.dart_tool/flutter_gen/gen_l10n` directory.

shihaohong:
Hey @drekka, sorry about the delay with getting back to this. Could you share which part of the Flutter docs are out of date? I'd like to go and sort that out if that's still a problem. If it's something we can fix on the framework side of things, file an issue at [flutter/flutter](https://github.com/flutter/flutter/issues/new/choose) and cc/ me in on it!

drekka:
A couple of places on https://flutter.dev/docs/development/accessibility-and-localization/internationalization

First, second code example under setup which looks like this:

```dart
import 'package:flutter_localizations/flutter_localizations.dart';

// ...

MaterialApp(
 localizationsDelegates: [
   // ... app-specific localization delegate[s] here
   AppLocalizations.delegate,
   GlobalMaterialLocalizations.delegate,
   GlobalWidgetsLocalizations.delegate,
   GlobalCupertinoLocalizations.delegate,
 ],
 supportedLocales: [
    const Locale('en', ''), // English, no country code
    const Locale('ar', ''), // Arabic, no country code
    const Locale.fromSubtags(languageCode: 'zh'), // Chinese *See Advanced Locales below*
    // ... other locales the app supports
  ],
  // ...
)
```

Should be something like this:

```dart
import 'package:flutter_gen/gen_l10n/app_localizations.dart';

// ...

@override
  Widget build(BuildContext context) {
    return MaterialApp(
      localizationsDelegates: AppLocalizations.localizationsDelegates,
      supportedLocales: AppLocalizations.supportedLocales,
      onGenerateTitle: (context) => AppLocalizations.of(context).#localised-app-key#,
      home: // ...
      );
  }
```

[...]

PR #5135, opened 2020-12-28
Address i18n doc fixes

shihaohong:
Fixes remaining issues addressed in https://github.com/flutter/website/issues/5062.

1. Remove unnecessary import in the app. Since the import is only needed for additional customization in the flutter app, it's not necessary for the purposes of the tutorial.
2. Added a link and snippet regarding `onGenerateTitle`, which is helpful for Android apps when the task manager needs a localized name to use to refer to the Flutter app.


sfshaza2 (approved):
LGTM!

PR #5142, opened 2020-12-30
Update devtools documentation for timeline->performance and performance->cpu-profiler renames

kenzieschmoll:
The Performance page in DevTools has been renamed to CPU profiler and the Timeline page has been renamed to Performance. This change is currently in master and has not been publicly released. This PR updates the documentation to reflect the rename.

NOTE: this CL should not be landed until the next release of DevTools goes out (sometime in January). If this is landed before the next release of DevTools, links to documentation from the released version of DevTools will be broken.

@sfshaza2 

cc @filiph in case any hyperlinks need to be updated in videos / articles.


src/docs/development/tools/devtools/cpu-profiler.md
```diff
@@ -0,0 +1,131 @@
...
+
+{{site.alert.note}}
+  **If you are running a Flutter application,
+  use a profile build to analyze performance.**
+  CPU profiles are not indicative of release performance
+  unless your Flutter application is run in profile mode.
+{{site.alert.end}}
+
+<!-- TODO(kenz): this is the same information that is in the Timeline page.
+Consolidate somehow. -->
```
sfshaza2, line 25: You can always create a _definitions.md or _profiler-def.md page (or something appropriately named, but the filename should begin with an underbar).. Then just include the file in both places.

src/docs/development/tools/devtools/performance.md
```diff
@@ -5,71 +5,147 @@ description: Learn how to use the DevTools performance view.
...
+60 FPS (frames per second), each frame must render in
+~16 ms or less. When this target is missed, you may
+experience UI jank or dropped frames.
+
+For more information on how to analyze your app's performance,
+see [Flutter performance profiling][].
+
+## Timeline events chart
+
+The timeline events chart shows all event trace from your application.
```
sfshaza2, line 90: => all event tracing?

sfshaza2 (approved):
LGTM, with a couple comments. I know you don't want to merge this until Friday. I won't merge it, but if you wanted to make sure, you could always set this as a WIP or Draft, and then it *can't* be published until that designation is removed.

src/docs/development/tools/devtools/performance.md
kenzieschmoll: done

src/docs/development/tools/devtools/cpu-profiler.md
kenzieschmoll: ptal - not sure if the way I included the file is correct. followed the guide here: https://developer.lightbend.com/docs/paradox/current/directives/includes.html#include-block

kenzieschmoll: PTAL at the include format - `{% include _profiler.md %}` - is that correct? I cannot build the website on my machine without some configuration tasks so I don't always for "safe" .md changes.

sfshaza2: Sorry I didn't see this before, @kenzieschmoll. Yes, that include is correct. Can we go ahead and merge this?

sfshaza2:
Actually, we don't generally put includes like this into the _includes directory. For example, see /src/docs/get-started/install for a slew of include files that really only apply to the install instructions. It's not a huge issue, though.

sfshaza2 (approved):
lgtm

PR #5151, opened 2021-01-04
add riverpod to state management options page

buncis:
add riverpod to state management options page

### DIFF
--- a/src/docs/development/data-and-backend/state-mgmt/options.md
+++ b/src/docs/development/data-and-backend/state-mgmt/options.md
@@ -35,14 +35,6 @@ A recommended approach.
 * [You might not need Redux: The Flutter edition][], by Ryan Edge
 * [Making sense of all those Flutter Providers][]
 
-## Riverpod
-
-A provider rewrite by the same author, compiled safe, more testable, more flexible
-
-* [Riverpod Homepage][]
-* [Riverpod Github][]
-* [Motivation for provider rewrite][]
-
 ## setState
 
 The low-level approach to use for widget-specific, ephemeral state.
@@ -145,9 +137,15 @@ A simplified reactive state management solution.
 * [Complete GetX State Management][], a video by Tadas Petra
 * [GetX Flutter Firebase Auth Example][], by Jeff McMorris
 
-[Motivation for provider rewrite]: {{site.github}}/rrousselGit/river_pod#motivation 
-[Riverpod Homepage]: https://riverpod.dev/
-[Riverpod Github]: {{site.github}}/rrousselGit/river_pod
+## Riverpod
+
+An approach similar to Provider that is compile-safe and testable. It doesn't have a dependency on the Flutter SDK.
+
+* [Riverpod][]homepage
+* [Getting started with Riverpod][]
+
+[Getting started with Riverpod]: https://riverpod.dev/docs/getting_started
+[Riverpod]: https://riverpod.dev/
 [Flutter-Movie]: {{site.github}}/o1298098/Flutter-Movie
 [Fish-Redux-Source]: {{site.github}}/alibaba/fish-redux
 [Fish-Redux-Library]: {{site.pub-pkg}}/fish_redux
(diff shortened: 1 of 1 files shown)


sfshaza2:
@filiph, please review

src/docs/development/data-and-backend/state-mgmt/options.md
```diff
@@ -35,6 +35,14 @@ A recommended approach.
 * [You might not need Redux: The Flutter edition][], by Ryan Edge
 * [Making sense of all those Flutter Providers][]
 
+## Riverpod
+
+A provider rewrite by the same author, compiled safe, more testable, more flexible
```
filiph, line 40: This reads like an advert. Please rewrite so that:

1. This can stand on its own (not "rewrite of the above") — people scan these pages, and the order of the items can change at any time.
2. The sentence or two distill the essence of the package. 

So maybe something like:

> An approach similar to Provider that is compile-safe and testable. It doesn't have a dependency on the Flutter SDK.

Also, nit: There should be a period at the end of the sentence, for consistency.

src/docs/development/data-and-backend/state-mgmt/options.md
```diff
@@ -35,6 +35,14 @@ A recommended approach.
...
+
+* [Riverpod Homepage][]
+* [Riverpod Github][]
```
filiph, line 43: This is redundant. From this page, we're linking to resources (like a homepage or an article or a video). You could link to pub.dev here, but since you're already linking to homepage above (which I think is the right choice), I suggest removing this line.

Alternatively, add a link to some tutorial (video or article).

src/docs/development/data-and-backend/state-mgmt/options.md
```diff
@@ -35,6 +35,14 @@ A recommended approach.
 * [You might not need Redux: The Flutter edition][], by Ryan Edge
 * [Making sense of all those Flutter Providers][]
 
+## Riverpod
+
+A provider rewrite by the same author, compiled safe, more testable, more flexible
+
+* [Riverpod Homepage][]
```
filiph, line 42: The "h" in "homepage" should be lowercase, imho.

Also, consider moving the "homepage" outside the link, like this:

```suggestion
* [Riverpod][] homepage
```

I think it's cleaner. Of course, you'll need to update the named link below if you do this.

src/docs/development/data-and-backend/state-mgmt/options.md
```diff
@@ -35,6 +35,14 @@ A recommended approach.
...
+* [Riverpod Github][]
+* [Motivation for provider rewrite][]
```
filiph, line 44: I think this is redundant because we get a good idea about the motivation from the homepage above. If you feel strongly about keeping it, please try to format the item similarly to other items on this page.

src/docs/development/data-and-backend/state-mgmt/options.md
```diff
@@ -35,6 +35,14 @@ A recommended approach.
...
+
```
filiph, line 45: Please put this section at the end of the page. We might move things around at some point but right now we're adding approaches in "first come, first serve" order.

filiph (changes requested):
Hi, thanks for the contribution! I can't believe we don't have a riverpod entry yet.

I made some comments inline.

src/docs/development/data-and-backend/state-mgmt/options.md
```diff
@@ -145,9 +137,15 @@ A simplified reactive state management solution.
...
 * [GetX Flutter Firebase Auth Example][], by Jeff McMorris
 
-[Motivation for provider rewrite]: {{site.github}}/rrousselGit/river_pod#motivation 
-[Riverpod Homepage]: https://riverpod.dev/
-[Riverpod Github]: {{site.github}}/rrousselGit/river_pod
+## Riverpod
+
+An approach similar to Provider that is compile-safe and testable. It doesn't have a dependency on the Flutter SDK.
+
+* [Riverpod][]homepage
```
filiph, line 144:
```suggestion
* [Riverpod][] homepage
```

buncis:
I have applied it

thanks for the suggestions @filiph

filiph (changes requested):
Thank you! One missing space and I'm happy.

filiph (approved):
LGTM, thanks!

PR #5166, opened 2021-01-09
Update index.md

omartinma:
If we use  IntegrationTestWidgetsFlutterBinding.ensureInitialized(); inside the setUpAll all the test will fail. Issue opened here: https://github.com/flutter/flutter/issues/72063

Fixes [72063](https://github.com/flutter/flutter/issues/72063)


omartinma:
> @googlebot I signed it!

@googlebot I signed it!

dnfield (changes requested):
I don't think this is what we want to say as the most up to date way to do this. @jiahaog has worked on some things to redo the way reporting is handled. @jiahaog can you take a look at this file?

src/docs/testing/integration-tests/index.md
```diff
@@ -258,9 +258,9 @@ import 'package:integration_test/integration_test.dart';
 import 'package:integration_test_example/main.dart' as app;
 
 void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
   group('end-to-end test', () {
     setUpAll(() async {
-      IntegrationTestWidgetsFlutterBinding.ensureInitialized();
     });
```
jiahaog, line 264: This should be removed as well

src/docs/testing/integration-tests/index.md
```diff
@@ -258,9 +258,9 @@ import 'package:integration_test/integration_test.dart';
 import 'package:integration_test_example/main.dart' as app;
 
 void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
```
jiahaog, line 261: nit: newline

```suggestion
  IntegrationTestWidgetsFlutterBinding.ensureInitialized();

```

jiahaog (commented):
@dnfield this change is correct. The documentation on that page looks like it's specific to the version of `package:integration_test` from pub, which does not support the newer way to do test reporting.

The newer way (wrapping tests with `run`) to do test reporting is only supported when `package:integration_test` is loaded from the Flutter SDK, so we'll have to update the documentation when that happens.

jiahaog:
@omartinma looks like you'll need to run a script to update some other artifacts before the build passes, please see https://travis-ci.org/github/flutter/website/jobs/753716327

sfshaza2:
Hi folks! I need some context. Is this updated code valid as of our current stable release? Or is it updated for master channel? We test the Travis build against the latest stable, which is currently 1.22. If it's for a future stable release, then we have two options:
1. Delay addressing this until the changes hit stable.
2. Document BOTH on the website, but make sure it's clear what is stable code vs what is future code.

If the decision is to postpone documenting this, I would prefer to close this because that could take a long time.

jiahaog:
Sorry, I missed this. This change is independent of the SDK branch, as it is [published in Pub](https://pub.dev/packages/integration_test). We really should try and land this as soon as possible – it is quite confusing to users.

@omartinma will you be making the requested changes? If not I'm happy to take over this PR and see it through :)

sfshaza2:
@jiahaog, thanks for that context. Let me know when it is ready for final review and merging. We will need it to pass Travis, however.

omartinma:
@jiahaog i will be happy to do whatever change needed, but actually I dont know what these changes are. I mean, this PR is just for changing the documentation(as it appears in pub.dev)

jiahaog:
@omartinma Looking at the [CI failure](https://travis-ci.org/github/flutter/website/jobs/753716325), you'll need to run https://github.com/flutter/website/blob/master/tool/refresh-code-excerpts.sh locally to update your branch. I'm not familiar with this repo but it's probably a mechanism for additional checks to ensure that the code is buildable. There are also some suggested edits above which you'll want to apply to your PR before running the script.

jiahaog:
I pushed the relevant changes :) @sfshaza2 can you take a look again please?

sfshaza2 (approved):
Just to give you context, @omartinma, this is the repo for the flutter.dev website. And, yes, it contains code that we test and, if a PR doesn't pass the code check, we can't merge it. It's how we make sure that the code passes the latest stable release. At any rate, thanks for your PR and LGTM.

PR #5176, opened 2021-01-12
Public listing of deprecations removed after 1.22

Piinks:
This adds a doc link to the public listing of the deprecated api we are removing for the next stable release


darrenaustin (approved):
LGTM.

PR #5181, opened 2021-01-13
Language pass on use-maxLengthEnforcement-instead-of-maxLengthEnforced

Hixie:
Just some editing to clean up some of the language.

Also adds a section that talks about why assuming English is dangerous.


sfshaza2 (approved):
Thanks, Hixie! More merges are happening before I have a chance to review.  :). LGTM

PR #5189, opened 2021-01-15
Fix warning in validation cookbook showSnackBar

louisrli:
Gives a linter warning:
```
 [dart deprecated_member_use] [I] 'showSnackBar' is deprecated and shouldn't
  be used. Use ScaffoldMessenger.showSnackBar. This feature was deprecated
 after v1.23.0-14.0.pre..
Try replacing the use of the deprecated member with the replacement
```

### DIFF
--- a/src/docs/cookbook/forms/validation.md
+++ b/src/docs/cookbook/forms/validation.md
@@ -219,7 +219,7 @@ class MyCustomFormState extends State<MyCustomForm> {
                 // otherwise.
                 if (_formKey.currentState.validate()) {
                   // If the form is valid, display a Snackbar.
-                  Scaffold.of(context)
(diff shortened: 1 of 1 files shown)


louisrli:
@googlebot I signed it!

sfshaza2:
Hi, @louisrli! We test the website against Flutter's latest stable build, which is 1.22. This change occurred in 1.23, which has not reached stable yet.

So, we have a couple options:
1. Wait to make this change until our next stable release.
2. Update the PR to offer both solutions. After the next stable release, we can then delete the deprecated solution.

I could go either way. Do you have a preference? If we choose to wait, I will likely close this PR as it could be some time before our next stable release.

louisrli:
I'll close it in that case :)

PR #5192, opened 2021-01-17
Fix trailing commas example

mernen:

* The code in the screenshots had an extra "s" and would write "Button tapped 5 timess".


sfshaza2 (approved):
Good catch! LGTM.

PR #5196, opened 2021-01-18
Fix wrong parameter name

Numoy:
In the create method of the NativeViewFactory a wrong parameter name is used (id instead of viewId).

*  Correct name of parameter id -> viewId


sfshaza2 (approved):
LGTM

PR #5199, opened 2021-01-20
TextEditingController.buildTextSpan migration guide

Jjagg:
Adds the migration guide for flutter/flutter#73510.


Hixie:
Make sure to include the error message that users will see, so if they google the error message the guide comes up.

justinmc (approved):
LGTM 👍 

This is great, thanks for taking the time to write this.  I will point any Google Flutter users to this when I migrate their code.

sfshaza2:
Kicked off the build again. It complained about the null safety check.

sfshaza2 (approved):
LGTM

PR #5201, opened 2021-01-20
Moving image files and also updating shared repos.

sfshaza2:
Moving image files so they can be used in DartPad instances as well as markdown files.

Also syncing the shared repos.


kwalrath:
I expected some .md files in this PR, too. Are these files not referred to by anything on flutter.dev? (What _does_ refer to them?)

sfshaza2:
DartPad. As it is now, they can't use these images from DartPad. They have
to be moved out of _assets.

sfshaza2:
The md files are still to come. We are still reviewing them.  :)

sfshaza2:
The Travis failure is due to out of date code excerpts, which this PR doesn't involve.

PR #5202, opened 2021-01-20
Pass a URI to package:http API

natebosch:
Updates example to be forwards compatible with the breaking change in
https://github.com/dart-lang/http/pull/507

### DIFF
--- a/src/docs/cookbook/networking/fetch-data.md
+++ b/src/docs/cookbook/networking/fetch-data.md
@@ -242,7 +242,7 @@ import 'package:http/http.dart' as http;
 
 Future<Album> fetchAlbum() async {
   final response =
-      await http.get('https://jsonplaceholder.typicode.com/albums/1');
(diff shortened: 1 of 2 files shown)


sfshaza2:
Hi, @natebosch! It's unclear to me. Will this update work now when analyzed against the stable channel? If so, let's go for it. If not, we'll have to wait for this change to hit stable. The website is analyzed against stable. Alternatively, we could document both for now, but annotate the not-yet-working code with the  tag which tells Travis to not analyze it.

natebosch:
This will work correctly when analyzed against the current package versions.

The existing published API is statically typed `dynamic` and at runtime it allows either `String` or `Uri`. The upcoming API is statically typed `Uri`. This PR does not break backwards compatibility, it adds forwards compatibility.

sfshaza2:
Sounds great!

sfshaza2 (approved):
LGTM

sfshaza2:
@RedBrogdon fixed the cookbook code check errors in another PR. Forcing a merge.

Issue #5074, opened 2020-12-11
'Return data from a screen' page issue


349989153:
Page URL: https://flutter.dev/docs/cookbook/navigation/returning-data.html
Page source: https://github.com/flutter/website/tree/master/src/docs/cookbook/navigation/returning-data.md

Description of issue:
I try to follow the guide but get the error: `The method 'ElevatedButton' isn't defined for the type 'SelectionButton'`.

And the version is :
```
flutter --version
Flutter 1.17.5 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 8af6b2f038 (5 months ago) • 2020-06-30 12:53:55 -0700
Engine • revision ee76268252
Tools • Dart 2.8.4
```

So I do some extra search and find that **To use ElevatedButton without warnings or errors, you should update to Flutter 1.22.0 or newer versions. The simplest way to do that is to run this command: flutter upgrade**

**But what really bother me is I can't get that impotant INFO ABOUT VERSION on either https://api.flutter.dev/flutter/material/ElevatedButton-class.html or https://flutter.dev/docs/cookbook/navigation/returning-data, I think it's not convenient for developers. Can you optimize the doc, please?**

VarunSAthreya:
If the issue is still open I would like to work on it.

PR #5205, opened 2021-01-21
Added comment in index.md

Abhishek01039:
Fixes https://github.com/flutter/website/issues/5074.

### DIFF
--- a/src/docs/index.md
+++ b/src/docs/index.md
@@ -92,6 +92,7 @@ You might also find these docs useful:
 * [State management][]
 * [Animations][]
 
+The documentation follows the latest stable version of flutter
 
 [A tour of the Flutter widget framework]: /docs/development/ui/widgets-intro
 [Adding assets and images]: /docs/development/ui/assets-and-images
--- a/src/docs/index.md
+++ b/src/docs/index.md
@@ -92,7 +92,7 @@ You might also find these docs useful:
 * [State management][]
 * [Animations][]
 
-The documentation follows the latest stable version of flutter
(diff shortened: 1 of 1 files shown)


src/docs/index.md
```diff
@@ -92,6 +92,7 @@ You might also find these docs useful:
 * [State management][]
 * [Animations][]
 
+The documentation follows the latest stable version of flutter
```
sfshaza2, line 95: Sorry to be so nit-picky, but please make the f in Flutter uppercase. Also, add a period at the end. Thx!

Hixie:
maybe "refers to" rather than "follows"? But in general this is fine by me. @sfshaza2 the context here is that it turns out we don't seem to actually say anywhere that the website is intended to document the latest stable, rather than, say, the latest dev build, and it would be useful to have something to point to when confusion about this arises.

sfshaza2:
Yes, I think adding this comment (or a version thereof) is an excellent idea. Lately I've been barraged by PRs where some eager beavers (whom we love) are trying to update code/text on flutter.dev to post 1.22 changes/deprecations/etc.  I've repeatedly had to explain that we document against stable.

I just want Flutter to be upper case and for the sentence to end in a period.  :). Also, a more accurate line might be:

The documentation on this site reflects the latest stable release of Flutter.

or

Unless otherwise noted, the code on this website is tested against the latest stable release of Flutter.

Abhishek01039:
```The documentation on this site reflects the latest stable release of Flutter.```
This seems fine.
should I add this @sfshaza2 ?

sfshaza2 (approved):
LGTM

PR #5207, opened 2021-01-22
Fixes the URL link.

chenglu:
The link of `MaterialApp.onGenerateTitle` is missed, this PR is trying to fix this.

cc/ @sfshaza2, Long time no see ;-)


sfshaza2 (approved):
Ah, the way we usually fix this is to add the code font keys to the URL description. But this is ok for now. Thanks, Luke! I hope you are doing well!

PR #5195, opened 2021-01-18
Mark `dart:ffi` as stable

dcharkes:
Should be merged when https://dart-review.googlesource.com/c/sdk/+/179765 is merged.

PR #5216, opened 2021-01-26
Add site.alert.version-note

kwalrath:
As promised in #5195, this change enables flutter.dev to support version notes, like dart.dev does. I tested the change locally.


sfshaza2 (approved):
LGTM!

PR #5222, opened 2021-01-27
Update sidenav.yml

timsneath:

*  Switch from the rather old Dart overview page to the new, vibrant, tripartite technical intro that has just gone live on dart.dev!

/cc @kwalrath


sfshaza2 (approved):
LGTM!

PR #5227, opened 2021-01-28
Migrate Layout Basics Codelab to Null Safety

miquelbeltran:
@Sfshaza @RedBrogdon 

In this PR I have migrated the Layout Basics codelab following the steps in the migration plan document.

The PR is split in commits in a way that hopefully makes easier to follow the changes!

Improvements:
- Dartpads are now embedded in markdown rather than on gists.
- Code is migrated to null safety, and it is enabled by default on the dartpads. All changes happen in the test code almost.
- Upgraded text styles to the current material design spec in the "Putting it all together" chapter.

Live version here: https://mb-dart-dev-1.firebaseapp.com/docs/codelabs/layout-basics_migrated

The "Name non-constant identifiers using lowerCamelCase" warning that appears is an issue on DartPad. I have already reported to Andrew.

Please feel free to edit anything as necessary before merging :)

### DIFF
--- a/src/docs/codelabs/layout-basics_migrated.md
+++ b/src/docs/codelabs/layout-basics_migrated.md
@@ -978,10 +978,111 @@ wrap a widget and force the widget to fill extra space.
   **3.** Select the **Format** button to properly format the code,
          and run again.
 {{site.alert.end}}
-{% comment %}
-  Gist: https://gist.github.com/c4dfa9058f803dea1cff4fca2532977a
-{% endcomment %}
-<iframe src="{{site.custom.dartpad.embed-flutter-prefix}}?id=c4dfa9058f803dea1cff4fca2532977a&amp;theme=dark&amp;split=60" width="100%" height="400px"></iframe>
+
+```run-dartpad:theme-dark:mode-flutter:run-true:width-100%:height-400px:split-60:null_safety-false
+{$ begin main.dart $}
+import 'dart:async';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+class MyWidget extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Row(
+      children: [
+        BlueBox(),
+        BlueBox(),
+        BlueBox(),
+      ],
+    );
+  }
+}
+
+class BlueBox extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+      width: 50,
+      height: 50,
+      decoration: BoxDecoration(
+        color: Colors.blue,
+        border: Border.all(),
+      ),
+    );
+  }
+}
+{$ end main.dart $}
+{$ begin test.dart $}
+class MyApp extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Directionality(
+      textDirection: TextDirection.ltr,
+      child: Container(
+        color: Color(0xffeeeeee),
+        child: Center(
+          child: Container(
+            child: MyWidget(),
+            color: Color(0xffcccccc),
+          ),
+        ),
+      ),
+    );
+  }
+}
+
+Future<void> main() async {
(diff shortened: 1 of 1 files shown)


sfshaza2 (commented):
It LGTM, @miquelbeltran! But I'll let @RedBrogdon also take a look.

src/docs/codelabs/layout-basics_migrated.md
```diff
@@ -0,0 +1,3340 @@
...
+import 'package:flutter_test/flutter_test.dart';
+
+class MyWidget extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Row(
+      children: [
+        Padding(
+            padding: const EdgeInsets.all(8.0),
+            child: Icon(Icons.account_circle, size: 50)),
```
RedBrogdon, line 2166: In general, we always try to use trailing commas to keep things indented by two characters, in part because DartPad's width is so constrained when shown in an iframe.

Here, if there was a comma in between the two parentheses at the end of the line, the last one would be moved to the next line, and both parameters for `Padding` would be indented two character instead of four.

src/docs/codelabs/layout-basics_migrated.md
```diff
@@ -0,0 +1,3340 @@
...
+        Column(
+          mainAxisSize: MainAxisSize.min,
+          crossAxisAlignment: CrossAxisAlignment.start,
+          children: [
+            Text('Flutter McFlutter',
+                style: Theme.of(context).textTheme.headline5),
```
RedBrogdon, line 2383: Here as well, a comma after `headline5` will help with indenting.

src/docs/codelabs/layout-basics_migrated.md
```diff
@@ -0,0 +1,3340 @@
...
+{:.no_toc}
+  The following example displays the differences between
+  a `Row` and `Column`.
+
+  **1.** Click the **Run** button.
+
+  **2.** In the code, change the `Row` to a `Column`, and run again.
+{{site.alert.end}}
+
+```run-dartpad:theme-dark:mode-flutter:run-true:width-100%:height-400px:split-60:null_safety-true
```
RedBrogdon, line 55: `run-true` should be taken out here and in the others. Right now it's causing every DartPad embed on the page to execute at the same time, which is a CPU hit for both the browser and our compilation backend. 😄

RedBrogdon (commented):
The updated code looks great. My only nits are about indenting and the injection code itself. I'm happy to see these stop relying on gists as well!

src/docs/codelabs/layout-basics_migrated.md
miquelbeltran: Ouch! I see, removed now.

src/docs/codelabs/layout-basics_migrated.md
miquelbeltran: Agreed, I think this was present in the original Gist, I will fix it now before merge.

sfshaza2 (commented):
You rock, @miquelbeltran! @RedBrogdon, I'd merge this now, but it's failing code check. Is the null safe code triggering that?  https://travis-ci.org/github/flutter/website/jobs/756691978

miquelbeltran:
I see, looks like the CI check that extracts and analyzes code snippets in md files does not like the embedded codelab code :) I will research why, but if you know how to fix it please let me know!

miquelbeltran:
I think it may be because the dartpad code contains a section separating the main/solution/test and the code analyzer doesn't like that. As far as I understand, adding `
` over each dartpad snippet would skip the checks. Is that correct?

RedBrogdon (approved):

RedBrogdon:
LGTM. I've restarted the travis build, which had previously errored out.

sfshaza2:
@RedBrogdon has fixed the cookbook errors in another PR, so I'm forcing a merge.

PR #5240, opened 2021-02-01
Un-deprecate scrollable alert dialog by default

shihaohong:
Since scrollable AlertDialogs by default are not backwards-compatible, we're un-deprecating the `scrollable` parameter and keeping the existing behavior. 

Update to design doc: flutter.dev/go/scrollable-alert-dialog
Link to framework PR: https://github.com/flutter/flutter/pull/75138
Link to website PR (self-link): #5240

### DIFF
--- a/src/docs/release/breaking-changes/scrollable-alert-dialog.md
+++ b/src/docs/release/breaking-changes/scrollable-alert-dialog.md
@@ -1,9 +1,14 @@
 ---
-title: Scrollable AlertDialog
+title: Scrollable AlertDialog (No longer deprecated)
 description: AlertDialog should scroll automatically when it overflows.
 ---
 
 ## Summary
+
+Note: `AlertDialog.scrollable` is no longer deprecated since there is
+no backwards compatible way to have AlertDialog be scrollable by default.
+Instead, the parameter will remain and users will be able to set `scrollable`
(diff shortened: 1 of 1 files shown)


src/docs/release/breaking-changes/scrollable-alert-dialog.md
```diff
@@ -1,10 +1,15 @@
...
 description: AlertDialog should scroll automatically when it overflows.
 ---
 
 ## Summary
 
+Note: `AlertDialog.scrollable` is no longer deprecated since there is
+no backwards compatible way to have AlertDialog be scrollable by default.
+Instead, the parameter will remain and users will be able to set `scrollable`
+to true if they want a scrollable `AlertDialog`.
+
```
sfshaza2, line 12:
```suggestion
{{{site.alert.note}}
  `AlertDialog.scrollable` is no longer deprecated because there is
  no backwards-compatible way to make `AlertDialog` scrollable by default.
  Instead, the parameter will remain and you can set `scrollable`
  to true if you want a scrollable `AlertDialog`.
{{site.alert.end}}
```

sfshaza2 (approved):
LGTM

shihaohong:
Thank you Shams!

PR #5242, opened 2021-02-01
Fixes for dryLayout breaking change guide

goderbauer:
Currently, the guide doesn't render properly, see the list of links in https://flutter.dev/docs/release/breaking-changes/renderbox-dry-layout. This PR fixes that.

It also puts the `Stable channel link:` and `Master channel link:` line into comments in the template because those should never be included in the rendering output.


sfshaza2 (approved):
LGTM!  It looks like this is ready to land, yes, @goderbauer?

goderbauer:
Yes, ready to land!

sfshaza2:
The changes in this PR are not related to the Travis build, which I've informed @RedBrogdon about.

I'm forcing a merge.

PR #5245, opened 2021-02-02
Removing submodules from format/analyze/test checks

RedBrogdon:
Removes submodules from the normal format/analyze/test done on all projects in the `examples` folder. This work makes sense for local examples, but it's wasted effort on the submodules, which have their own CI in their repos.

It also sidesteps the issue caused by the new, null-safe cookbook code in flutter/codelabs, which caused tests to fail in #5242.


sfshaza2 (approved):
Brilliant!!! Thanks, @RedBrogdon! LGTM

PR #5248, opened 2021-02-03
Update flutter artifacts bucket

godofredoc:
Flutter artifacts are moving to a different bucket owned by the flutter team.

Fixes go/fxb/64576

*  Updates GCS bucket used to create the download links.


sfshaza2 (approved):
Thanks for this, @godofredoc! LGTM

PR #5254, opened 2021-02-03
Updating banner

sfshaza2:
Sign up for Flutter Engage!


kwalrath:
We're still ironing out the exact words, but aside from that... should we make the link target bigger?

kwalrath:
@sfshaza2 how does https://kw-flutter-1.firebaseapp.com look?

sfshaza2 (commented):
No comment on link target, but looks good! lgtm

PR #5262, opened 2021-02-05
Fixing image links in layout basics codelab.

RedBrogdon:
DartPad was recently upgraded to use CanvasKit rendering, which has implications for how images are loaded (see https://flutter.dev/docs/development/platform-integration/web-images#cross-origin-images for info).

As a result, one of the exercises in the Layout Basics codelab broke. This PR changes the image sources used in that exercise in both the migrated and unmigrated versions, fixing the problem.


sfshaza2 (approved):
LGTM. Thanks, @RedBrogdon!

PR #5275, opened 2021-02-11
Add file_selector to Desktop

domesticmouse:
Adding `file_selector` to desktop's list of supported packages


stuartmorgan (commented):
I don't think we should do this; `file_selector` itself doesn't support desktop yet. There are "unofficial" (I wrote them, but they are in FDE, not flutter/plugins) unendorsed implementations which have to be explicitly added by users. I think listing this here will cause confusion.

PR #5276, opened 2021-02-11
Add notes to macOS distribution section

domesticmouse:
I totally missed that there is a section on Hardened Runtime while attempting to build a macOS Flutter desktop app.

### DIFF
--- a/src/desktop.md
+++ b/src/desktop.md
@@ -321,7 +321,10 @@ and the other DLLs, and bundle them together in a zip file.
 
 #### macOS
 
-The `.app` is self-contained, and can be distributed as-is.
+The `.app` is self-contained, and can be distributed as-is. However, you should 
+read through the [macOS-specific support](#macos-specific-support) section below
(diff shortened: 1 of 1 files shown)


src/desktop.md
```diff
@@ -321,7 +321,10 @@ and the other DLLs, and bundle them together in a zip file.
 
 #### macOS
 
-The `.app` is self-contained, and can be distributed as-is.
+The `.app` is self-contained, and can be distributed as-is. However, you should 
+read through the [macOS-specific support](#macos-specific-support) section below
+to understand about how Entitlements, the App Sandbox, and the Hardened Runtime
```
stuartmorgan, line 326: Nit: Apple capitalizes App Sandbox and Hardened Runtime since they are their own specific terms, but don't capitalize "entitlements" in their docs as it's a more widespread term.

stuartmorgan (approved):
LGTM.

sfshaza2 (approved):
LGTM! Thanks, Brett.

PR #5280, opened 2021-02-11
Migrate Web Codelab to Null Safety

miquelbeltran:
@Sfshaza @RedBrogdon

In this PR I migrate the starter web codelab to null safety.

Some considerations regarding the code changes:

- Instead of using `MaterialStateColor` I had to switch to `MaterialStateProperty` because `MaterialStateColor` does not allow null values for Color.
- Seems that `Navigator.of` returns nullable in older versions of Flutter, which is the one DartPad uses, so I need to add a nullcheck here. Newer versions of Flutter `Navigator.of` is not nullable anymore. This may need to be updated once DartPad is updated.
- The issue `Animation<Color?>` not being supported by `LinearProgressIndicator` has been fixed by creating a custom `Tween` that does not allow null Color values. Once `LinearProgressIndicator` accepts `Animation<Color?>` then this can be replaced by `ColorTween` back again. The issue has been reported and fixed in https://github.com/flutter/flutter/pull/75794

### DIFF
--- a/src/docs/get-started/codelab-web_migrated.md
+++ b/src/docs/get-started/codelab-web_migrated.md
@@ -349,10 +349,14 @@ Fix the error reported by the analyzer that `_showWelcomeScreen`
 is not defined. Directly above the `build()` method,
 add the following function:
 
+{%- comment %}
+TODO Remove `!` when no longer necessary
+{% endcomment -%}
+
 <!-- skip -->
 ```dart
 void _showWelcomeScreen() {
-  Navigator.of(context).pushNamed('/welcome');
+  Navigator.of(context)!.pushNamed('/welcome');
 }
 ```
 </li>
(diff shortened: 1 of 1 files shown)


sfshaza2:
@miquelbeltran, I just noticed you used @Sfshaza. I no longer use that (I got locked out of it some years ago). Please use @sfshaza2. I can merge this now and then you can "undo" it later, or we can just wait. Which do you prefer?

miquelbeltran:
Sorry about the `@`! I noticed in another places in GitHub but I knew I was going to make that mistake someday :)
I am fine to wait until the other fix is merged, the change will be quick from my side.

sfshaza2:
SG! Yeah, I had no way to recover that stuff. That's why it says on sfshaza2 to ignore Sfshaza.  :D

I'll close this one, but we can re-open if needed.

PR #5298, opened 2021-02-15
Documenting macOS Flutter app distribution options

domesticmouse:
Adding documentation on how to distribute a macOS Flutter application on either the macOS App Store or as a notarized app.


sfshaza2 (approved):
LGTM. Thanks again, @domesticmouse!

PR #5302, opened 2021-02-16
Update create instructions

stuartmorgan:
Makes the command to add desktop support more targeted using the new `--platforms` switch. That way it won't add or alter other platform directories.


sfshaza2 (approved):
LGTM

PR #5313, opened 2021-02-18
Migrate Web Codelab to Null Safety

miquelbeltran:
@sfshaza2  @RedBrogdon @theacodes 

In this PR I migrate the starter web codelab to null safety.

Some considerations regarding the code changes:

- An info message appears on the first DartPad, that's because one of the variables could be `final`, but it is going to be used as `var` later on, so I decided to leave it like that.
- Instead of using MaterialStateColor I had to switch to MaterialStateProperty because MaterialStateColor does not allow null values for Color.
- As you can see in the last commit, some of the changes I did originally have been reverted because DartPad was updated to the latest Beta.
- The code in the screenshots have not been updated. If you want me to redo all screenshots, let me know!

Live version available here: https://mb-dart-dev-2.firebaseapp.com/docs/get-started/codelab-web_migrated

### DIFF
--- a/src/docs/get-started/codelab-web_migrated.md
+++ b/src/docs/get-started/codelab-web_migrated.md
@@ -349,10 +349,14 @@ Fix the error reported by the analyzer that `_showWelcomeScreen`
 is not defined. Directly above the `build()` method,
 add the following function:
 
+{%- comment %}
+TODO Remove `!` when no longer necessary
+{% endcomment -%}
+
 <!-- skip -->
 ```dart
 void _showWelcomeScreen() {
-  Navigator.of(context).pushNamed('/welcome');
+  Navigator.of(context)!.pushNamed('/welcome');
 }
 ```
 </li>
@@ -879,7 +883,7 @@ class _SignUpFormState extends State<SignUpForm> {
   }
 
   void _showWelcomeScreen() {
-    Navigator.of(context).pushNamed('/welcome');
+    Navigator.of(context)!.pushNamed('/welcome');
   }
 
   @override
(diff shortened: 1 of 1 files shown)


theacodes:
> An info message appears on the first DartPad, that's because one of the variables could be final, but it is going to be used as var later on, so I decided to leave it like that.

Makes me wonder if it's possible for us to tell Dart Pad to ignore that. For now, can you change:

```
The starting app is displayed in the following DartPad.
```

to

```
The starting app is displayed in the following DartPad. You can ignore the "Private field could be final" message- that will be addressed in a later step.
```

> Instead of using MaterialStateColor I had to switch to MaterialStateProperty because MaterialStateColor does not allow null values for Color.

That's fine, it's the way it's demonstrated in the [documentation for ButtonStyle](https://api.flutter.dev/flutter/material/ButtonStyle-class.html).

> The code in the screenshots have not been updated. If you want me to redo all screenshots, let me know!

It doesn't look like the code in the screenshot matters much, as they're just there to point out specific UI elements. So this seems fine (but someone else is welcome to disagree)

RedBrogdon:
There is indeed a way you can get DartPad to ignore that lint. Because you're using an embed, you have the option of including a test.dart file that will be appended to the user's code when executed or analyzed. Even if you don't want to put any code in that file, you can still add analyzer directives to it:

```
// ignore_for_file: prefer_final_locals
```

When that's appended to the user's code, the analyzer will ignore the lint.

FWIW, you can also just add this to the code the user can see, but we've had data from UXR suggesting it's better to leave the user code as uncluttered as possible.

miquelbeltran:
Thank you both, adding a test.dart file annotation did the trick!

I've updated the firebase hosted version as well.

sfshaza2 (commented):
This LGTM, but I'm going to let @theacodes or @RedBrogdon take a look.

RedBrogdon (approved):
Code LGTM. I'm glad the ignore trick worked out!

PR #5327, opened 2021-02-19
Update submodules

kwalrath:
examples/codelabs
site-shared
(looks like flutter doesn't get auto-updated)

/cc @kevmoo @domesticmouse


sfshaza2 (approved):
That is correct, we aren't autoupdated. I didn't even know that was possible. And, lately, I pretty much have to sync every day.  LGTM

PR #5332, opened 2021-02-20
Fix link section in Linux Snap page

domesticmouse:
The links section at the bottom of the page had a single broken link which broke most of the links in the page.


domesticmouse:
TBR: @sfshaza2 

Making sure this lands before freeze

PR #5333, opened 2021-02-20
Document Linux distribution requirements

domesticmouse:
Adding notes to the Desktop page detailing how the Linux desktop app is compiled, and the requirements for using the compiled binary on other systems.

### DIFF
--- a/src/desktop.md
+++ b/src/desktop.md
@@ -359,8 +359,8 @@ impact your distributable application.
 
 The executable can be found in your project under `build/linux/<build mode>/bundle/`. Alongside your 
 executable binary in the `bundle` directory there are two directories:
- - `lib` which contains required `.so` library files, and 
- - `data` which contains the data assets that your application relies on.
+ * `lib` contains the required `.so` library files
+ * `data` contains the application's data assets, such as fonts or images
 
 In addition to these files, your application also relies on various operating system libraries that your
 application has been compiled against. You can see the full list by running `ldd` against your application.
--- a/src/desktop.md
+++ b/src/desktop.md
@@ -357,14 +357,36 @@ impact your distributable application.
 
 #### Linux
 
+The executable can be found in your project under `build/linux/<build mode>/bundle/`. Alongside your 
+executable binary in the `bundle` directory there are two directories, `lib` which contains required `.so`
+library files, and the `data` directory that contains data assets that your application relies on.
+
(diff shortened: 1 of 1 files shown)


src/desktop.md
```diff
@@ -357,14 +357,36 @@ impact your distributable application.
 
 #### Linux
 
+The executable can be found in your project under `build/linux/<build mode>/bundle/`. Alongside your 
+executable binary in the `bundle` directory there are two directories, `lib` which contains required `.so`
```
stuartmorgan, line 361: Nit: s/,/:/

src/desktop.md
```diff
@@ -357,14 +357,36 @@ impact your distributable application.
...
+library files, and the `data` directory that contains data assets that your application relies on.
```
stuartmorgan, line 362: Nit: s/that/which/ for parallel structure.

stuartmorgan (approved):
LGTM with nits. Thanks!

src/desktop.md
domesticmouse: Re-worded, but I'm unsure if I like the new structure. I'll let @sfshaza2 word smith my prose =)

src/desktop.md
domesticmouse: Refactored the two points into a list, following the new `:` in the text.

src/desktop.md
```diff
@@ -358,8 +358,9 @@ impact your distributable application.
 #### Linux
 
 The executable can be found in your project under `build/linux/<build mode>/bundle/`. Alongside your 
-executable binary in the `bundle` directory there are two directories, `lib` which contains required `.so`
-library files, and the `data` directory that contains data assets that your application relies on.
+executable binary in the `bundle` directory there are two directories:
+ - `lib` which contains required `.so` library files, and 
+ - `data` which contains the data assets that your application relies on.
```
sfshaza2, line 363:
```suggestion
executable binary in the `bundle` directory there are two directories:
 * `lib` contains the required `.so` library files
 * `data` contains the application's data assets, such as fonts or images
```

sfshaza2 (changes requested):
Hopefully the suggested assets are good examples. If not, feel free to replace.

src/desktop.md
domesticmouse: Yup, much betterer

sfshaza2 (approved):
LGTM

PR #5378, opened 2021-03-02
Temporarily pin stable SDKs

athomas:


.github/workflows/build.yml
```diff
@@ -90,19 +90,28 @@ jobs:
...
           git pull
+      - name: Select flutter 1.22.6
+        if: matrix.sdk == 'stable'
+        run: |
+          cd flutter
+          git fetch --depth 1 origin 1.22.6:1.22.6
+          git checkout 1.22.6 --
+          git remote set-branches origin stable
+          git fetch --depth 1 origin stable
+          git branch --set-upstream-to origin/stable
```
filiph, line 114: [nit] I don't understand this. Why is it needed? A comment would be great.

filiph (approved):
LGTM with 2 nits.

.github/workflows/build.yml
```diff
@@ -90,19 +90,28 @@ jobs:
...
         with:
-          sdk: stable
+          sdk: 2.10.5
       - name: Install node dependencies
         run: npm install
       - name: Install dart dependencies
         run: dart pub get
       - name: Select flutter
-        if: matrix.sdk != 'submodule-commit'
+        if: matrix.sdk != 'submodule-commit' && matrix.sdk != 'stable'
```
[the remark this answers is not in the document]
athomas: I did add some comments, I hope that helps.

Though probably there is a better way. Flutter's tooling relies on clues in the git checkout to figure out which version it is and on what branch it is. GitHub Actions does a shallow, single-branch checkout which doesn't have all the metadata that is needed to switch Flutter channels properly. Normally, we'd have to fetch the entire repo. This code tries to preserve the shallow checkout and checkout only what's needed. Perhaps this could be simpler if we removed the flutter submodule, and then we could directly checkout the right branch instead.

On Travis, we did roughly this:
* Checkout stable (full checkout that fetches everything)
* Run flutter doctor to get the flutter tool.
* Use the flutter tool to switch channels (which internally does similar things as these git commands).
* Then, when flutter was used somewhere, it would fetch additional artifacts for the selected release channel.

Perhaps there's a better way to get a specific Flutter version, but with the tight integration with a git checkout I don't know any. For the Dart SDK, it's just "forge a URL to the zip file, then download and unpack". It would be possible to go back to the Travis approach, but that is rather inefficient.

.github/workflows/build.yml
athomas: I added some comments that hopefully make this clearer. The point of this is to ensure the local branch for the pinned version has the proper upstream. The Flutter tool runs `git rev-parse --abbrev-ref --symbolic @{u}` to figure out that it is on the stable channel, and that requires the right upstream.

sfshaza2:
Can you take another look, @filiph?

PR #5388, opened 2021-03-03
Remove outdated lockdownd error note

jmagman:
`ERROR: Could not connect to lockdownd, error code -17` was replaced with a much nicer error as of https://github.com/flutter/flutter/pull/49854 in Flutter v1.16.3.


sfshaza2 (approved):
LGTM

PR #5389, opened 2021-03-03
Remove "flutter build ios" from archive build instructions

jmagman:
These instructions are no longer necessary as of Flutter 2.0 with `flutter build ipa`.


sfshaza2 (approved):
LGTM

PR #5391, opened 2021-03-03
Remove Xcode 10 signing reference

jmagman:
Xcode 11 is the minimum allowed version as of https://github.com/flutter/flutter/pull/50315 in Flutter v1.16.3.  Remove Xcode 10 reference.


sfshaza2 (approved):
LGtm

PR #5399, opened 2021-03-04
Update firebase.json to add Deprecation StatusTransitionWidget

Sameerkash:
This adds the proposal for deprecation of StatusTransitionWidget.


sfshaza2 (approved):
LGTM

PR #5406, opened 2021-03-04
Fixing some errors on the breaking change page for buttons

sfshaza2:
Staged here:  https://sz-flutter.firebaseapp.com/docs/release/breaking-changes/buttons


HansMuller (approved):
Thank you for the corrections and the all the reformatting! LGTM

PR #5414, opened 2021-03-05
Migration of "internationalization" examples to null safety

miquelbeltran:
In this PR I migrate the small projects inside the internationalization folder in the examples to null safety.

Most code changes are due to the fact that `Localizations.of<..>` returns a nullable type. I decided to use the "bang operator" to force it to not being nullable, since the original code wasn't taking care of null cases, but also IMO a null value in localizations would mean a bug in code, so failing silently would be counterproductive. I want to hear your opinion on this :)

To test the code, I run the examples as a Flutter web app, as there were no tests in the files.

@sfshaza2 @RedBrogdon

### DIFF
--- a/null_safety_examples/internationalization/.gitignore
+++ b/null_safety_examples/internationalization/.gitignore
@@ -1,0 +1,8 @@
+.DS_Store
+pubspec.lock
+
+*/**/android
+*/**/ios
+*/**/web
+*/**/.gitignore
+*/**/.metadata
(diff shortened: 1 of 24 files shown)


miquelbeltran:
Saw the warning, working on it. Somehow it did not appear here locally. sorry!

domesticmouse:
Can you please bring this PR up to date with master? It's Github Actions is currently pinned to an old version of Flutter, which is why the tests are failing.

domesticmouse (approved):
LGTM, once tests pass

sfshaza2:
Thanks, @domesticmouse!

PR #5416, opened 2021-03-05
Add "required" to example for null-safety

hungrybluedev:
The second example in this tutorial is almost entirely correct except for the compiler complaining of the title being possibly null. Simply adding the `required` keyword fixes the issue.

### DIFF
--- a/src/docs/development/ui/widgets-intro.md
+++ b/src/docs/development/ui/widgets-intro.md
@@ -98,7 +98,7 @@ Below are some simple widgets that combine these and other widgets:
 import 'package:flutter/material.dart';
 
 class MyAppBar extends StatelessWidget {
-  MyAppBar({this.title});
+  MyAppBar({required this.title});
 
   // Fields in a Widget subclass are always marked "final".
 


hungrybluedev:
@googlebot I signed it!

domesticmouse (approved):

kwalrath:
Not sure why this broke the build, but if it's a good change, we should figure out how to reinstate it and make the build _not_ break.

miquelbeltran:
The problem is that markdown documents like **src/docs/development/ui/widgets-intro.md** are not being checked with the null safety flag, so we cannot simply add null safety to the code samples.

What we did before was to add the `<!-- skip -->` tag to the code snippet so it won't be checked by CI.

kwalrath:
I'm wondering what we can do to make these snippets work. Just `<!-- skip -->` is sad, since it's unclear why it's being skipped, and the code doesn't get tested.

Can we somehow add a marker for null-safe or un-null-safe code, so that snippets are tested the right way? Or should we stage a major migration to using code-excerpt (which is what dart.dev does, but it does make contributing more complicated).

hungrybluedev:
I'm happy (and quite surprised) that such a minor addition can potentially lead to a more robust documentation framework. Good luck to everyone assigned to this issue! 😄

sfshaza2:
I have to agree that there is too much reliance on the skip tag. I know it bothers @RedBrogdon, too. The less we use it, the better.

PR #5421, opened 2021-03-06
Clarify & add detail for Windows distribution

Merrit:
Clarify where the dlls for the Visual C++ Redistributable should go (current wording can be read to say they should be in their own folder under the executable..)

Also quick mention about using regular Windows installers being a possibility. (I just did this with Inno Setup, and it was trivially easy - definitely less hassle than when I tried MSIX, so seems worth a mention as an option.)


Merrit:
@googlebot I signed it!

sfshaza2 (approved):
This is a nice addition. Thx!  LGTM

PR #5442, opened 2021-03-09
update web support info

mariamhas:
in the IDE section it still said you had to enable web support, but it is now a stable platform

### DIFF
--- a/src/desktop.md
+++ b/src/desktop.md
@@ -242,8 +242,7 @@ desktop, make sure you restart the IDE if it was
 already running.
 
 Create a new app in your IDE and it automatically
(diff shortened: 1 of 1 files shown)


src/desktop.md
```diff
@@ -242,8 +242,7 @@ desktop, make sure you restart the IDE if it was
 already running.
 
 Create a new app in your IDE and it automatically
-creates iOS, Android, and desktop versions of your app.
-(And web, too, if you've enabled [web support][].)
+creates iOS, Android, web and desktop versions of your app.
```
sfshaza2, line 245:
```suggestion
creates iOS, Android, web, and desktop versions of your app.
```

PR #5443, opened 2021-03-09
Add IDs to the install cards

jayoung-lee:
I'd like to add an ID to each of the OS cards on the `https://flutter.dev/docs/get-started/install` page, so that I can track the click events on these cards more precisely via Google Tag Manager. 

image

For instance, "Install On Windows Button" tag will be triggered if there's a click on class=`card` with id=`install-windows`. The current setup only fires when users click on the text. 

Reference: "Clicks on any element" section on [this GTM document](https://support.google.com/tagmanager/answer/6106716?hl=en).


sfshaza2 (approved):
LGTM

PR #5450, opened 2021-03-10
Fix the build (URI deprecation)

mit-mit:


sfshaza2 (approved):
Thanks, @mit-mit! LGTM

PR #5463, opened 2021-03-11
Set experimental flag to true for dev CI jobs

miquelbeltran:
In this PR I have set the flag `experimental` to `true` for the `dev` null check CI job.

For some reason, it was set to false, this caused that new APIs would break the job when running null checks, for example here: https://github.com/flutter/website/runs/2083002981 in the `master` branch.

Potentially the `beta` null check CI job should also have it to `true`

Both the `beta` and `dev` jobs have `experimental` to `true` for non-null safety jobs.

@sfshaza2 @RedBrogdon

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,7 +47,7 @@ jobs:
           - name: "Dev channel: null safety code check"
             sdk: dev
             task: tool/check-code.sh --null-safety
-            experimental: false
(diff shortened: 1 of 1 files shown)


kwalrath (commented):
LGTM. Thank you, @miquelbeltran!

kwalrath (approved):

PR #5464, opened 2021-03-11
Remove ' ' from install card IDs

jayoung-lee:
Need to remove ' ' from the install card IDs


kwalrath (approved):
LGTM. The build break isn't related to this change, so I'm going to merge.

/cc @sfshaza2

PR #5478, opened 2021-03-12
Removed obsolete ButtonBar widget from the widget catalog

HansMuller:
The ButtonBar widget is obsolete, best not to mention it here.

Fixes https://github.com/flutter/flutter/issues/77706


sfshaza2 (approved):
Yes!  Buh-bye!

PR #5482, opened 2021-03-13
Explain why the Debugger tab is missing.

pcamp:
Add "Note: DevTools hides the Debugger tab if the app was launched from VS Code because VS Code has a built-in debugger."


sfshaza2 (commented):
LGTM, but I'd like @kenzieschmoll to verify.

kenzieschmoll (approved):

sfshaza2 (approved):
LGTM!

Issue #5488, opened 2021-03-14
'Mock dependencies using Mockito' page issue


i3games:
Page URL: https://flutter.dev/docs/cookbook/testing/unit/mocking.html
Page source: https://github.com/flutter/website/tree/master/src/docs/cookbook/testing/unit/mocking.md

Description of issue:

This page neeeds to be updated. 

(1) references a previous version of the https://flutter.dev/docs/cookbook/networking/fetch-data recipe. The current recipe fetches `Album`not `Post`

(2) does not contain workable code. It introduces a code snippet where `http.Client`provided as a parameter to the `fetchPost` function. This breaks the old code and it should show how and where the client is created.

(3) uses an outdated `http` package with substantial brealking changes to the currrent version.

The recipe needs a complete overhaul adressing (1) to (3).

sfshaza2:
@miquelbeltran, would you mind adding this to your list?

miquelbeltran:
Yep, the mockito codelab will need some work indeed, because Mockito works differently as it is currently with null safety and requires to do code generation.

miquelbeltran:
Hey @i3games I have submitted a PR addressing your points, you can take a look at #5492 

Regarding point 2:

> This breaks the old code and the recipe does not show how and where the http client is created.

I think that this is out of the scope of the cookbook recipe, as the goal is to explain how mocks work rather how the `http` package works. In this case developers would have to use `http.Client()` to create a Client instance. But if it not obvious and @sfshaza2 agrees I can add a short explanation for that in the cookbook recipe.

sfshaza2:
I agree that this recipe is not about how to use the http package, but I also don't object to adding a single sentence or two.

miquelbeltran:
thanks! I will update the PR then

i3games:
I would recommend to add a full working example at the end as it is common in other Flutter recipes. The goal of a recipe should be self-contained and working code that I can take without having to understand how to build contextual dependencies.

sfshaza2:
This makes a lot of sense to me. What do you think, @miquelbeltran?

miquelbeltran:
I've added now the complete example, copying over the parts from the fetch-data recipe but modified to be testable

miquelbeltran:
this can be closed now! :)

PR #5492, opened 2021-03-16
Migrated "mocking" cookbook to null-safety

miquelbeltran:
In this PR I have both fixed the issues reported in #5488 as well as migrated the example to null safety.

The cookbook recipe required updating also because Mockito now uses code generation to support null safety, and some of the code samples were incomplete.

Please feel free to modify the text copy :)

@sfshaza2 @RedBrogdon


sfshaza2:
I'm not yet asking Brett or John to review because you might add a complete example, yes?

miquelbeltran:
I've added now a complete example, which includes the original code from the fetch-data cookbook recipe, but modified to be testable, and then the mock test at the end.

sfshaza2 (commented):
LGTM with one question. We usually try NOT to skip analysis on complete examples. Can we avoid this?

It looks good to me otherwise, but can you review this, @johnpryan?

miquelbeltran:
@sfshaza2 I've got an idea, I think I could move the full example into a project in the `null_safety_examples` and use the code excerpt html tag to integrate the sample. What do you think?

sfshaza2:
Yes, I like that!

johnpryan (approved):

PR #5494, opened 2021-03-16
Migrate the fetch-data cookbook to null-safety

miquelbeltran:
When I was fixing the Mocking cookbook recipe in https://github.com/flutter/website/pull/5492 I noticed that the "fetch-data" cookbook also had some issues.

In this PR I have fixed the outdated http code, as well as migrated the code example to null-safety.

I had to add `<!-- skip -->` annotation in one code example because we don't have a way to check null safe code in markdown yet. :cry: 

I've also updated the explanation on why we should throw an exception if the fetch data fails.

@sfshaza2 @RedBrogdon

### DIFF
--- a/src/docs/cookbook/networking/fetch-data.md
+++ b/src/docs/cookbook/networking/fetch-data.md
@@ -95,7 +95,7 @@ class Album {
   final int id;
   final String title;
 
-  Album({this.userId, this.id, this.title});
+  Album({required this.userId, required this.id, required this.title});
 
   factory Album.fromJson(Map<String, dynamic> json) {
     return Album(
@@ -129,7 +129,7 @@ function to return a `Future<Album>`:
 import 'dart:convert';
 
 Future<Album> fetchAlbum() async {
-  final response = await http.get('https://jsonplaceholder.typicode.com/albums/1');
+  final response = await http.get(Uri.https('jsonplaceholder.typicode.com', 'albums/1'));
 
   if (response.statusCode == 200) {
     // If the server did return a 200 OK response,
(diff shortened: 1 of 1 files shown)


sfshaza2:
Can you review this, @johnpryan?

src/docs/cookbook/networking/fetch-data.md
```diff
@@ -190,18 +190,22 @@ You must provide two parameters:
...
+Otherwise, the spinner will be displayed.
 
 <!-- skip -->
 ```dart
 FutureBuilder<Album>(
   future: futureAlbum,
   builder: (context, snapshot) {
     if (snapshot.hasData) {
-      return Text(snapshot.data.title);
+      return Text(snapshot.data!.title);
```
johnpryan, line 208: If only type flow analysis was smart enough to know `data` can't be null here :)

src/docs/cookbook/networking/fetch-data.md
```diff
@@ -161,7 +161,7 @@ See [`State`][] for more details.
 <!-- skip -->
 ```dart
 class _MyAppState extends State<MyApp> {
-  Future<Album> futureAlbum;
+  late Future<Album> futureAlbum;
```
johnpryan, line 164: nit: this could be `late final` I think

src/docs/cookbook/networking/fetch-data.md
```diff
@@ -281,7 +286,7 @@ class MyApp extends StatefulWidget {
 }
 
 class _MyAppState extends State<MyApp> {
-  Future<Album> futureAlbum;
+  late Future<Album> futureAlbum;
```
johnpryan, line 289: nit: this could be `late final` here too

johnpryan (approved):
LGTM

sfshaza2:
Thanks!

PR #5504, opened 2021-03-17
Revert "Add libblkid / liblzma / libgcrypt to Linux requirements"

MarcusTomlinson:
Following on from [this conversation](https://github.com/flutter/flutter/pull/77926#issuecomment-799189661):

Please accept my sincerest apologies for this! The last few updates to `desktop.md` are actually only [required for the snap package](https://github.com/canonical/flutter-snap/commit/e2e65acc4516b03b396a1bc09dfdfe567b68960d) and thus should not have been made upstream. This PR reverts those changes.

See also: https://github.com/flutter/flutter/pull/78415

### DIFF
--- a/src/desktop.md
+++ b/src/desktop.md
@@ -109,8 +109,6 @@ you need the following in addition to the Flutter SDK:
 * [GTK development headers][]
 * [Ninja build][]
 * [pkg-config][]
-* libblkid
-* liblzma
 
 The easiest way to install the Flutter SDK along with these
 dependencies is by using [snapd][].
@@ -127,7 +125,7 @@ If `snapd` is unavailable on the Linux distro you're using,
 you might use the following command:
 
 ```terminal
-$ sudo apt-get install clang cmake ninja-build pkg-config libgtk-3-dev libblkid-dev liblzma-dev
+$ sudo apt-get install clang cmake ninja-build pkg-config libgtk-3-dev
 ```
 
 [Clang]: https://clang.llvm.org/
--- a/src/desktop.md
+++ b/src/desktop.md
@@ -111,7 +111,6 @@ you need the following in addition to the Flutter SDK:
 * [pkg-config][]
 * libblkid
 * liblzma
-* libgcrypt
 
 The easiest way to install the Flutter SDK along with these
 dependencies is by using [snapd][].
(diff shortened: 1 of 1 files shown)


sfshaza2 (commented):
Ah, thanks for reverting this. I am not familiar with snapd and Linux tooling. As it's in draft mode, I won't otherwise comment on it right now.

MarcusTomlinson:
> Ah, thanks for reverting this. I am not familiar with snapd and Linux tooling. As it's in draft mode, I won't otherwise comment on it right now.

Yeah, let's wait to see if [this](https://github.com/flutter/flutter/pull/78415) gets approved before making changes to the docs. Thanks!

sfshaza2:
SG

MarcusTomlinson:
@sfshaza2 this can merge. Thanks!

sfshaza2 (approved):
LGTM

stuartmorgan:
We should probably temporarily revert this change; I didn't think about it at the time, but many users are now using desktop in the beta and stable channels, where the other libraries are still necessary. I've seen a few issues come in where prior followed these instructions, but then we're missing libraries, because they were in stable.

This should be changed only once the Flutter changes that go with it reach stable.

MarcusTomlinson:
> We should probably temporarily revert this change

The only library that legitimately should be on this list is `liblzma` (indefinitely - not temporarily). This is because some distros (such as Fedora 33+ & Ubuntu 20.04+) have since removed `liblzma` from their default installed packages.

I've reverted this [here](https://github.com/flutter/flutter/pull/84710) and [here](https://github.com/flutter/website/pull/5931).

The others (`libblkid` & `libgcrypt`) should never have been added upstream as those only ever showed up in snap usage.

stuartmorgan:
> > We should probably temporarily revert this change
> 
> The only library that legitimately should be on this list is `liblzma`

Well, now that 2.2 has shipped my comment is moot.
 
> The others (`libblkid` & `libgcrypt`) should never have been added upstream as those only ever showed up in snap usage.

This statement is not correct, which is why the instructions should not actually have been changed while 2.0 was the stable release. The *motivation* to add them was for snap usage, but they were made [requirements in every single developer's app templates](https://github.com/flutter/flutter/blob/2.0.6/packages/flutter_tools/templates/app/linux.tmpl/flutter/CMakeLists.txt#L27-L28) regardless of whether they were using the snap or not. Which means that not having them installed prevented compilation.

MarcusTomlinson:
Yes, they were made requirements. This is what I was referring to when I said they “should never have been added”.

PR #5515, opened 2021-03-19
Migrated forms validation cookbook to null safety

miquelbeltran:
In this PR I have migrated the Forms Validation to null safety.

The code has been moved into the null_safety_examples and imported into the file using code-excerpt extraction to support null safety.

As well, the dart-pad has null safety enabled by default too.

@sfshaza2 @RedBrogdon


sfshaza2:
@johnpryan, can you review?

null_safety_examples/cookbook/forms/validation/lib/main.dart
```diff
@@ -0,0 +1,80 @@
...
+            // #docregion ElevatedButton
+            child: ElevatedButton(
+              onPressed: () {
+                // Validate returns true if the form is valid, or false otherwise.
+                if (_formKey.currentState!.validate()) {
+                  // If the form is valid, display a snackbar. In the real world,
+                  // you'd often call a server or save the information in a database.
+                  ScaffoldMessenger.of(context)
+                      .showSnackBar(SnackBar(content: Text('Processing Data')));
+                }
```
johnpryan, line 70: (nitpick) you could use type flow analysis instead of `!`:

```suggestion
              onPressed: () {
                var state = _formKey.currentState;
                // Validate returns true if the form is valid, or false
                // otherwise.
                if (state != null && state.validate()) {
                  // If the form is valid, display a snackbar. In the real
                  // world, you'd often call a server or save the information in
                  // a database.
                  ScaffoldMessenger.of(context)
                      .showSnackBar(SnackBar(content: Text('Processing Data')));
                }
```

johnpryan (approved):
LGTM

sfshaza2:
Thanks @miquelbeltran and @johnpryan. @miquelbeltran, I'll leave it to you to commit John's suggestions if you agree. Otherwise, we can talk here...

sfshaza2:
@miquelbeltran, just let me know when you are ready to merge.

sfshaza2:
OK, I committed @johnpryan's suggestion. I hope that's ok, @miquelbeltran!

sfshaza2:
Ah, I was hoping to merge this, but it looks like it broke on John's suggested changes. So I'll just leave it here and let miquel sort it out on Monday. Sorry!

null_safety_examples/cookbook/forms/validation/lib/main.dart
miquelbeltran: The reason I prefer the `!` instead of checking for null here is because a null `currentState` means that there is a bug in the code (e.g. the developer forgot to set the `_formKey` to the `Form` widget) and so the code should fail with an error rather than silently.

So I would write in production code is indeed `!`, the same way we didn't check for null `currentState` before null-safety.

miquelbeltran:
@sfshaza2 the reason your change didn't work is because the code excerpts need to be updated too, but IMO there's shouldn't be a null check there, so I will revert the commit meanwhile.

miquelbeltran:
The current CI issue doesn't seem to be related. Is it possible to re-run the workflow?

johnpryan:
Sounds good, re-running the workflow

johnpryan (approved):

PR #5518, opened 2021-03-19
Make doc compatible with null safety

ColonisationCaptain:
The key parameter should be nullable


sfshaza2:
I don't think this can be fixed unless we migrate this file (and it's a huge one) to use code-excerpts. Right, @miquelbeltran? I'm not sure that's such a priority right now. Or I suppose we could use the skip tag, but I doubt that this is the only code in this file that doesn't use null safety.

sfshaza2:
Given that this is a much bigger issue than what is covered by this PR, I'm closing this.

PR #5520, opened 2021-03-19
Correct PATH modification instructions

cbracken:
Previously we asked users to modify their path as follows:

    export PATH="$PATH:[PATH_TO_FLUTTER_GIT_DIRECTORY]/flutter/bin"

The correct location is just `bin` relative to the main Flutter
directory, so the correct instructions should be:

    export PATH="$PATH:[PATH_TO_FLUTTER_GIT_DIRECTORY]/bin"

I've also made some minor wording changes to remove any ambiguity that
`PATH_TO_FLUTTER_GIT_DIRECTORY` is actually the path of the cloned
`flutter` repo and not the path into which the clone was landed.

This was a potential source of confusion/ambiguity since the initial
install instructions do suggest they run the following command to
temporarily add Flutter to their PATH after the initial install.

    export PATH="$PATH:`pwd`/flutter/bin"

Credit for spotting this error goes to [christiannc on StackOverflow][so].

[so]: https://stackoverflow.com/questions/66713077/setting-up-flutter-in-terminal-yields-a-command-not-found-message/66716562#66716562

### DIFF
--- a/src/docs/get-started/install/_path-linux-chromeos.md
+++ b/src/docs/get-started/install/_path-linux-chromeos.md
@@ -10,7 +10,7 @@ all terminal sessions are machine-specific.
 Typically you add a line to a file that is executed
 whenever you open a new window. For example:
 
- 1. Determine the directory where you placed the Flutter SDK.
+ 1. Determine the path of your clone of the Flutter SDK.
     You need this in Step 3.
  2. Open (or create) the `rc` file for your shell.
     For example, Linux uses the Bash shell by default,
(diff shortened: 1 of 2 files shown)


sfshaza2 (approved):
Mucho excellente. Thanks, @cbracken! LGTM

src/docs/get-started/install/_path-linux-chromeos.md
```diff
@@ -10,19 +10,19 @@ all terminal sessions are machine-specific.
 Typically you add a line to a file that is executed
 whenever you open a new window. For example:
 
- 1. Determine the directory where you placed the Flutter SDK.
+ 1. Determine the path of your clone of the Flutter SDK.
```
christopherfujino, line 13: Note that the user may have downloaded the prepackaged archive, and not cloned it themselves.

src/docs/get-started/install/_path-linux-chromeos.md
```diff
@@ -10,19 +10,19 @@ all terminal sessions are machine-specific.
...
     and filename will be different on your machine.
  3. Add the following line and change
-    `[PATH_TO_FLUTTER_GIT_DIRECTORY]` to be
-    the path where you cloned Flutter's git repo:
+    `[PATH_OF_FLUTTER_GIT_DIRECTORY]` to be
+    the path of your clone of the Flutter git repo:
 
     ```terminal
-    $ export PATH="$PATH:[PATH_TO_FLUTTER_GIT_DIRECTORY]/flutter/bin"
+    $ export PATH="$PATH:[PATH_OF_FLUTTER_GIT_DIRECTORY]/bin"
```
christopherfujino, line 25: d'oh!

src/docs/get-started/install/_path-linux-chromeos.md
cbracken: Ah that's a good point, the surrounding directions refer to it as a git clone, but we should probably fix that. (go for it!)

PR #5521, opened 2021-03-20
add null check in "Build a form with validation"

G1Joshi:


sfshaza2 (commented):
Thanks for this PR, @G1Joshi! @miquelbeltran is in the process of migrating this code to use our code-excerpts process and it can then be updated to null safety and also be tested. Closing this PR.

PR #5531, opened 2021-03-23
Typo fix in plugin-api-migration.md

geisterfurz007:
`onAttachToEngine` won't be called because the method doesn't exist.


geisterfurz007:
@googlebot I signed it!

sfshaza2 (approved):
lgtm

PR #5532, opened 2021-03-23
Migrated Internationalization documentation to null safety

miquelbeltran:
This was a tough one :) 

In this PR I have migrated the internationalization.md to null safety, as the projects that this document uses were already migrated the work has been mostly using code-excerpts to extract the right lines of code.

In some cases I had to add extra code to show different examples. One section had to be reworded a bit to show the example.
Mostly all the code is extracted in some way.

There is one part in which the doc asks the reader to switch the device to Arabic to see the locale at work, but there was no demo built showing that.

After chatting with @RedBrogdon about this, we decided to remove the mention to Arabic and instead mention that you can switch the device language to Spanish to check that the app text changes, as Spanish is already provided in the sample code.

@sfshaza2

Live version: https://mb-flutter-dev-1.firebaseapp.com/docs/development/accessibility-and-localization/internationalization

### DIFF
--- a/null_safety_examples/internationalization/add_language/lib/main.dart
+++ b/null_safety_examples/internationalization/add_language/lib/main.dart
@@ -5,6 +5,7 @@ import 'nn_intl.dart';
 
 void main() {
   runApp(
+    // #docregion MaterialApp
     MaterialApp(
       localizationsDelegates: [
         GlobalWidgetsLocalizations.delegate,
@@ -15,6 +16,7 @@ void main() {
         const Locale('en', 'US'),
         const Locale('nn'),
       ],
+      // #enddocregion MaterialApp
       home: Home(),
     ),
   );
--- a/null_safety_examples/internationalization/add_language/lib/main.dart
+++ b/null_safety_examples/internationalization/add_language/lib/main.dart
@@ -16,9 +16,9 @@ void main() {
         const Locale('en', 'US'),
         const Locale('nn'),
       ],
-      // #enddocregion MaterialApp
       home: Home(),
     ),
+    // #enddocregion MaterialApp
   );
 }
 
--- a/src/docs/development/accessibility-and-localization/internationalization.md
+++ b/src/docs/development/accessibility-and-localization/internationalization.md
@@ -711,6 +711,8 @@ MaterialApp(
     const Locale('en', 'US'),
     const Locale('nn'),
   ],
+  home: Home(),
+),
 ```
 
 <a name="alternative-internationalization-workflows">
--- a/src/docs/development/accessibility-and-localization/internationalization.md
+++ b/src/docs/development/accessibility-and-localization/internationalization.md
@@ -136,6 +136,9 @@ following instructions to add localized text to your application.
 
 1. Add the `intl` package to the `pubspec.yaml` file:
 
+{% comment %}
+RegEx removes "# Add this line" from lines "flutter_localizations:" and "sdk: flutter"
+{% endcomment %}
    <?code-excerpt "gen_l10n_example/pubspec.yaml (Intl)" replace="/(?<!0) # Add this line//g" ?>
    ```yaml
    dependencies:
@@ -599,16 +602,24 @@ The date patterns and symbols of the locale will also need to
 be specified. In the source code, the date patterns and symbols
 are defined like this:
 
-<?code-excerpt "add_language/lib/nn_intl.dart (Date)"?>
+{% comment %}
+RegEx adds last two lines with commented out code and closing bracket.
+{% endcomment %}
+<?code-excerpt "add_language/lib/nn_intl.dart (Date)" replace="/  'LLL': 'LLL',/  'LLL': 'LLL',\n  \/\/ ...\n}/g"?>
 ```dart
 const nnLocaleDatePatterns = {
   'd': 'd.',
   'E': 'ccc',
   'EEEE': 'cccc',
   'LLL': 'LLL',
+  // ...
+}
 ```
 
-<?code-excerpt "add_language/lib/nn_intl.dart (Date2)"?>
+{% comment %}
+RegEx adds last two lines with commented out code and closing bracket.
+{% endcomment %}
+<?code-excerpt "add_language/lib/nn_intl.dart (Date2)" replace="/  ],/  ],\n  \/\/ ...\n}/g"?>
 ```dart
 const nnDateSymbols = {
   'NAME': 'nn',
@@ -616,6 +627,8 @@ const nnDateSymbols = {
     'f.Kr.',
     'e.Kr.',
   ],
+  // ...
+}
 ```
 
 These will need to be modified for the locale to use the correct
--- a/src/docs/development/accessibility-and-localization/internationalization.md
+++ b/src/docs/development/accessibility-and-localization/internationalization.md
@@ -637,14 +637,27 @@ not share the same flexibility for number formatting, the formatting
 for an existing locale will have to be used as a substitute in
 `_NnMaterialLocalizationsDelegate`:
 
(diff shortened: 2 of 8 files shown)


sfshaza2 (commented):
@shihaohong, @miquelbeltran has updated this page to null safety. It LGTM, but can you cast an eye over this?

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -636,11 +630,6 @@ class _NnMaterialLocalizationsDelegate extends LocalizationsDelegate<MaterialLoc
         // for 'en_US' instead.
         decimalFormat: intl.NumberFormat('#,##0.###', 'en_US'),
         twoDigitZeroPaddedFormat: intl.NumberFormat('00', 'en_US'),
-        //...
```
shihaohong, line 639: Same question about removing the closing brackets

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -619,11 +611,13 @@ not share the same flexibility for number formatting, the formatting
 for an existing locale will have to be used as a substitute in
 `_NnMaterialLocalizationsDelegate`:
 
-<!-- skip -->
+<?code-excerpt "add_language/lib/nn_intl.dart (Delegate)" replace="/@(.|\n)*';\n\n//g;/    final (.|\n)*\);\n\n//g"?>
```
shihaohong, line 614: Could you add a comment about what the `replace` here does?

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -135,29 +132,32 @@ following instructions to add localized text to your application.
 
 1. Add the `intl` package to the `pubspec.yaml` file:
 
+   <?code-excerpt "gen_l10n_example/pubspec.yaml (Intl)" replace="/(?<!0) # Add this line//g" ?>
```
shihaohong, line 135: Add a comment on what this `replace` does

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -69,31 +72,25 @@ dependencies:
...
+  ],
+  supportedLocales: [
     const Locale('en', ''), // English, no country code
-    const Locale('ar', ''), // Arabic, no country code
-    const Locale.fromSubtags(languageCode: 'zh'), // Chinese *See Advanced Locales below*
-    // ... other locales the app supports
+    const Locale('es', ''), // Spanish, no country code
   ],
-  // ...
-)
```
shihaohong, line 96: Should the closing brackets be removed?

shihaohong (commented):
The NNBD migration parts of the code look fine. 

I'm just unsure about some of the code snippet sections that were removed. I'm also not sure what `#docregion` does. One part was also modified incorrectly (ie `MaterialApp` became `Material` in one of the doc samples.

null_safety_examples/internationalization/add_language/lib/main.dart
```diff
@@ -5,6 +5,7 @@ import 'nn_intl.dart';
 
 void main() {
   runApp(
+    // #docregion MaterialApp
```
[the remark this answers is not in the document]
miquelbeltran: It is part of the code-excerpt logic to copy sections of code into markdown documents. This allows to keep the documents updated with code that stays on example projects. docregion defines where a snippet starts and end. You can learn more about it here: https://github.com/chalin/code_excerpt_updater

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -653,19 +642,18 @@ Here's some code that sets the app's language to Nynorsk and
 adds the `NnMaterialLocalizations` delegate instance to the app's
 `localizationsDelegates` list:
 
-<!-- skip -->
+<?code-excerpt "add_language/lib/main.dart (MaterialApp)"?>
```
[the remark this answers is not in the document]
miquelbeltran: When the `./tool/refresh-code-excerpts.sh` script runs, it copies over the section of code marked between `docregion` and `enddocregion` named `MaterialApp` from the `main.dart`.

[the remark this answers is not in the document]
shihaohong: Ahh got it, that makes sense. Thanks!

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -69,31 +72,25 @@ dependencies:
...
 ```dart
 import 'package:flutter_localizations/flutter_localizations.dart';
-// TODO: uncomment the line below after codegen
-// import 'package:flutter_gen/gen_l10n/app_localizations.dart';
+```
 
-// ...
+<?code-excerpt "gen_l10n_example/lib/main.dart (LocalizationDelegates)" replace="/App.*//g"?>
+```dart
+return Material
```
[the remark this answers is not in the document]
shihaohong: I wondered if the `replace` fixed this, but it shows in https://mb-flutter-dev-1.firebaseapp.com/docs/development/accessibility-and-localization/internationalization

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -69,31 +72,25 @@ dependencies:
...
+  title: 'Localizations Sample
```
shihaohong, line 83: Looks like the title is cut off as well: 
```suggestion
    title: 'Localizations Sample App',
```

src/docs/development/accessibility-and-localization/internationalization.md
```diff
@@ -579,38 +572,37 @@ like `r'About $applicationName'`,
...
+<?code-excerpt "add_language/lib/nn_intl.dart (Date2)"?>
+```dart
 const nnDateSymbols = {
   'NAME': 'nn',
   'ERAS': <dynamic>[
     'f.Kr.',
     'e.Kr.',
   ],
-  // ...
-}
```
shihaohong, line 613: If someone copy/pastes this, it'll result in an error since it's an incomplete map. Is there a way we can preserve the correctness of the objects here and elsewhere? I feel like this will likely be a source of friction for our users.

src/docs/development/accessibility-and-localization/internationalization.md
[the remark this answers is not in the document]
miquelbeltran: Was an error on my replace RegEx, I have simplified the code-excerpt a bit now.

src/docs/development/accessibility-and-localization/internationalization.md
miquelbeltran: fixed now on the latest commit

src/docs/development/accessibility-and-localization/internationalization.md
miquelbeltran: I've been trying to improve this a little bit more but it's very complicated to extract the section of code from the `_NnMaterialLocalizationsDelegate`, so I thought also that it would be beneficial to show the whole class instead instead of doing a very complex RegEx replace. What do you think?

src/docs/development/accessibility-and-localization/internationalization.md
miquelbeltran: I changed the script to extract the whole MaterialApp class to avoid confusion.

miquelbeltran:
Thanks for the review @shihaohong ! I have addressed your points. Also changed the way the `_NnMaterialLocalizationsDelegate` is extracted into the markdown document, let me know what you think :)

shihaohong (approved):
LGTM! I think leaving in the full sample despite it making it more verbose is better than leaving it short and perhaps confusing. Thanks for the PR :)

sfshaza2 (approved):
A huge thanks to you both, @miquelbeltran and @shihaohong!

PR #5543, opened 2021-03-25
add interactive example

G1Joshi:
add an interactive example for "Create and style a text field" cookbook

Issue #5534, opened 2021-03-23
'Visual Studio Code' page issue


RandalSchwartz:
Page URL: https://flutter.dev/docs/development/tools/vs-code.html
Page source: https://github.com/flutter/website/tree/master/src/docs/development/tools/vs-code.md

Description of issue:

Doesn't seem to mention Flutter Desktop yet.  I know, mere days after #flutterengage. :)

sfshaza2:
Yup! Also, the web note is now wrong. You no longer have to enable web, you just get it. As always, PRs welcome.  ;)

PR #5549, opened 2021-03-26
Removed web note from android-studio.md and vs-code.md

Tapaswi846580:
This PR removes the **"Web Note"** from **"Selecting a Target"** section in [Android Studio](https://flutter.dev/docs/development/tools/android-studio#selecting-a-target) and [VS-Code](https://flutter.dev/docs/development/tools/vs-code#selecting-a-target-device) tools.

Before change there was a note saying **"You need to enable web if Chrome(web) target doesn't exist"**  but now with flutter 2.0 you no longer have to enable web as it is already enabled.

This PR resolves [#5543 issue comment.](https://github.com/flutter/website/issues/5534#issuecomment-807901959)

### DIFF
--- a/src/docs/development/tools/android-studio.md
+++ b/src/docs/development/tools/android-studio.md
@@ -130,13 +130,6 @@ Flutter-specific buttons on the right-hand side of the toolbar.
     When you connect devices, or start simulators,
     additional entries appear.
 
-{{site.alert.note}}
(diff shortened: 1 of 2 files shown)


sfshaza2 (approved):
LGTM. Thanks for catching that, @Tapaswi846580!

PR #5550, opened 2021-03-26
Correct integration_test pub dependency

ened:
integration_test is now shipped as part of the SDK


johnpryan:
It looks like we need to update the docs for that package as well.

PR #5561, opened 2021-03-30
Add note that insecure http ban has been reverted

xster:
https://github.com/flutter/flutter/issues/72723

This is maybe a case study for a new use case. How should we keep track of breaking change reverts and the version it got reverted in.


sfshaza2:
Thanks for raising my awareness about this, @xster. I will add a "reverted" section to the index.

sfshaza2 (approved):
LGTM

PR #5566, opened 2021-03-31
Migrated widgets-intro.md to null safety

miquelbeltran:
In this PR I have migrated the widgets-intro.md to null safety.

All code samples have been moved into a project in the `null_safety_examples` folder.

After a chat with @RedBrogdon we decided that it would be a good idea to use DartPads instead of plain code to show the readers the examples in action.

To do so, I had to adapt the code samples a bit, adding a main method, missing code, etc. as well as adjust the code wrapping so it looks good in the DartPad. I also set the dartpads to be "run false" so they don't run when the page loads automatically, like we did on other pages with lots of dartpads.

cc. @sfshaza2 

Live version: https://mb-flutter-dev-1.web.app/docs/development/ui/widgets-intro

### DIFF
--- a/null_safety_examples/ui/widgets_intro/lib/main.dart
+++ b/null_safety_examples/ui/widgets_intro/lib/main.dart
@@ -1,0 +1,12 @@
+import 'package:flutter/material.dart';
+
+void main() {
+  runApp(
+    Center(
+      child: Text(
+        'Hello, world!',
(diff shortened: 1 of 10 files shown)


sfshaza2:
Can you review, @domesticmouse?

domesticmouse (approved):

PR #5600, opened 2021-04-08
fix null-safety on hot-reload.md

miquelbeltran:
Here I fixed the hot-reload.md for null safety, I didn't create an external code project since the code is minimal. The changes has been done on-file and the short snippets where already "skipped".

On the changes, I think that `?` for the Generics example is fine, as it matches the previous code functionality, alternatives could be to mark them as `late` or a constructor (but I didn't want to add too much noise to the code snippet).

The `get bar => foo` complained that required a return type, so I used `int` as `foo` is declared as an int in a previous example.

cc. @sfshaza2 @RedBrogdon

### DIFF
--- a/src/docs/development/tools/hot-reload.md
+++ b/src/docs/development/tools/hot-reload.md
@@ -155,7 +155,7 @@ Before the change:
 <!-- skip -->
 ```dart
 class A<T> {
-  T i;
+  T? i;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
LGTM! It's true that the code snippets on this page were tiny, so I am really fine with your approach.

PR #5601, opened 2021-04-08
Fix Material Design color link

hrishikesh-kadam:
https://material.io/guidelines/style/color.html shows 404 page.
https://material.io/design/color seems to be the appropriate.


sfshaza2 (approved):
Thanks for catching and fixing that, @hrishikesh-kadam! LGTM

PR #5603, opened 2021-04-09
Update references in layout tutorial to null safety examples

miquelbeltran:
I updated the references to the null safety examples in the layout tutorial document (/docs/development/ui/layout/tutorial), the layout index (/docs/development/ui/layout) and the "interactive" tutorial (src/docs/development/ui/interactive.md).

I also had to copy a small examples into the null_safety_examples because it was referencing the startup_namer codelab and causing issues with a diff, so I removed that dependency and instead made a copy of the file into the examples.

There's no code changes after running the refresh-code-excerpts script, but the references are correct now.

cc. @sfshaza2 @RedBrogdon


sfshaza2:
LGTM! @domesticmouse, can you cast your eye over this PR?

domesticmouse (approved):

PR #5610, opened 2021-04-11
Web support is now present in stable.

jayeshmann:


jayeshmann:
@googlebot I signed it!

sfshaza2 (changes requested):
Hi, @jayeshmann! Thanks for this PR, but it doesn't go far enough. When you create a new project in Flutter 2, the web is just there. So you don't really need those instructions at all. I'd delete pretty much all of that info.

sfshaza2 (approved):
Perfect!

PR #5612, opened 2021-04-12
Update errors.md with info about Zones errors handling

blazern:
Some errors that occur in a Flutter app cannot be caught by the Flutter framework.
When I first faced it, it took me some time to figure out that these errors can be caught by a Zone and how a Zone should be initialized.

Part of the added text is taken from this issue: https://github.com/flutter/flutter/issues/10761

### DIFF
--- a/src/docs/testing/errors.md
+++ b/src/docs/testing/errors.md
@@ -33,10 +33,16 @@ You can customize these behaviors,
 typically by setting them to values in
 your `void main()` function.
 
-## Quit application on encountering an error
+Below each error type handling is explained. At the bottom
+there's a code snippet which handles all types of errors. Even
+though you can just copy-paste the snippet, we recommend you
+to first get acquainted with each of the error types.
+
+## Errors caught by Flutter
 
 For example, to make your application quit immediately any time an
-error occurs in release mode, you could use the following handler:
+error is caught by Flutter in release mode, you could use the
+following handler:
 
 <!-- skip -->
 ```dart
@@ -65,7 +71,7 @@ This handler can also be used to report errors to a logging service.
 For more details, see our cookbook chapter for 
 [reporting errors to a service][].
 
-## Define a custom error widget
+## Define a custom error widget for build phase errors
 
 To define a customized error widget that displays whenever
 the builder fails to build a widget, use [`MaterialApp.builder`][].
@@ -90,11 +96,12 @@ class MyApp extends StatelessWidget {
 }
 ```
 
-## Set up error handling for a Zone
+## Errors not caught by Flutter
 
 Consider an `onPressed` callback that invokes an asynchronous function,
 such as `MethodChannel.invokeMethod` (or pretty much any plugin).
 For example:
+
 <!-- skip -->
 ```dart
 OutlinedButton(
@@ -105,26 +112,88 @@ OutlinedButton(
   },
 ),
 ```
+
 If `invokeMethod` throws an error, it won't be forwarded to `FlutterError.onError`.
 Instead, it's forwarded to the `Zone` where `runApp` was run.
 
 To catch such an error, use [`runZonedGuarded`][].
+
 <!-- skip -->
 ```dart
 import 'dart:async';
 
 void main() {
-  runZonedGuarded(() async {
-    WidgetsFlutterBinding.ensureInitialized();
-    runApp(MyApp()));
+  runZonedGuarded(() {
+    runApp(MyApp());
   }, (Object error, StackTrace stack) {
     myBackend.sendError(error, stack);
   });
 }
 ```
-Note that `WidgetsFlutterBinding.ensureInitialized()` is
-called within `runZonedGuarded`; error handling wouldn't work if
-the function were called from the outside.
+
+Note that if in your app you call `WidgetsFlutterBinding.ensureInitialized()`
+manually to perform some initialization before calling `runApp` (e.g.
+`Firebase.initializeApp()`), you **must** call
+`WidgetsFlutterBinding.ensureInitialized()` inside `runZonedGuarded`:
+
+<!-- skip -->
+```dart
+runZonedGuarded(() async {
+  WidgetsFlutterBinding.ensureInitialized();
+  await Firebase.initializeApp();
+  runApp(MyApp());
+}
+```
+
+{{site.alert.note}}
+    Error handling wouldn't work if `WidgetsFlutterBinding.ensureInitialized()`
+    was called from the outside.
+{{site.alert.end}}
+
+## Handling all types of errors
+
+Say you want to exit application on any exception and to display
+a custom error widget whenever a widget building fails - you can base
+your errors handling on next code snippet:
+
+<!-- skip -->
+```dart
+import 'dart:io';
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+
+void main() {
+  runZonedGuarded(() async {
+    WidgetsFlutterBinding.ensureInitialized();
+    await myErrorsHandler.initialize();
+    FlutterError.onError = (FlutterErrorDetails details) {
+      FlutterError.dumpErrorToConsole(details);
+      myErrorsHandler.onError(details);
+      exit(1);
+    };
+    runApp(MyApp());
+  }, (Object error, StackTrace stack) {
+    myErrorsHandler.onError(error, stack);
+    exit(1);
+  });
+}
+
+class MyApp extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return MaterialApp(
+      builder: (BuildContext context, Widget widget) {
+        Widget error = Text('...rendering error...');
+        if (widget is Scaffold || widget is Navigator)
+          error = Scaffold(body: Center(child: error));
+        ErrorWidget.builder = (FlutterErrorDetails errorDetails) => error;
+        return widget;
+      },
+    );
+  }
+}
+```
 
 [`ErrorWidget.builder`]: {{site.api}}/flutter/widgets/ErrorWidget/builder.html
 [`FlutterError.onError`]: {{site.api}}/flutter/foundation/FlutterError/onError.html
--- a/src/docs/testing/errors.md
+++ b/src/docs/testing/errors.md
@@ -5,9 +5,11 @@ description: How to control error messages and logging of errors
 
 The Flutter framework catches errors that occur during callbacks
 triggered by the framework itself, including errors encountered
-during the build, layout, and paint phases.
+during the build, layout, and paint phases. Errors that don't occur
+within Flutter's callbacks can't be caught by the framework,
+but you can handle them by setting up a [`Zone`][].
 
-All of these errors are routed to the
+All errors caught by Flutter are routed to the
 [`FlutterError.onError`][] handler. By default,
 this calls [`FlutterError.dumpErrorToConsole`][],
 which, as you might guess, dumps the error to the device logs.
@@ -22,6 +24,10 @@ invoked to build the widget that is used
 instead of the one that failed. By default,
 in debug mode this shows an error message in red,
 and in release mode this shows a gray background.
+
+When errors occur without a Flutter callback on the call stack,
+they are handled by the `Zone` where they occur. By default,
+a `Zone` only prints errors and does nothing else.
 
 You can customize these behaviors,
 typically by setting them to values in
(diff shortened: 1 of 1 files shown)


blazern:
@googlebot I signed it!

src/docs/testing/errors.md
```diff
@@ -5,9 +5,11 @@ description: How to control error messages and logging of errors
 
 The Flutter framework catches errors that occur during callbacks
 triggered by the framework itself, including errors encountered
-during the build, layout, and paint phases.
+during the build, layout, and paint phases. Errors which occur
+not within Flutter's callbacks cannot be caught by the framework,
+but you can handle them by setting up a Zone.
```
sfshaza2, line 10:
```suggestion
during the build, layout, and paint phases. Errors that don't occur
within Flutter's callbacks can't be caught by the framework,
but you can handle them by setting up a [`Zone`][].
```

src/docs/testing/errors.md
```diff
@@ -5,9 +5,11 @@ description: How to control error messages and logging of errors
...
 
-All of these errors are routed to the
+All of errors caught by Flutter are routed to the
```
sfshaza2, line 12:
```suggestion
All errors caught by Flutter are routed to the
```

src/docs/testing/errors.md
```diff
@@ -23,6 +25,10 @@ instead of the one that failed. By default,
 in debug mode this shows an error message in red,
 and in release mode this shows a gray background.
 
+When errors occur without a Flutter callback on the call
+stack they are handled by the Zone they occur in. By default
+Zones only print errors and do nothing else.
```
sfshaza2, line 30:
```suggestion
When errors occur without a Flutter callback on the call stack,
they are handled by the `Zone` where they occur. By default,
a `Zone` only prints errors and does nothing else.
```

src/docs/testing/errors.md
```diff
@@ -84,9 +90,45 @@ class MyApp extends StatelessWidget {
 }
 ```
 
+## Set up a Zone errors handling
```
sfshaza2, line 93:
```suggestion
## Set up error handling for a Zone
```

src/docs/testing/errors.md
```diff
@@ -84,9 +90,45 @@ class MyApp extends StatelessWidget {
...
+
+Consider an `onPressed` callback that invokes an asynchronous function,
+such as `MethodChannel.invokeMethod` (or pretty much any plugin):
```
sfshaza2, line 96:
```suggestion
such as `MethodChannel.invokeMethod` (or pretty much any plugin).
For example:
```

src/docs/testing/errors.md
```diff
@@ -84,9 +90,45 @@ class MyApp extends StatelessWidget {
...
+<!-- skip -->
+```dart
+OutlinedButton(
+  child: Text('Click me!'),
+  onPressed: () async {
+    final channel = const MethodChannel('crashy-custom-channel');
```
sfshaza2, line 102: This made me LOL.  :D

src/docs/testing/errors.md
```diff
@@ -84,9 +90,45 @@ class MyApp extends StatelessWidget {
...
+    await channel.invokeMethod('blah');
+  },
+),
+```
+If `invokeMethod` throws an error, it will not be forwarded to `FlutterError.onError`.
```
sfshaza2, line 107:
```suggestion
If `invokeMethod` throws an error, it won't be forwarded to `FlutterError.onError`.
```

src/docs/testing/errors.md
```diff
@@ -84,9 +90,45 @@ class MyApp extends StatelessWidget {
...
+Instead, it will be forwarded to the `Zone` where `runApp` was run.
```
sfshaza2, line 108:
```suggestion
Instead, it's forwarded to the `Zone` where `runApp` was run.
```

src/docs/testing/errors.md
```diff
@@ -84,9 +90,45 @@ class MyApp extends StatelessWidget {
...
+    WidgetsFlutterBinding.ensureInitialized();
+    runApp(MyApp()));
+  }, (Object error, StackTrace stack) {
+    myBackend.sendError(error, stack);
+  });
+}
+```
+Please note that `WidgetsFlutterBinding.ensureInitialized()` is
+called within `runZonedGuarded` - error handling wouldn't work if
+the function would be called from the outside.
```
sfshaza2, line 126:
```suggestion
Note that `WidgetsFlutterBinding.ensureInitialized()` is
called within `runZonedGuarded`; error handling wouldn't work if
the function were called from the outside.
```

sfshaza2 (commented):
What a great tip, @blazen! Thanks for updating the docs with this valuable insight. With my suggested changes, it LGTM, but I would like @domesticmouse to also take a look.

src/docs/testing/errors.md
```diff
@@ -84,9 +90,45 @@ class MyApp extends StatelessWidget {
...
+
 [`ErrorWidget.builder`]: {{site.api}}/flutter/widgets/ErrorWidget/builder.html
 [`FlutterError.onError`]: {{site.api}}/flutter/foundation/FlutterError/onError.html
 [`FlutterError.dumpErrorToConsole`]: {{site.api}}/flutter/foundation/FlutterError/dumpErrorToConsole.html
 [`kReleaseMode`]:  {{site.api}}/flutter/foundation/kReleaseMode-constant.html
 [`MaterialApp.builder`]: {{site.api}}/flutter/material/MaterialApp/builder.html
 [reporting errors to a service]: /docs/cookbook/maintenance/error-reporting
+[`runZonedGuarded`]: {site.api}}/flutter/dart-async/runZonedGuarded.html
```
sfshaza2, line 134: Add a [`Zone`] link to this section.

Also, the `runZonedGuarded` link is causing the build failure. It should be ``{{site.api}}``, not `{site.api}``.

domesticmouse (approved):
LGTM once @sfshaza2's concerns are dealt with =)

esDotDev:
I'm not clear from the docs whether Zone catches all errors, or if we still need to FlutterError.onError callback.

Also would be nice if they cut to the chase a bit, and led off with Global Error Handling section. Something like:

To catch all errors that can occur in your FlutterApp, use the following code:
```
void main() {
  void handleError(Object error, StackTrace stack) => myBackend.sendError(error, stack);
  FlutterError.onError = (FlutterErrorDetails details) => handleError(details.exception, details.stack)
  runZonedGuarded(() async {
    WidgetsFlutterBinding.ensureInitialized();
    runApp(MyApp()));
  }, handleError);
}
```

Then get into the various implementation details and explanations.

src/docs/testing/errors.md
```diff
@@ -23,6 +25,10 @@ instead of the one that failed. By default,
 in debug mode this shows an error message in red,
 and in release mode this shows a gray background.
 
+When errors occur without a Flutter callback on the call stack,
+they are handled by the <span class="x x-first x-last">`</span>Zone<span class="x x-first x-last">` where </span>they occur. By default<span class="x x-first x-last">,</span>
```
blazern, line 29: That doesn't seem good :scream:

blazern:
> I'm not clear from the docs whether Zone catches all errors, or if we still need to FlutterError.onError callback.

Both are needed :)
Could you please suggest a text change which would reflect that?

> Also would be nice if they cut to the chase a bit, and led off with Global Error Handling section. Something like:
> 
> To catch all errors that can occur in your FlutterApp, use the following code:
> 
> ```
> void main() {
>   void handleError(Object error, StackTrace stack) => myBackend.sendError(error, stack);
>   FlutterError.onError = (FlutterErrorDetails details) => handleError(details.exception, details.stack)
>   runZonedGuarded(() async {
>     WidgetsFlutterBinding.ensureInitialized();
>     runApp(MyApp()));
>   }, handleError);
> }
> ```
> 
> Then get into the various implementation details and explanations.

@domesticmouse, @sfshaza2, what do you think?
I tried to make as few changes as possible to retain original structure and spirit of the document as I'm just a passersby and not an owner.
But such "Global Error Handling" section does look useful to me.

src/docs/testing/errors.md
```diff
@@ -84,9 +90,47 @@ class MyApp extends StatelessWidget {
...
+To catch such an error, use [`runZonedGuarded`][].
+<!-- skip -->
+```dart
+import 'dart:async';
+
+void main() {
+  runZonedGuarded(() async {
+    WidgetsFlutterBinding.ensureInitialized();
+    runApp(MyApp()));
+  }, (Object error, StackTrace stack) {
```
kaboc, line 120: Is `WidgetsFlutterBinding.ensureInitialized()` really necessary here? If I'm not mistaken, it is also called from `runApp()`. I don't see why it needs to be called twice in a row in the same zone.

src/docs/testing/errors.md
```diff
@@ -84,9 +90,47 @@ class MyApp extends StatelessWidget {
...
+If `invokeMethod` throws an error, it won't be forwarded to `FlutterError.onError`.
+Instead, it's forwarded to the `Zone` where `runApp` was run.
+
+To catch such an error, use [`runZonedGuarded`][].
+<!-- skip -->
+```dart
+import 'dart:async';
+
+void main() {
+  runZonedGuarded(() async {
```
kaboc, line 117: nit:
```suggestion
  runZonedGuarded(() {
```

src/docs/testing/errors.md
```diff
@@ -84,9 +90,47 @@ class MyApp extends StatelessWidget {
...
+    WidgetsFlutterBinding.ensureInitialized();
+    runApp(MyApp()));
```
kaboc, line 119:
```suggestion
    runApp(MyApp());
```

src/docs/testing/errors.md
blazern: It is indeed not necessary it this example.

But in many cases `WidgetsFlutterBinding.ensureInitialized()` needs to be called manually if a developer wants to initialize something before `runApp` (e.g. Crashlytics).

`WidgetsFlutterBinding.ensureInitialized()` is called manually in this example to emphasize that it needs to be called from inside of the `Zone`, otherwise errors handling won't work.

kaboc: @blazern
Yes, but does it have to be inside `runZonedGuarded()` even if the initialization you mentioned is outside of it, or do you mean all initialization should be done inside `runZonedGuarded()` like below?

```dart
void main() {
  runZonedGuarded(() async {
    WidgetsFlutterBinding.ensureInitialized();
    await Firebase.initializeApp();
    runApp(MyApp());
  }, ...);
}
```

blazern: @kaboc
I mean that if `WidgetsFlutterBinding.ensureInitialized()` is to be called, it must always be called inside `runZonedGuarded()`.

And if some initialization (like `await Firebase.initializeApp()`) must be called after `WidgetsFlutterBinding.ensureInitialized()`, then such initialization must also be called from `runZonedGuarded()`.

I thought that it can be easily understood from [the note below](https://github.com/flutter/website/pull/5612/files#diff-913f1e4f31d4c1700b06e4c3c4653f4470fe6c8f088b258fac6143a944767d3aR125)... Do you think the note should be reworded?

kaboc: @blazern 
Now that I understand what you meant, the note looks easy to understand, but it was confusing at the first glance because I'd never seen pieces of code where initialization was done inside `runZonedGuarded()`.

How about adding some comment about initialization between `WidgetsFlutterBinding.ensureInitialized();` and `runApp(MyApp());` in the example code? I guess it'll help readers understand it more clearly.

sfshaza2:
I absolutely love this discussion. I'm ok with larger edits if it clarifies things. There is no need to treat the existing page as precious. Just like we deprecate APIs, we can restructure/rewrite docs. You just have to be careful about it because I am kind of picky about things. Let me know when you are ready for another review. I appreciate all the community members who have weighed in! It makes me feel like this info is in good hands.

esDotDev:
I'm not sure how I can suggest a whole-scale edit to the file, but in general I'd move the details that pertain to each section, into that section, which leaves the intro quite short, then get into global error handling right off the top.

Attached is a sample .md file that shows what I mean, I didn't add much new content, just moved things around (and added `code` tags) 
[errors.md](https://github.com/flutter/website/files/6306191/errors.md)

blazern:
Added a commit which tries to address all suggestions and issues pointed by @esDotDev and @kaboc.
Now I'm much less sure if the changes are ok though, because the document became 2x larger than it originally was.

sfshaza2 (approved):
This looks ok to me, though I admit I have not staged it. I am going to merge it. If the resulting page still looks problematic, either I, or involved community members (love you folks), can revisit.

blazern:
Hooray!

PR #5613, opened 2021-04-12
Update box-constraints.md as per Effective Dart Style

hrishikesh-kadam:
Reference - https://dart.dev/guides/language/effective-dart/style#prefer-using-lowercamelcase-for-constant-names


sfshaza2 (approved):
Oooh, good catch, @hrishikesh-kadam!

PR #5615, opened 2021-04-12
Migrate json.md to null safety

miquelbeltran:
Here I have migrated the json.md to null safety. Almost all code examples have been moved to a project and imported using code-excerpts. There's one thing I am not sure if I like, which I am addressing below in the comments.

cc. @sfshaza2 @RedBrogdon

### DIFF
--- a/src/docs/development/data-and-backend/json.md
+++ b/src/docs/development/data-and-backend/json.md
@@ -3,6 +3,8 @@ title: JSON and serialization
 description: How to use JSON with Flutter.
 ---
 
+<?code-excerpt path-base="../null_safety_examples/development/data-and-backend/json/"?>
+
 It is hard to think of a mobile app that doesn't need to communicate with a
 web server or easily store structured data at some point. When making
 network-connected apps, the chances are that it needs to consume some good old
@@ -105,6 +107,7 @@ decoder.
 
 The following sample JSON implements a simple user model.
 
+<?code-excerpt "lib/manual/main.dart (JSON)" skip="1" take="4"?>
 ```json
 {
   "name": "John Smith",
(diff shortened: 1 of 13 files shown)


src/docs/development/data-and-backend/json.md
```diff
@@ -278,18 +280,18 @@ from the previous samples.
...
 
-{% prettify dart %}
+<?code-excerpt "lib/serializable/user.dart"?>
+```dart
 import 'package:json_annotation/json_annotation.dart';
 
 /// This allows the `User` class to access private members in
 /// the generated file. The value for this is *.g.dart, where
 /// the star denotes the source file name.
-part '[[highlight]]user[[/highlight]].g.dart';
```
miquelbeltran, line 287: I am not sure how to add the `[[highlight]]` code in here, using a "replace" regex in the code-excerpt seems very complicated. Alternatively I could revert this change and leave this code sample without a code-excerpt. I am open to ideas!

sfshaza2: Yes, the highlight directive is only supported in `prettify` sections. In this case, just remove the highlight. I don't want you jumping hoops with regex. Just please make sure that the introductory text to this section doesn't refer to the "highlighted" code, since it won't be.

sfshaza2:
Other than that, it LGTM. Can you take a look, @domesticmouse?

domesticmouse (approved):

src/docs/development/data-and-backend/json.md
miquelbeltran: Got it! I will double check that

miquelbeltran: I didn't see any references to highlighted code, so it should be good to go :)

PR #5624, opened 2021-04-14
Migrate text-field-changes to null safety

miquelbeltran:
Here I have migrated the text-field-changes.md as part of migrating all the priority P1 documents to null safety.

I added a couple of inline comments to help with the review.

cc. @sfshaza2 @RedBrogdon

### DIFF
--- a/src/docs/cookbook/forms/text-field-changes.md
+++ b/src/docs/cookbook/forms/text-field-changes.md
@@ -12,6 +12,8 @@ js:
     url: https://dartpad.dev/inject_embed.dart.js
 ---
 
+<?code-excerpt path-base="../null_safety_examples/cookbook/forms/text_field_changes/"?>
+
 In some cases, it's useful to run a callback function every time the text
 in a text field changes. For example, you might want to build a search
 screen with autocomplete functionality where you want to update the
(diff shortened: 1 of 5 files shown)


src/docs/cookbook/forms/text-field-changes.md
```diff
@@ -131,22 +133,32 @@ Begin listening for changes when the
...
+```
+
+<?code-excerpt "lib/main.dart (dispose)"?>
+```dart
+@override
+void dispose() {
+  // Clean up the controller when the widget is removed from the widget tree.
+  // This also removes the _printLatestValue listener.
+  myController.dispose();
+  super.dispose();
```
miquelbeltran, line 154: because the previous text tells to "stop listening when disposed" I added an example of the dispose method here.

src/docs/cookbook/forms/text-field-changes.md
```diff
@@ -131,22 +133,32 @@ Begin listening for changes when the
 `_MyCustomFormState` class is initialized,
 and stop listening when the `_MyCustomFormState` is disposed.
 
-<!-- skip -->
+<?code-excerpt "lib/main.dart (initState)"?>
 ```dart
-class _MyCustomFormState extends State<MyCustomForm> {
```
miquelbeltran, line 136: It was difficult to include the code excerpt with the class name around it, so I am extracting here only the method `initState`, I hope that's OK

sfshaza2 (commented):
This looks good to me, @miquelbeltran (and I'm OK with extracting just initState), but can you take a look, @domesticmouse?

domesticmouse (approved):

PR #5630, opened 2021-04-14
Add member variable to cached FlutterEngine Java example

ryansama:
The Kotlin example has the `lateinit var flutterEngine` member variable shown but the Java version does not.

Also uncomments `code-excerpts` in `add-flutter-screen.md`.

### DIFF
--- a/src/docs/development/add-to-app/android/add-flutter-screen.md
+++ b/src/docs/development/add-to-app/android/add-flutter-screen.md
@@ -58,7 +58,7 @@ Make sure to use the following import:
 
 {% samplecode default-activity-launch %}
 {% sample Java %}
-<!--code-excerpt "ExistingActivity.java" title-->
+<? code-excerpt "ExistingActivity.java" title ?>
 ```java
 myButton.setOnClickListener(new OnClickListener() {
   @Override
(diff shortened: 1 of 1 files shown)


ryansama:
@googlebot I signed it!

sfshaza2 (commented):
@miquelbeltran, is this PR correct with the code excerpt?

miquelbeltran:
The original code-excerpt code seems to be wrong, because they are commented out, for example:
```
<!--code-excerpt "MyApplication.java" title-->
```
instead of
```
<? code-excerpt "MyApplication.java" title ?>
```
so the refresh-code-excerpts script does nothing.

You can merge this change, but the original code excerpt tags need to be fixed as well.

sfshaza2:
Oh, I hadn't noticed that! @ryansama, we certainly can't comment out the code excerpt portion. Can you fix this? Or should I close this PR?

sfshaza2 (changes requested):
Reinstate code excerpts.

ryansama:
> Oh, I hadn't noticed that! @ryansama, we certainly can't comment out the code excerpt portion. Can you fix this? Or should I close this PR?

Hi @sfshaza2 I uncommented the line above my change -- should I also uncomment every other code-excerpt in the file?

sfshaza2 (changes requested):
@ryansama, I'm not sure why any code excerpts would be commented out. I searched and found just one other instance where it was commented out. I'd say, yes, remove that.

sfshaza2 (approved):
Thanks, @ryansama! LGTM

sfshaza2:
@miquelbeltran, uncommenting the code excerpts (which the original poster didn't comment out), is causing breakage. Can you take a quick looksie?

sfshaza2:
OK, this build is breaking on a reverted change (required version of english_words package). This PR doesn't even modify that text. Merging manually.

PR #5634, opened 2021-04-15
Update the title `web-images.md`

Abhishek01039:
image

- Here  tag is not visible to the title


sfshaza2 (approved):
Actually, it's against the Google Developer Documentation Guidelines to use code font in a header, but I'm going to merge this anyway, since codefont is better than "invisible" text. However, I would have changed this to "Use the image tag in a platform view" perhaps. But it's not that big of a deal. LGTM

Abhishek01039:
Thanks, @sfshaza2 
if you want then I would like to do that change.

PR #5642, opened 2021-04-16
sort go links

xster:


sfshaza2:
LGTM. :). It wasn't that big of a deal, but thx.

PR #5647, opened 2021-04-17
Fix `@required` annotation in code snippet 

satyambnsal:
With dart version `2.12.0`, we don't need to use `@required` annotation. we can simply use the `required` keyword. using `@required` in the latest version of Dart results in an Error.


sfshaza2 (approved):
LGTM

PR #5648, opened 2021-04-17
Fixed Response isn't a type

abughalib:
The package http.dart is imported as http,
the above can be fixed just by importing import 'package:http/http.dart' but that would create duplicate import.


abughalib:
Should I open issues instead of these small change pull requests?

PR #5661, opened 2021-04-21
Fix typo in layout-basics.md

brunobely:
Remove comma between "meaning" and "that".

From [this helpful article](https://jakubmarian.com/comma-before-that-and-which/):

> A good way to recognize [a dependent clause] is to try to enclose the clause in parentheses; if the sentence still makes sense, you should use commas (or parentheses) to separate the clause from the rest[.]

In the sentence:

> You’ll also learn that Flutter widgets are composable, meaning, *that you can combine existing widgets to make more sophisticated widgets*.

The sentence would not make sense if the part in italics was in parantheses.


sfshaza2 (approved):
LGTM

PR #5663, opened 2021-04-21
Fix formatting of url-strategies.md

hrishikesh-kadam:
1. Fix ordered list formatting.
2. Add a `<br>`.
3. Add parenthesis for setUrlStrategy method.
4. Remove master-api references.

### DIFF
--- a/src/docs/development/ui/navigation/url-strategies.md
+++ b/src/docs/development/ui/navigation/url-strategies.md
@@ -7,11 +7,11 @@ Flutter web apps support two ways of configuring URL-based navigation on the
 web:
 
 **Hash (default)**: Paths are read and written to the [hash fragment][].
-  For example, `flutterexample.dev/#/path/to/screen`.
+  For example, `flutterexample.dev/#/path/to/screen`. <br>
 **Path**:  Paths are read and written without a hash. For example,
   `flutterexample.dev/path/to/screen`.
   
-These are set using the [setUrlStrategy][] API with either a [HashUrlStrategy][]
+These are set using the [setUrlStrategy()][] API with either a [HashUrlStrategy][]
 or [PathUrlStrategy][].
   
 ## Configuring the URL strategy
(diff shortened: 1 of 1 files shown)


sfshaza2 (commented):
I'm ok with these changes, but I want to make sure you are, too, @johnpryan.

src/docs/development/ui/navigation/url-strategies.md
```diff
@@ -7,11 +7,11 @@ Flutter web apps support two ways of configuring URL-based navigation on the
 web:
 
 **Hash (default)**: Paths are read and written to the [hash fragment][].
-  For example, `flutterexample.dev/#/path/to/screen`.
+  For example, `flutterexample.dev/#/path/to/screen`. <br>
```
johnpryan, line 10: why do we need a `<br>` tag?

johnpryan: can we just use a newline instead?

src/docs/development/ui/navigation/url-strategies.md
```diff
@@ -26,50 +26,52 @@ or [PathUrlStrategy][].
   setup.
 {{site.alert.end}}
 
-The `setUrlStrategy` API can only be called on the web. The following
+The `setUrlStrategy()` API can only be called on the web. The following
```
johnpryan, line 29: I don't think we typically refer to function names using `()` in our docs, but I might be wrong.

sfshaza2:
I'll be honest. When I looked at the current page yesterday, I wanted to make formatting changes, but not quite like the ones in this PR. For example, I'd prefer that we use a definition list when describing the two navigation approaches. I appreciate that you brought this page to my attention, @hrishikesh-kadam, but I'm going to close this PR and submit a different PR.

src/docs/development/ui/navigation/url-strategies.md
hrishikesh-kadam: [Google developer documentation style guide](https://developers.google.com/style/code-in-text#method-names) says "Put an empty pair of parentheses after a method name to indicate that it's a method."

sfshaza2: :D  We generally only do that when it's otherwise unclear. But saying that you call (or invoke) `blank` makes it pretty clear it's a method. I'm sorry for any confusion! Sometimes the guidelines change over time. But, yeah, that's a totally fair comment.  :)

PR #5664, opened 2021-04-21
Move animated-container to code excerpts and enable null safety

miquelbeltran:
Here I moved the example from the cookbook recipe animated-container.md to a project, used code excerpts to extract the code and enabled null safety in the dartpad.

cc. @sfshaza2 @RedBrogdon

### DIFF
--- a/src/docs/get-started/codelab.md
+++ b/src/docs/get-started/codelab.md
@@ -223,7 +223,7 @@ as well as many other open source packages, on [pub.dev][].
        flutter:
          sdk: flutter
        cupertino_icons: ^1.0.2
-    +  english_words: ^4.0.0
+    +  english_words: ^4.0.0-0
     ```
 
  2. While viewing the `pubspec.yaml` file in Android Studio's editor view,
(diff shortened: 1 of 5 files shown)


src/docs/get-started/codelab.md
```diff
@@ -223,7 +223,7 @@ as well as many other open source packages, on [pub.dev][].
        flutter:
          sdk: flutter
        cupertino_icons: ^1.0.2
-    +  english_words: ^4.0.0
+    +  english_words: ^4.0.0-0
```
miquelbeltran, line 226: For some reason this is changed by the `./tool/refresh-code-excerpts.sh` script, even if I didn't change the original file

sfshaza2: Yes, this is from a PR that @domesticmouse created and I merged and it broke the build. I'm not quite sure why it broke the build, as 4.0.0 *is* the latest release for english_words. But, in any case, it's nothing to do with your PR.  :)

sfshaza2:
Please review, @domesticmouse!

src/docs/get-started/codelab.md
domesticmouse: The issue is that I didn't use `./tool/refresh-code-excerpts.sh` to update this file. I just updated it directly.

Please the original file to have `english_words: ^4.0.0` as the dep constraint

domesticmouse (approved):

PR #5677, opened 2021-04-23
migrate page-route-animation.md to null safety

miquelbeltran:
Same procedure as with the other pages :)

cc. @sfshaza2 @RedBrogdon


sfshaza2:
Can you take a look, @domesticmouse?

domesticmouse (approved):

PR #5679, opened 2021-04-23
Remove references to deprecated package integration_test

gaetschwartz:
Removed all links to https://pub.dev/packages/integration_test as this package is deprecated and can confuse people.

### DIFF
--- a/src/docs/testing/integration-tests/index.md
+++ b/src/docs/testing/integration-tests/index.md
@@ -3,7 +3,7 @@ title: Integration testing
 description: Learn how to write integration tests
 ---
 
-This page describes how to use the [integration_test][] package to run
+This page describes how to use the `integration_test` package to run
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
Hi there, @gaetschwartz! I'm going to go ahead and land this, though I will open a follow up PR. This PR doesn't go far enough for me.  :). But, thanks for bringing this to my attention! I also had the pub.dev page for integraton_test updated because, before, it just said "deprecated" which is misleading.  LGTM

gaetschwartz:
@sfshaza2 Indeed, but the other changes need to make decisions and I was unsure which way you wanted to go to. In example, the are still references to parts of the package's documentation (https://pub.dev/packages/integration_test) in the page that aren't anywhere else as far as I remember.

PR #5682, opened 2021-04-24
fix typo

osama96gh:


sfshaza2 (changes requested):
Hi, @osama96gh! Thanks so much for correcting this typo! Your PR broke the build because your fix was incomplete and is now causing a broken link.  Can you also change the following line?

Original line:
```
[`AndroidViewService`]: {{site.api}}/flutter/widgets/AndroidViewSurface-class.html
```

Updated line:
```
[`AndroidViewSurface`]: {{site.api}}/flutter/widgets/AndroidViewSurface-class.html
```

That should fix the build and then I can land the fix!

PR #5685, opened 2021-04-25
Remove Documentation Link from Bottom of Widget Cards

### LINKED ISSUE #1495: Limit height of widget index cards
Widget index cards:

- [ ] Use ellipses and hide excess description.
- [x] Make the image a link too. Related: #1485
- [x] With the image and card header as links, consider dropping the "Documentation" link - #5685

@Sfshaza - what do you think?

---

This is what the current site looks like (on Chrome):

> screen shot 2018-10-18 at 09 01 44

And these are our new cards:

> screen shot 2018-10-18 at 09 02 03

micpap25:
Partially fixes #1495
Removes Documentation Footer on Widget Cards. Redundant as both the Image and the Title are already links to the documentation page.

### DIFF
--- a/src/_includes/catalogpage.html
+++ b/src/_includes/catalogpage.html
@@ -30,9 +30,6 @@
                         <a href="{{comp.link}}"><header class="card-title">{{comp.name}}</header></a>
                         <p class="card-text">{{comp.description}}</p>
                     </div>
-                    <div class="card-footer card-footer--transparent">
-                        <a href="{{comp.link}}">Documentation</a>
-                    </div>
                 </div>
             {% endif %}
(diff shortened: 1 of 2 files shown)


sfshaza2 (approved):
I like this improvement on many levels! LGTM

chalin:
Thanks for this contribution @micpap25!

FYI: I'd suggest writing "Contributes to #1495" rather than "Partially fixes #1495"; otherwise, when GitHub finds text like "fixes #XYZ" in a PR comment, it automatically closes #XYZ when the PR is merged. We want to avoid that when a PR is only a partial fix. Other than that, keep those PRs coming!

PR #5691, opened 2021-04-26
Migrated orientation.md to null safety

miquelbeltran:
Migrated the orientation.md to null safety. Same as all the other cookbook recipes :)

There's a couple of small code snippets skipped inside, but since they don't change I did not extract them.

cc. @sfshaza2 @RedBrogdon

### DIFF
--- a/null_safety_examples/cookbook/design/drawer/pubspec.yaml
+++ b/null_safety_examples/cookbook/design/drawer/pubspec.yaml
@@ -1,5 +1,5 @@
-name: orientation
-description: Sample code for orientation cookbook recipe.
+name: drawer
+description: Sample code for drawer cookbook recipe.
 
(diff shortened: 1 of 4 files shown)


null_safety_examples/cookbook/design/drawer/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
-name: themes
-description: Sample code for themes cookbook recipe.
+name: drawer
+description: Sample code for drawer cookbook recipe.
```
miquelbeltran, line 2: I noticed that this was wrong when copy/pasting it

sfshaza2:
Please take a look, @domesticmouse!

domesticmouse (approved):

PR #5692, opened 2021-04-26
Migrated snackbars.md to null safety

miquelbeltran:
Same as with the other cookbook recipes. No code changes happened in this case but I extracted the main code since I was at it.

cc. @sfshaza2 @RedBrogdon


sfshaza2:
Please take a look, @domesticmouse. Thx!

domesticmouse (approved):

PR #5696, opened 2021-04-27
Enable null-safety on tabs.md

miquelbeltran:
No code changes required for null safety support, so simply adding the null_safety flag to the dartpad and extracting the code snippet to a project.

cc. @sfshaza2 @RedBrogdon


sfshaza2 (approved):
LGTM

PR #5697, opened 2021-04-27
Migrated focus.md to null safety

miquelbeltran:
Same as the other cookbooks, here there was a variable that needed to be "late"

cc. @sfshaza2 @RedBrogdon


sfshaza2:
Please take a look, @domesticmouse.

domesticmouse (approved):

PR #5707, opened 2021-04-28
Update _initial-route-cached-engine.md

igorassis:
There is no method called **initialRoute** in the **navigationChannel**. In Kotlin should be called the **setInitialRoute**, passing the route as parameter as you can see below:

```
 public void setInitialRoute(@NonNull String initialRoute) {
    Log.v(TAG, "Sending message to set initial route to '" + initialRoute + "'");
    channel.invokeMethod("setInitialRoute", initialRoute);
  }
```


igorassis:
@googlebot I signed it!

sfshaza2 (approved):
LGTM

PR #5709, opened 2021-04-28
Adds workshop links

RedBrogdon:
Adds three redirect links for workshops. Right now they just point to DartPad so we can verify they aren't broken. I'll add the correct URLs in a subsequent PR.


RedBrogdon:
CC: @johnpryan.

Piinks:
I think the actually links for these can be subbed in, I was just about to do the same. 😜

Piinks:
I believe links are listed in the description for https://github.com/flutter/codelabs/pull/101

sfshaza2 (approved):
Thanks for not calling them codelabs!!! LGTM

PR #5722, opened 2021-05-03
Update editor.md

johnsenner:
Nothing says "Configure" but it does say "Preferences". I think it may have been this way for a while.

Your docs are great, thank you!


johnsenner:
@googlebot I signed it!

sfshaza2 (approved):
Thanks for that correction! LGTM

PR #5726, opened 2021-05-03
Added deprecated ThemeData accent properties breaking-changes doc

HansMuller:
I moved the contents of the existing document called theme-data-accent-properties.md to a new documented called _fab_-theme-data-accent-properties.md because its scope is limited to the FloatingActionButton.  The original document's contents have been replaced by material about deprecating the four theme accent properties: accentColor, accentColorBrightness, accentIconTheme, accentTextTheme.

The contents of the theme-data-accent-properties.md document also appear in the pull request that introduced the deprecations: https://github.com/flutter/flutter/pull/81336.


sfshaza2 (approved):
Excellent and better late than never!!

PR #5747, opened 2021-05-06
Migrate delete-data.md to null safety

miquelbeltran:
Migrated page to null safety.

I am not super convinced about the original solution which relied on having the Album attributes as null to show it as deleted, but I decided to not to change that now.

### DIFF
--- a/null_safety_examples/cookbook/networking/delete_data/lib/main.dart
+++ b/null_safety_examples/cookbook/networking/delete_data/lib/main.dart
@@ -1,0 +1,126 @@
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:flutter/material.dart';
+import 'package:http/http.dart' as http;
+
+Future<Album> fetchAlbum() async {
+  final response = await http.get(
+    Uri.parse('https://jsonplaceholder.typicode.com/albums/1'),
+  );
+
+  if (response.statusCode == 200) {
+    // If the server did return a 200 OK response, then parse the JSON.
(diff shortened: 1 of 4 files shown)


null_safety_examples/cookbook/networking/delete_data/lib/main.dart
```diff
@@ -0,0 +1,130 @@
...
+                  return Column(
+                    mainAxisAlignment: MainAxisAlignment.center,
+                    children: <Widget>[
+                      Text('${snapshot.data?.title ?? 'Deleted'}'),
+                      ElevatedButton(
+                        child: Text('Delete Data'),
+                        onPressed: () {
+                          setState(() {
+                            _futureAlbum =
+                                deleteAlbum(snapshot.data!.id.toString());
```
miquelbeltran, line 110: because data will never be null in this case

null_safety_examples/cookbook/networking/delete_data/lib/main.dart
```diff
@@ -0,0 +1,130 @@
...
+    // If the server did not return a "200 OK response",
+    // then throw an exception.
+    throw Exception('Failed to delete album.');
+  }
+}
+// #enddocregion deleteAlbum
+
+class Album {
+  final int? id;
+  final String? title;
```
miquelbeltran, line 47: because the deleteAlbum method relies of parsing a null response, needs to have these two properties as nullable

domesticmouse (approved):

PR #5752, opened 2021-05-07
Update windows.md

ArkyTech-Inc:
added a full stop on line 17

### DIFF
--- a/src/docs/get-started/install/windows.md
+++ b/src/docs/get-started/install/windows.md
@@ -14,7 +14,7 @@ next:
 To install and run Flutter,
 your development environment must meet these minimum requirements:
 
-- **Operating Systems**: Windows 7 SP1 or later (64-bit), x86-64 based
+- **Operating Systems**: Windows 7 SP1 or later (64-bit), x86-64 based.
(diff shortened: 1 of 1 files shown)


ArkyTech-Inc:
@googlebot I signed it!

sfshaza2 (approved):
You get points for consistency in the list but, actually, none of these items are complete sentences, so they really shouldn't have punctuation. But... LGTM

ArkyTech-Inc:
Thank you for the clarification. Appreciate it.

sfshaza2 (changes requested):
Hi, @ArkyTech-Inc! You are right that there is an error in this list, but your fix is not correct. The punctuation on a list _should_ be parallel, but are not in this list (and sublist). In this case, these are not full sentences, so they shouldn't have punctuation. So, I'd remove the periods, not add more.

You can see the rule in the [Google developer documentation style guide](https://developers.google.com/style/lists?hl=en#parallel-syntax). I'll leave this PR open so you can decide if you want to update it, otherwise I'll close.

sfshaza2 (approved):
LGTM

PR #5760, opened 2021-05-08
Update navigate-with-arguments.md

GoooGu:
!. in dart is not a thing. 

I'm sure the original author meant ?. to create null-aware access here.


GoooGu:
Never mind. It was my misunderstanding.

PR #5762, opened 2021-05-10
Update interactive.md

pinguin999:
Clean up diff

Adding FavoriteWidget(), would not result in adding ); and } at the end.


pinguin999:
@googlebot I signed it!

sfshaza2 (commented):
Thanks for this PR, @pinguin999! But it looks to me like this has already been fixed. Closing.

PR #5508, opened 2021-03-18
Update fetch-data.md

olahat90:

PR #5763, opened 2021-05-10
Migrate to null safety

hrishikesh-kadam:
Maybe #5508 should not have been merged.

### DIFF
--- a/src/docs/cookbook/networking/fetch-data.md
+++ b/src/docs/cookbook/networking/fetch-data.md
@@ -95,7 +95,7 @@ class Album {
   final int id;
   final String title;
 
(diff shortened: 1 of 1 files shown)


sfshaza2:
Yes, it looks like you're right. @miquelbeltran, I could revert that merge but I don't want to step on your toes. (Back when I merged it in mid March I don't think you had started migrating this part of the cookbook.)

miquelbeltran:
Oh I see what happened, fetch-data.md does not have the examples extracted so CI didn't catch the issue in the other PR.

@sfshaza2 let's merge this one, and I will extract the examples on fetch-data.md to code-excerpts so it should not happen again.

PR #5767, opened 2021-05-11
Migrate to null safety

hnishi:
I followed the instruction and got the following error.

```
> flutter run -d chrome
Launching lib\main.dart on Chrome in debug mode...
lib/main.dart:135:20: Error: The parameter 'value' can't have a value of 'null' because of its type 'double', but theimplicit default value is 'null'.
Try adding either an explicit non-'null' default value or the 'required' modifier.
    @required this.value,
                   ^^^^^
Waiting for connection from debug service on Chrome...             17.1s
Failed to compile application.
```

This modification solved this error.

### DIFF
--- a/src/docs/get-started/codelab-web.md
+++ b/src/docs/get-started/codelab-web.md
@@ -713,7 +713,7 @@ class AnimatedProgressIndicator extends StatefulWidget {
   final double value;
 
   AnimatedProgressIndicator({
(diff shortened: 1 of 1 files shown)


sfshaza2:
Thanks for this PR, @hnishi! And thanks for that heads up, @miquelbeltran! I will leave this to you to fix.

PR #5770, opened 2021-05-11
Adding a basic /embed page

sfshaza2:
Do we consider embedded to be fully released and just an advanced topic?

### DIFF
--- a/src/embedded.md
+++ b/src/embedded.md
@@ -4,13 +4,20 @@ description: "Details of how Flutter supports the creation of embedded experienc
 ---
 
 If you would like to embed Flutter engine into a car,
-a refrigerator, a thermostat... you CAN! 
+a refrigerator, a thermostat... you CAN! For example,
+you might embed Flutter in the following situations:
 
+* Use Flutter on an "embedded device",
+  typically a low-powered hardware device
+  such as a smart-display, a thermostat, or similar.
+* Embed Flutter into a new operating system or
+  environment, for example a new mobile platform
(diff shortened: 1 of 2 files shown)


src/embedded.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Embedded support for Flutter
```
mit-mit, line 2: There are two potentially separate topics here:

1) Using Flutter on a "embedded device", typically a low-powered hardware device such as a smart-display, a thermostat, or similar

2) Embedding Flutter into a new operating system or environment, for example a new mobile platform or a new operating system

These two things are connected, but not necessarily the same, right?

sfshaza2: Correct. I can list both out but is there anything else we want to say about this?

csells: I think listing both of those scenarios is sufficient for now.

src/embedded.md
```diff
@@ -0,0 +1,34 @@
...
+description: "Details of how Flutter supports the creation of embedded experiences."
+---
+
+If you would like to embed Flutter engine into a car,
+a refrigerator, a thermostat... you CAN! 
+However, this feature, while stable,
+uses low-level API and is _not_ for beginners. 
+In addition to the following resources, you
+might consider joining Discord, where Flutter
```
domesticmouse, line 11: This feels like a CTA without a link target

domesticmouse (approved):

zanderso (approved):

PR #5771, opened 2021-05-11
Drag gestures migration guide

Piinks:
This adds a migration guide for the new `ScrollBehavior.dragDevices` and deprecated `GestureDetector.kind`


Piinks:
Thanks @sfshaza2! It's rolled in, but it has not gotten a build tag yet. I'm ok to wait for the tag and keep an eye on it. I'll update once I have a version #. :)

jonahwilliams (approved):
LGTM!

sfshaza2:
@Piinks, would you mind merging the conflicts? I am a bit sleep deprived at the moment.

domesticmouse:
@Piinks & @sfshaza2 is this ready to land now?

creativecreatorormaybenot (commented):
I feel like this guide is not really addressing how to migrate an app on a **high-level**, i.e. this is a breaking change and how do we best deal with it for whole apps?

Let's say, we have many horizontal scroll views in an app. What should be done in the migration process then? I would think it would benefit to address this choice.

1. Not doing anything is not an option because it will break for many users.
2. Should we override the `MaterialApp` scroll behavior to allow touch drag devices everywhere? If so, we lose all benefit of this change.
3. Should we add the touch drag device for every horizontal scroll view?


* Potentially useful as well: why is the touch drag device not allowed for horizontal scroll views by default.

sfshaza2 (changes requested):
This can't be published until a release version number is entered under "Timeline" and "Landed in version".

Piinks:
@sfshaza2 I have updated now that this has a build tag. :) 🚀

sfshaza2 (approved):
LGTM!

PR #5773, opened 2021-05-12
Changed order to match VS Code Steps

Publicker:
I was following the [guide](https://flutter.dev/docs/get-started/test-drive?tab=vscode#create-app) to create my first app using VS Code and I noticed that the order of some steps were switched.

This PR attends to set the order that VS Code actually followed.


sfshaza2 (approved):
Thx, @Publicker! LGTM

PR #5774, opened 2021-05-12
Update url-strategies.md

minikin:
Regarding [<base>: The Document Base URL element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base#examples) the base href has to end with a "/".


minikin:
@googlebot I signed it!

sfshaza2 (approved):
LGTM

PR #5732, opened 2021-05-04
Add Google APIs content

kevmoo:
Adds initial content towards https://github.com/flutter/website/issues/5637

PR #5793, opened 2021-05-14
fixes url link for go/a11y-text-attributes/

chunhtai:


sfshaza2 (approved):
LGTM

mono0926:
@chunhtai @sfshaza2 

This broke intended change at #5732 (@sfshaza2 reiewed) and make me inaccessible from it.
If this change is applied, changing sharing setting of the document is needed. Until then, this should be reverted.

PR #5795, opened 2021-05-15
Update download-button.md

AugustDG:
Added @ just before "required" keyword in constructor parameter, as it is required :D

Just a typo fix as otherwise the code does not compile!

### DIFF
--- a/src/docs/cookbook/effects/download-button.md
+++ b/src/docs/cookbook/effects/download-button.md
@@ -76,7 +76,7 @@ enum DownloadStatus {
 class DownloadButton extends StatefulWidget {
  const DownloadButton({
    Key? key,
-   required this.status,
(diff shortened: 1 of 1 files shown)


AugustDG:
@googlebot I signed it!

sfshaza2:
@AugustDG, are you using Flutter 1.x or Flutter 2.x? The @ sign is no longer required in Flutter 2 due to null safety.

sfshaza2:
@miquelbeltran

AugustDG:
Oh, I see. I'm using 2.x, but without null-safety....

Probably why it was complaining 👌🏽

sfshaza2:
Thanks for clearing that up, @AugustDG! At some point, null safety will no longer be optional.

PR #5807, opened 2021-05-17
Make README.md symbolic link as per examples/

hrishikesh-kadam:
Two reasons for this PR -

1. To make the repo a bit lighter (9177 bytes) using symbolic link.

2. Running `tool/code-check.sh` and `tool/code-check.sh --null-safety`
hydrates the samples of `examples/` and `null_safety_examples/` if it
contains test folder.
`flutter create --no-overwrite $sample` creates the template README.md
file if not already present.
To avoid tracking of the newly created README.md, it is better to add
symbolic link to the `examples/` and `null_safety_examples/` folder's
README.md


hrishikesh-kadam:
@miquelbeltran FYI

miquelbeltran:
I think we could get rid of those README.md, looks like they are just legacy from when the examples where created.

hrishikesh-kadam:
No @miquelbeltran, we need those README.md. I'll try to explain you again more deeply.
In [tool/build_check_deploy][] command, if the sample contains a test folder then it will create a new template README.md if not already present.
If the new README.md files are created by `flutter create --no-overwrite $sample` command then, it will be a new file for git which will be untracked.
So to avoid tracking of the newly created README.md file, it will be better that we ourselves create the least meaningful 
 README.md file.

To understand more follow the below steps - 
1. `git clone --recurse-submodules https://github.com/flutter/website.git`
2. `cd webiste`
3. `git checkout 3b01462` (Optional - HEAD at this point of time. Just to make sure we are in sync)
4. `source ./tool/env-set.sh`
5. `./tool/before-install.sh`
6. `./tool/install.sh` (May be optional for this PR)
7. `tool/check-code.sh --null-safety`
8. `git status`

You would find at least 2 README.md files untracked.
- null_safety_examples/codelabs/add_flutter_to_android_app/add_flutter_to_android/README.md
- null_safety_examples/cookbook/persistence/reading_writing_files/README.md
Those got created because they contain a test folder and there was no existing symbolic link file to the null_safety_examples/README.md file.

For the rest of the existing README.md files, I just made a symbolic link to the null_safety_examples/README.md file.

Do let me know if you need more clarity on this.

[tool/build_check_deploy]: https://github.com/flutter/website/blob/3b01462762d260111b739bfb23627ad30a121791/tool/build_check_deploy.sh#L191-L198

miquelbeltran:
Oh understood now, thanks! Yes, this sounds like a good solution for now.

sfshaza2 (approved):
LGTM

PR #5815, opened 2021-05-18
Banner schedule update and CTA update

btobin:
@Sfshaza and @johnpryan - ready to go! Thanks.


sfshaza2 (approved):
LGTM (Though I thought that folks might use a difference codebase for, say, a watch app.)

PR #5819, opened 2021-05-18
Unbreaks tests for stable channel

RedBrogdon:
* Updates pubspec used for bare snippet tests to remove dependency on `package:test`.
  - This should be `flutter_test` now, which is already included.
* Adds skips to a couple integration test cookbook articles which are out of date.
* Removes the old, unmigrated `internationalization` example, which has been replaced with a migrated version at this point.
* Adds skips to two snippets in an old breaking changes page.


sfshaza2:
LGTM

PR #5830, opened 2021-05-20
update faq.md with valid CodePen URL

ymmtyuhei:
`https://codepen.io/flutter` does not seems to be about Flutter. (It seems to be a personal page of `goodo`.)
`https://codepen.io/topic/flutter` is a topic page for Flutter.


ymmtyuhei:
@googlebot I signed it!

sfshaza2 (approved):
Thanks for catching that, @ymmtyuhei! LGTM

PR #5834, opened 2021-05-21
Prefer using Uri.parse instead of Uri.http/Uri.https

jamesderlin:
Using `Uri.http`/`Uri.https` are *much* more error-prone than using
`Uri.parse`, and judging from various StackOverflow questions, quite
a number of people either misunderstand how to use them or would have
avoided their problem by using `Uri.parse`.  Examples:

* https://stackoverflow.com/q/66619895/
* https://stackoverflow.com/a/66608366/
* https://stackoverflow.com/q/67535694/
* https://stackoverflow.com/q/67628555/
* https://stackoverflow.com/q/67451865/

IMO, we should advocate using `Uri.parse` when possible, particularly
in tutorials, where using `Uri.http`/`Uri.https` directly is a
premature optimization.

### DIFF
--- a/null_safety_examples/cookbook/networking/fetch_data/lib/main.dart
+++ b/null_safety_examples/cookbook/networking/fetch_data/lib/main.dart
@@ -7,7 +7,7 @@ import 'package:http/http.dart' as http;
 // #docregion fetchAlbum
 Future<Album> fetchAlbum() async {
   final response =
-      await http.get(Uri.https('jsonplaceholder.typicode.com', 'albums/1'));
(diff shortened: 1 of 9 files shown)


jamesderlin:
And note that prior to https://github.com/flutter/website/pull/5202, some of these sites used `String`s directly and formerly would have ended up using `Uri.parse` internally in `package:http` anyway.

sfshaza2 (approved):
Yaaaay! You also fixed it in the accompanying Dart files, so the analyzer is happy. LGTM

Issue #5821, opened 2021-05-19
'Internationalizing Flutter apps' page issue


Reksa97:
Page URL: https://flutter.dev/docs/development/accessibility-and-localization/internationalization.html
Page source: https://github.com/flutter/website/tree/master/src/docs/development/accessibility-and-localization/internationalization.md

Description of issue:

The links to 'Sample internationalized apps' are not working:
Minimal internationalization
https://github.com/flutter/website/tree/master/examples/internationalization/minimal
Internationalization based on the intl package
https://github.com/flutter/website/tree/master/examples/internationalization/intl_example

piedcipher:
They are under `null_safety_examples`.

https://github.com/flutter/website/tree/master/null_safety_examples/internationalization/minimal
https://github.com/flutter/website/tree/master/null_safety_examples/internationalization/intl_example

PR #5838, opened 2021-05-21
Fix links to internationalization example

goderbauer:
The example was removed from its old location in https://github.com/flutter/website/pull/5819, but is still referenced on https://flutter.dev/docs/development/accessibility-and-localization/internationalization at that old location. This updates the page to the new location.

/cc @RedBrogdon @sfshaza2 

Also, I thought we had a link checker? Should it have caught this?


piedcipher:
Fixes: #5821?

sfshaza2 (approved):
Yes, the link checker should have caught that. LGTM

PR #5839, opened 2021-05-21
docs/deployment/cd:  Add usage of --dart-define

KShivendu:
Environment variables are important while building the apps for production but the use of --dart-define isn't documented on the main website. This can lead to a bad developer experience. 

This issue was reported in https://github.com/flutter/flutter/issues/81283

### DIFF
--- a/src/docs/deployment/cd.md
+++ b/src/docs/deployment/cd.md
@@ -117,7 +117,8 @@ untrusted, you won't be leaving your credentials like your Play Store service
 account JSON or your iTunes distribution certificate on the server.
 
 Continuous Integration (CI) systems generally support encrypted environment 
-variables to store private data.
+variables to store private data. You can pass these environment variables 
+using `--dart-define MY_VAR=MY_VALUE` while building the app.
 
 **Take precaution not to re-echo those variable values back onto the console in
 your test scripts**. Those variables are also not available in pull requests


src/docs/deployment/cd.md
```diff
@@ -117,7 +117,8 @@ untrusted, you won't be leaving your credentials like your Play Store service
 account JSON or your iTunes distribution certificate on the server.
 
 Continuous Integration (CI) systems generally support encrypted environment 
-variables to store private data.
+variables to store private data. You can pass these environment variables 
+using `--dart-define MY_VAR=MY_VALUE` while building the app.
```
domesticmouse, line 121: Can we link to further documentation on this? I'm curious what I can pass `--dart-define` to.

sfshaza2: Yeah, this is a deficiency of our docs! `dart-define` is a way to pass compile-time variables to the `flutter run` or `flutter build` commands. I gather that you can even use it to pass values down to the native level. Please file a bug about this omission on the flutter.dev repo. In the short term, maybe provide a full example *or* link to the only page on the site where dart-define is used, [Report errors to a service](https://flutter.dev/docs/cookbook/maintenance/error-reporting).

sfshaza2:
Hi @KShivendu! That bug was filed on the wrong repo, so I've [moved it to our repo](https://github.com/flutter/website/issues/5852). So, you don't need to do anything for that. (Issues on flutter.dev docs go on the flutter/website repo, not the flutter/flutter repo, which I don't even look at.)

This PR is a good start, but there needs to be more info.

src/docs/deployment/cd.md
domesticmouse: Raised https://github.com/flutter/website/issues/5853 =)

sfshaza2 (approved):
Given that @domesticmouse has filed a new issue for adding further docs about the dart-define tag. I'm going to land this. Thx!

PR #5864, opened 2021-05-27
Extracted code excerpts from testing/widget/tap-drag.md

miquelbeltran:
No changes necessary for null safety support, but I extracted the code anyways and fixed some lint errors.

### DIFF
--- a/src/docs/cookbook/testing/widget/tap-drag.md
+++ b/src/docs/cookbook/testing/widget/tap-drag.md
@@ -6,6 +6,8 @@ prev:
   path: /docs/cookbook/testing/widget/finders
 ---
 
+<?code-excerpt path-base="../null_safety_examples/cookbook/testing/widget/tap_drag/"?>
+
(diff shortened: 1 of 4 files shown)


src/docs/cookbook/testing/widget/tap-drag.md
```diff
@@ -232,6 +238,8 @@ void main() {
 }
 
 class TodoList extends StatefulWidget {
+  const TodoList({Key? key}) : super(key: key);
```
domesticmouse, line 241: fwiw, I'd add `use_key_in_widget_constructors: false` to the linter rules above so you don't need to introduce this cconstructor here.

domesticmouse: @sfshaza2 if @miquelbeltran isn't around to make these changes, feel free to land as is.

sfshaza2:
I'm not completely sure if he's used up his hours, so I will wait a bit before landing. Thanks! :D

Issue #5885, opened 2021-06-01
Flutter on Windows getting started directions should not suggest C:\src\flutter as an install path


jonahwilliams:
https://github.com/flutter/flutter/issues/83721

c.f. https://flutter.dev/docs/get-started/install/windows#get-the-flutter-sdk

A user that installs Flutter to this location will almost certainly fail to run it, because that directory will not have user read/run/write permissions by default. A more reasonable location would C:\Users\ <Username> \Documents which has all of the correct permissions by default.

Marking as high priority because the top crash on 2.2 (and previous releases) appears to be improperly installed Flutter SDKs

PR #5888, opened 2021-06-02
Tweak Windows install instructions

sfshaza2:
Fixes https://github.com/flutter/website/issues/5885

### DIFF
--- a/src/docs/get-started/install/_get-sdk-win.md
+++ b/src/docs/get-started/install/_get-sdk-win.md
@@ -9,7 +9,7 @@
     [SDK releases][] page.
  1. Extract the zip file and place the contained `flutter`
     in the desired installation location for the Flutter SDK
-    (for example, `C:\Users\Documents`).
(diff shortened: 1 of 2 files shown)


src/docs/get-started/install/_get-sdk-win.md
```diff
@@ -9,7 +9,7 @@
     [SDK releases][] page.
  1. Extract the zip file and place the contained `flutter`
     in the desired installation location for the Flutter SDK
-    (for example, `C:\src\flutter`).
+    (for example, `C:\Users\Documents`).
```
jonahwilliams, line 12: For me this would be something like `C:\Users\Jonah\Documents`, this would be for a user named 'Documents'. Is there a pattern for pointing the user towards this instead?

sfshaza2: D'OH. Of course. Fixed.

jonahwilliams:
Thanks!

PR #5897, opened 2021-06-03
Update Github Actions node version and firebase-tools/superstatic

parlough:
This is a follow-up to https://github.com/flutter/website/commit/1840d4a2b16176824df4abcec76baee727334b01

It may be worth exploring updating the other npm dependencies since npm finds some vulnerabilities with them. But for now, I'm limiting the scope to what I know works and corresponds to the similar change in https://github.com/dart-lang/site-www/commit/77c6fef7b0875e7847a4085bf74f26f5175f1511

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,7 +86,7 @@ jobs:
       - run: mkdir -p $TMP
       - uses: actions/setup-node@v1
         with:
-          node-version: 12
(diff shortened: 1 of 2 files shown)


sfshaza2 (approved):
Thanks, @parlough! It looks good to me, but we'll see if it passes the checks.

PR #5908, opened 2021-06-08
Update firebase.json, adding go/retiring-tests

christopherfujino:
Open sourcing a previously private doc on deleting tests.

cc @yjbanov


sfshaza2 (changes requested):
Please update link to not require Google login.

sfshaza2 (commented):
I just tried opening this URL in regular browser and I got that I need access. So, I'm closing this PR. Re-open when the doc has more open permissions.

PR #5909, opened 2021-06-09
fixed broken link in README.md within example

al919:
Fixed the broken link(extract.dart) in README.md within example. The example.g link in the file is still broken and I was unable to find the file in the repo to fix the broken link. Maybe the file is renamed?


sfshaza2 (approved):
Yes, that link is wrong. In the REPO, the file is under `/tool` but, because the tooling folder is not under `/src`, then it is not put into the generated site. This workaround (linking to the file on GitHub) is fine.

sfshaza2 (approved):
It makes no sense that the build breaks on this PR. Manually merging. LGTM

PR #5928, opened 2021-06-16
Migration Guide for Flutter web slotted rendering.

ditman:
This change adds a Migration Guide for the new Flutter web slotted rendering, explaining how the code is changing and what users need to do for their web apps to keep working as normal.

Fixes: https://github.com/flutter/flutter/issues/80524

### DIFF
--- a/src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
+++ b/src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
@@ -1,0 +1,196 @@
+---
+title: Using HTML slots to render platform views in the web
+description: iframes in Flutter web used to reload, because of the way some DOM operations were made. A change in the way Flutter web apps render Platform Views makes them much more stable (preventing iframe reloads, and other problems with video tags or forms potentially losing their state).
+---
+
+## Summary
+
+Flutter now renders Platform Views using [slot elements][] inside of a single,
+app-wide [Shadow Root][]. Slot elements can be added/removed/moved around the
+Shadow DOM without affecting the underlying slotted content (which is rendered
+in a constant location)
+
+## Context
+
+The Flutter framework frequently tweaks its Render Tree to optimize the paint
+operations that are ultimately made per frame. In the web, these Render Tree
+changes often result in DOM operations.
+
+Flutter web used to render its Platform Views (`HtmlElementView` widgets)
+directly into its corresponding position of the DOM.
+
+Using certain DOM elements as the "target" of some DOM operations causes those
+elements to lose their internal state. In practice this means that `iframe` tags
+are going to reload, `video` players might restart, or a form being edited might
+lose the edits made to it.
+
+This change was made to:
+
+* make Platform Views behave in a stable way in Flutter web.
+* unify the way Platform Views were rendered in the web for both rendering
+backends (`html` and `canvaskit`).
+
+## Description of change
+
+A Flutter web app is now rendered inside a common Shadow Root in which slot
+elements are used to represent platform views. The actual content of each
+platform view is rendered *outside* of the Shadow Root, as a child of the Shadow
+Root host.
+
+### Before
+
+<!-- skip -->
+```html
+...
+
+<flt-glass-pane>
+  ...
+  <div id="platform-view">Contents</div> <!-- canvaskit -->
+  <!-- OR -->
+  <flt-platform-view>
+    #shadow-root
+    | <div id="platform-view">Contents</div> <!-- html -->
+  </flt-platform-view>
+  ...
+</flt-glass-pane>
+
+...
+```
+
+### After
+
+<!-- skip -->
+```html
+...
+
+<flt-glass-pane>
+  #shadow-root
+  | ...
+  | <flt-platform-view-slot>
+  |   <slot name="platform-view-1" />
+  | </flt-platform-view-slot>
+  | ...
+  <flt-platform-view slot="platform-view-1">
+    <div id="platform-view">Contents</div>
+  </flt-platform-view>
+  ...
+</flt-glass-pane>
+
+...
+```
+
+Now, when the framework needs to move DOM nodes around, it'll operate over
+`flt-platform-view-slot`s, which only contain the `slot` tag that "projects" the
+content defined in the `flt-platform-view` tags outside of the Shadow Root
+(which never move), thus preventing the reloads.
+
+From an app's perspective, this change is transparent, **however** this was
+deemed a _breaking change_ because tests that make assumptions about the DOM of
+a Flutter web app can break.
+
+## Migration guide
+
+### Code
+
+The content returned by the users' Platform View Factory used to be resized and
+positioned by the framework, but not anymore. Ensure that your returned
+`html.Element` takes the whole space allocated for it by the framework, by
+setting its `style.width` and `style.height` properties to `'100%'`:
+
+<!-- skip -->
+```dart
+ui.platformViewRegistry.registerViewFactory(viewType, (int viewId) {
+  final html.Element htmlElement = html.DivElement()
+    // ..other props
+    ..style.width = '100%'
+    ..style.height = '100%';
+  // ...
+  return htmlElement;
+});
+```
+
+Otherwise the engine will print a warning message to the console similar to:
+
+<!-- skip -->
+```bash
+Height of Platform View type: [$viewType] may not be set. Defaulting to `height: 100%`.
+Set `style.height` to any appropriate value to stop this message.
+```
+
+or:
+
+<!-- skip -->
+```bash
+Width of Platform View type: [$viewType] may not be set. Defaulting to `width: 100%`.
+Set `style.width` to any appropriate value to stop this message.
+```
+
+### Tests
+
+Unfortunately, there are no selectors that let one _pierce_ through the Shadow
+DOM boundary to locate elements underneath it, so code that once peeked inside
+of the `flt-glass-pane`, needs to be made Shadow DOM aware. Using the html
+markup from above:
+
+<!-- skip -->
+```dart
+// Find #some-element inside the app:
+final element = document.querySelector('#some-element');
+```
+
+Code after migration:
+
+<!-- skip -->
+```dart
+final shadowRoot = document.querySelector('flt-glass-pane')?.shadowRoot!;
+final element = shadowRoot.querySelector('#some-element');
+```
+
+It's probably a good practice to not assume `document` is always going to be the
+root element of searches, so a helper like this might be convenient:
+
+<!-- skip -->
+```dart
+/// Locate elements in the correct root of the application, whether it is
+/// `document` or the new `shadowRoot` of `flt-class-pane`.
+List<Node> findElements(String selector) {
+  final ShadowRoot? shadowRoot = document.querySelector('flt-glass-pane')?.shadowRoot;
+  return (shadowRoot != null) ?
+    shadowRoot.querySelectorAll(selector):
+    document.querySelectorAll(selector);
+}
+```
+
+(See Relevant PRs below for multiple examples of the "migrations" described
+above).
+
+## Timeline
+
+Landed in version: 2.3.0-16.0.pre<br>
+In stable release: not yet
+
+## References
+
+Design Doc:
+
+* [Using slot to embed web Platform Views][design doc]
+
+Relevant issues:
+
+* [Issue #80524][issue-80524]
+
+Relevant PRs:
+
+* [flutter/engine#25747][pull-25747]: Introduces the feature.
+* [flutter/flutter#82926][pull-82926]: Tweaks `flutter` tests.
+* [flutter/plugins#3964][pull-3964]: Tweaks to `plugins` code.
+* [flutter/packages#364][pull-364]: Tweaks to `packages` code.
+
+[design doc]: https://flutter.dev/go/web-slot-content
+[issue-80524]: {{site.github}}/flutter/flutter/issues/80524
+[pull-25747]: {{site.github}}/flutter/engine/pull/25747
+[pull-364]: {{site.github}}/flutter/packages/pull/364
+[pull-3964]: {{site.github}}/flutter/plugins/pull/3964
+[pull-82926]: {{site.github}}/flutter/flutter/pull/82926
+[Shadow Root]: https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot
+[slot elements]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot
--- a/src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
+++ b/src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
@@ -84,18 +84,21 @@ Now, when the framework needs to move DOM nodes around, it'll operate over
 content defined in the `flt-platform-view` tags outside of the Shadow Root
 (which never move), thus preventing the reloads.
 
-From an app's perspective, this change is transparent, **however** this was
-deemed a _breaking change_ because tests that make assumptions about the DOM of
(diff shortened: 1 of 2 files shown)


src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
+---
+title: Using HTML slots to render platform views in the web
+description: iframes in Flutter web used to reload, because of the way some DOM operations were made. A change in the way Flutter web apps render Platform Views makes them much more stable (preventing iframe reloads, and other problems with video tags or forms potentially losing their state).
```
yjbanov, line 3: Remove "much more". It wasn't stable and now it simply is :)

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+---
+
+## Summary
+
+Flutter now renders Platform Views using [slot elements][] inside of a single,
+app-wide [Shadow Root][]. Slot elements can be added/removed/moved around the
+Shadow DOM without affecting the underlying slotted content (which is rendered
+in a constant location)
```
yjbanov, line 11: This discusses an implementation detail, which may be useful in the "Context" section below. In the summary I think the developers would be most interested in what happens to the HTML DOM that _they_ created, so I'd suggest that we talk about how we put their platform views into the root element, instead of injecting it into the Flutter's internal DOM structures.

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Context
+
+The Flutter framework frequently tweaks its Render Tree to optimize the paint
```
yjbanov, line 15: I'm not sure if "render tree" is worth capitalizing.

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+Using certain DOM elements as the "target" of some DOM operations causes those
+elements to lose their internal state. In practice this means that `iframe` tags
+are going to reload, `video` players might restart, or a form being edited might
+lose the edits made to it.
+
+This change was made to:
+
+* make Platform Views behave in a stable way in Flutter web.
+* unify the way Platform Views were rendered in the web for both rendering
+backends (`html` and `canvaskit`).
```
yjbanov, line 31: And:

* provide a predictable location in the DOM to allow developer to reliably use CSS to style HTML inside platform views, and use other standard DOM API, such as `querySelector`, and `getElementById`.

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+Shadow DOM without affecting the underlying slotted content (which is rendered
+in a constant location)
+
+## Context
+
+The Flutter framework frequently tweaks its Render Tree to optimize the paint
+operations that are ultimately made per frame. In the web, these Render Tree
+changes often result in DOM operations.
+
+Flutter web used to render its Platform Views (`HtmlElementView` widgets)
```
yjbanov, line 19: I'd probably also not capitalize "platform views"

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+This change was made to:
+
+* make Platform Views behave in a stable way in Flutter web.
+* unify the way Platform Views were rendered in the web for both rendering
+backends (`html` and `canvaskit`).
+
+## Description of change
+
+A Flutter web app is now rendered inside a common Shadow Root in which slot
```
yjbanov, line 35: I'd not capitalize "shadow root", but I would link to MDN or some other reliable source that explains what it is.

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+elements are used to represent platform views. The actual content of each
+platform view is rendered *outside* of the Shadow Root, as a child of the Shadow
+Root host.
```
yjbanov, line 38: I think we can gzip this sentence a bit: "...is rendered as a sibling of the shadow root"

yjbanov: Or, since we're keeping `<flt-glass-pane>` we could just say "direct child of `<flt-glass-pane>`"

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+From an app's perspective, this change is transparent, **however** this was
+deemed a _breaking change_ because tests that make assumptions about the DOM of
+a Flutter web app can break.
+
+## Migration guide
+
+### Code
+
+The content returned by the users' Platform View Factory used to be resized and
```
yjbanov, line 95: nit: user's

yjbanov, line 95: nit: s/resized/sized/

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+positioned by the framework, but not anymore. Ensure that your returned
```
yjbanov, line 96: I'd insert another sentence between "but not anymore." and "Ensure...":

...but not anymore. Instead, Flutter sizes and positions `<flt-platform-view>`, which is the parent of the content. Ensure...

yjbanov: I'd also replace "Ensure" with "To ensure". Maybe the developer doesn't want to fill the whole space. They only need to set size to 100% if they want to.

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+  return htmlElement;
+});
+```
+
+Otherwise the engine will print a warning message to the console similar to:
+
+<!-- skip -->
+```bash
+Height of Platform View type: [$viewType] may not be set. Defaulting to `height: 100%`.
+Set `style.height` to any appropriate value to stop this message.
```
yjbanov, line 117: Hmm, do we need this? Why can't I use relative layout (flexbox, etc) to fill the space?

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+```
+
+or:
+
+<!-- skip -->
+```bash
+Width of Platform View type: [$viewType] may not be set. Defaulting to `width: 100%`.
+Set `style.width` to any appropriate value to stop this message.
```
yjbanov, line 125: same here

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+```
+
+### Tests
+
+Unfortunately, there are no selectors that let one _pierce_ through the Shadow
```
yjbanov, line 130: s/Unfortunately/By design/ (this is why we use it 😈)

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+DOM boundary to locate elements underneath it, so code that once peeked inside
+of the `flt-glass-pane`, needs to be made Shadow DOM aware. Using the html
```
yjbanov, line 132: nit: remove comma before "needs"

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+// Find #some-element inside the app:
+final element = document.querySelector('#some-element');
+```
+
+Code after migration:
+
+<!-- skip -->
+```dart
+final shadowRoot = document.querySelector('flt-glass-pane')?.shadowRoot!;
+final element = shadowRoot.querySelector('#some-element');
```
yjbanov, line 146: Why is this necessary? After the migration `#some-element` will no longer be inside the shadow DOM, only the slot. We don't want people looking into Flutter's private DOM structure. It's unpredictable and unstable.

ditman: This is what most of the tests that failed were doing. We can document the fact that now they shouldn't drill inside the shadowDOM, as mentioned above, that's probably better!

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
ditman: If you use any other way of filling the space, you need to set height/width to `auto`, or anything that is not empty. In the engine the only thing we look at is that width/height is set.

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
ditman: Hah, yep. Removed!

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
sfshaza2, line 38: I'm suggesting a change that corresponds to Yegor's comment However, he might have intended for that second sentence to be deleted? @yjbanov?

```suggestion
A Flutter web app is now rendered inside a common [shadow root][] in which slot
elements represent platform views. The actual content of each
platform view is rendered *outside* of the shadow root, as a child of the shadow
root host.

[shadow root]: https://www.html5rocks.com/en/tutorials/webcomponents/shadowdom-201/
```

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+  ...
+</flt-glass-pane>
+
+...
+```
+
+Now, when the framework needs to move DOM nodes around, it'll operate over
+`flt-platform-view-slot`s, which only contain the `slot` tag that "projects" the
+content defined in the `flt-platform-view` tags outside of the Shadow Root
+(which never move), thus preventing the reloads.
```
sfshaza2, line 85:
```suggestion
After this change, when the framework needs to move DOM nodes around,
it operates over`flt-platform-view-slot`s, which only contain the `slot` tag that "projects" the
content defined in the `flt-platform-view` tags outside of the shadow root
(which never move), thus preventing the reloads.
```

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
ditman: Capitalization Of Things, The Bane Of My Existence :P

ditman: (lower-cased :P)

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+From an app's perspective, this change is transparent, **however** this was
+deemed a _breaking change_ because tests that make assumptions about the DOM of
+a Flutter web app can break.
```
sfshaza2, line 89:
```suggestion
From an app's perspective, this change is transparent. **However**, this is
considered a _breaking change_ because some tests make assumptions
about how the DOM of a Flutter web app might break.
```

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
ditman: I moved this to the `context` section.

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Migration guide
+
+### Code
+
+The content returned by the users' Platform View Factory used to be resized and
+positioned by the framework, but not anymore. Ensure that your returned
+`html.Element` takes the whole space allocated for it by the framework, by
+setting its `style.width` and `style.height` properties to `'100%'`:
```
sfshaza2, line 98:
```suggestion
Previously, the content returned by [`PlatformViewFactory`][] was resized and
positioned by the framework. Instead, Flutter now sizes and positions
`<flt-platform-view>`, which is the parent of the content.
To ensure the `html.Element` that you return takes the whole space
allocated to it, set its `style.width` and `style.height` properties to `'100%'`:

[`PlatformViewFactory`]: {{site.api}}/javadoc/index.html?io/flutter/plugin/platform/PlatformView.html
```

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+  final html.Element htmlElement = html.DivElement()
+    // ..other props
+    ..style.width = '100%'
+    ..style.height = '100%';
+  // ...
+  return htmlElement;
+});
+```
+
+Otherwise the engine will print a warning message to the console similar to:
```
sfshaza2, line 112:
```suggestion
Otherwise, the engine prints a warning message to the console similar to:
```

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
sfshaza2, line 132:
```suggestion
By design, there are no selectors that let one _pierce_ through the shadow
DOM boundary to locate elements underneath it, so code that once peeked inside
of the `flt-glass-pane` needs to be made shadow DOM aware. Using the html
```

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
```diff
@@ -0,0 +1,196 @@
...
+above).
+
+## Timeline
+
+Landed in version: 2.3.0-16.0.pre<br>
+In stable release: not yet
+
+## References
+
+Design Doc:
```
sfshaza2, line 174:
```suggestion
Design doc:
```

sfshaza2 (changes requested):
Yegor made some comments that I turned into official GitHub suggestions. He also had some questions as to the design, and I will leave that to the two of you to resolve.  :)

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
ditman: I'll add a "But I don't want to set width: 100% in my platform view" section

ditman:
Thanks @sfshaza2 for the suggestions! I was halfway through applying by hand Yegor's suggestions so I took the ones from you that wouldn't directly conflict with my changes :P I've manually applied your changes to the first half of the doc, I think I got'em'all!

sfshaza2 (approved):
LGTM, but I will let Yegor have the final say.

src/docs/release/breaking-changes/platform-views-using-html-slots-web.md
ditman: Removed this, warned people against looking into the shadowDom of the app.

ditman:
I've added a few more lines here and there, I think I addressed all of the comments. @yjbanov I've re-tagged you to PTAL. Once we're happy with the content, I'll ping @sfshaza2 for a last review!

yjbanov (approved):
lgtm

PR #5939, opened 2021-06-20
s/--remote/--recursive in git submodule update

hrishikesh-kadam:
[`--remote`][1] option will pull all the submodule's remote master
branch HEAD and not the current branch's recorded SHA-1.

~~So, instead of `--remote` option [`--recursive`][2] option is
appropriate because this repo has `examples/codelabs` submodule in the
depth deeper than it's root.~~

So, instead of `--remote` option [`--recursive`][2] option can be more 
appropriate as in future submodules may add another submodules.

~~In case of `dart-lang/site-www`, `git submodule update --init` is
sufficient because it doesn't have any submodule deeper than it's 
root.~~

In case of `dart-lang/site-www` also, 
`git submodule update --init --remote` should be 
`git submodule update --init --recursive`

[1]: https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt---remote
[2]: https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt---recursive

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,13 +120,13 @@ submodule-cloning techniques:
   this command from the repo root:<br>
 
   ```console
-  $ git submodule update --init --remote
+  $ git submodule update --init --recursive
   ```
 
 > NOTE: At any time during development you can use the `git submodule` command to
 > refresh submodules:<br>
 > ```console
-> $ git pull; git submodule update --init --remote
+> $ git pull; git submodule update --init --recursive
 > ```
 
 ### 3. Run installation scripts


hrishikesh-kadam:
Related to dart-lang/site-www#3349

kwalrath:
I wonder if this was why I was having trouble maintaining copies of flutter repos but not dart repos. Thank you, @hrishikesh-kadam!

I'll look into the build break... I don't think it's new for this PR.

kevmoo:
🤷 – I'm not super sure about this option. Would have to dig in a bit.

kwalrath:
I have a single alias I use for both dart.dev & flutter.dev, and it seems to work fine on both. (But I haven't made any tricky changes.)

```
alias gp='git pull; git submodule update --init --recursive'
```

hrishikesh-kadam:
@kwalrath For dart-lang/site-www `git submodule update --init --recursive`, adding `--recursive` will not harm at all but can be avoided at the moment.

hrishikesh-kadam:
@kevmoo Just for evidence, `--recursive` is used instead of `--remote` in the [GitHub Actions Checkout](https://github.com/flutter/website/pull/5939/checks?check_run_id=2869250140#step:2:511) code of this Actions Workflow.

hrishikesh-kadam:
Quoting from my [first comment](https://github.com/flutter/website/pull/5939#issue-674041951).

> So, instead of `--remote` option [`--recursive`](https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt---recursive) option is
> appropriate because this repo has `examples/codelabs` submodule in the
> depth deeper than it's root.
> 
> In case of `dart-lang/site-www`, `git submodule update --init` is
> sufficient because it doesn't have any submodule deeper than it's root.

I think I got the [`--recursive`](https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt---recursive) meaning wrong. It is not recursive in means of submodule present in depth deeper than the repo root but actually the submodules present in the submodules.

So, it is always safe to add `--recursive` option.

Removed commit message and forced pushed same code.

hrishikesh-kadam:
Added `--recursive` option again in dart-lang/site-www#3353

sfshaza2:
I'm going to LGTM and merge this, unless @kevmoo or @kwalrath objects.

kevmoo:
Go for it!

I don't have any objections. If it breaks your workflow, I think it'll be
obvious!

sfshaza2 (approved):
LGTM

PR #5941, opened 2021-06-20
Fix code example var name in building-adaptive-apps.md

imagentleman:
The boolean name seems to have an incorrect name in a code example, since the variable referenced later when setting the direction is different (line 269). This PR changes the variable name so they match (when initialized and used).


sfshaza2 (approved):
LGTM. The code isn't being analyzed or this would have surfaced.

PR #5964, opened 2021-06-26
Minor formatting changes in docs

hrishikesh-kadam:
- Minor formatting changes in docs
- Correct ordered list formatting
The reason to correct with \<ol\> instead of markdown syntax is to have
fewer diffs


sfshaza2 (approved):
LGTM

PR #5968, opened 2021-06-28
Make the buildTypes block directions more clear

suragch:
I taught a Flutter class last term and I helped several students prepare their apps for publishing. The "Replace the buildTypes" instructions were confusing for them. It looks like the given block is what you are supposed to copy and use to replace something (especially since there is a copy button). However, the actual meaning is that you are supposed to find that block and then replace it with the next given block. This PR seeks to make that meaning more clear.


sfshaza2 (approved):
Definitely an improvement! Thanks, @suragch!

PR #5973, opened 2021-06-29
Add links to 2 design docs

Piinks:
This adds links to docs for adding and reducing framework tests.


justinmc (approved):
LGTM 👍

Piinks:
@sfshaza2 I can't merge this, would you when you have a chance?

sfshaza2 (approved):
@Piinks, some PRs use Google doc links that require logging into a Google account, but your links don't. Do you know what the magic is to avoid that? Anyway, LGTM

Piinks:
> @Piinks, some PRs use Google doc links that require logging into a Google account, but your links don't. Do you know what the magic is to avoid that? Anyway, LGTM

For public design docs we can't use our `@google` addresses, exactly because of this access issue. I have a separate `@gmail` account that I just use for design docs and create them with that account. I set it up to forward any messages to my `@google`, so I don't have to switch back and forth. :)
There is another option where you can set up a flutter contrib account, I am not sure how to do that though.

Thanks for merging!

PR #5978, opened 2021-06-30
Update firebase.json

kf6gpe:
Modify go links from KF6GPE's personal account to kf6gpe@fcontrib.org.


sfshaza2 (commented):
@kf6gpe, I opened both of these links in an incognito window and both required a Google sign-on. Can that be avoided?

kf6gpe:
@sfshaza2 Looks like I muffed the sharing settings in Google docs... good catch. Can you try again, please?

sfshaza2 (approved):
Perfection! Thx!

PR #5980, opened 2021-07-01
Start of the great `const`ification

domesticmouse:
This is the start of the fixes I think are required for `lints` and `flutter_lints`.

I'm unsure if I'm putting fixes in the right place (i.e. are these files downstream of some other source of truth). Also it seems the backend takes a while to analyze the samples when updated, so these fixes may not be final. Or totally correct. It looked correct at the time of editing ;-)


domesticmouse:
Silly question, where is the source of truth for these code samples?

PR #5986, opened 2021-07-02
Remove non-migrated examples, replace with null_safety_examples

RedBrogdon:
* Deletes the old `/examples` directory
* Removes an old copy of some `flutter/codelabs` code that was copy/pasted into `/null_safety_examples` to avoid having a second submodule during the migration process.
* Renames `/null_safety_examples` to `/examples`
* Updates `/tool/build_check_deploy.sh` so that code in `/examples` is only tested with null safety opted-in.
* Updates code excerpts that used to reference `/null_safety_examples` to reference the same code in its new, `/examples`, location.

A subsequent PR will enable flutter_lints for all of this code.


RedBrogdon:
Not to be reviewed before Wednesday, @Sfshaza! 😄

sfshaza2 (approved):
Thanks so much for this PR, @RedBrogdon! LGTM

piedcipher:
This PR created an issue: #5998

PR #5991, opened 2021-07-04
Fixed a spelling mistake

nawalhmw:
Changed the comment: "Create an store the VideoPlayerController. The VideoPlayerController" to "Create and store the VideoPlayerController. The VideoPlayerController"

### DIFF
--- a/src/docs/cookbook/plugins/play-video.md
+++ b/src/docs/cookbook/plugins/play-video.md
@@ -120,7 +120,7 @@ class _VideoPlayerScreenState extends State<VideoPlayerScreen> {
 
   @override
   void initState() {
-    // Create an store the VideoPlayerController. The VideoPlayerController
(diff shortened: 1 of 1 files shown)


nawalhmw:
@googlebot I signed it!

sfshaza2 (approved):
LGTM

sfshaza2:
This didn't break the build. Forcing a merge.

RedBrogdon:
This did actually break the build (though it may already have been broken for another reason). Because the error was in a code excerpt sourced from one of the projects in /examples, fixing the excerpt rather than the underlying code caused a mismatch, which the tests will detect (and ask you to run /tool/refresh_code_excerpts.sh to fix).

It's not a huge deal, but definitely something to bear in mind when one of the excerpts needs a fix.

sfshaza2:
Ah,I must be more careful.

PR #5993, opened 2021-07-06
Fixed reference links for iOS + Android implementation of Switch class

zyruva:
The links for each implementation of the Material Switch class and the Cupertino Switch class were mixed up


zyruva:
@googlebot I signed it!

sfshaza2 (approved):
LGTM

PR #6005, opened 2021-07-09
internationalization: fix links from null_safety_examples to examples

rubensdemelo:
fix minimal internationalization link


sfshaza2 (approved):
Thanks for this PR, @rubensdemelo!

PR #6020, opened 2021-07-12
Prefer `const` literals as parameters of constructors on immutable classes

asashour:
The current code samples give `Prefer const literals as parameters of constructors on @immutable classes.`.

This PR updates those warnings accordingly.

### DIFF
--- a/examples/internationalization/add_language/lib/main.dart
+++ b/examples/internationalization/add_language/lib/main.dart
@@ -7,14 +7,14 @@ void main() {
   runApp(
     // #docregion MaterialApp
     MaterialApp(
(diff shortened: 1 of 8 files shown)


sfshaza2:
@johnpryan, this is a matter of best practices. This PR adds `const` to an array of literals and also moves a `const`, applied to each value in another array, to the entire array.

Can you weigh in?

johnpryan (approved):
LGTM

johnpryan (approved):
LGTM

johnpryan:
Thanks for updating the code snippets too, LGTM

PR #6026, opened 2021-07-13
Random updates

domesticmouse:
Some random changes to make this a better sample:

  - Sign in automatically, if possible
  - Put everything in a single `ListView` so it has usable UX on iOS
  - Traverse the whole list, not just the first page of five videos
  - Put 250ms between page requests to reduce the risk of getting throttled
  - Filter out deleted videos to prevent null check failures (deleted videos don't have thumbnails)

### DIFF
--- a/examples/googleapis/lib/main.dart
+++ b/examples/googleapis/lib/main.dart
@@ -51,7 +51,7 @@ class _LikedVideosWidgetState extends State<_LikedVideosWidget> {
         _downloadLikedList();
       }
     });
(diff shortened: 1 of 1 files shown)


sfshaza2:
@domesticmouse, it looks like code excerpts needs to be called.

domesticmouse:
Actually, it looks like my re-writing of the sample needs to be done with an understanding of the article it is being used by. I don't have that, so I'll close this PR out.

PR #6028, opened 2021-07-13
Updating banner and fixing broken links

sfshaza2:
As it says.

### DIFF
--- a/src/_includes/banner.html
+++ b/src/_includes/banner.html
@@ -5,7 +5,7 @@
     https://github.com/flutter/website/issues/3897
   {% endcomment %}
   <div class="site-banner site-banner--default" role="alert">
-    <a href="https://medium.com/flutter/announcing-flutter-2-2-at-google-i-o-2021-92f0fcbd7ef9">Get started with Flutter 2.2.</a>. See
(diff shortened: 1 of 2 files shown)


src/_includes/banner.html
```diff
@@ -5,7 +5,7 @@
     https://github.com/flutter/website/issues/3897
   {% endcomment %}
   <div class="site-banner site-banner--default" role="alert">
-    Flutter 2.2 is here. <a href="https://medium.com/flutter/announcing-flutter-2-2-at-google-i-o-2021-92f0fcbd7ef9">Learn more</a>. Also, see
-   <a href="/docs/whats-new#may-18-2021-google-io-edition">What's new on this site</a>, which includes a list of the new instructor-led video workshops!
+    <a href="https://medium.com/flutter/announcing-flutter-2-2-at-google-i-o-2021-92f0fcbd7ef9">Get started with Flutter 2.2.</a>. See
```
kwalrath, line 8: There are two dots in a row.

```suggestion
    <a href="https://medium.com/flutter/announcing-flutter-2-2-at-google-i-o-2021-92f0fcbd7ef9">Get started with Flutter 2.2.</a> See
```

kwalrath (commented):
It took a while because I staged it locally (and my local repo was broken, so I had to start from scratch). But worth it! I found something that would bug us both!

PR #6030, opened 2021-07-14
Update router sample link

johnpryan:
This sample has been replaced, so the link needs to be fixed.


sfshaza2 (approved):
Thanks for keeping this up to date!

PR #6033, opened 2021-07-15
Using Country / Region.

chenglu:
Hi team:

Here we should use "Country / Region" ;-)

cc/ @InMatrix @sfshaza2 

Thanks!

### DIFF
--- a/src/subscribe/index.html
+++ b/src/subscribe/index.html
@@ -39,7 +39,7 @@ js:
                     <select id="language" required></select>
                 </div>
                 <div class="group">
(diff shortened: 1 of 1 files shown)


InMatrix:
I support this change. The wording "Country/Region" is consistent with internal guidelines.

InMatrix:
Cc: @jaydchang

jaydchang:
@InMatrix I agree as well the label should be Country / Region

sfshaza2 (approved):
Hiiiiii, Luke!!!!!!!!!!  LGTM!

PR #6045, opened 2021-07-19
Changed pubspec.yaml? Restart program on device.

alex-kaa:
The image asset wouldn't show up, got me stuck googling and looking for typos (of which there were none) for quite some time. The advice would've been really helpful to me then.

Also: some advices on the internet, like to necessarily have a tab but not a space in front of the filename, or that the file extension should be jpeg, but in no case jpg, do not apply in my case. I can imagine how people could be confused into thinking of tabs/spaces and such being important while restarting, in passing, the simulation.

### DIFF
--- a/src/docs/development/ui/layout/tutorial.md
+++ b/src/docs/development/ui/layout/tutorial.md
@@ -371,8 +371,8 @@ leaving only the image. Add the image file to the example:
   * The pubspec file is also sensitive to white
     space, so use proper indentation.
   * You might need to restart the running program
-    (in simulation or on connected device) for the
-    changes in the file to take effect.
+    (either on the simulator or a connected device) for the
+    pubspec changes to take effect.
 {{site.alert.end}}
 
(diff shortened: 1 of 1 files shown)


alex-kaa:
@googlebot I signed it!

src/docs/development/ui/layout/tutorial.md
```diff
@@ -370,6 +370,9 @@ leaving only the image. Add the image file to the example:
     as shown above.
   * The pubspec file is also sensitive to white
     space, so use proper indentation.
+  * You might need to restart the running program
+    (in simulation or on connected device) for the
+    changes in the file to take effect.
```
sfshaza2, line 375:
```suggestion
  * You might need to restart the running program
    (on the simulator or a connected device) for the
    pubspec changes to take effect.
```

sfshaza2 (approved):
LGTM (with minor edits)

PR #6059, opened 2021-07-22
Remove "!" operator for currentState field

sergiocasero:
I've created this PR because I believe that this shouldn't be in the official documentation. The reason is that as we know, use `!.` implies that we tell the compiler that we take responsibility that this attribute is not null. In my opinion this is a behavior that we always have to try to avoid. What is your opinion? I suggest to use the other solution `currentState?.validate ?? false`, by setting a default value in case of null

### DIFF
--- a/src/docs/cookbook/forms/validation.md
+++ b/src/docs/cookbook/forms/validation.md
@@ -220,7 +220,7 @@ class MyCustomFormState extends State<MyCustomForm> {
             child: ElevatedButton(
               onPressed: () {
                 // Validate returns true if the form is valid, or false otherwise.
-                if (_formKey.currentState!.validate()) {
+                if (_formKey.currentState?.validate() ?? false) {
                   // If the form is valid, display a snackbar. In the real world,
                   // you'd often call a server or save the information in a database.
                   ScaffoldMessenger.of(context)


sergiocasero:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here with `@googlebot I signed it!` and we'll verify it.
> 
> #### What to do if you already signed the CLA
> ##### Individual signers
> * It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
> ##### Corporate signers
> * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google/docs/cla/#troubleshoot)).
> * The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Fflutter%2Fwebsite%2Fpull%2F6059) for more info**.

@googlebot I signed it!

sfshaza2:
@domesticmouse, can you weigh in on this PR?

domesticmouse (approved):

domesticmouse:
Actually, I need to withdraw my approval, because it looks like the `GlobalKey`'s `currentState` should always be populated if the key is associated with a `Widget`.

@RedBrogdon can you confirm my thinking here?

RedBrogdon:
> Given this exception thrown here is not due to a runtime state, but a coding time error, I'm happy with this ! in the code sample.

AFAICT, the code as currently written shouldn't ever crash due to NPE, as I can't imagine a way for the button's onPressed handler to be fired _after_ the Form's element is unmounted from the element tree.

That said, it's not like the suggested change is *wrong*, either. I do like the pattern of avoiding assertion operators anywhere you can, and here it could theoretically help you avoid an issue if the code were refactored by a second author later on.

To sum up, I'd probably say "Tie goes to the runner" and land it. The code does appear in two places on the page (see line 133), and this PR only addresses one, though, so the author still has some work to do. 😄

domesticmouse:
Cool, sounds good to me.

@sergiocasero Can you please look into the issue raised in https://github.com/flutter/website/pull/6059#issuecomment-890083934

Thanks!

davidmigloz:
I don't think the proposed change is better than the current implementation because:

- It goes against the "Fail Fast" principle. Hiding the coding error and making it much more difficult to spot.
- It implies that if `currentState` is null, the form is invalid. But that may not be the case, the form data may be valid but you are not able to verify it due to a coding error.

In Kotlin, I would use `requireNotNull(currentState).validate()` which gives a bit more context than `!!` and at the same time promotes the type to non-null. But as we don't have anything like that in Dart, using the bang operator achieves the same result at the end.

Otherwise, if we really want to avoid `!` using the proposed solution, it would be extremely helpful to add `assert(currentState != null)` before the if to, at least, avoid hiding the coding error.

sfshaza2 (commented):
I really appreciate the thoughtful conversation around this issue! Thanks, all!

But, it seems to me that the consensus is to leave the code as-is. If I am wrong, please re-open the PR and address the second use of ! on the page. (Pointed out by @RedBrogdon.)

PR #6062, opened 2021-07-23
Update link to engine size dashboard to point to new storage bucket

jmagman:
Storage URL was changed from `https://storage.googleapis.com/flutter_infra` to `https://storage.googleapis.com/flutter_infra_release`, see https://github.com/flutter/flutter/issues/75363.

Updated the example SHA to point to the last stable engine. Also clarified the SHA is for the engine repository, not the framework.


christopherfujino (approved):
Thanks, LGTM

sfshaza2 (approved):
LGTM. Thx!

PR #6068, opened 2021-07-25
Corrected Platform View Integration import

thertzelle:
### What
- Corrected the tutorial import line to use the correct `widgets.dart` file.


sfshaza2 (approved):
Good catch! LGTM

PR #6075, opened 2021-07-27
adding states_rebuilder as one of the states management solution

mabujaber:


mabujaber:
@googlebot  I signed it!

src/docs/development/data-and-backend/state-mgmt/options.md
```diff
@@ -152,6 +152,13 @@ An approach similar to Provider that is compile-safe and testable. It doesn't ha
 * [Riverpod][] homepage
 * [Getting started with Riverpod][]
 
+## states_rebuilder
+is a Flutter state management combined with a dependency injection solution and an integrated router to get the best state management experience and speed up your development.
```
sfshaza2, line 156:
```suggestion
An approach that combines state management with a dependency injection solution
and an integrated router. For more information, see the following info:
```

sfshaza2 (approved):
I modified the sentence to be a full sentence (rather than a fragment) and removed the more "marketing" hype from it. Otherwise, LGTM

PR #6099, opened 2021-08-03
Add issue templates

kwalrath:
These are the same as the dart-lang/site-www issue templates, but with dart.dev changed to flutter.dev and a sample URL changed.


sfshaza2 (approved):
LGTM!

PR #5954, opened 2021-06-24
Add iOS + Firebase TestLab instructions to "Integration testing" page.

johnpryan:
closes #5645

PR #6106, opened 2021-08-06
Correct spelling of Xcode in integration test instructions

jmagman:
Introduced in #5954.

### DIFF
--- a/src/docs/testing/integration-tests/index.md
+++ b/src/docs/testing/integration-tests/index.md
@@ -234,13 +234,13 @@ you can view the output by selecting the red icon:
 See the [Firebase Test Lab section of the README][]
 for instructions on uploading the APKs from the command line.
 
-### Uploading XCode tests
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
LGTM. thanks for keeping us honest!

jmagman:
> LGTM. thanks for keeping us honest!

I couldn't find how https://github.com/flutter/website/blob/master/.spelling is used or if it's used, but there doesn't seem to be a way in that kind of config file to make suggestions XCode->Xcode, it seems you can only add exceptions?

johnpryan:
I think `.spelling` was added by accident in https://github.com/flutter/website/commit/2714d9f2e575ffd683b736744a0e1a5812c0abde - it doesn't look like any of our site infrastructure uses it

johnpryan:
I filed an issue for the `spelling`. file: https://github.com/flutter/website/issues/6109

PR #6108, opened 2021-08-06
Change link to Flutter web samples

mit-mit:
Make it a deep link that actually filters to web samples and not all samples


mit-mit:
Actually, it looks like that might not be the right filter, as there are other web samples like Web Dashboard that are tagged differently?

sfshaza2 (approved):
LGTM!

PR #6137, opened 2021-08-11
Add filter to prevent running no-response on forks.

godofredoc:
The no-response plugin was running on all the forks in the website
repository. This change still runs the workflow but it exits immediately
if the repo is not flutter/website.

Bug: https://github.com/flutter/flutter/issues/87573

**IMPORTANT:** Due to work on the flutter.dev infrastructure, all open pull requests will be **closed August 16**.


sfshaza2 (approved):
LGTM. Thx!

PR #6145, opened 2021-08-13
Update FAQ with ABI compatibility info

### LINKED ISSUE #6125: Document that we don't consider ABI compatibility a current goal
### What information is missing?

Flutter/Dart does not offer ABI compatibility, except at the engine level. This has several implications, including that libraries built with one toolchain are not guaranteed to be binary-compatible with another runtime. (For example, I can't embed a library into an existing app without its source code). 

This is not uncommon -- Swift did not reach ABI stability until Swift 5, and there are tradeoffs of ABI stability that restrict rapid evolution in ways that customers may appreciate. Nevertheless, we should document it so that folk can understand the implications of this.

### How would you like us to fix this problem?

The obvious place to document this is in the FAQ.

This came up from a customer who assumed ABI compatibility and was disappointed to discover this was not in fact the case.

micpap25:
Fixes #6125

Adds a section about ABI compatibility to the FAQ. Hopefully this is a small enough change that we can merge it today. Otherwise I'm fine recreating this on the 23rd.

### DIFF
--- a/src/docs/resources/faq.md
+++ b/src/docs/resources/faq.md
@@ -680,6 +680,12 @@ We have seen Flutter apps run on some Chromebooks.
 We are tracking [issues related to running Flutter on
 Chromebooks][].
 
+### Is Flutter ABI compatible?
+
+Flutter and Dart do not offer application binary interface (ABI)
+compatibility. Offering ABI compatability is not a current
+goal for Flutter or Dart.
+
 ## Framework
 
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):

domesticmouse:
@sfshaza2 is this safe to land?

sfshaza2:
LGTM

PR #6220, opened 2021-09-01
Fix location of trackpad gestures proposal

moffatman:
It should be under /go rather than the root


sfshaza2 (approved):
D'OH. Yes, definitely. LGTM

PR #6226, opened 2021-09-02
Fixes robots.txt production version during Docker build

btobin:
- Ensures correct robots.txt content for production during build process
- Updates check-links robots reformatting
- Squeezes in reformatting side TOC on docs pages

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -92,6 +92,13 @@ RUN cd flutter && \
       git pull
 
 RUN flutter doctor
+
+# NOTE this is a bit sneaky and we could be more clear about it 
+# by having an actual production template for robots.txt rather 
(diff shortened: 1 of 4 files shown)


sfshaza2 (approved):
LGTM! (Of course, I trust Brian, because I am assuming this is right.  :) )

Dockerfile
```diff
@@ -92,6 +92,13 @@ RUN cd flutter && \
       git pull
 
 RUN flutter doctor
+
+# NOTE this is a bit sneaky and we could be more clear about it 
+# by having an actual production template for robots.txt rather
```
domesticmouse, line 97: I'd do prefer having a template. Then it is obvious.

domesticmouse (approved):

PR #6186, opened 2021-08-23
Correct misspelled awhile

vkuhlmann:
The adverbial awhile can't be used in the expression "Take a while"; the space is required.

See this website for an explanation:
https://blog.esllibrary.com/2017/05/04/a-while-vs-awhile/

PR #6232, opened 2021-09-04
Correct misspelled awhile

vkuhlmann:
The adverbial awhile can't be used in the expression "Take a while"; the space is required.

See this website for an explanation:
https://blog.esllibrary.com/2017/05/04/a-while-vs-awhile/

This change was previously proposed in #6186 but was discarded there because of the website upgrade


domesticmouse (approved):

sfshaza2 (approved):
LGTM

PR #6235, opened 2021-09-05
Removed extra spaces and number sign (#)

Pradumnasaraf:
Removed extra spaces and number sign (#)


domesticmouse:
This formatting is intentional, it is one comment, not two.

PR #6244, opened 2021-09-08
Fix some npm vulnerabilities, and sh issue with Dockerfile

gspencergoog:
When building the site for testing locally, I noticed that NPM printed the following:

```
8 vulnerabilities (2 low, 2 moderate, 2 high, 2 critical)
```

So I updated the `package.json` dependencies to skip past the vulnerabilities (it now reports zero).

I also noticed that the `Dockerfile` had an error in it that likely was causing it to not switch to different flutter branches. It was saying:

```
Step 24/29 : RUN if [[ -z $FLUTTER_BRANCH && "$FLUTTER_BRANCH" != "stable" ]]; then       cd flutter;       git fetch;       git remote set-branches origin $FLUTTER_BRANCH;       git fetch --depth 1 origin $FLUTTER_BRANCH;       git checkout $FLUTTER_BRANCH --;       git pull;     fi
 ---> Running in 646a6cf1ad7d
/bin/sh: 1: [[: not found
```

So I fixed the `Dockerfile` too. It also probably was continuing to run subsequent steps in the command even if one of them failed, so I switched it from using `;` to `&&`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,12 +46,12 @@ ENV PATH="/app/flutter/bin:$PATH"
 # Used if wanting to build the container with a different branch
 # e.g. `make FLUTTER_BRANCH=dev build`
 # This is not to be confused with the $TEST_TARGET_CHANNEL
-RUN if [[ -z $FLUTTER_BRANCH && "$FLUTTER_BRANCH" != "stable" ]]; then \
-      cd flutter; \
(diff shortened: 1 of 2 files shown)


sfshaza2 (approved):
LGTM. Thanks, @gspencergoog! Wow, that's a significant update for diff2html.

PR #6252, opened 2021-09-09
Add `Gemfile.lock` and `package-lock.json` to the repo

hrishikesh-kadam:
- Add `Gemfile.lock` to the repo which will eliminate the need to run
`bundle install` twice. Before rearch to docker setup `Gemfile.lock`
was present in the repo.
- Add `package-lock.json` to the repo to avoid any inconsistency in the
future.


hrishikesh-kadam:
Disclaimer - I hardly know about Ruby but have tried playing with it to solve some minor issues.

sfshaza2 (commented):
This looks good to me, but can you cast your eyes over it, @btobin?

hrishikesh-kadam:
Andrew is making necessary changes to not just pass the CI but to make it less hard.
Till then I will convert this to a draft.
Once Andrew's PR gets merged into main, I will delete https://github.com/flutter/website/pull/6252/commits/fcf96ddd1ca989103307d2c1e968825a2cbe8067 and make this PR open for review again.

@btobin You can still review https://github.com/flutter/website/pull/6252/commits/a0eefd2095341574d8bb6b53fce7af2bc8e736f1 commit.

sfshaza2:
SG!

btobin:
@Sfshaza @RedBrogdon 

Yes, let's go ahead and commit the Gemfile.lock. It saves the time of installing deps twice on local builds (no mounted volume on build) and it's the right thing to do at this point. @hrishikesh-kadam we're also going to need to commit the package-lock.json as well. If you are not seeing this locally then run `docker compose run site npm i`. The pubspec.lock is different in this case because of testing. 

NOTE that if a contributor decides they do not want to use docker/compose and does a local install, there may be conflicts between platform-specific dependencies (e.g. x86_64-linux vs darwin), and/or overwrites, etc. That may be a nonissue, but something we should pay attention to.

- Installing npm globally actually upgrades the existing npm to the latest version, so I would leave that. No harm. 
- `git` and `diffutils` are redundant but were added to be explicit about those deps. They can most definitely be removed, and should have been commented as why they are there. But again, no harm except a bit more time for the build. 

Thanks!

domesticmouse (approved):

hrishikesh-kadam:
> @hrishikesh-kadam we're also going to need to commit the package-lock.json as well. If you are not seeing this locally then run `docker compose run site npm i`. The pubspec.lock is different in this case because of testing.

@btobin Yes, even I feel `package-lock.json` should be committed. Although currently not causing any issues but to avoid any issues in the future we should commit it.

@Sfshaza `dart-lang/site-www` haven't committed `package-lock.json`. Should I commit it for `flutter/website`?
Also as a question, I believe we don't commit `pubspec.lock` may be so that we always fetch the latest dependencies. Am I right?

> NOTE that if a contributor decides they do not want to use docker/compose and does a local install, there may be conflicts between platform-specific dependencies (e.g. x86_64-linux vs darwin), and/or overwrites, etc. That may be a nonissue, but something we should pay attention to.

@btobin Yes, even I have the same doubt about the `PLATFORMS` key in `Gemfile.lock`.
Because I am currently on x86_64-linux, I am unable to check for other platforms.
But I feel it should not cause any issues as the transitive dependencies are platform-specific and not the direct dependencies.
Transitive dependencies like `nokogiri` and `ffi` do correctly install x86_64-linux versions even if `PLATFORMS` key contains just `ruby` in `Gemfile.lock`.
Can you check once by pulling this PR?

btobin:
@hrishikesh-kadam This site does not have to follow the Dart site, that will be updated separately at a future date and it's not a shared file any longer. So no concern there. We can go ahead and add the package-lock.json. 

Regarding protecting for other platform installs - I don't think we're concerned about that. This repo should be used as specified (Docker) and lock files are for the build, which is always going to be `x86_64-linux`. I was saying that we should protect against accepting accidental changes to lock files IF a contributor decides to install locally sans Docker. 

I've done a lot of testing yesterday with and without lock files, with full env clean, etc. Everything looks good to me as you have it.

hrishikesh-kadam:
> * Installing npm globally actually upgrades the existing npm to the latest version, so I would leave that. No harm.
> * `git` and `diffutils` are redundant but were added to be explicit about those deps. They can most definitely be removed, and should have been commented as why they are there. But again, no harm except a bit more time for the build.

@btobin Reverted these 2 changes as it doesn't harm anything.

sfshaza2 (approved):
LGTM

sfshaza2 (approved):
Lockfiles good. Thanks for weighing in @btobin. LGTM

sfshaza2:
@btobin, can this be merged?

PR #6124, opened 2021-08-10
Adding homebrew installation to documentation

NathanFallet:
I added a simple paragraph in the SDK page to show how to install Flutter using Homebrew, which is a preferred method for many developers using macOS.

See https://formulae.brew.sh/cask/flutter for details about the Homebrew cask.

PR #6263, opened 2021-09-11
remove homebrew instructions

christopherfujino:
This removes the instructions added in https://github.com/flutter/website/pull/6124/files. The validity of this cask is uncertain, so we should not be officially recommending it unless its security has been vetted.

### DIFF
--- a/src/docs/get-started/install/_get-sdk.md
+++ b/src/docs/get-started/install/_get-sdk.md
@@ -81,18 +81,6 @@ process.
 Once you have installed any missing dependencies, run the `flutter doctor`
 command again to verify that you’ve set everything up correctly.
 
-### Downloading and installing with Homebrew
-
-If you have Homebrew installed on your machine,
-you can install Flutter using the following command:
-
-```terminal
-$ brew install --cask flutter
-```
-
-Then, run `flutter doctor`. That lets you know if there are
-other dependencies you need to install to use Flutter, such as the Android SDK.
-
 ### Downloading straight from GitHub instead of using an archive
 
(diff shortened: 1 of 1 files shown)


christopherfujino:
cc @jmagman

jmagman (approved):
LGTM

zanderso (approved):
lgtm

tvolkert (approved):

NathanFallet:
I know a lot of developers working on macOS (including me) prefer to install things using Homebrew (like it can be done with package managers on many linux distributions) to make it simpler and quicker, and allow to update automatically. That's why I opened #6124

christopherfujino:
> I know a lot of developers working on macOS (including me) prefer to install things using Homebrew (like it can be done with package managers on many linux distributions) to make it simpler and quicker, and allow to update automatically. That's why I opened #6124

This was removed from the website not because homebrew isn't useful, but because we hadn't vetted that flutter homebrew package for validity and security.

Hixie:
If you would like to help us support homebrew officially, please [volunteer on Discord](https://github.com/flutter/flutter/blob/master/CONTRIBUTING.md) to help us fix https://github.com/flutter/flutter/issues/14050. A solution here would need to be secure and verified so that we know we're not shipping the actual Flutter binaries to people, rather than, as with an unofficial package, having the risk that the package is being manipulated by whoever is setting it up. (Not that I have any reason to believe that that is happening, but I certainly would want to be sure before telling people to use it.) Setting this up will probably require learning how our CI works and working with our infra team to figure out how to do this securely.

PR #6264, opened 2021-09-11
Add Android job listing.

tvolkert:
- Also, remove some positions from the index for which we've
  recently filled the roles (but leave the job details page up
  in case anyone's still referencing it).


tvolkert:
@kwalrath @sfshaza2

domesticmouse (approved):

PR #6188, opened 2021-08-24
Rearch integration

btobin:

PR #5893, opened 2021-06-03
Add Flutter macOS build and release guide

craastad:

PR #6288, opened 2021-09-16
Add macOS deployment link to side nav

jmagman:
The macOS deployment instructions were added to the side nav in #5893, but was lost in #6188.
Add it back.


sfshaza2 (approved):
LGTM! Thanks!

PR #6318, opened 2021-09-24
Adding a Firebase redirect for "/go/android-plugin-migration."

khanhnwin:
As discussed with @kwalrath and @sfshaza2 - this path is often showing up as a 404, so we're adding a redirect to the same place as "/go/android-plugin-migration" -> "https://flutter.dev/docs/development/packages-and-plugins/plugin-api-migration"

Confirmed the redirect by staging in my personal Firebase hosting environment.


domesticmouse (approved):

sfshaza2 (approved):
I wish we knew where this was coming from. LGTM

PR #6321, opened 2021-09-26
Adding the missing "ipa" line 47.

sirMolex:
In this section, this sentence lists the targets that support obfuscation: 
The apk, appbundle, ipa, ios, and ios-framework targets are currently supported.
Adding missing "ipa".


sfshaza2 (approved):
LGTM

PR #6322, opened 2021-09-26
Fix some unexpected text in short-titles

sirMolex:
Fix some unexpected text in short-titles. 

a11y &amp; 'i18n'
when it should show
'accessibility & localization'

the same with the value
'i18n'
that should be shown as
'localization'


sfshaza2 (commented):
@sirMolex, we prefer the shorter text that will appear in the tab. It's especially useful when you have many tabs open.

sirMolex:
@sfshaza2, the related task should be closed? # 6132

PR #6359, opened 2021-10-06
Fix link error in memory.md

MeandNi:
the link was not parsed successfully in the html tag.


domesticmouse:
I suspect [this](https://python-markdown.github.io/extensions/md_in_html/) is the context I needed.

PR #6371, opened 2021-10-08
Adding API docs redirect

redbrogdon:
Adds redirects for docs-related paths to api.flutter.dev.


RedBrogdon:
Actually, having said that, I'm not sure that Chrome or Firebase hosting isn't just fixing my mistake for me. I'll double check.

Issue #6281, opened 2021-09-15
Make commands not working


johnpryan:
### Describe the problem

Certain `make` commands aren't working for me:

```
make refresh-code-excerpts
docker-compose exec site tool/refresh-code-excerpts.sh --keep-dart-tool
make: *** [refresh-code-excerpts] Error 1
```

```
make check-code
docker-compose exec site tool/check-code.sh --refresh --null-safety
make: *** [check-code] Error 1
```

johnpryan:
This is working for me now, but I don't know what fixed it.

Issue #6277, opened 2021-09-14
Needed to run 'npm install' on first setup


johnpryan:
### Describe the problem

I followed the steps from the README, and ran into a SASS compilation error when I ran `make up`:

```
site_1  | /app/src/_sass/_vendor.scss:10: Error: File to import not found or unreadable: ../../node_modules/bootstrap-scss/bootstrap. (SassC::SyntaxError)
site_1  |         on line 10:1 of src/_sass/_vendor.scss
site_1  |         from line 7:1 of src/_sass/main.scss
site_1  |         from line 1:1 of main.scss
site_1  | >> @import '../../node_modules/bootstrap-scss/bootstrap';
site_1  | 
site_1  |    ^
```

I had to run `npm install` to get it working.

### Expected fix

Run `npm install` from the Docker container when `make setup` runs

johnpryan:
FYI @sfshaza2

btobin:
@johnpryan This is fixed now when reverting the setup. Thanks.

PR #6373, opened 2021-10-08
Make Docker better

hrishikesh-kadam:
The motivation behind this change was to avoid redundant steps.

I would like to admit my mistake while removing redundant steps in flutter/website#6252.
I didn't run the steps on the clean working directory. The persistence of the `node_modules` folder gave me a false impression that I was able to build the site without any errors.
This happened because the `make clean` didn't had `node_modules` in it.

I picked the Docker skill after the rearchitecting of this website was done.
I was a novice in Docker and trying to understand all the steps.
As I got better with Docker, I started working on removing the redundant steps with small changes first.
But, I got caught in this for the last 24 days and couldn't stop making changes.
I am Sorry but It ended up being this large.

Please don't hesitate to turn this PR down. I can totally understand.
I took this as a challenge for myself first to improve my Docker skills.
In case if all the reviewers are happy with these changes, It would be very helpful for my career to showcase this PR to my future employers.
Also, I will be happy to assist if in case anything breaks after these changes are accepted.

Following is the detailed explanation of the changes - 

- Bundler and Dart Dependencies are now bind-mounted.
This will help to persist the dependencies and make the subsequent serve faster in the development process.
They are persisted in the working directory folder.
`bundler` folder in the host bind mounts to `/usr/local/bundle` in the container, 
`.pub-cache` folder in the host bind mounts to `$HOST_USER_HOME/.pub-cache` in the container.
`node_modules` folder is already part of the bind mount of the working directory.
This is done because `bundle install` and `dart pub get` installs dependencies in the location other than the working directory unlike `npm install`.

- Added host user in the image to avoid file ownership issues on the host side.
As Docker containers run as root by default, files created from the container create issues of file ownership on the host side due to bind mounting.
I have written scripts for Linux and Darwin Kernels, tested on Linux but couldn't test for Darwin as I don't have any Darwin machine.
I have also tested this on Windows, both with and without WSL setup.
On Windows without WSL setup, the script will not add the host user as I have added the necessary checks.
On Windows with WSL setup, the script will add the host user.
Yes, we can change the file ownership using the `chown` command after `make down` but this is done to make the development smooth.
This will also help to use flutter/bin/flutter from both local and Docker development because the path stored in places like `flutter/packages/flutter_tools/.dart_tool/package_config.json` will be the same. Local development will use actual `$HOME/.pub-cache` while Docker development will use the bind-mounted `.pub-cache` folder.

- Multi layering Dockerfile with [BuildKit enabled](https://docs.docker.com/develop/develop-images/build_enhancements/#to-enable-buildkit-builds).
Dockerfile has been layered as follows - 
`base` - Contains Bundler, Node, and common other dependencies.
`user-base` - Contains host user addition setup and is mainly used as a base for Docker Compose development.
`dev` - Base image for the Docker development which doesn't have any bind mounting.
`test` - Test layer for the overall test in the Docker development.
`builder` - Jekyll build layer to generate `_site` folder with the production environment.
`deploy` - To deploy from the Docker development but haven't used anywhere in the Makefile.
To skip the obvious not-required layers, BuildKit should be enabled. Plain Docker doesn't support this.
Also, on the Windows Docker installation, BuildKit is enabled by default but not on Linux.
I hope on the macOS Docker installation, BuildKit is enabled by default.
Also after enabling BuidKit, `touch /.dockerenv` in the `base` layer was required to get away with the Flutter binaries superuser warning.

- Removed `DISABLE_TESTS` variable as it is not required anymore.
As we have better control on the layer of the Dockerfile now, we can get away the `DISABLE_TESTS` variable from everywhere.

- Removed unused build arguments from the Makefile commands
Some build arguments were not used before these changes and after these changes, more unused build arguments were removed.

- Made most of the commands in the Makefile usable on both local and Docker development.
Deliberately created useful phony targets so that they can be used on both local and Docker development.
Local development here by means if the developer decides to build the site without the Docker setup.

- Added three variants of Clean.
I have added the Tip about this in the README.md file.
Names of the targets have been taken from the [GNU Make manual](https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html).

- Edited README.md _Setting up your local environment and serving changes_ steps.
Based on the changes made, edited the necessary steps.

- Edited README.md _Deploying to a staging site_ steps.
`firebase use -add` modifies the existing `.firebaserc` file.
Also added necessary steps for the novice Firebase users.

- Removed FLUTTER_ROOT and FLUTTER_BIN environment variables.
They were not being used anywhere.
As the current invocation Flutter and Dart binaries rely on the PATH variable, I personally feel they should be invoked with the help of the above variables like before the rearchitecting change so that it can help in both the local and Docker development.
Can be thought of later if these changes are accepted.

*List which issues are fixed by this PR. You must list at least one issue.*
- flutter/website#6281
- flutter/website#6277

I have made a copy of this branch - [make-docker-better-copy](https://github.com/hrishikesh-kadam/flutter-website/tree/make-docker-better-copy) to get the results of the build.yml workflow.
Result - https://github.com/hrishikesh-kadam/flutter-website/actions/runs/1329715145
Staging URL - https://hrk-flutter-website.web.app
You can clearly see that the `deploy` job takes less time than the current workflow.

For anything that I have not explained above or something that requires more explanation, please do let me know, I will try my best to justify the changes.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat


sfshaza2:
I defer to @btobin on this. I see you've already cc'd him.  :)

PR #6428, opened 2021-10-29
Fix directory reference

jgfoster:
It couldn't find lake.jpg in `images` but it found it in `assets/images`.


sfshaza2 (commented):
Hi @jgfoster! Thanks for this PR, but putting images in an "assets" folder is optional. I'm closing this PR.

jgfoster:
While it may be the intent is that it is optional, it didn't work to have the images at the top level. I'll open an issue instead.

PR #6436, opened 2021-11-02
Document how to rename the generated Windows application

domesticmouse:
This PR documents the process for renaming the generated Windows application for a Flutter Windows desktop app.

fixes: https://github.com/flutter/flutter/issues/92376

## Presubmit checklist

### DIFF
--- a/src/desktop.md
+++ b/src/desktop.md
@@ -425,6 +425,21 @@ Here are some options:
 * Collect all of the necessary pieces
   and build your own zip file.
 
+#### Changing the name of the generated binary
+
+To change the name of the generated Windows application, edit the 
+`BINARY_NAME` variable set on line 4 of `windows/CMakeLists.txt` in
+your Flutter project.
+
+```cmake
+cmake_minimum_required(VERSION 3.15)
+project(windows_desktop_app LANGUAGES CXX)
+
+set(BINARY_NAME "YourNewApp")  # Change this line
+
+cmake_policy(SET CMP0063 NEW)
+```
+
 #### MSIX packaging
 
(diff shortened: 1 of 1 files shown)


src/desktop.md
```diff
@@ -425,6 +425,21 @@ Here are some options:
...
+`BINARY_NAME` variable set on line 4 of `windows/CMakeLists.txt` in
+your Flutter project.
+
+```cmake
+cmake_minimum_required(VERSION 3.15)
+project(windows_desktop_app LANGUAGES CXX)
+
+set(BINARY_NAME "YourNewApp")  # Change this line
+
+cmake_policy(SET CMP0063 NEW)
```
stuartmorgan, line 440: I would omit this line of context since it's much more likely to become totally wrong than the lines above.

stuartmorgan (approved):
LGTM as an isolated change, but:
- Shouldn't we have a https://flutter.dev/docs/deployment/windows at this point (in which case this should live there)? We have macOS and Linux, so it seems very strange to not have Windows.
- We should also, either in this change or a follow-up, talk about how to change the version (and maybe description, given the discussion in the issue) in the resource file.

sfshaza2 (commented):
I defer to Stuart on this change.

domesticmouse:
I'd suggest we land this change as is and put a task against me to pull out a windows page. I'll get to it after I return from carers leave.

stuartmorgan:
Works for me. I'll file it now.

sfshaza2 (approved):
lgtm. Thx, @domesticmouse and @stuartmorgan!

PR #6457, opened 2021-11-12
Update dates for the Q4 survey

jayoung-lee:
Please merge by Nov 19 (so that we can safely launch the survey the following week).

### DIFF
--- a/src/f/flutter-survey-metadata.json
+++ b/src/f/flutter-survey-metadata.json
@@ -4,9 +4,9 @@
     "title should not exceed 45 characters.",
     "startDate and endDate should follow ISO 8601 standard with a timezone offset."
   ],
-  "uniqueId": "2021Q3",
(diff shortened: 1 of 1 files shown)


sfshaza2:
Doing it by Nov 19th should be possible, if all goes as planned.  :)

sfshaza2 (approved):
lgtm

sfshaza2:
This build is failing almost immediately with the following error:
Screen Shot 2021-11-15 at 5 32 03 PM

I re-ran the jobs and it did the same. @johnpryan, @khanhnwin, could this be because  the PR was created before the switch over to docs.flutter.dev? (Though the file location is fine.)

I don't believe that this PR would actually fail the build. So, please advise!

sfshaza2:
I forcibly merged it as I don't believe it's really breaking the build.

PR #6490, opened 2021-11-19
[Proposal] update the URLs of the doc site to absolute path

chenglu:
Hello team:

Congratulations on the launch of flutter's newly designed site, we've also launched the CN site, thanks to the team!

For the documentation site, we currently plan to keep it at flutter.cn/docs, so the proposal in this PR is to replace all relative links in the documentation site with full links by adding {{site.url}}.
This change will allow us to handle the .CN sites more easily, and hopefully this proposal will be reviewed and approved!

We did this by globally searching for `]: /` and `](/`, replacing them with `]: {{site.url}}/` and `]({{site.url}}/`.

Thanks!
Luke

### DIFF
--- a/src/_includes/docs/performance.md
+++ b/src/_includes/docs/performance.md
@@ -5,4 +5,4 @@
   see [Using the Performance view][].
 {{site.alert.end}}
 
-[Using the Performance view]: /development/tools/devtools/performance
+[Using the Performance view]: {{site.url}}/development/tools/devtools/performance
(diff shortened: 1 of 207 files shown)


sfshaza2 (approved):
Wow, thatsalongPR! Thanks for this, @chenglu! It does add some overhead for us, but I guess that's ok.  :)

AlexV525:
We should apply more checks for this, to make sure new docs follow the rule by default.

PR #6498, opened 2021-11-20
Fix #5356: the dollar sign won't be copied

### LINKED ISSUE #5356: 'Linux install' page issue
Page URL: https://flutter.dev/docs/get-started/install/linux.html
Page source: https://github.com/flutter/website/tree/master/src/docs/get-started/install/linux.md

Description of issue:
there is no space between $ sign and "sudo snap install flutter --classic" the whole page have spaces between $ sign and code which is not acceptable on Ubuntu.

MrCsabaToth:

Although #5356 is closed the Linux install page has two types of shell command sections: majority are green on black and the dollar sign is not selectable / not copiable. However there are two blocks which are light themed and the dollar sign is copiable and selectable. The #5356 could have been about those. Upon inspection it looks like the `sh` source code type lets everything to be copied while the `terminal` source code type is intelligent enough to not let the shell dollar sign selected or copied. The `sh`'s look is light while the terminal is the dark themed.

The `sh` is used at a dozen more places though, although most of the time the dollar sign is not used along with it so that doesn't create copy problem. In my opinion it would be great if all `sh`s would be replaced with the `terminal` just in case someone in the future may introduce the dollar signs.

I don't want to churn too much code though so I'll only correct those additional sets of shell code sections which use a dollar sign. The only possible downside of this is that the theme of the code changes from light do dark. In my opinion that sets apart the regular source code from the shell code which can even help a little when skimming through pages.

_Issues fixed by this PR (if any):_ #5356

## Presubmit checklist

### DIFF
--- a/src/get-started/install/_get-sdk-linux.md
+++ b/src/get-started/install/_get-sdk-linux.md
@@ -12,14 +12,14 @@ Once you have snapd, you can
 [install Flutter using the Snap Store][],
 or at the command line:
 
-```sh
+```terminal
 $ sudo snap install flutter --classic
 ```
 
 {{site.alert.note}}
   Once the snap is installed, you can use the following command to display your Flutter SDK path:
 
-  ```sh
+  ```terminal
   $ flutter sdk-path
   ```
 {{site.alert.end}}
--- a/src/perf/deferred-components.md
+++ b/src/perf/deferred-components.md
@@ -293,7 +293,7 @@ You can find another example of deferred import loading in
 Use the following `flutter` command to build a deferred
 components app:
 
-```sh
+```terminal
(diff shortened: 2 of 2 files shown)


domesticmouse (approved):

MrCsabaToth:
Post release verification: I viewed the two affected pages in production and they look good to me:
* https://docs.flutter.dev/get-started/install/linux#install-flutter-using-snapd
* https://docs.flutter.dev/perf/deferred-components#step-3-building-the-app
* https://docs.flutter.dev/perf/deferred-components#running-the-app-locally

kwalrath:
Thanks for making sure this worked!

PR #6501, opened 2021-11-21
Fixes #6500 by using terminal code style

### LINKED ISSUE #6500: Issue: standardize shell command / shell code appearance
### Describe the problem

The source code type can be specified right after the triple back tick source code marker. In my opinion `sh` and `shell` type source code should mark codes which are shell snippets (part of or complete shell scripts designated to a script file) while `terminal` supposed to mark actual commands the reader should execute manually at the command prompt.

In concert with this the `sh` and `shell` code styles are light similarly as `dart` or other language styles, while the `terminal` is dark themed like a classic terminal window. An example which demonstrates both `shell` and `terminal` styles: Update path directly section of https://github.com/flutter/website/blob/41a6566e6543101751e8e1ca8b958b07cdd9148c/src/get-started/install/_path-linux-chromeos.md

While dealing with [PR#6498](https://github.com/flutter/website/pull/6498) (#5356) I did some research and found some sections which are inconsistent files.

### Expected fix

Align the currently inconsistent markings and declare the advised styling in some meta documentation about the site if there's any. I'll craft a PR with my findings.

### Additional context

_No response_

MrCsabaToth:

At some places where the user supposed to execute some command at the command prompt manually the documentation usually uses `terminal` code style. When the shell command section is just a snippet or meant to be part of a shell command file then `shell` or `sh` styles are used. I found a few places which I think should be `terminal`. The `terminal` has the extra advantage that it is able to ignore the optional dollar sign prefix (representing the command prompt) from selection or copy. I didn't go as far with this PR to supply dollar signs where they are not there. Majority of the documentation uses the dollar sign together with the `terminal` though. Let me know if I should add dollar signs as well.

#6500

## Presubmit checklist


kwalrath (commented):
I suspect we should add the `$`s, but I'll let @sfshaza2 rule on that when she returns.

sfshaza2 (approved):
lgtm

PR #6519, opened 2021-12-01
Update cookbook performance profiling to use integration_test

miquelbeltran:

In this PR I update the performance profiling guide to show package:integration_test instead of Flutter driver.

(original page: https://docs.flutter.dev/cookbook/testing/integration/profiling)

In my code I have moved the traceAction method from the flutter_driver implementation to the integration_test one, and it works.

I have also added a short explanation about the `integrationDriver` method, which already captures the performance, but that can be customized like in the example.

I have also added an alert note explaining that if you run the profiling command on a mobile device or emulator, you need to add the `--no-dds` parameter.

none

## Presubmit checklist

### DIFF
--- a/examples/cookbook/testing/integration/profiling/analysis_options.yaml
+++ b/examples/cookbook/testing/integration/profiling/analysis_options.yaml
@@ -1,0 +1,9 @@
+# Take our settings from the example directory's main analysis_options.yaml
+# file. If necessary for a particular example, this file can also include
+# overrides for individual lints.
+
+include: ../../../../analysis_options.yaml
+
+linter:
+  rules:
+    avoid_print: false
--- a/examples/cookbook/testing/integration/profiling/analysis_options.yaml
+++ b/examples/cookbook/testing/integration/profiling/analysis_options.yaml
@@ -1,9 +1,1 @@
-# Take our settings from the example directory's main analysis_options.yaml
-# file. If necessary for a particular example, this file can also include
-# overrides for individual lints.
-
 include: ../../../../analysis_options.yaml
-
-linter:
-  rules:
-    avoid_print: false
--- a/examples/cookbook/testing/integration/profiling/test_driver/perf_driver.dart
+++ b/examples/cookbook/testing/integration/profiling/test_driver/perf_driver.dart
@@ -6,7 +6,7 @@ Future<void> main() {
   return integrationDriver(
     responseDataCallback: (data) async {
       if (data != null) {
-        final timeline = driver.Timeline.fromJson(data['timeline']);
+        final timeline = driver.Timeline.fromJson(data['scrolling_timeline']);
 
         // Convert the Timeline into a TimelineSummary that's easier to
         // read and understand.
--- a/examples/cookbook/testing/integration/profiling/test_driver/perf_driver.dart
+++ b/examples/cookbook/testing/integration/profiling/test_driver/perf_driver.dart
@@ -1,0 +1,29 @@
+// #docregion Timeline
+import 'package:integration_test/integration_test_driver.dart';
+import 'package:flutter_driver/flutter_driver.dart' as driver;
+
+Future<void> main() {
+  return integrationDriver(
+    responseDataCallback: (data) async {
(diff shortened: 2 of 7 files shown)


miquelbeltran:
cc. @RedBrogdon

sfshaza2:
@johnpryan @khanhnwin, can either of you answer @miquelbeltran's question?

johnpryan:
> Timeline and TimelineSummary classes: Those seem to exist only on flutter_driver and I could not find an equivalent on integration_test. Do you know if there's some equivalent I am missing? or should I use the ones in flutter_driver to show the summary?

I don't think there's an equivalent in integration_test, so the approach you are using here looks correct. The [testing_app README](https://github.com/flutter/samples/blob/master/testing_app/README.md?plain=1#L33) and the [the docs for traceAction](https://api.flutter.dev/flutter/package-integration_test_integration_test/IntegrationTestWidgetsFlutterBinding/traceAction.html) have  instructions similar to yours. More context: https://github.com/flutter/flutter/issues/74930

examples/cookbook/testing/integration/profiling/analysis_options.yaml
```diff
@@ -0,0 +1,9 @@
+# Take our settings from the example directory's main analysis_options.yaml
+# file. If necessary for a particular example, this file can also include
+# overrides for individual lints.
+
+include: ../../../../analysis_options.yaml
+
+linter:
+  rules:
+    avoid_print: false
```
domesticmouse, line 9: Any chance of removing dependence on `print`?

domesticmouse (approved):

examples/cookbook/testing/integration/profiling/analysis_options.yaml
miquelbeltran: thanks, well spotted, print is not even used in the sample

miquelbeltran:
Thanks @johnpryan ! I did a couple of adjustments and the PR is ready for review now.

examples/cookbook/testing/integration/profiling/test_driver/perf_driver.dart
```diff
@@ -0,0 +1,29 @@
...
+
+        // Convert the Timeline into a TimelineSummary that's easier to
+        // read and understand.
+        final summary = driver.TimelineSummary.summarize(timeline);
+
+        // Then, write the entire timeline to disk in a json format.
+        // This file can be opened in the Chrome browser's tracing tools
+        // found by navigating to chrome://tracing.
+        // Optionally, save the summary to disk by setting includeSummary
+        // to true
```
johnpryan, line 19: This looks great. I'm wondering why [the sample in flutter/samples](https://github.com/flutter/samples/blob/master/testing_app/integration_test/perf_driver.dart) uses `data.entries` but your example uses `data.timeline`  - is it because there's [more than one test](https://github.com/flutter/samples/blob/master/testing_app/integration_test/perf_test.dart)? Dealing with these traces has been tricky in the past.

johnpryan (approved):

examples/cookbook/testing/integration/profiling/test_driver/perf_driver.dart
miquelbeltran: yes, good question, the [traceAction method](https://api.flutter.dev/flutter/package-integration_test_integration_test/IntegrationTestWidgetsFlutterBinding/traceAction.html) stores the timeline by default in the `data['timeline']` entry, so if you have more than one perf test, your should change the `reportKey` parameter to something else. I could make the parameter explicit in this example to something like `scrolling_timeline` to avoid confusion.

miquelbeltran: I pushed a change where I specify the `reportKey` instead of using the default one, and added a short explanation about it.

domesticmouse:
@miquelbeltran is this ready to land?

miquelbeltran:
@domesticmouse yes!

domesticmouse:
Thanks @miquelbeltran!

PR #6592, opened 2021-12-20
Fix documentation about dartPluginClass

stuartmorgan:
The current docs say that Dart-only plugin implementations are supported for mobile in 2.8+, but it turns out that there is a bug that wasn't caught until after 2.8 reached stable, and that's not the case. This corrects the docs to reflect the actual state, which is that while `dartPluginClass` works to register Dart code for those platforms as of 2.8, native code is currently still required.

See https://github.com/flutter/flutter/issues/95012

## Presubmit checklist


khanhnwin (approved):
LGTM!

PR #6696, opened 2022-01-19
Push back DevTools survey dates

kenzieschmoll:
There is a url parsing bug in DevTools that will prevent this survey from being able to be opened. Pushing back the date for now so that users do not see the prompt. Once we know when the DevTools fix can get into dart / flutter stable, we can update the dates accordingly.


kenzieschmoll:
CC @cobblest

sfshaza2 (approved):
lgtm

PR #6725, opened 2022-01-26
Remove unnecessary null safety query parameters from DartPads

parlough:
DartPad only supports null safety now and this query parameter was removed, so it is not doing anything and can be removed.


domesticmouse (approved):

domesticmouse:
PTAL @johnpryan

sfshaza2:
I loooove this PR! The more cleanup the better. But, as @domesticmouse requested, will wait for @johnpryan's review.

sfshaza2:
@johnpryan, please review.

sfshaza2:
OK, John is swamped and Brett has approved, so I am merging. We can revert if needed.

PR #6734, opened 2022-01-27
Revert "Update to Ruby 3.1"

khanhnwin:
Reverts flutter/website#6726

Ruby 3.1 and the new dependency versions seem to be breaking the build. This previously happened over the holidays [context](https://github.com/flutter/website/pull/6616). cc: @btobin was going to restructure things to match up with dart.dev? Reverting for now to unbreak the build.


parlough:
It seems this is due to the i18n gem version 0.9.0 being removed from rubygems today: https://rubygems.org/gems/i18n

parlough (approved):

domesticmouse (approved):

khanhnwin:
Hey @parlough! Yeah, that's really weird :/ I noticed that 1.9.0 was gone and opted for the revert option as I recall Ruby 3.1 being an issue as well. However, if 3.1 isn't an issue and we can just stick to a different version of i18n, that's totally good with me as well!

parlough:
@khanhnwin I was able to fix the 3.1 issues in that PR, but for now you can revert these changes. I can follow up with an update once I'm sure i18n won't be retracted again :)

khanhnwin:
Awesome, sounds good to me. Thank you @parlough :)

Issue #6803, opened 2022-02-10
[PAGE ISSUE]: 'Adding assets and images'


jaded0:
### Page URL

https://docs.flutter.dev/development/ui/assets-and-images/

### Page source

https://github.com/flutter/website/tree/main/src/development/ui/assets-and-images.md

### Describe the problem

Following the exact directions under 'Bundling of Package Assets' doesn't do the trick. The assets are unable to be found using this method, and are inaccessible by apps using the package

danagbemava-nc:
Hi @jaded0, can you share the errors you run into as well your flutter doctor -v?

jaded0:
They were the errors that come when the asset cannot be found. 
The solution was that I needed to also reference the files in the same way that they were referenced in my pubspec.yaml. eg:
`AssetImage('packages/fancy_backgrounds/backgrounds/background1.png'),`
Based on the way the documentation is worded, I might suggest an edit to clarify that the solution under 'Bundling of Package Assets' doesn't imply that you can still refer to that asset as you would in an app and not a package.

danagbemava-nc:
Thanks for the feedback

danagbemava-nc:
Since the PR has been merged, I'm going to close this as fixed.

PR #6805, opened 2022-02-10
Clarify bundling of package assets.

jaded0:
Clarify that the solution under 'Bundling of Package Assets' doesn't imply that you can still refer to that asset as you would in an app and not a package. It's a tad unclear:
https://stackoverflow.com/q/54740732/5229301

 #6803


jaded0:
Did it!

PR #6809, opened 2022-02-11
Code excerpts for cookbook/effects/staggered-menu-animation

khanhnwin:
_Description of what this PR is changing or adding, and why:_ Just what the the title says. I also fixed some of the excerpts that didn't match up with the DartPad at the bottom, like using a RaisedButton instead of an ElevatedButton.

## Presubmit checklist
- [X] This PR uses [semantic line breaks](https://github.com/dart-lang/site-shared/blob/master/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks) of 80 characters or fewer.


albertodev01 (commented):
I'd like to leave the same comment I made in the other PR (https://github.com/flutter/website/pull/6810). It would be better if you replaced functions with actual widgets here instead of doing it later 🙂 

That creates confusion to new Flutter developers because they see examples using functions to return widgets while this is not recommended by the [Stateless](https://github.com/flutter/flutter/blob/db747aa133/packages/flutter/lib/src/widgets/framework.dart#L441) / [Stateful](https://github.com/flutter/flutter/blob/db747aa133/packages/flutter/lib/src/widgets/framework.dart#L663) documentation and the official [YouTube](https://www.youtube.com/watch?v=IOyq-eTRhvo) channel

khanhnwin:
Noted, @albertodev01! I opened issue #6813 for it!

PR #6810, opened 2022-02-11
Code excerpts for cookbook/effects/shimmer-loading

khanhnwin:
_Description of what this PR is changing or adding, and why:_ More code excerpts migrations! These cookbook projects are getting more and more complex so I'm opening separate PRs for each one.


domesticmouse (approved):

examples/cookbook/effects/shimmer_loading/lib/shimmer_loading_items.dart
```diff
@@ -0,0 +1,203 @@
...
+        children: [
+          _buildImage(),
+          const SizedBox(height: 16),
+          _buildText(),
+        ],
+      ),
+    );
+  }
+
+  Widget _buildImage() {
```
domesticmouse, line 55: Consider breaking these private build methods out as their own stand alone widgets

albertodev01 (commented):
Since this code is being added, would there be the possibility to replace helper methods (functions) with widget at this stage?

Updating code excerpts and the source code is very complicated and may also break the website build for some time while synching repos (see https://github.com/flutter/website/pull/6763)

khanhnwin:
Hey y'all - We definitely do want to refactor the code to use standalone widgets instead of helper methods. Though, at the moment our goal is to get them in to a project in the examples/ directory just to make sure they're being analyzed in the build tests. I've been updating small things along the way like replacing deprecated widgets and such, but completely refactoring the code architecture will be a bit of a bigger effort. We can make a note to loop back and review all of the excerpt code to make sure that they're up to date with our recommendations.

@albertodev01 Sorry for the confusion! Updating the code excerpts for these cookbook examples are not as complex as the codelab ones. The codelab code changes are difficult because they're a submodule. These cookbook projects as in this case are in the flutter/website repo so it's much less complex.

PR #6882, opened 2022-03-07
Update top-menu link for What's New

mit-mit:
This now has its own page, so we should link to that directly


khanhnwin (approved):

khanhnwin:
LGTM! Thanks @mit-mit!

PR #6940, opened 2022-03-21
Avoid some unnecessary checks during CI tasks

parlough:
_Description of what this PR is changing or adding, and why:_ Actions are taking a long time. It would be nice to decrease checks and it's also good to reduce unnecessary compute work. These changes avoid some unnecessary checks.

There's not as large of an improvement as I hoped, but doesn't hurt. 

_Issues fixed by this PR (if any):_ N/A

## Presubmit checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,6 +14,7 @@ on:
 permissions: read-all
 
 env:
+  CI: true
   NODE_ENV: production
   NODE_VERSION: 17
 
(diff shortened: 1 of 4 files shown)


domesticmouse:
I'm concerned about people cargo culting these into their own CI systems. Is the speed gain enough to warrant these potential adverse down stream effects?

I'd also strongly suggest marking up the instances with comments that explain why the arguments are there.

parlough:
That definitely makes sense. Since the improvements aren't as large as I hoped, I'll close this for now :)

PR #6950, opened 2022-03-22
update committee chair email

itsjustkevin:
Updated committee chair email to point to the correct place.


sfshaza2 (approved):
Great catch, @itsjustkevin!

PR #7012, opened 2022-04-12
Fastlane: build Xcode project only once

maxlapides:
Fortunately, we don't have to rebuild the entire iOS project anymore :)

It actually looks like we will be able to update these docs again in the near future after @jmagman's PR https://github.com/flutter/flutter/pull/97672 lands on stable. At that point, most projects can skip the `fastlane build_app` step altogether.

It may be worth mentioning that `build_app` can be used to set Xcode values like `configuration` which is really helpful if you are using fastlane match and you want to have separate configurations for ad-hoc vs app store builds (without authoring a plist file).

## Presubmit checklist

### DIFF
--- a/src/deployment/cd.md
+++ b/src/deployment/cd.md
@@ -109,8 +109,7 @@ process to a continuous integration (CI) system.
 
 1. Build the release mode app.
     * ![Android]({{site.url}}/assets/images/docs/cd/android.png) `flutter build appbundle`.
-    * ![iOS]({{site.url}}/assets/images/docs/cd/ios.png) `flutter build ios --release --no-codesign`.
-    No need to sign now since fastlane will sign when archiving.
+    * ![iOS]({{site.url}}/assets/images/docs/cd/ios.png) `flutter build ipa`.
 1. Run the Fastfile script on each platform.
     * ![Android]({{site.url}}/assets/images/docs/cd/android.png) `cd android` then
     `fastlane [name of the lane you created]`.
--- a/src/deployment/cd.md
+++ b/src/deployment/cd.md
@@ -53,7 +53,7 @@ Visit the [fastlane docs][fastlane] for more info.
     (This is required for the scripts that deploy for iOS.)
 1. Create your Flutter project, and when ready, make sure that your project builds via
     * ![Android]({{site.url}}/assets/images/docs/cd/android.png) `flutter build appbundle`; and
-    * ![iOS]({{site.url}}/assets/images/docs/cd/ios.png) `flutter build ios --release --no-codesign`.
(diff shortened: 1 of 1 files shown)


sfshaza2:
@jmagman, this lgtm, and I see that the reference PR is merged. Can we go ahead and land this?

jmagman (approved):
I'm not a fastlane expert but this LGTM from my reading of https://docs.fastlane.tools/actions/build_ios_app/.  Thanks for the contribution @maxlapides!

maxlapides:
> @maxlapides below under `Running deployment locally` it says:
> 
> > `flutter build ios --release --no-codesign`.  No need to sign now since fastlane will sign when archiving.
> 
> Based on this lane, is that still the right command? I didn't add `--no-codesign` to `flutter build ipa` but maybe I should add that.

Great catch! I just updated that command to be `flutter build ipa`.

But, to answer your question, I do think it would be _great_ if we could have the `--no-codesign` flag on the `flutter build ipa` step. It would simplify our CI process a little bit :)

jmagman:
Since you only need the `xcarchive` and not `ipa` would it also be useful to have a `flutter build xcarchive --skip-ipa` flag that would create the xcarchive but not try to build the IPA?
Thanks again for your insights here.  And please file feature requests at https://github.com/flutter/flutter/issues if you have further suggestions for improving these commands.

maxlapides:
@jmagman Yes I think that  `--skip-ipa` flag would also be helpful for us :) Our use case is we want to run `flutter build ipa` to build an xcarchive, then run fastlane's `build_app` twice: once for an ad-hoc build, then once for an app store build.

I haven't tested out our build after your PR that adds automatic settings for the export options plist, but I assume that we will still want to use fastlane's `build_app` after `flutter build ipa` because it gives us easy control over the Xcode settings.

maxlapides:
Or, perhaps `--no-codesign` can automatically skip the IPA step, and then we might not need the `--skip-ipa` flag 🤔

jmagman (approved):
> @jmagman Yes I think that `--skip-ipa` flag would also be helpful for us :) Our use case is we want to run `flutter build ipa` to build an xcarchive, then run fastlane's `build_app` twice: once for an ad-hoc build, then once for an app store build.
> 
> I haven't tested out our build after your PR that adds automatic settings for the export options plist, but I assume that we will still want to use fastlane's `build_app` after `flutter build ipa` because it gives us easy control over the Xcode settings.

Okay, let's take the conversation to the Flutter GitHub issue tracker instead of this website PR.

@sfshaza2 this still LGTM

sfshaza2 (approved):
As per Jenn, lgtm. Thanks so much, @maxlapides!

PR #7055, opened 2022-04-21
Fix nokogiri install error

btobin:
- fixes nokogiri error thrown when running Jekyll serve after setup on aarch64 (and possibly x86)
   `ERROR: It looks like you're trying to use Nokogiri as a precompiled native gem on a system with glibc < 2.17: ...`


MiraMarshall (approved):
LGTM! I no longer see the error and have the repo running locally.

khanhnwin:
Thanks Brian!

PR #7128, opened 2022-05-12
Add a note about the widgets binding issue people are reporting

Hixie:
Adds some documentation regarding variants of https://github.com/flutter/flutter/issues/103561.

## Presubmit checklist

### DIFF
--- a/src/development/tools/sdk/release-notes/release-notes-3.0.0.md
+++ b/src/development/tools/sdk/release-notes/release-notes-3.0.0.md
@@ -9,6 +9,67 @@ For information about subsequent bug-fix releases, see
 
 [Hotfixes to the Stable Channel]: https://github.com/flutter/flutter/wiki/Hotfixes-to-the-Stable-Channel
 
+## If you see warnings about bindings
+
+You may see warnings like this:
+```
+Warning: Operand of null-aware operation '!' has type 'SchedulerBinding' which excludes null.
+```
+...when migrating to Flutter 3.
+
+These are caused by a simplification of our API (the `instance` property on bindings is now non-nullable),
+combined with a very eager compiler which wants to report any case where redundant null-aware operators
+(such as `!` and `?.`) are used when they not necessary.
+
+If this happens, there may be several causes with different solutions:
+
+### Dependencies
+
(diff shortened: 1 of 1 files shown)


christopherfujino (approved):
lgtm

src/development/tools/sdk/release-notes/release-notes-3.0.0.md
```diff
@@ -9,6 +9,67 @@ For information about subsequent bug-fix releases, see
 
 [Hotfixes to the Stable Channel]: https://github.com/flutter/flutter/wiki/Hotfixes-to-the-Stable-Channel
 
+## If you see warnings about bindings
+
+You may see warnings like this:
+```
+Warning: Operand of null-aware operation '!' has type 'SchedulerBinding' which excludes null.
+```
+...when migrating to Flutter 3.
```
domesticmouse, line 18: @sfshaza2 I'm curious what your preferred way of handling a block of code in the middle of a sentence.

domesticmouse (approved):
PTAL @sfshaza2

Hixie:
(FWIW, I'm happy to make grammatical fixes in follow-up PRs, but would really like to land this ASAP since it's high-profile.)

JacoFourie:
Dart 2.17 has new lint rules. I think it is those rules that are breaking Flutter 3. I also have this issue. I have been looking around and found this.

https://github.com/dart-lang/lints/blob/main/CHANGELOG.md#200

I think it may be this one

unnecessary_null_aware_assignments

My code will compile if I remove the !. But I have third-party libs that also have them in that I do not have control over. I downgraded my Flutter to 2.10.5 again so I don't have time to look at this now but if somebody can override those lint rules and check if the app compiles.

ekuleshov:
I'm wondering why such backward-incompatible API had been introduced in Flutter 3? It is affecting a lot of plugins and essentially with most of those 3rd party plugins you can't use the same version of plugin between Flutter 2 and Flutter 3.

Wouldn't it be better to keep the old method with existing nullability marked as deprecated in Flutter 3 and have a new method with new nullability added in Flutter 3?

Issue #7172, opened 2022-05-20
[PAGE ISSUE]: 'Adding Flutter to Android'


fhsbatista:
### Page URL

https://docs.flutter.dev/development/add-to-app/android/

### Page source

https://github.com/flutter/website/tree/main/src/development/add-to-app/android/index.md

### Describe the problem

The page is empty

sfshaza2:
@khanhnwin, this page IS empty (as is the markdown file). Many of our index pages are auto-generated and this appears to be one of those. Can you take a look?

khanhnwin:
Looking!

khanhnwin:
Merged and Fixed!

PR #6454, opened 2021-11-11
Refactor docs to live on their own

btobin:
- Removes marketing pages
- Removes all unused marketing assets
- Moves all docs pages to src/
- Updates all references to /docs/
- Updates all marketing pages for new url
- Adds config file for staging urls and testing across 2 sites, including updated docker for passing configs to build
- Adds extends main nav with bootstrap dropdown menus
- Adds staging action for testing
- Adds new Firebase default project 
- Removes unused NPM package
- Upgrade Node to v17

PR #7197, opened 2022-05-26
fix TOC on auto generated index pages

khanhnwin:
_Description of what this PR is changing or adding, and why:_ When the website was refactored to live on their own in #6454, the code that autogenerates the TOC on index pages wasn't updated. As a result, the index pages were empty since the code assumed that the TOC wasn't relevant at the top most level. 

**Example**
Paths used to be: `/docs/development/...`
Paths now: `/development/...`

_Issues fixed by this PR (if any):_ #7172 

This PR also reverts the changes the changes that I made to sidenav.yml in #6785 thinking that I was solving the problem by deleting the empty index pages. (The root cause was the code for the TOC itself, which is now fixed in this PR.)

## Presubmit checklist

### DIFF
--- a/src/_data/sidenav.yml
+++ b/src/_data/sidenav.yml
@@ -47,6 +47,7 @@
       permalink: /reference/tutorials
 
 - title: Development
+  permalink: /development
   expanded: true
   children:
     - title: User interface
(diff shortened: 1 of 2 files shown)


khanhnwin:
This same issue also came up in #6565 - which we also tried to fix with #6648 by dropping the index page. 

CC: @domesticmouse

domesticmouse (approved):

PR #7243, opened 2022-06-08
Update `breaking-changes/toggleable-active-color.md`

TahaTesser:
This PR updates migration guide for https://github.com/flutter/flutter/pull/97972

cc: @Piinks 

## Presubmit checklist

### DIFF
--- a/src/release/breaking-changes/toggleable-active-color.md
+++ b/src/release/breaking-changes/toggleable-active-color.md
@@ -47,7 +47,52 @@ Code after migration:
 final ThemeData theme = ThemeData();
 MaterialApp(
   theme: theme.copyWith(
-    colorScheme: theme.colorScheme.copyWith(secondary: myColor),
+    switchTheme: SwitchThemeData(
+      thumbColor: MaterialStateProperty.resolveWith<Color?>(
+          (Set<MaterialState> states) {
+        if (states.contains(MaterialState.disabled)) {
+          return null;
+        }
+        if (states.contains(MaterialState.selected)) {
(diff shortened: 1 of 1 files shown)


Piinks (commented):
Thanks for the updates @TahaTesser! I will review this once the change lands, I am still migrating customers and so we may still get some insights on migration gotchas that we'll want to include here.

domesticmouse (approved):

Piinks (commented):
Somewhere we should add mention how the default value of toggleableActiveColor was resolved.

I have seen some customers that do not set toggleableActiveColor, or any of the component themes (switch/checkbox/radio). In this case they are broken because ColorScheme.secondary is a different default. 

toggleableActiveColor would default to...
light mode: ThemeData.accentColor ?? ThemeData.primarySwatch[600]
dark mode: Colors.tealAccent[200]

and ThemeData.primarySwatch would default to Colors.blue if null.

These defaults aren't super apparent unless you go digging around in the source code. It would be helpful to include here for folks that need to migrate that way. :)

sfshaza2:
@TahaTesser, can you address @Piinks comments?

TahaTesser:
> @TahaTesser, can you address @Piinks comments?

Thank you, I'm looking into the details to update this soon.

Piinks:
Thanks for checking @TahaTesser! I was actually wondering if it is a bug with https://github.com/flutter/flutter/pull/97972
I will follow up there with a test case

sfshaza2 (approved):
lgtm!

PR #7266, opened 2022-06-14
remove concurrency property from deploy workflow

khanhnwin:
_Description of what this PR is changing or adding, and why:_ #7179 added the concurrency flag to test. However, "When concurrency is specified at the job level, order is not guaranteed for jobs or runs that queue within 5 minutes of each other." Also, instead of being a queue there's only an "active" job and a single "pending" job. This has the potential to prevent newer PRs from being deployed when multiple are approved + merged in a short time (a few seconds/minutes): [Example Run](https://github.com/flutter/website/actions/runs/2497217013)

cc: @btobin 

## Presubmit checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,8 +42,6 @@ jobs:
 
   deploy:
     needs: test
-    concurrency: 
-      group: prod-deploy-group
     runs-on: ubuntu-latest
     if: |
       github.event_name == 'push'


domesticmouse (approved):
We are going to need an alternate fix for the clashing builds problem, yeah?

sfshaza2:
I read through this and I am afeard to approve it. I leave it to @domesticmouse @btobin @johnpryan, etc.

Issue #7211, opened 2022-05-30
[PAGE ISSUE]: 'Build and release a macOS app'


mmattes:
### Page URL

https://docs.flutter.dev/deployment/macos/

### Page source

https://github.com/flutter/website/tree/main/src/deployment/macos.md

### Describe the problem

In Create a build archive with Codemagic CLI tools Point 13. Package the App the following code snipped is mentioned. But jq '.[0] only selects the first entry in the keychain which does not need to be the one with the common_name containing "Mac Developer Installer"

```
APP_NAME=$(find $(pwd) -name "*.app")
PACKAGE_NAME=$(basename "$APP_NAME" .app).pkg
xcrun productbuild --component "$APP_NAME" /Applications/ unsigned.pkg

INSTALLER_CERT_NAME=$(keychain list-certificates \
          | jq '.[0]
            | select(.common_name
            | contains("Mac Developer Installer"))
            | .common_name' \
          | xargs)
xcrun productsign --sign "$INSTALLER_CERT_NAME" unsigned.pkg "$PACKAGE_NAME"
rm -f unsigned.pkg 
```

### Expected fix

Changing it to jq '.[] will iterate over the whole array and will find the right certificate no matter if it is the first or second. 

```
APP_NAME=$(find $(pwd) -name "*.app")
PACKAGE_NAME=$(basename "$APP_NAME" .app).pkg
xcrun productbuild --component "$APP_NAME" /Applications/ unsigned.pkg

INSTALLER_CERT_NAME=$(keychain list-certificates \
          | jq '.[]
            | select(.common_name
            | contains("Mac Developer Installer"))
            | .common_name' \
          | xargs)
xcrun productsign --sign "$INSTALLER_CERT_NAME" unsigned.pkg "$PACKAGE_NAME"
rm -f unsigned.pkg 
```

MiraMarshall:
Closing, this issue has been fixed.

PR #7290, opened 2022-06-21
Fix macOS deployment Codemagic CLI tools Point 13 snippet

VeArnold:

The snippet update assures that the right certificate is found in the case of multiple certificates. The fix was suggested in the issue marked below and was deemed to be accurate.

[Issue 7211](https://github.com/flutter/website/issues/7211)

### DIFF
--- a/src/deployment/macos.md
+++ b/src/deployment/macos.md
@@ -382,7 +382,7 @@ PACKAGE_NAME=$(basename "$APP_NAME" .app).pkg
 xcrun productbuild --component "$APP_NAME" /Applications/ unsigned.pkg
 
 INSTALLER_CERT_NAME=$(keychain list-certificates \
-          | jq '.[0]
+          | jq '.[]
             | select(.common_name
             | contains("Mac Developer Installer"))
             | .common_name' \


jmagman (approved):
This matches https://docs.codemagic.io/yaml-code-signing/signing-macos/#packaging-the-application LGTM.

Also, I wish we didn't have a copy of these instructions and would just point to https://docs.codemagic.io/yaml-code-signing/signing-macos, so we're never out of date...

domesticmouse:
I'd be happy to just refer to codemagic's documentation. I strongly dislike duplicating external docs.

Thoughts @sfshaza2?

VeArnold:
> I'd be happy to just refer to codemagic's documentation. I strongly dislike duplicating external docs.
> 
> Thoughts @sfshaza2?

I think the main issue with this here is that the guide here instructs users to use Codemagic CLI tools independent of a CI/CD pipeline, while the scripts presented in Codemagic documentation are part of the `codemagic.yaml` structure, and thus perhaps a source of some confusion.

jmagman:
> I think the main issue with this here is that the guide here instructs users to use Codemagic CLI tools independent of a CI/CD pipeline, while the scripts presented in Codemagic documentation are part of the `codemagic.yaml` structure, and thus perhaps a source of some confusion.

I can see that being confusing, and our docs interleave Codemagic and `flutter` commands.

sfshaza2:
Yes, I agree with @jmagman. When I looked at the codemagic page a week or two ago, it was NOT an easy fix to just redirect to that page. It was quite confusing, in fact. I like the idea of referring to that page, but it has to be clear and coherent. And, yes, I'm on vacation this week.  :)

PR #7306, opened 2022-06-24
Clean up supported platform descriptions

stuartmorgan:
Fixes inconsistencies in the listings:
- The initial listing now includes both Google-tested and best-effort support tiers for all platforms; it included both for some platforms, but only Google-tested for others.
- Fixes the Unsupported platforms section for Android; it was listing best-effort platforms as unsupported.
- Adds a missing version number to the Linux best-effort entry.
- Adds 32-bit Linux to the Unsupported section for consistency with the Windows entry.
- Uses ranges for Android versions instead of listing each one, as we do for iOS.

Other minor changes:
- Explicitly lists 64-bit as a requirement for Windows in the main listing, instead of relying on people finding the note about 32-bit in the Unsupported section.
- Explicitly lists 64-bit as a requirement for Linux in the main listing.
- Adds OS version numbers for macOS, since it's hard to remember the name->version mapping for every macOS version.
- Adds an Android OS version number in the mail listing, to make it easy to map the support to user-visible OS numbers instead of needing to know the SDK->OS version mapping.

Fixes https://github.com/flutter/flutter/issues/106301

## Presubmit checklist

### DIFF
--- a/src/development/tools/sdk/release-notes/supported-platforms.md
+++ b/src/development/tools/sdk/release-notes/supported-platforms.md
@@ -11,25 +11,18 @@ Flutter supports the following platforms:
 
 |Platform|Version                       |Channels |
 |--------|------------------------------|---------|
-|Android | API 19 & above               | All     |
+|Android | API 16 (Android 4.1) & above | All     |
 |iOS     | iOS 9 & above                | All     |
-|Linux   | Debian 10 & above            | All     |
-|macOS   | El Capitan & above           | All     |
+|Linux   | Debian, 64-bit               | All     |
+|macOS   | El Capitan (10.11) & above   | All     |
 |Web     | Chrome 84  & above           | All     |
 |Web     | Firefox 72.0 & above         | All     |
 |Web     | Safari on El Capitan & above | All     |
 |Web     | Edge 1.2.0 & above           | All     |
-|Windows | Windows 10 & above           | All     |
+|Windows | Windows 7 & above, 64-bit    | All     |
 
 All channels include master, beta,
 and stable channels. 
-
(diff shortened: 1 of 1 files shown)


stuartmorgan:
There's also an inconsistency in the table ordering: the main table uses alphabetical ordering, while the others use: {mobile}, {desktop in most-to-least market share order}. Should we standardize on one or the other?

sfshaza2 (commented):
@stuartmorgan, this mostly looks great! But why did you remove the note about the dev channel? The 2.8 release wasn't that long ago.

src/development/tools/sdk/release-notes/supported-platforms.md
```diff
@@ -75,28 +59,29 @@ Note that Android SDK 20 is covered by
 testing Android SDK 19, as the differences
 between the two platform versions are minimal.
```
Hixie, line 60: we could just remove this paragraph, i'm not sure what it really adds

Hixie:
> why did you remove the note about the dev channel? The 2.8 release wasn't that long ago.

This is a weird page on which to draw attention to it, I support not mentioning it here. We have mentions of it in lots of other places still.

Hixie (approved):
Standardising on one order seems reasonable.

The changes all LGTM.

fyi @timsneath in case this suggests any policy changes we should be considering (the changes here are for accuracy and represent what we are actually doing today, but that may be in conflict with what we are intending to do)

LGTM

johnpryan (approved):

Hixie:
I'd go with alphabetical.

sfshaza2:
lgtm

PR #7329, opened 2022-06-30
Add desktop to plugin example

stuartmorgan:
Adds explanations of how to edit plugin code with native tooling for Linux, macOS, and Windows.

Part of https://github.com/flutter/flutter/issues/79204

## Presubmit checklist


stuartmorgan:
The Linux instructions here are a little rougher than the other platforms since unlike on most of our other platforms our build system isn't designed around something that has its own IDE. But this will still let someone get a setup that does C++ autocomplete, jump-to-definition, etc. even though building has to be done separately.

sfshaza2 (approved):
lgtm. Thanks, @stuartmorgan!

Issue #7233, opened 2022-06-05
Clarify plugin doc discussion of Dart-only plugins


stuartmorgan:
[This section](https://docs.flutter.dev/development/packages-and-plugins/developing-packages#dart-only-platform-implementations) of the plugin docs doesn't make it clear that it only applies to non-web platforms. Since the difference is non-obvious, we should clearly highlight it at the start of the section.

PR #7349, opened 2022-07-12
Improve discussion of Dart in non-web plugins

stuartmorgan:
- Clarifies that the Dart section is about non-web plugins, since this has caused some confusion (due to web plugins also being Dart-only, but having a different system).
- Adds discussion of hybrid plugins, as this structure is now widely used in 1P plugins.
- Removes obsolete comment about Dart-only plugins not working on mobile, and removes discussion of minimum versions that are several stable releases old.

Fixes https://github.com/flutter/website/issues/7233

## Presubmit checklist


sfshaza2 (approved):
lgtm! Thanks, @stuartmorgan!

PR #7376, opened 2022-07-18
Move "Our Values" section from individual job descriptions into the overview page

mkustermann:
This makes individual job postings more concise and avoids differences
in the "Our Values" section across postings.

### DIFF
--- a/src/jobs/_template.md
+++ b/src/jobs/_template.md
@@ -20,37 +20,6 @@ toc: false
 
 <TODO: Role Description>
 
-## Our values
-
-### Mentorship
-
-Upon joining Google, you will be paired with a formal mentor,
-who will help guide you in the process of ramping up, forging relationships,
-and learning the systems you’ll need to do your job.
-Your manager can also help you find mentors who can coach you
-as you navigate your career at Google. In addition to formal mentors,
-we work and train together so that we are always learning from one another,
-and we celebrate and support the career progression of our team members.
-
-### Inclusion
-
-Here on the Flutter team and at Google, we embrace our differences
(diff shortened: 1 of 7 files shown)


khanhnwin (approved):

domesticmouse:
> My only concern is that many people might skip the index page and go straight to the job posting.

Can this be placed into a template and included in each post?

mkustermann:
> Can this be placed into a template and included in each post?

It is in the template (see the change in `src/jobs/_template.md`), but it's not in every existing posting. The phrasing of it is also flutter specific, which may be a reason it wasn't included in some postings - since we have flutter and dart postings.

> My only concern is that many people might skip the index page and go straight to the job posting.

It seems to me that main landing pages are flutter.dev/jobs & dart.dev/jobs which is the index page. When searching for `flutter team jobs` etc one ends up also in the index landing page.

If there's strong opinions on not doing this, I can also drop this PR, though IMHO it would make sense to do it.

sfshaza2 (changes requested):
I prefer one of the following solutions:
- Update the template to be more Dart inclusive and less Flutter-focused.
- Update each job listing to refer to the index page for that info. You see, people DO tweet the job postings, and could easily miss that info.

Of the two approaches, I prefer the former. But I totally understand if you would rather not expend that effort. I don't mind doing it.

Also, just a little FYI, but dart.dev/jobs just redirects to our site.

mkustermann:
> Update the template to be more Dart inclusive and less Flutter-focused.
> 
> I don't mind doing it.

That works as well - though the section should then be consistently there for all postings.

@sfshaza2  Thanks for volunteering, then I'll close this PR.

PR #7464, opened 2022-08-19
Updated readMe for releases

kahdichienja:
Capitalise the first letter on the table.

**IMPORTANT:** Due to work on the docs.flutter.dev infrastructure, this repo is **not accepting pull requests**.

Instead of creating a PR, please file an issue (https://github.com/flutter/website/issues/new/choose) about the needed change.

We expect to start accepting PRs again the week of August 31.

### DIFF
--- a/src/development/tools/sdk/releases.md
+++ b/src/development/tools/sdk/releases.md
@@ -1,8 +1,8 @@
 ---
-title: Flutter SDK releases
-short-title: Releases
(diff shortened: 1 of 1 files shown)


domesticmouse:
@kahdichienja I'm curious why you are making this change? The keys for that map are supposed to be lower case, if I understand correctly.

sfshaza2 (commented):
The YAML matter should be lower case. Closing.

PR #7207, opened 2022-05-27
migration guide for loadBuffer

jonahwilliams:
This can't land yet as the PR in question hasn't rolled into google

PR #7480, opened 2022-08-30
Missing `ImageProvider.loadBuffer` BC from list

AlexV525:
https://github.com/flutter/website/pull/7207 doesn't bring it to the list. Also, fix the trailing format issues.

image


AlexV525:
It should be okay if we land this tomorrow, after the freeze, and before the new version is released.

domesticmouse (approved):

PR #7580, opened 2022-09-22
Infrastructure and build optimizations

btobin:
- Optimizes build stages in Dockerfile, makes test stage an entrypoint
- Cleans up Makefile, adds new utility commands
- Removes unnecessary env vars and var files
- Upgrades check-links.sh script, makes this easy to run in Docker container and as part of build
- Removes null-safety tests 
- Removes unnecessary NPM packages
- Updates action build file, removing unnecessary installs and simplifying flow
- Updates README

@khanhnwin Take a look please and we can discuss!


sfshaza2 (commented):
LGTM. This sounds great, @btobin! I'm not merging because I want @khanhnwin to review.

domesticmouse (approved):

Dockerfile
```diff
@@ -1,40 +1,34 @@
...
-COPY ./ ./
+# ============== INSTALL FLUTTER ==============
+# TODO recommended to move to a SHA based install for security reasons
+FROM base AS flutter
 
-# So that we have the most up to date submodules
-RUN git submodule update --init --recursive
+# TODO to avoid issues/clashes with the host repo, and to one day remove 
+# the flutter submodule in the host repo, test with clone. Only issue we need 
+# more information on is why the local version would ever be detached or pinned
```
khanhnwin, line 24: This shouldn't happen anymore since we introduced Dependabot to keep all of the submodules up to date. So for most of the time, the flutter submodule be up to date with the latest on flutter/flutter, with the exception of maybe a ~24 hour delay if we haven't gottena round to reviewing the Dependabot PRs.

Dockerfile
```diff
@@ -44,7 +38,7 @@ ENV PATH="/app/flutter/bin:$PATH"
 
 # Used if wanting to build the container with a different branch
 # e.g. `make FLUTTER_BRANCH=dev build`
-# This is not to be confused with the $TEST_TARGET_CHANNEL
+# This is not to be confused with the $FLUTTER_TEST_BRANCH
 RUN if test -n "$FLUTTER_BRANCH" -a "$FLUTTER_BRANCH" != "stable" ; then \
```
khanhnwin, line 42: This has been around for a while but I'll just call it out - it might be better to make it `make FLUTTER_BRANCH=beta build` since dev doesn't exist anymore.

Dockerfile
```diff
@@ -44,7 +38,7 @@ ENV PATH="/app/flutter/bin:$PATH"
 
 # Used if wanting to build the container with a different branch
 # e.g. `make FLUTTER_BRANCH=dev build`
```
khanhnwin, line 40: This has been around for a while but I'll just call it out - it might be better to make it make FLUTTER_BRANCH=beta build since dev doesn't exist anymore.

Dockerfile
```diff
@@ -55,59 +49,97 @@ RUN if test -n "$FLUTTER_BRANCH" -a "$FLUTTER_BRANCH" != "stable" ; then \
...
-# -- Build target
-FROM test AS builder
+# ============== BUILD PROD JEKYLL SITE ==============
+FROM node AS build
 
 ENV JEKYLL_ENV=production
+RUN gem install bundler
+COPY Gemfile Gemfile.lock ./
+RUN bundle config set force_ruby_platform true
+RUN BUNDLE_WITHOUT="test development" bundle install --jobs=4 --retry=2 --quiet
```
khanhnwin, line 126: could we potentially want logs here? Is bundle install too noisy which is why --quiet is being used here?

tool/check-code.sh
```diff
@@ -87,86 +84,62 @@ function check_formatting() {
...
 
+# Extract snippets, analyze and check formatting
+# Null safety flag set
+EXAMPLE_ROOT="examples"
+echo "=> ANALYZING and testing apps in $EXAMPLE_ROOT/* --null-safety"
+
+for sample in $EXAMPLE_ROOT/*/*{,/*}; do
+
+  # Does it have a pubspec?
+  if [[ -d "$sample" && -e "$sample/pubspec.yaml" ]]; then
```
khanhnwin, line 95: Looks like GitHub Actions has 2-core CPUs. Could we try to run 2 of these sample tests in parallel at the same time? Something along the lines of.

test Sample1 &
test sample2 &
wait

Dockerfile
btobin: This was formerly to ensure that if the repo was pulled fresh and the submodules were NOT initialized, they would be correctly set up (should have been included in the build stage this round, so incomplete anyway!). This line though is likely why the current `build.yml` is not broken right now for builds with the `submodules: recursive` missing in the `deploy` job. I agree though, removing (next commit) in favor of it be performed by build file and manually via host setup for local dev.

Dockerfile
btobin: Yes, for the production build. Makes sense, I'll remove that with this commit.

tool/check-code.sh
btobin: Yes, absolutely - this would be the place to run these these on multiple threads. I suggest we get this new baseline in place, evaluate, then come back in and test these additional optimizations.

khanhnwin: Got it! You mentioned did that in the email, we can loop back on this one!

Makefile
```diff
@@ -74,121 +82,62 @@ emulate:
...
-		--build-arg RUBY_VERSION=${RUBY_VERSION} \
-		--build-arg NODE_VERSION=${NODE_VERSION} -t ${BUILD_TAG}:${BUILD_COMMIT} .
-	docker run --rm -it \
-		-v ${PWD}/tool:/app/tool ${BUILD_TAG}:${BUILD_COMMIT} bash
-	docker rmi -f ${BUILD_TAG}:${BUILD_COMMIT}
-
-# Build the production image
+# Usage `make debug-tests`
+debug-tests:
+	DOCKER_BUILDKIT=1 docker build --rm --target flutter -t flt-test .
```
khanhnwin, line 104: DOCKER_BUILDKIT=1 vs docker build is a performance improvement change right?

Makefile
```diff
@@ -74,121 +82,62 @@ emulate:
...
-# WARNING this can take long while to run!
-# # Usage: `make test TEST_TARGET_CHANNEL=<channel>`
+# WARNING this can take a while to run!
+# Usage: `make test FLUTTER_TEST_BRANCH=<channel>`
 test:
-	docker build --rm --target test \
-		--build-arg TEST_TARGET_CHANNEL=${TEST_TARGET_CHANNEL} \
-		--build-arg RUBY_VERSION=${RUBY_VERSION} \
-		--build-arg NODE_VERSION=${NODE_VERSION} -t ${BUILD_TAG}:${BUILD_COMMIT} .
+	DOCKER_BUILDKIT=1 time docker build --rm --target tests -t flt-test .
```
khanhnwin, line 92: time is so helpful! I'm thinking we can leave it even after improving the optimizations are implemented. It'll give us a good baseline in case the build time starts increasing again.

Makefile
```diff
@@ -205,26 +154,20 @@ else
 		--non-interactive \
 		--token ${FIREBASE_TOKEN} \
 		--project ${FIREBASE_ALIAS} \
-		--debug \
-		--json
```
khanhnwin, line 209: Why was this specifying `--json` before?

Makefile
```diff
@@ -205,26 +154,20 @@ else
...
+		--debug
 endif
 
-
 # All in one command to stage your build to a Firebase 
 # channel on your currently selected project
 # Usage: `make stage STAGE_NAME=foo`
-STAGE_NAME ?= docs
 stage:
```
khanhnwin, line 163: this doesn't need to be addressed now and can be a todo for later on, but it'd be nice to integrate automatic staging on PRs within `build.yml`!

tool/check-code.sh
```diff
@@ -87,86 +84,62 @@ function check_formatting() {
   fi
 }
 
+# Extract snippets, analyze and check formatting
+# Null safety flag set
+EXAMPLE_ROOT="examples"
+echo "=> ANALYZING and testing apps in $EXAMPLE_ROOT/* --null-safety"
```
khanhnwin, line 90: `--null-safety` in this line should be removed since it's removed from the test?

khanhnwin:
LGTM. This is looking so much better already. Thank you, @btobin!! @johnpryan Can you please give this PR a quick pass? Once John gets a chance to review, I think we're good to land!

Dockerfile
btobin: It's in the README, so I feel like that is probably enough?

Makefile
btobin: I had added it long ago for CI purposes, but the regular output is better I feel.

Makefile
btobin: Yes, mirror the Dart site? Great idea.

tool/check-code.sh
btobin: Another followup optimization might be not to use the local Docker container at all for tests, since it now _only_ uses Flutter. Using a 3rd party action might just be faster and worth testing for the CI side of things, and the container is great for local testing and completeness:

```
steps:
      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
         with:
            submodules: recursive
      - uses: subosito/flutter-action@v2
         with:
            channel: ${{ matrix.channel }}
      - name: Test
         run: tool/test.sh --target ${{ matrix.channel }}
         continue-on-error: ${{ matrix.experimental }}
```

khanhnwin: the third party action would be good to look at after we parallelize the tests. Looks like the tests are responsible for majority of the build time, so breaking it out might only save us a few minutes?

Makefile
khanhnwin: gotcha

Makefile
khanhnwin: understood!

Makefile
khanhnwin: Yes!

Dockerfile
```diff
@@ -55,59 +50,94 @@ RUN if test -n "$FLUTTER_BRANCH" -a "$FLUTTER_BRANCH" != "stable" ; then \
...
+RUN npm install -g npm firebase-tools@11.0.1
+
+
+
+# ============== FLUTTER CODE TESTS ==============
+FROM flutter AS tests
+
+COPY ./ ./
+
+ARG FLUTTER_TEST_BRANCH=stable
```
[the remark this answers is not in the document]
khanhnwin: That env variable should specify whether stable / beta to run within test.sh. 

Though, I'm wondering - does specifying `run: make test FLUTTER_TEST_BRANCH=${{ matrix.branch }}` in build.yml override line 82 which says`ARG FLUTTER_TEST_BRANCH=stable`?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,8 @@
+{
+	"name": "Flutter.dev",
```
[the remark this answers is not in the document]
btobin: I don't know much about that, .devcontainers are just a normal option these days for Dockerized environments - some VSCode devs appreciate them because it's easy to use and keeps them in the Editor. It's not required.

.github/workflows/link-validate.yml
```diff
@@ -20,8 +20,6 @@ jobs:
           submodules: "recursive"
       - name: Build
         run: make build
-        env:
-          DISABLE_TESTS: 1
```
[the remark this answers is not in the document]
btobin: This is removed in favor of optimized Docker build stages rather than using an ENV var to skip tests.

Dockerfile
```diff
@@ -1,50 +1,45 @@
...
       diffutils \
+      git \
+      lsof \
+      vim-nox \
       xdg-user-dirs \
-      nodejs && \
-      rm -rf /var/lib/apt/lists/*
+    && rm -rf /var/lib/apt/lists/*
 
+RUN echo "alias lla='ls -lAhG --color=auto'" >> ~/.bashrc
```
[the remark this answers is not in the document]
btobin: This is a convenience alias for anyone who is used to using `lla` to list files in color, sizes, etc - not required.

Dockerfile
```diff
@@ -1,50 +1,45 @@
...
-RUN bundle install
 
-# Install Node deps
-ENV NODE_ENV=development
-COPY package.json package-lock.json ./
-RUN npm ci
 
-COPY ./ ./
+# ============== INSTALL FLUTTER ==============
+# TODO recommended to move to a SHA based install for security reasons
```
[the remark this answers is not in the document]
btobin: That's different I believe, and I left that usage in play even though I'm not aware of why you are pinning a particular base image. I'm talking about [what's being done in Dart](https://github.com/dart-lang/site-www/blob/main/Dockerfile#L33).

Dockerfile
```diff
@@ -1,50 +1,45 @@
...
+FROM base AS flutter
+
+# TODO to avoid issues/clashes with the host repo, and to one day remove 
+# the flutter submodule in the host repo, test with clone. Only issue we need 
+# more information on is why the local version would ever be detached or pinned 
+# at a specific commit/hash???
```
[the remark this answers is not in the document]
btobin: There is no issue if the container is used as documented. Happy to remove if you think this commentary is confusing.

Dockerfile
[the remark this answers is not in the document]
btobin: Yes @khanhnwin, that Make arg is passed to the Docker build args, which then sets the container env var `FLUTTER_TEST_BRANCH` - which is the target "channel" to be used for running tests.

Dockerfile
```diff
@@ -1,50 +1,45 @@
...
 ENV FLUTTER_BIN=flutter/bin
 ENV PATH="/app/flutter/bin:$PATH"
 
-# Used if wanting to build the container with a different branch
-# e.g. `make FLUTTER_BRANCH=dev build`
-# This is not to be confused with the $TEST_TARGET_CHANNEL
+# Used if wanting to build the container with a different branch, this 
+# would change the current branch of and update the mirrored submodule
+# e.g. `make build FLUTTER_BRANCH=beta`
+# This is not to be confused with the $FLUTTER_TEST_BRANCH
```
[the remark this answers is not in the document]
btobin: Maybe it would be good to rename `FLUTTER_BRANCH` to `FLUTTER_BUILD_BRANCH`?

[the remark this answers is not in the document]
johnpryan: SGTM

Dockerfile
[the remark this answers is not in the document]
johnpryan: Is there an issue for this? That might be easier to track than a comment.

Dockerfile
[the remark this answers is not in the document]
johnpryan: Could you file an issue for this and link it here?

Dockerfile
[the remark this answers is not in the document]
johnpryan: Consider leaving this out...

johnpryan (approved):
LGTM overall - I would consider filing an issue for any TODOs and adding a link as part of the comment

khanhnwin:
Landed! Thanks y'all! @btobin only remaining AI on your end seems to be filing the Issues? I'll take a crack at parallelizing the tests!

Dockerfile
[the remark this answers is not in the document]
btobin: @johnpryan Removing this commentary in this [followup PR](https://github.com/flutter/website/pull/7737) rather than adding concern for a non-issue.

Dockerfile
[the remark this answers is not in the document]
btobin: Updated [here](https://github.com/flutter/website/pull/7737)

Dockerfile
btobin: Added comment in this [PR](https://github.com/flutter/website/pull/7737)

Dockerfile
```diff
@@ -55,59 +50,94 @@ RUN if test -n "$FLUTTER_BRANCH" -a "$FLUTTER_BRANCH" != "stable" ; then \
...
+# TODO recommended to move to a SHA based install for security reasons
+FROM flutter AS node
+
+RUN curl -sL https://deb.nodesource.com/setup_17.x -o node_setup.sh && \
+      bash node_setup.sh 1> /dev/null
+RUN apt-get update -q && apt-get install -yq --no-install-recommends \
+      nodejs \
+    && rm -rf /var/lib/apt/lists/*
+# Ensure latest NPM, install global Firebase CLI
+RUN npm install -g npm firebase-tools@11.0.1
```
ditman, line 73: It appears this change has broken the `check_link_references` in my pull request? `Node 17` seems to not be supported by `firebase-tools@11.0.1` (if I'm reading the error right)?

@btobin, can you take a look at: https://github.com/flutter/website/pull/7740 ? Thanks!

PR #7737, opened 2022-11-08
Adds optimizations to Docker Build

btobin:
Followup resolutions from merged infra [PR](https://github.com/flutter/website/pull/7580) comments, plus additional optimizations:

- Removes `firebase-tools` and `superstatic` package.json installs. Neither are required any longer, and installing them slows down docker builds. Updates outdated comment in link check config. Updates package-lock.json after local setup.
- Removes comments about host user because they were confusing, and it’s not really important for this usage. 
- Updates naming of `FLUTTER_BUILD_BRANCH` env var for clarity
- Removes comment referencing —null-safety flag
- Removes unnecessary docker compose Firebase env vars
- Removes comments regarding TODO future usage of SHA installs for Flutter/Node as this would increase maintenance cost of the Dockerfile. If we still feel it’s worth documenting, we can file a separate issue for clarity on that potential future upgrade, similar to [what’s being done in the Dart website](https://github.com/dart-lang/site-www/blob/main/Dockerfile#L33)

@khanhnwin @johnpryan I'll add a reference to the appropriate comment(s) in the above PR. 

## Presubmit checklist


sfshaza2:
@khanhnwin, can you review?

ditman:
@btobin it seems that the node+firebase-tools combo installed by the Dockerfile has started blowing up? I've just noticed a couple of PRs broken because of it:

https://github.com/flutter/website/issues/7741

khanhnwin (approved):

khanhnwin:
Thanks Brian! LGTM.

@domesticmouse @sfshaza2 let's wait for #7742 to land before we merge this one?

domesticmouse:
Sgtm!

khanhnwin:
@btobin John was able to put in the fix with #7742!

btobin:
@khanhnwin Okay, awesome - that works as well. Great to check it regardless, and more stable to lock it. I think in this case that the problem really was that we are upgrading npm to the latest version, and it clashed because of the recent LTS bump to 18. 

I'll remove that and also update the build.yml file so it's consistent with those new versions, then submit another PR. 

Thanks!

PR #7739, opened 2022-11-09
Changed Docker settings to allow Dart and Flutter websites to both run locally

atsansone:
@khanhnwin : I got both sites to build local and staged. You can checkout this PR to try the local builds, but the staged builds are:

- [Dart](https://dart--staging.web.app/)
- [Flutter](https://flutter-docs-prod.web.app/)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,12 +97,12 @@ RUN npm install
 COPY ./ ./
 
 # Jekyl ports
-EXPOSE 35729
+EXPOSE 35730
 EXPOSE 4002
 
 # Firebase emulator port
 # Airplay runs on :5000 by default now
-EXPOSE 5500
+EXPOSE 5502
 
 
 
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,7 @@ serve:
 		--port ${JEKYLL_SITE_PORT} \
 		--config _config.yml,_config_dev.yml \
 		--livereload \
+		--livereload_port 37530 \
 		--incremental \
 		--trace
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,8 +13,8 @@ services:
       - JEKYLL_ENV=development
     ports:
       - "4002:4002"
-      - "35729:35729"
-      - "5500:5500"
+      - "35730:35730"
+      - "5502:5502"
     volumes:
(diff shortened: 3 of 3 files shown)


sfshaza2:
@khanhnwin, can you review?

Makefile
```diff
@@ -18,6 +18,8 @@ FLUTTER_BRANCH ?= stable
 FLUTTER_TEST_BRANCH ?= stable
 JEKYLL_SITE_HOST ?= 0.0.0.0
 JEKYLL_SITE_PORT ?= 4002
+JEKYLL_LIVERELOAD_PORT ?= 37530
+FIREBASE_EMU_PORT ?= 5502
```
khanhnwin, line 22: are we using this variable anywhere?

Dockerfile
```diff
@@ -105,12 +105,13 @@ RUN npm install
...
 
 # Jekyl ports
-EXPOSE 35729
+EXPOSE 35730
 EXPOSE 4002
 
 # Firebase emulator port
 # Airplay runs on :5000 by default now
-EXPOSE 5500
+EXPOSE 5502
```
khanhnwin, line 113: We'll want to update this port value in the [firebase.json](https://github.com/flutter/website/blob/main/firebase.json#L447)

khanhnwin:
Thanks @atsansone! Were the submodule updates an intentional change with this PR? Otherwise, just left a minor comments.

khanhnwin:
Since the submodule updated, this build is now failing because the code excerpts haven't been refreshed.

Dockerfile
atsansone: Oddly enough, I don't. When I do that, the staging build fails.

khanhnwin: I just look a deeper dive and I think this port 5502 doesn't actually need to change to run Jekyll concurrently. 5500 is used for the Firebase emulator which is only used by the link checker (testing) Do you visit port 5502 for any reason? Otherwise, we might just be able to leave this port alone.

khanhnwin (approved):

khanhnwin:
We're good to land once the build is green! I'll look more into why port 5500 is being used for a jekyll build.... Thanks @atsansone!

PR #7829, opened 2022-11-16
Tweaking the banner

sfshaza2:
Temporarily removing the job listings link.


khanhnwin (approved):

atsansone:
> Makes sense - Should we remove the link on the Dart side for now too?
> 
> This also makes me question if the jobs currently [listed in /jobs](https://docs.flutter.dev/jobs) are still available since they haven't been updated in a while? \cc @tvolkert

I would say go ahead, @parlough . I'll review and approve on Dart.

Issue #7733, opened 2022-11-08
[PAGE ISSUE]: 'Animations tutorial' error word


nickppbd:
### Page URL

https://docs.flutter.dev/development/ui/animations/tutorial/

### Page source

https://github.com/flutter/website/tree/main/src/development/ui/animations/tutorial.md

### Describe the problem

The current value of the Animation object can be found in the .value **method**

### Expected fix

The current value of the Animation object can be found in the .value

### Additional context

I do not know well what it is supposed to be.

danagbemava-nc:
Hi @nickppbd, can you point out the exact places in the documentation that is not clear to you?

Thank you

nickppbd:
@danagbemava-nc 

https://docs.flutter.dev/development/ui/animations/tutorial#tween

A Tween object does not store any state. Instead, it provides the evaluate(Animation<double> animation) method that applies the mapping function to the current value of the animation. The current value of the Animation object can be found in the .value method

danagbemava-nc:
Thanks for the clarification

sfshaza2:
Perhaps it would be clearer if the text said "`Animation.value` method" and linked to its [API page](https://api.flutter.dev/flutter/animation/Animation/value.html).

nickppbd:
@sfshaza2 I did not realize that .value is a method. I am sorry!

sfshaza2:
no problem!

PR #7842, opened 2022-11-18
Clarify .value method

MiraMarshall:
Clarifies Tween object wording by specifying that `.value` is a method and by adding a link to API docs.
_Issues fixed by this PR (if any):_ Fixes #ISSUE-NUMBER
#7733 
## Presubmit checklist


src/development/ui/animations/tutorial.md
```diff
@@ -204,8 +204,8 @@ colorTween = ColorTween(begin: Colors.transparent, end: Colors.black54);
 ```
 
 A `Tween` object does not store any state. Instead, it provides the
-`evaluate(Animation<double> animation)` method that
-applies the mapping function to the current value of the animation.
+[`Animation.value`][] method that applies the mapping function 
+to the current value of the animation.
```
atsansone, line 208: **question:** What do we mean by mapping function? Does the `Tween` object's value map to an `Animation.value` at a given point in time? If so, that would be the more clear way to say this.

sfshaza2: cc @HansMuller, I think that Tony's statement is correct, but can you confirm?

sfshaza2:
@HansMuller, how about the following:

```
A `Tween` object doesn't store any state. Instead, it provides the
`evaluate(Animation<double> animation)` method that uses
the `transform` function to map the current value of the animation
(between 0.0 and 1.0), to the actual animation value.
 ```

Please review, @HansMuller. If it's ok, @MiraMarshall can land it.

HansMuller (approved):
LGTM

sfshaza2:
@miramarshall, Hans has approved my suggested change. Please update and land the PR. thx!

sfshaza2 (approved):
lgtm

PR #7855, opened 2022-11-21
Main branch cleanup

mit-mit:
Just a bit of cleanup wrt. master vs main branch (we should avoid the former term when possible).


mit-mit:
This is very incomplete. Would be great if someone else could scrub the rest.

mit-mit:
I think it would go a long way of we replaced links like`{{site.repo.flutter}}/blob/master/` with something like `{{site.repo.flutter_blob}}`?

sfshaza2 (approved):
Very good catches, Mit! Sure, that macro would make the text more readable. Meanwhile, this is a great start.

PR #7885, opened 2022-11-29
Updating the memory view page

sfshaza2:
Staged:  https://sz-flutter-2.web.app/development/tools/devtools/memory


polina-c:
Minor structural change.

This is the suggested text right after the title:

> The memory screen provides insights into details of the application’s memory allocation and gives tools to detect and narrow down specific issues.
> 
> To better understand the insights found on this page, the first section explains how Dart manages memory. If you already understand Dart’s memory management, you can skip to the Memory page guide. 

And the current text should go under the header 'Memory view guide':

Screenshot 2022-11-30 at 2 19 44 PM

polina-c:
Let's color the code examples as dart code. Example: https://sz-flutter-2.web.app/development/tools/devtools/inspector#slow-animations

You need to add word 'dart' after ``` to make it happen:

Screenshot 2022-11-30 at 2 25 00 PM

polina-c:
Do we really want 'provides' two times in this sentense:

The memory view provides insights into details of the application’s memory allocation and provides tools to detect and debug specific issues.

Everything else looks great.

khanhnwin (approved):

khanhnwin:
Landed! Thanks y'all!

PR #7957, opened 2022-12-19
Optimize table of contents generation

parlough:
_Description of what this PR is changing or adding, and why:_ I'm investigating areas where we can reduce local build time and since the TOC is generated twice on almost every page, this was an important area to make improvements.

The primary improvement is moving out the `{% assign toc_content = content | toc_only -%}` so jekyll_toc only only parses the content and builds the initial TOC once. 

Then the two TOCs (side vs inline) are quite different, so separates them into two different includes to avoid multiple checks.

Finally, uncomments the logic for the "Page top" button which is hidden with `display: none` in CSS anyway. This also slightly reduces built HTML size.

I've tested locally, but I **recommend verifying** the both TOCs continue to work as you expect. Side when wide screen and top of page when thinner(?) screen.

Staged (updated as of 12/19/22): https://parlough-docs-flutter-dev.web.app/

## Presubmit checklist


src/assets/js/main.js
```diff
@@ -20,15 +20,15 @@ $(function () {
...
-// TODO(chalin): Copied (& tweaked) from site-www, consider moving into site-shared
 function adjustToc() {
   // Adjustments to the jekyll-toc TOC.
 
   var tocId = '#site-toc--side';
-  var tocWrapper = $(tocId);
-  $(tocWrapper).find('.site-toc--button__page-top').click(function () {
-    $('html, body').animate({ scrollTop: 0 }, 'fast');
-  })
+  // Uncomment to enable 'page top' button
```
domesticmouse, line 27: Do you want to delete this, or is intended to be brought back in at some point?

khanhnwin: @sfshaza2 Have we ever used a back to page top button on the site? It's interesting to me that the button is hidden.

khanhnwin:
Poked around the stage site and this looks great! Thanks @parlough! I'm just going to wait for @sfshaza2 on the back to page top button.

khanhnwin (approved):

PR #7964, opened 2022-12-19
Move sidenav code and styles to repo

parlough:
_Description of what this PR is changing or adding, and why:_ I'm going to work on some fixes and optimizations to the sidenav generation (to reduce fresh build time), then later some improvements. I'd rather do them in isolation on the Flutter site for easier review and I expect the Dart and Flutter sites to diverge in sidenav behavior anyway. So this PR moves all related files from `dart-lang/site-shared` to `flutter/website`. We can always move it back (to site-shared) in the future if needed.

### DIFF
--- a/src/_includes/header.html
+++ b/src/_includes/header.html
@@ -36,7 +36,7 @@
       <div class="site-header__sheet">
         <ul class="navbar-nav">
           <div class="site-sidebar site-sidebar--header d-md-none">
-            {% include shared/sidenav-level-1.html nav=site.data.sidenav %}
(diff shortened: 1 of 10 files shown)


domesticmouse (approved):

atsansone:
**question:** @parlough would this be reversed once changes have been made or is this a permanent split?

parlough:
@atsansone It depends on what functional and style changes are made down the the road, but it is an option if we end up wanting similar behavior on the Dart site.

PR #7970, opened 2022-12-20
Remove outdated pages

jmagman:
_Description of what this PR is changing or adding, and why:_ Remove outdated pages. These listings are no longer open.

## Presubmit checklist


src/_includes/banner.html
```diff
@@ -8,9 +8,5 @@
     See how Flutter is pushing UI development at Flutter Forward;<br>
     streaming live from Kenya on January 25, 2023.
     <a href="https://flutter.dev/events/flutter-forward">Register now</a>
-    {% comment %}
-    <br>The Flutter and Dart teams are hiring.
-    <a href="https://docs.flutter.dev/jobs">Learn more</a><br>
-    {% endcomment %}
```
sfshaza2, line 14: This is already commented out, so it doesn't even show up in the HTML as a comment. I feel no need to delete it.

sfshaza2 (changes requested):
These files are not being displayed currently. Why do we need to remove them from the repo? Is it a case of people following these links and finding the pages? If so, I prefer a different solution. I would, for now, add a redirect that goes to a "we are not currently hiring." page.

jmagman:
> These files are not being displayed currently. 

I navigated to them from https://docs.flutter.dev/jobs. I guess why would we not remove them from the repository since it's source controlled and we can bring it back whenever we want?

However you want to solve this is fine, but those postings should be removed.

sfshaza2:
Closing. Replaced by PR https://github.com/flutter/website/pull/7973

jmagman:
Thanks @sfshaza2!

PR #7980, opened 2022-12-22
Remove filled job pages

parlough:
_Description of what this PR is changing or adding, and why:_ As otherwise these show up in the sitemap and search engines or other tools may surface them, causing confusion for users when they click the link and then the site redirects them to `/jobs`. This PR keeps the [`_template.md`](https://github.com/flutter/website/blob/main/src/jobs/_template.md) file for future use and the other entries can be accessed in Git history if needed.


khanhnwin:
Hey @parlough thanks for this! I was chatting with @sfshaza2 yesterday and _I think_ she wants to keep these pages around. Correct me if I'm wrong, Shams?

atsansone:
@parlough : Removing the pages may not be our best option. Can we change them to something generic, like "No open positions exist at this time", while moving the content to a hidden directory?

parlough:
@atsansone Shams already redirected all entries under /jobs to the /jobs index page which was updated to say that in https://github.com/flutter/website/commit/5c1be34c1aa690df928b756eb5d76c17cde1de55.

The goal here is to avoid confusion as these pages still show up in the sitemap but are not accessible. We could make these pages hidden and not render but the jobs are already filled and the [template](https://github.com/flutter/website/blob/main/src/jobs/_template.md) is still in the directory for future reference.

sfshaza2 (approved):
Never have I seen so much concern/confusion over so little. I am ok leaving them here, when you can't actually call them up, unless you are intent on downloading the website source. But, ok.

PR #3893, opened 2020-04-03
Turning banner back on due to tabs not working.

sfshaza2:
Fixes https://github.com/flutter/website/issues/3889

https://sz-flutter.firebaseapp.com/docs/get-started/editor?tab=vscode#vscode

Issue #3889, opened 2020-04-03
MacOS and Linux tabs are disabled on Flutter SDK releases page


TahaTesser:
When opening the links, it only lands to Windows tab and clicking on MacOS and Linux tab doesn't do anything

https://flutter.dev/docs/development/tools/sdk/releases#macos
https://flutter.dev/docs/development/tools/sdk/releases#linux

eseidelGoogle:
@Sfshaza @RedBrogdon

pcsosinski:
I used the site yesterday afternoon (PDT) and it worked fine FYI

legalcodes:
We're looking into this issue -- many thanks for raising @TahaTesser.

PR #7985, opened 2022-12-24
Fix JS error in topOffset calculation

### LINKED ISSUE #3897: Presence of banner should not impact tab behavior
We should investigate why the presence of the banner impacts how nav-tabs function, see https://github.com/flutter/website/issues/3889 and https://github.com/flutter/website/pull/3893.

filiph:
Top offset was being calculated assuming that the site banner exists on the page. This lead to a fatal error in `main.js` when banner was _not_ present, which in turn prevented other JavaScript to execute. This includes the tab behavior (see #3897, #3889, #3893).

Fixes #3897.

## Presubmit checklist


filiph:
It might be good to check if this change also fixes other JavaScript-related issues on the website. The bug prevented most of the page's JS to execute whenever the banner was missing (but only when the fixed column was shown, e.g. on desktop).

domesticmouse (approved):
PTAL @khanhnwin

sfshaza2:
OMG, Filip! THANKS!!! This bug has been there forever. I'm only aware of one other JS bug and I'll test it out. (I doubt it's related.)

PR #7995, opened 2022-12-27
Speedup and cleanup widget catalog page generation

parlough:

The categorical widget index pages (such as [Material Components widgets](https://docs.flutter.dev/development/ui/widgets/material)) were being generated with a very large amount of whitespace and generated HTML divs and unordered lists even when there were no children present.

This PR improves the situation by:
- Using Liquid templating whitespace modifiers to limit whitespace generated
- Finds the specified category only once and store it in variable to avoid looping twice
- Pulls out the categories and subcategories to variables to avoid rendering empty divs and unordered lists

Notice the size of the generated files being cut in half:

**Before:**

| Size | Time |
|----|----|
| **327.49K** |  0.027 |

**After:**

| Size | Time |
|----|----|
| **162.31K** |  0.021 |

### DIFF
--- a/src/_includes/docs/catalogpage.html
+++ b/src/_includes/docs/catalogpage.html
@@ -1,25 +1,29 @@
-<div class="catalog">
-    {% for section in site.data.catalog.index %}
-        {% if section.name contains include.category %}
-            <div class="category-description"><p>{{section.description}}</p></div>
-        {% endif %}
-    {% endfor %}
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):

khanhnwin (approved):

khanhnwin:
Thanks again @parlough!

PR #7998, opened 2022-12-27
Optimize PNG images on the site

parlough:
This PR optimizes the PNGs on the site through lossless compression completed by [OxiPNG](https://github.com/shssoichiro/oxipng). The goal here is to reduce clone time as well as build time, since Jekyll needs to copy these images on some file systems. As another benefit, these images will load slightly faster for users visiting pages where they are used.


domesticmouse (approved):

khanhnwin (approved):
Thanks @parlough

sfshaza2:
NICE, Sir @parlough!

PR #8044, opened 2023-01-04
Avoid extra border on embedded DartPads

parlough:
_Description of what this PR is changing or adding, and why:_ Embedded DartPads had an extra border, most clearly seen on the bottom side, causing an awkward look. These come from borders used for all code-excerpt `div`s. This fixes the issue by moving the code excerpt border to the `pre` elements used by snippets rather than the `code-excerpt__code` `div` itself.

**Before (with extra border):**
Before (with extra border)

**After (without extra border):**
After (without extra border)

## Presubmit checklist

### DIFF
--- a/src/_sass/components/_code.scss
+++ b/src/_sass/components/_code.scss
@@ -49,11 +49,12 @@ pre > span.strike { text-decoration: line-through; }
 
 .code-excerpt {
   &__code {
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
Thanks for the graphic! LGTM!

PR #8049, opened 2023-01-05
Add big juicy link to new codelab

### LINKED ISSUE #6938: [PAGE ISSUE]: 'Write your first Flutter app, part 1'
### Page URL

https://docs.flutter.dev/get-started/codelab/

### Page source

https://github.com/flutter/website/tree/main/src/get-started/codelab.md

### Describe the problem

[Step 4: Create an infinite scrolling ListView](https://docs.flutter.dev/get-started/codelab#step-4-create-an-infinite-scrolling-listview) has a number of confusing, let's say, continuity issues imo, making it difficult to peruse other than just skipping ahead and copying the working code from the last steps:
1. In step 1, the text says "It increments twice for every suggested word pairing: once for the ListTile, and once for the Divider", but the reader does not yet know about the ListTitle and Divider because they are not introduced until step 2 (note `/*1*/`, "For even rows, the function adds a ListTile row for the word pairing. For odd rows, the function adds a Divider widget to visually separate the entries.")
2. Step 2 says "Add a ListView.builder widget to the build method of the _RandomWordsState class", but then gives a snippet of code that will eventually (in step 4) turn out to be part of the statement that starts with `return Scaffold(`. So, this snippet can not be added to the method in any meaningful way at this time.
3. Step 3 says to "Add a ListTile in the itemBuilder body of the ListView.builder in _RandomWordsState". The reader has the same problem here of not knowing where to add this, but in addition to that, the snippet that is supposed to be added is already included in the larger snippet in step 2. 
4. In step 4 the `Scaffold` is moved into the widget state, and is introduced for the second time but without giving a reason why it is being moved here. Maybe the point of this Step (showing a stateful list instead of a single centered name) could be presented more clearly by having the widget only render the `ListView`, that is, by returning the `ListView` from the `build()` function and adding the `RandomWords()` call to the `body` field instead of the `home` field.

### Expected fix

1. Add a line introducing the ListTile and the Divider, such as "We will present each word pairing inside a ListTile, and alternate them with Divider widgets in order to visually separate the entries", after the line "This method builds the ListView that displays the suggested word pairing.". Also maybe consider having step 2 start after this line, that is, including the next paragraph in step 2.
2. Either return the `ListView` widget from the `build()` function here, or add the `Scaffold`... scaffolding to the code in step 1.
3. Step 3 can be removed altogether, as it seems to be left over from the last refactoring, before which it was a separate function. As it is, it has already been added in step 2. Also the phrase "in the next step" can be removed from step 2. Alternatively, only return the `Text` in step 2, and add the `ListTile` here (to "make the rows more attractive"); this may just be confusing though.
4. Explain why the `Scaffold` is moved here, or move it to step 1, or just return the `ListView`. In the latter case, the code in step 5 should be updated accordingly. Either way, step 4 seems to be obsolete as well.

### Additional context

I'll try to come up with a PR, maybe after I learn a little more about Flutter first.

### LINKED ISSUE #7533: [PAGE ISSUE]: 'Write your first Flutter app, part 1'
### Page URL

https://docs.flutter.dev/get-started/codelab/

### Page source

https://github.com/flutter/website/tree/main/src/get-started/codelab.md

### Describe the problem

Source code and instructions for Step 2 (using external package) are out of date, and inconsistent with the Hello World code that the user is instructed to copy from Step 1.

- main dart before wordPair

- Inserting the Step 2 code after the corresponding lines from the Step 1 code results in an error in the child field.
   main dart after wordPair

### Expected fix

_No response_

### Additional context

_No response_

### LINKED ISSUE #7932: [PAGE ISSUE]: 'Write your first Flutter app, part 1'
### Page URL

https://docs.flutter.dev/get-started/codelab/

### Page source

https://github.com/flutter/website/tree/main/src/get-started/codelab.md

### Describe the problem

The code diff in Step 4.5 looks misaligned. For example, these two lines were deleted from line 30-21 and then added at line 54-55:
```dart
class RandomWords extends StatefulWidget
  const RandomWords({super.key});
```

These two lines didn't actually change between step 3 and step 4 as far as I can tell. 

### Expected fix

_No response_

### Additional context

_No response_

filiph:
Replaces the old "first Flutter app" codelab with a prominent link to the new one. Also removes all referenced files of the old codelab (such as images) in this repo.

Screenshot 2023-01-05 at 09 49 41

Staged: https://flutter-website-filiph-staging.web.app/get-started/codelab

Introduces a new CSS class `juicy-button-container` for big links on top of pictures. This can be reused on other pages in the future by just creating a `<div>` with `class="juicy-button-container"` and populating it with an image (the background) and a button (`<a class="btn">`).

Closes #7932, closes #7533, fixes #6938. Later, this work also de-facto closes [at least 3 other issues](https://github.com/flutter/website/issues?q=is%3Aissue+is%3Aopen+%22first-flutter-app-pt2%22+OR+%22first-flutter-app-pt1%22+) as the old codelab gets removed (or unlisted) on Google Codelabs as well.

## Presubmit checklist


filiph:
I just added a fly-by fix of two broken links that `linkcheck` caught during site-build ([here](https://github.com/flutter/website/pull/8049/files#diff-e04f708d756459fc5000c549f622f956d912c13d45b24ac4ad7698f4602db426) and [here](https://github.com/flutter/website/pull/8049/files#diff-2ed7b88885e225f65ccf489640094c26101ef4d03566ffb4131583ef36db38c7)). Hope that's okay. Wanted to mention it here because otherwise it might look weirdly out of place. It _is_ completely unrelated. But it's such a small change, and it makes `linkcheck` happy. (Which, in turn, makes _me_ happy.)

sfshaza2 (approved):
This is great, @filiph! I love it. (Though I haven't actually seen a staged version, but it looks correct.) 

And, thanks for fixing those two broken anchors!

lgtm

cc @domesticmouse

sfshaza2:
@parlough, you might also take a looksie. I'm excited to land this. :D

src/get-started/codelab.md
```diff
@@ -1,842 +1,52 @@
...
+</div>
 
 {{site.alert.tip}}
-  If you see a warning on a hot reload that you might
-  need to restart the app, consider restarting it.
-  The warning might be a false positive, but
-  restarting your app ensures that
-  your changes are reflected in the app's UI.
+  The codelab above walks you through writing your first Flutter
+  app for all platforms &mdash; mobile, desktop and web.
```
parlough, line 37: To be consistent with the other oxford comma you added :)

```suggestion
  app for all platforms &mdash; mobile, desktop, and web.
```

parlough (approved):
This looks great to me. The button is super juicy!

parlough:
One question, how did you generate the image? 

Just wondering if there is a source diagram file or something, in-case we want to instead generate a webp image later.

sfshaza2:
Great question, @parlough! I did ask @filiph how he made his doodles and here is his reply:

```
As for the doodles, I have an iPad with an Apple Pen, and I use Procreate (an iPad app). It feels like using a cannon to kill a mosquito (please replace with an appropriate English saying), considering the "beauty" of my output. But I already have an iPad with a pen, and Procreate is $13.
```

filiph:
Thanks both!

The image is painted by me on an iPad, then post-processed to be brighter and a bit blurry (so it's less distracting behind the big button). I will work with Brett and Shams to make a Google Drive folder where I'll put all the intermediaries, so that you can recreate different versions of it if needed.

src/get-started/codelab.md
filiph: Gah! I didn't make it in time, sorry!

Issue #3435, opened 2019-12-17
Flutter API website sidebar


mingsai:
First I just want to say thanks for the quick response to the API website mentioned earlier (in #3362). I live in these docs, and the quick action on the requests are welcome. 

When selecting a document to view the sidebar always resets to the top. This is a bit of a hassle if one is going through the docs sequentially. Better behavior might be to have the sidebar scroll to the currently selected topic area. Love that little link back to code by the way. That rocks and helps one comprehend many of these new ideas.

sfshaza2:
I totally agree, @mingsai!!! Unfortunately, we don't have infrastructure support atm, so PRs welcome.  :D

shivam15:
@sfshaza2 Can i pick this?

sfshaza2:
Absolutely, @shivam15, I've assigned it to you!  :)

PR #8050, opened 2023-01-05
Scroll active sidenav entry into view on load

parlough:
Right now the sidenav always resets to scrolling to the top when loading the page, interrupting the user flow, especially when the active entry is below their view port. This is becoming an increasingly apparent issue as the sidenav and its categories gain more entries, particularly with future IA work.

This PR adds JS code to automatically scroll the fixed sidenav on load so that deepest active entry is in view.

Fixes https://github.com/flutter/website/issues/3435

Try it on this [staged site](https://parlough-docs-flutter-dev.web.app/), with links far down in the sidenav and/or reducing the height of your browser window.

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -1,6 +1,7 @@
 $(function () {
   adjustToc();
   initFixedColumns();
+  scrollSidebarIntoView();
   initVideoModal();
   initCarousel();
   initSnackbar();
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):

sfshaza2 (approved):
Oh, wow, @parlough, this is GREAT!

PR #8068, opened 2023-01-06
Fix sidebar collapse icon vertical alignment

parlough:
The inner collapse icon was not vertically centered with the included text.

**After:**
Fixed vertical alignment

**Before:**
Before fix of vertical alignment


sfshaza2 (approved):
lgtm

PR #8082, opened 2023-01-08
Fix multiple API doc links

parlough:
_Description of what this PR is changing or adding, and why:_ Fixes some broken API docs links. A few due to the targets being removed and others due to being rendered enums instead of classes.

## Presubmit checklist

### DIFF
--- a/src/release/breaking-changes/eliminating-nullok-parameters.md
+++ b/src/release/breaking-changes/eliminating-nullok-parameters.md
@@ -54,7 +54,7 @@ return a non-nullable value:
 * [`FocusTraversalOrder.of`][]
 * [`FocusTraversalGroup.of`][]
 * [`Focus.of`][]
-* [`Shortcuts.of`][]
+* `Shortcuts.of`
 * [`Actions.handler`][]
 * [`Actions.find`][]
 * [`Actions.invoke`][]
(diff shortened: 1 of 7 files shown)


src/release/breaking-changes/eliminating-nullok-parameters.md
```diff
@@ -54,7 +54,7 @@ return a non-nullable value:
 * [`FocusTraversalOrder.of`][]
 * [`FocusTraversalGroup.of`][]
 * [`Focus.of`][]
-* [`Shortcuts.of`][]
```
domesticmouse, line 57: Does `Shortcuts.of` not have a target page like everything else in this list? That seems... unexpected?

domesticmouse (approved):

src/release/breaking-changes/eliminating-nullok-parameters.md
parlough: The `Shortcuts` methods have been removed in a more recent version. Perhaps I should add a small note in parenthesis providing some extra context.

domesticmouse: SGTM. I'm assuming if it is removed in `master` that it was deprecated previously. Maybe Something along the lines of "Deprecated, to be removed in version XXX" or similar.

sfshaza2 (approved):
lgtm!

PR #8083, opened 2023-01-09
Make the example visually explicit

davidgaspardev:
Make the example visually explicit with the [native android reference](https://developer.android.com/training/multiscreen/screendensities#TaskProvideAltBmp).

For people coming from native Android, I believe that information is absorbed faster with this naming reference (`xxxhdpi`, `xxhdpi` ... etc)


sfshaza2 (approved):
lgtm

PR #8092, opened 2023-01-09
Replace outdated links to site-shared writing wiki

parlough:
This content is now found within the directory structure of site-shared itself.


sfshaza2 (approved):
lgtm!

Issue #1176, opened 2018-08-20
Fix page descriptions


chalin:
_From @chalin on July 24, 2018 19:30_

The meta properties `description` and `og:description` use `page.description` or the initial text from the page. In some cases, the latter isn't very pretty; e.g. from `_site/web-analogs/index.html`:

> \<meta name="description" content="Contents:* TOC Placeholder{:toc}This page is for users who are familiar with the HTML and CSS syntax forarranging components of an application's UI. It maps HTML/CSS code snippets totheir Flutter/Dart code equivalents.The examples assume:* The HTML document starts with a modern HTML DOCTYPE, and the CSS box model for all HTML...">

Maybe all pages should be required to have a description attribute in the Jekyll front matter, or the description could be dropped otherwise.

cc @Sfshaza @filiph 

_Copied from original issue: flutter/flutter#19723_

chalin:
_From @Sfshaza on July 25, 2018 13:16_

Yes, we should add that and be consistent. The lawless days of chaos will soon be over.  ;)

Sfshaza:
Just want to confirm I support adding a sensible description to the front jekyll matter. I can do this as part of the content migrations.

PR #8196, opened 2023-02-01
Throw an error if a page is missing a description

parlough:
Throw an error if a page is missing a description with a new Liquid `throw_error` filter. This will prevent us adding new pages without descriptions in the future. As a result, I added descriptions to all pages currently missing one to get the build to pass.

I considered doing this with a [Jekyll hook](https://jekyllrb.com/docs/plugins/hooks/) instead, but that was finicky due to it considering all content types. This gives us easy control over which layouts want to make checks like this and makes it easy to add similar checks in the future.

Fixes https://github.com/flutter/website/issues/1176


domesticmouse (approved):

src/development/ui/widgets/text.md
```diff
@@ -1,5 +1,8 @@
 ---
 title: Text widgets
 short-title: Text
+description: >
+  A catalog of Flutter's widgets for displaying and styling text.
```
sfshaza2, line 5: I've never seen using the arrow for breaking lines in yaml text. But, in this case, it doesn't seem necessary.

src/security/index.md
```diff
@@ -1,5 +1,7 @@
 ---
 title: Security
+description: >
+  An overview of the Flutter's team philosophy and processes for security.
```
sfshaza2, line 4: Same comment. Is the arrow needed here?

src/perf/power.md
```diff
@@ -1,5 +1,7 @@
 ---
 title: Preserving your battery
+description: Learn more about using limiting batter usage in your Flutter app.
+sitemap: false
```
sfshaza2, line 4: "batter usage" ? LOL

But, maybe we should just delete these TBD pages?

src/development/platform-integration/ios/index.md
```diff
@@ -1,4 +1,6 @@
 ---
 title: iOS
 layout: toc
+description: >
+  Content covering integration with iOS in Flutter apps.
```
sfshaza2, line 5: Same comment as elsewhere.

src/perf/power.md
parlough: I've been wanting to get into baking!

Yeah they've been sitting in this state for a while. I'll delete them and make an issue tracking interest for those pages.

sfshaza2 (approved):
Wow, what an ambitious PR! You know, some years ago I searched for every page in the repo that had no description and added one, EXCEPT for the pages that were autogenerated, etc. :D. 

I didn't know that you could break lines up in the yaml matter with an arrow. I see you did that in many files, which I guess is fine. :) But I do think that the TBD pages should be deleted. And the "batter" page should move to the cookbook.  LOL

src/security/index.md
parlough: In this case I used it to keep the line <= 80 characters.

PR #8201, opened 2023-02-02
Fix sidebar text overlaying banner

parlough:
Currently if you scroll the sidebar up, its text ends up overlaying the banner:

Sidebar text overlaying banner

This PR adds a `z-index` to the banner that is greater than the sidebar, but less than the top-nav to make sure it stays above the sidebar text.

Sidebar text no longer overlaying banner


domesticmouse (approved):

sfshaza2 (approved):
lgtm

Issue #8135, opened 2023-01-23
[PAGE ISSUE]: 'Simple app state management'


saborrie:
### Page URL

https://docs.flutter.dev/development/data-and-backend/state-mgmt/simple/

### Page source

https://github.com/flutter/website/tree/main/src/development/data-and-backend/state-mgmt/simple.md

### Describe the problem

The example unit test is not correct. It will pass even if cart.add is a no-op - if `notifyListeners()` is never called by the cart model, then the line `expect(cart.totalPrice, greaterThan(startingPrice));` will never be called, and this test will make 0 assertions, so it would still pass.

```dart
test('adding item increases total cost', () {
  final cart = CartModel();
  final startingPrice = cart.totalPrice;
  cart.addListener(() {
    expect(cart.totalPrice, greaterThan(startingPrice));
  });
  cart.add(Item('Dash'));
});
```

### Expected fix

Provide an example test which *does* verify the cart.add method - I'm very new to dart so I don't know the idiomatic way to do so.

domesticmouse:
Fixed by https://github.com/flutter/website/pull/8207

PR #8207, opened 2023-02-04
Fixes a potential success even if no-op in test.

nohe427:

This fixes a potential success in the test which could happen if the cart.add method fails to call
notifyListeners(). This just uses a simple interger counter. A better implementation would be to use the mockito package: https://pub.dev/packages/mockito however, I was attempting to keep the sample simple.

Fixes [PAGE ISSUE]: 'Simple app state management' #8135

## Presubmit checklist

### DIFF
--- a/examples/state_mgmt/simple/test/model_test.dart
+++ b/examples/state_mgmt/simple/test/model_test.dart
@@ -6,10 +6,13 @@ void main() {
   test('adding item increases total cost', () {
     final cart = CartModel();
     final startingPrice = cart.totalPrice;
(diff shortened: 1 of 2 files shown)


domesticmouse (changes requested):
This is an excerpt from a dart file. This PR will fail CI until the related file is also changed 😏

nohe427:
Forgot to read the `README`. I should be all set now.

domesticmouse (approved):

domesticmouse:
Good to land once CI turns green

PR #8232, opened 2023-02-09
Fixing a bad link

sfshaza2:
As it says


atsansone (approved):
LGTM as changed. Did you want to change "see" to "check out", per recent copy updates?

PR #8243, opened 2023-02-11
Update pigeon example to generate properly

parlough:
**Issue:** The pigeon example source file was running into multiple errors and failing to generate.

**Note:** We should document pigeon more which we have multiple issues mentioning already, but the goal here is to just get the example at least correct/functional in its current minimal form.

**Fixes and updates:** 

- Updates the source file to fix errors.
- Renames the source file to `pigeon_source.dart`
- Checks in the generated code so we know we're using it correctly in `generated_pigeon.dart`
- - Updates the usage of the generated code to account for those updates
- Updates to pigeon 8.0.0
- Removes the generated comments from `pubspec.yaml`


examples/development/platform_integration/lib/generated_pigeon.dart
```diff
@@ -1,16 +1,117 @@
-// #docregion Search
```
domesticmouse, line 1: Is this going to break a snippet somewhere?

domesticmouse (approved):

examples/development/platform_integration/lib/generated_pigeon.dart
parlough: I updated the contents in this file and moved them to `pigeon_source.dart` where the updated docregion is.

Issue #8221, opened 2023-02-08
[PAGE ISSUE]: 'Performance'


james2doyle:
### Page URL

https://docs.flutter.dev/perf/

### Page source

https://github.com/flutter/website/tree/main/src/perf/index.md

### Describe the problem

The link on "[Using memory wisely](https://docs.flutter.dev/perf/memory)" goes nowhere.

### Expected fix

 Link to a viable page or remove the link.

danagbemava-nc:
Thanks for reporting this.

Issue #8249, opened 2023-02-13
Create docs for memory and battery performance


parlough:
### What information is missing?

These documents were planned and linked to, but never implemented. 

### How would you like us to fix this problem?

- Write a "Using memory wisely" doc at `/perf/memory`
- Write a "Conserving battery" doc at `/perf/power`
- Update [`/perf`](https://docs.flutter.dev/perf) to link to these articles

parlough:
I've split this issue into more focused ones: https://github.com/flutter/website/issues/9400 and https://github.com/flutter/website/issues/9399, since they likely will be completed separately.

PR #8250, opened 2023-02-13
Temporarily comment out mentions to memory perf doc like battery entry

parlough:
We removed these docs as they were empty and causing confusion. This comments out the broken link to it, but we should write the articles and link to them. Tracking in https://github.com/flutter/website/issues/8249

Fixes https://github.com/flutter/website/issues/8221

### DIFF
--- a/src/perf/index.md
+++ b/src/perf/index.md
@@ -56,7 +56,7 @@ in some future pages.
 
 {% comment %}
 
-TODO: Reintroduce this article and add this link back.
+TODO(https://github.com/flutter/website/issues/8249): Reintroduce this article and add this link back.
 
 ## Memory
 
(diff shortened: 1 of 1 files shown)


src/perf/index.md
```diff
@@ -69,14 +74,15 @@ the quicker it is to download.
 
 {% comment %}
 
+TODO: Reintroduce this article and add this link back.
```
domesticmouse, line 77: Do you want to commit TODO items?

domesticmouse (approved):

src/perf/index.md
parlough: It's a comment and I've added links to the issues now, so I think it should be fine. It will make it easier to notice that it's the section that needs to be added back.

sfshaza2 (approved):
lgtm

PR #8267, opened 2023-02-15
Adding an image so I can refer to it without breaking the build.

sfshaza2:
Adding an image that shows the anatomy of an app as created by "flutter create".


atsansone (approved):
I would suggest including what's on the left inside box with bigger type, but that doesn't mean this can't merge as is. Your call.

sfshaza2:
Yeah, to do that, I'd have to pester the engineer who created it.

PR #8274, opened 2023-02-16
Adding info on how to remove flutter using snapd

Elio-Muculo:

This add a new heading explaining how flutter could be removed using snapd once it's installed.
comes in need when someone want to remove flutter on their OS that was causing a problem.

## Presubmit checklist


sfshaza2:
Hi, @Elio-Muculo! Thanks for this PR! Unfortunately, I'd rather not put this info right on the install page, which is already long enough. I'm going to close this for now.

PR #8289, opened 2023-02-22
Fix header active state

MeandNi:
Fix header active state

image

### DIFF
--- a/src/_includes/header.html
+++ b/src/_includes/header.html
@@ -90,13 +90,13 @@
                Docs
             </a>
             <div class="dropdown-menu" aria-labelledby="docs-navbar-dropdown">
-              <a class="dropdown-item {% if route == '/' %} active{% endif %}"
+              <a class="dropdown-item {% if route == '/whats-new' %} active{% endif %}"
                  href="/whats-new">What's new</a>
               <a class="dropdown-item {% if route contains '/get-started/editor' %} active{% endif %}"
                  href="/get-started/editor">Editor support</a>
               <a class="dropdown-item {% if route contains '/development/tools/hot-reload' %} active{% endif %}"
                  href="/development/tools/hot-reload">Hot reload</a>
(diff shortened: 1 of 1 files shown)


src/_includes/header.html
```diff
@@ -90,13 +90,13 @@
                Docs
             </a>
             <div class="dropdown-menu" aria-labelledby="docs-navbar-dropdown">
-              <a class="dropdown-item {% if route == '/' %} active{% endif %}"
+              <a class="dropdown-item {% if route == '/whats-new' %} active{% endif %}"
```
parlough, line 93: Nice catch!

src/_includes/header.html
```diff
@@ -90,13 +90,13 @@
...
                  href="/whats-new">What's new</a>
               <a class="dropdown-item {% if route contains '/get-started/editor' %} active{% endif %}"
                  href="/get-started/editor">Editor support</a>
               <a class="dropdown-item {% if route contains '/development/tools/hot-reload' %} active{% endif %}"
                  href="/development/tools/hot-reload">Hot reload</a>
-              <a class="dropdown-item {% if route contains '/perf/rendering/ui-performance' %} active{% endif %}"
+              <a class="dropdown-item {% if route contains '/perf' %} active{% endif %}"
                  href="/perf/rendering/ui-performance">Profiling</a>
```
parlough, line 100: Seems the problem here is that the profiling page changed locations and this old link is redirecting to just `/perf`. Perhaps let's just fix the link in this PR, so it matches flutter.dev and the active entry will show on the intended page:

```suggestion
              <a class="dropdown-item {% if route contains '/perf/ui-performance' %} active{% endif %}"
                 href="/perf/ui-performance">Profiling</a>
```

If someone thinks we should link to "Performance" and `/perf` instead, we can discuss that separately and make the change on both sites.

parlough (commented):
Thanks for noticing and fixing this! I have one potential suggestion so the behavior matches flutter.dev:

parlough (approved):
Thanks again, looks good to me!

Issue #8286, opened 2023-02-21
[PAGE ISSUE]: 'Copy command not working'


CaptainLazarus:
### Page URL

https://docs.flutter.dev/get-started/install/linux/

### Page source

https://github.com/flutter/website/tree/main/src/get-started/install/linux.md

### Describe the problem

The "flutter sdk-path" command cannot be copied to the clipboard without the preceding dollar symbol and whitespace.

### Expected fix

The copy command button should only copy the command without the preceding dollar and whitespace.

PR #8290, opened 2023-02-22
Allow leading spaces when stripping `$` from terminal clipboard content

parlough:
Kramdown (our Markdown processor) has different indentation rules than other Markdown processors, and in this case the indentation is tied to the start of the line, not the indentation of the opening tag (```). We could move all alert code blocks to the beginning, but it will result in inconsistency and a lot more changes. 

For now, this adjusts the script which handles removing the leading `$` when clicking the copy to clipboard button on terminal snippets, so that the `$` is removed even if there are extra leading spaces. This also removes those leading spaces when copying.

Fixes https://github.com/flutter/website/issues/8286

### DIFF
--- a/src/get-started/install/_get-sdk-linux.md
+++ b/src/get-started/install/_get-sdk-linux.md
@@ -17,7 +17,7 @@ $ sudo snap install flutter --classic
 ```
 
 {{site.alert.note}}
-  Once the snap is installed, you can
+  Once you install the snap,
   use the following command to display your Flutter SDK path:
 
   ```terminal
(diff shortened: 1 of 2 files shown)


domesticmouse (approved):

src/get-started/install/_get-sdk-linux.md
```diff
@@ -17,7 +17,8 @@ $ sudo snap install flutter --classic
 ```
 
 {{site.alert.note}}
-  Once the snap is installed, you can use the following command to display your Flutter SDK path:
+  Once the snap is installed, you can
+  use the following command to display your Flutter SDK path:
```
atsansone, line 21: **chore:** Change to Active Voice.
```suggestion
  Once you install the snap,
  use the following command to display your Flutter SDK path:
```

atsansone (approved):
LGTM % one chore. Thanks @parlough .

PR #8316, opened 2023-02-28
Fix link checker ignoring links during GitHub workflow

parlough:
The link checker has been ignoring local links (docs.flutter.dev to docs.flutter.dev), at least since https://github.com/flutter/website/pull/7580, since we use absolute URLs through `{{site.url}}`. When emulating the site locally, the url set as `docs.flutter.dev` was seen as external (from `localhost:5500`) which we currently don't check on CI, therefore we were not checking if links were valid.

This PR introduces an override to the `{{site.url}}` variable through an extra `_config_link_check.yml` when running the link checker on GitHub actions, so that it matches the emulator.

There may be a better way to fix this, but without restructuring some of the Docker workflow, this seems like the easiest way for now.

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -19,4 +19,6 @@ jobs:
         with:
           submodules: "recursive"
       - name: Build and check links
-        run: make build # Runs tool/check-links.sh as part of build
+        # Runs tool/check-links.sh as part of build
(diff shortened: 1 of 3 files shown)


domesticmouse (approved):

sfshaza2:
Nice! lgtm

PR #8318, opened 2023-02-28
Update and fix multiple redirected or outdated links

parlough:
These links were being redirected, sometimes to not the most accurate place. This PR updates all internal links to their current most-accurate location.

flutter.cn and perhaps other translated sites don't support the Firebase redirects currently, so these links would fail for them before this change.

### DIFF
--- a/src/development/platform-integration/ios/c-interop.md
+++ b/src/development/platform-integration/ios/c-interop.md
@@ -241,42 +241,6 @@ use the following instructions:
 1. Also add it to the **Linked Frameworks & Libraries**
    section of the target in Xcode.
 
-#### Compiled (dynamic) library (macOS)
-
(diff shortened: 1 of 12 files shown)


src/development/platform-integration/ios/c-interop.md
```diff
@@ -241,42 +241,6 @@ use the following instructions:
 1. Also add it to the **Linked Frameworks & Libraries**
    section of the target in Xcode.
 
-#### Compiled (dynamic) library (macOS)
```
parlough, line 244: I removed this section since its duplicated on the macOS c-interop page which seems where it's targeted to.

domesticmouse (approved):

sfshaza2 (approved):
lgtm

PR #8329, opened 2023-03-02
Standardize web document locations

parlough:
The "Building a web app" entry under "Get started" was alone there while every other platform's equivalent pages are under "Platform integration". This PR moves that page under "Platform integration > Web".

It also moves the orphaned `/web` page to be the Web "overview/homepage" at https://docs.flutter.dev/development/platform-integration/web which is currently an autogenerated TOC page.


sfshaza2 (approved):
lgtm

PR #8343, opened 2023-03-03
Fix entire section being wrapped in alert in Layout codelab

parlough:
Everything after **Exercise: Wrap the Column in a Row** of [Putting it all together](https://docs.flutter.dev/codelabs/layout-basics#putting-it-all-together) was being wrapped in an alert, since the alert on [L2094](https://github.com/flutter/website/blob/main/src/codelabs/layout-basics.md?plain=1#L2094) was unclosed.


domesticmouse (approved):

sfshaza2 (approved):
good catch!

PR #8361, opened 2023-03-07
Update more splash screen links

parlough:
This way we end up on more accurate pages and don't rely on redirects which don't work on some translated sites like flutter.cn.


khanhnwin (approved):
Thanks @parlough!

Issue #1435, opened 2018-10-10
[dash] Styling followup for active main-nav entries


chalin:
From https://github.com/flutter/website/pull/1434#discussion_r224178372:

> make the "hover" and "active" design look like https://material.io/develop

fertolg:
@chalin The mobile nav doesn't show the get started button because the button is not present in desktop. Should the button always be present in mobile? Currently the liquid template is set to add or remove the button markup entirely depending on the `show-nav-get-started-button` button property.

chalin:
> Should the button always be present in mobile?

I guess that that is a question for stakeholders.

Sfshaza:
I do not have strong feelings on this, but I do like the Get started button at the bottom of the docs page.  @maguinis, do you have a strong opinion?

fertolg:
@chalin there were some updates last week while I was gone with regards to the Get Started button. It seems now on desktop the 'get started' button is shown on all pages. But in mobile it has disappeared completely from all pages, is this intentional? cc @sfshaza2

sfshaza2:
It was intentional to add it to all pages on the web. I don't recall
talking about removing it from all pages on mobile, but that sounds like a
mistake.

chalin:
> @chalin there were some updates last week while I was gone with regards to the Get Started button. It seems now on desktop the 'get started' button is shown on all pages. 

Right.

> But in mobile it has disappeared completely from all pages, is this intentional?

Sorry, I had meant to open an issue about that. I wasn't sure whether we still wanted the Get started button on mobile, and if so, where it should appear. Can you look into that @fertolg?

PR #8370, opened 2023-03-08
Introduce basic hover states for sidenav entries

parlough:
- Uses primary color (like other links) when hovering any non-active entries
- Uses the link hover color when hovering the deepest active entry which is already the primary color

Try it out with different combinations on the staged site: https://parlough-style-flutter-dev.web.app

Closes https://github.com/flutter/website/issues/1435


domesticmouse (approved):

khanhnwin:
looks like there's a conflict @parlough - can you resolve? Once that's done, this is good to land! Thank you!

atsansone (approved):
Another good catch. Surprised that it wasn't in the sidenav to start.

Thanks @parlough .

PR #8371, opened 2023-03-08
Move 'Security False Positives' to references

parlough:
It's really a reference material when users come across security error messages, rather than a learning resource, and one of the main entry points is likely the [Security](https://docs.flutter.dev/security) doc anyway.


domesticmouse (approved):

khanhnwin:
@sfshaza2 ptal - related to docs structure!

khanhnwin (approved):

atsansone (approved):
Good catch @parlough

sfshaza2 (approved):
lgtm

PR #8376, opened 2023-03-10
Adding a page on testing plugins and doing some related cleanup.

sfshaza2:
@stuartmorgan, please review the new page

@parlough, in the process of adding a new page on testing plugins, I found an old (2.22) page on migrating plugins to use the "new" plugin API that supports federated plugins. Going forward, all migration docs should be tucked away with release notes or breaking changes. I moved the obsolete file to `breaking changes` and updated all the links that pointed to it. I also removed it from the sidebar, and added the new page to the sidebar. I made sure that the two "plugin testing" files point to each other and explained their different purposes. Finally (I think), I set up a redirect in firebase.json.


parlough:
I've updated the submodules to match the versions in `main`. If you pull down the changes, running `git submodule update` should update them locally too.

Thanks for moving the migration doc!

domesticmouse (approved):

PR #8378, opened 2023-03-10
Regenerate package-lock with latest updates and new format

parlough:
There's a new v3 lockfile format with npm 9+. This PR regenerates the lockfile for those changes as well as other dependency updates.


sfshaza2 (approved):
lgtm

Issue #8322, opened 2023-02-28
Web support information in the installation guides is out of date


InMatrix:
### Page URL

https://docs.flutter.dev/get-started/install/windows/

### Page source

https://github.com/flutter/website/tree/main/src/get-started/install/windows.md

### Describe the problem

> Flutter has support for building web applications in the stable channel. Any app created in Flutter 2 automatically builds for the web. To add web support to an app created before web was in stable, follow the instructions on [Building a web application with Flutter](https://docs.flutter.dev/get-started/web) when you’ve completed the setup above.

Does web support just work out of the box now? This instruction still refers to Flutter 2, which we can probably drop or significantly simplify. 

Cc: @kevmoo

kevmoo:
Thanks for flagging, @InMatrix !

PR #8408, opened 2023-03-20
Removing mention of compiling specifically for web

sfshaza2:
Fixes https://github.com/flutter/website/issues/8322

I don't see any reason to highlight web in these sections.

@kevmoo @InMatrix


kevmoo:
LGTM – seems like a non sequitur.

Issue #8427, opened 2023-03-23
☂ Add UI adaptation guides for mobile platforms


InMatrix:
### What information is missing?

A few of us are looking into UI adaptation between iOS and Android beyond automated behavior documented in [Platform-specific behaviors and adaptations](https://docs.flutter.dev/resources/platform-adaptations). 

### How would you like us to fix this problem?

We'd like to develop guides to help developers make informed UI design choices and provide example code to assist the implementation of adaptive UI. Here're some of the topics we're investigating:

- [x] Alert Dialog ([draft](https://github.com/InMatrix/veggieseasons_adaptive/discussions/15))
- [x] flutter/website#8428 
- [x] flutter/website#8744
- [x] flutter/website#8678
- [ ] Segmented Buttons
- [ ] Bottom sheets
- [ ] Buttons

The tentative plan is to add those guides to [Platform-specific behaviors and adaptations](https://docs.flutter.dev/resources/platform-adaptations) as new sections.

Comment on this issue to help us prioritize which topic we should spend time on. 

Cc: @leighajarett @nancyhu2023

PR #8429, opened 2023-03-23
Add guidance for adapting alert dialogs

InMatrix:

This PR adds guidance for alert dialogs to [Platform-specific behaviors and adaptations
](https://docs.flutter.dev/resources/platform-adaptations). This is the first of a few cross-platform adaptation guides we're developing. The goal is to address https://github.com/flutter/website/issues/8427 over time. 

https://github.com/flutter/website/issues/8427 

## Presubmit checklist


InMatrix:
I'm submitting this PR to get this on the radar of the DevRel team and other interested parties. There are some stylistic issues I need to address to satisfy the checklist. 

If this is not the right place for the proposed content, please suggest alternatives. Thanks!

Cc: @leighajarett @nancyhu @HansMuller

leighajarett:
Added @MitchellGoodwin to provide any eng guidance

InMatrix:
This PR is ready for review. I'd appreciate if one of the code owners can take a look.

src/resources/platform-adaptations.md
```diff
@@ -548,6 +552,34 @@ double tap and shows the selection toolbar.
...
+
+Because alert dialogs are tightly integrated with the operating system 
+and users have been trained to recognize and respond to them, 
+it's recommended to follow the platform conventions. 
+Specifically, you can use Flutter’s `AlertDialog` widget on Android 
+and the `CupertinoAlertDialog` widget on iOS. 
+Here’s a [code sample](https://gist.github.com/InMatrix/1a61ba065c0c36ae642ff934971ab97b) 
+you can use to achieve the recommended adaptation behavior. 
+Further details about adapting alert dialogs is available in 
+[this article](https://github.com/InMatrix/veggieseasons_adaptive/discussions/15).
```
sfshaza2, line 581:
```suggestion
Further details about adapting alert dialogs is available in
this [multiplatform version of the VeggieSeasons example][veggie-seasons].

[veggie-seasons]: https://github.com/InMatrix/veggieseasons_adaptive/discussions/15.
```

sfshaza2 (commented):
Thanks for updating this, @InMatrix! Ultimately, I prefer that we don't link to personal GitHub accounts from the official docs. I'll leave it for now, but maybe we could finalize the multiplatform version of Veggie Seasons and add it to the examples repo? (Without deleting the original Cupertino version, as I rely on that in the restore state docs?)

Also, we avoid linking to gist files, which aren't tested. Our samples repo undergoes testing on every CI (I believe). Also, our code excerpts are routinely tested on every submit.

Finally, I made the link text more descriptive.

src/resources/platform-adaptations.md
InMatrix: The URL points to a page more specific than the suggested label. Would something like "...available in this [preliminary alert dialog adaptation guide][alert-guide]" work?

InMatrix:
@sfshaza2 Thank you for your review!

> Ultimately, I prefer that we don't link to personal GitHub accounts from the official docs.

Yes, I'd like to move those platform adaptation guides to an official repo eventually. They're preliminary recommendations at the moment, so I considered two options 1) writing them in. the wiki section of the flutter/flutter repo, and 2) posting them to the Discussion section of the veggieseasons_adaptive repo under my personal account. I went with the second option,  because it includes an easy way to receive feedback. 

> maybe we could finalize the multiplatform version of Veggie Seasons and add it to the examples repo? (Without deleting the original Cupertino version, as I rely on that in the restore state docs?)

Yes, this is something I'd like to do in Q2. The multiplatform version won't use state restoration APIs for simplicity. 

> Also, we avoid linking to gist files, which aren't tested. 

That makes sense. Including the full snippet inline is a bit long. I considered using DartPad, but this piece of code doesn't work on the Web. Do you have any suggestions on where I should host this snippet?

src/resources/platform-adaptations.md
sfshaza2: Yes, that would be fine.

InMatrix:
Thanks @sfshaza2. I'm making some additional tweaks after getting feedback from @nancyhu2023. I'll ping this thread once it's ready.

InMatrix:
@sfshaza2 I updated the PR. Please take another look. Thank you for your help.

PR #8457, opened 2023-03-29
Misc 'Supported Platforms' changes and cleanup

parlough:
- Moves supported platforms from being deeply nested under release notes to being under `/reference`, since it is a sort of reference and can be linked to from multiple locations. Open to feedback here :)
- Removes the page's unrelated breadcrumbs (through the above change)
- Replaces `<center>` usage with a table class that centers all entries not in the first column.

**Staged:** https://parlough-docs-flutter-dev.web.app/reference/supported-platforms

### DIFF
--- a/src/reference/supported-platforms.md
+++ b/src/reference/supported-platforms.md
@@ -27,7 +27,7 @@ make sure to make a corresponding change in the get started pages:
   * /src/get-started/install/macos.md
   * /src/get-started/install/windows.md
   * /src/get-started/install/linux.md
-{% endcomment %}
+{% endcomment -%}
 
 
 ## Deploying Flutter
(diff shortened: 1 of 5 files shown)


src/reference/supported-platforms.md
```diff
@@ -27,7 +27,7 @@ make sure to make a corresponding change in the get started pages:
   * /src/get-started/install/macos.md
   * /src/get-started/install/windows.md
   * /src/get-started/install/linux.md
-{% endcomment %}
+{% endcomment -%}
```
domesticmouse, line 30: I'm curious what this change does

domesticmouse (approved):

src/reference/supported-platforms.md
parlough: It just removes some extra whitespace in the generated HTML around where the comment would be.

I primarily made the change to be consistent with comments elsewhere.

sfshaza2 (approved):
lgtm!

Issue #5777, opened 2021-05-13
Update Information Architecture


sfshaza2:
Since Flutter launched 1.0, the IA hasn't scaled well and I'd like to update it. Note that the IA doesn't have to reflect the directory structure.

One thing I'd like to do is to move the release notes, which don't really belong under docs/development/tools. When I do find/grep, it's unpleasant to include those.

parlough:
I'm going to close this in favor of the more specific issues, particularly since the specific concern listed here has been addressed.

PR #8469, opened 2023-03-31
[Proposal] Reorganize and streamline release content

parlough:
Screenshot of proposed releases top-level

Currently release information is hidden deep under **Development > Tools > SDK** (`/development/tools/sdk/<...>`) where users don't expect to find it, and there were other inconsistencies in terms of file location and breadcrumbs. For example, breaking changes were already under `/release`. This PR seeks to streamline various release reference material solving those and other related issues.

The primary change here where I'm **looking for the most feedback** is on the new top-level subnav category of **Releases** to make these docs more discoverable. How does this location seem? Should it be below **Reference** instead? Should it be a child of **Reference**? 

While these materials can be considered a reference and I originally had them under reference, releases seem like an isolated concept to me, and many other sites include releases separately. For example, Android Studio, where [release information](https://developer.android.com/studio/releases) has its complete own sidenav. 

----

Detailed changes if you're interested:
- Renames the "Flutter SDK releases" page to "Flutter SDK archive" to better fit its status
  - The name "Releases" implied it was where users should go to learn about releases, but that's not what the page is for. It's for downloading bundles of specific versions and it's not the primary way users should download or upload Flutter. Beyond that, with the new structure, "Releases" would show up twice in the sidenav, making it a bit awkward.
  - Adds a cross-link to the release notes from the page, primarily to handle users that do visit the page expecting to find release information.
  - Question: Should this be called "SDK archive" instead on the sidenav? Didn't seem as nice to me for some reason.
 - Moves what is now named "Flutter SDK archive" to `/release/archive` from `/development/tools/sdk/releases`
   - Adds the appropriate redirect
- Moves "Release notes" to `/release/release-notes` from `/development/tools/sdk/release-notes`
   - Adds the appropriate redirects for the index page and all children pages
 - Adds an empty index page to `/release` with the title "Releases". Uses this to enable proper breadcrumbs for the archive page, release notes, and breaking changes which didn't all have breadcrumbs enabled before.
   - Visiting this page redirects users to `/release/archive`.
 - Moves the archive, release notes, and breaking changes from under **Development > Tools > SDK** to a new top-level **Releases** section in the sidenav above **Reference** and below **Resources**.
   - Releases are a relatively isolated concepts and users did not expect to find release notes deep under tools.
   - Originally, I included some of the most recent releases in the sidenav too, but after some consideration, I think a one click into the index is sufficient, plus it has clear links to the blog posts too.
   - **Question:** Do you think the [Upgrading Flutter](https://parlough-docs-flutter-dev.web.app/development/tools/sdk/upgrading) doc should be with these other release-orientated docs or stay under tools/sdk?
- Adds a special redirect from `/release-notes/<version>` to `/release/release-notes/release-notes-<version>` 
  - The goal here is for shorter, more readable links that tools, social media, blogs, etc could link to.
  - Example: https://parlough-docs-flutter-dev.web.app/release-notes/3.7.0
  
 Contributes to https://github.com/flutter/website/issues/5777


parlough:
@InMatrix @sfshaza2 I wanted to figure out the release portion of changes, since they are rather independent of other changes. My goal is to separate work as much as possible, similar to the style, performance, and generation fixes I've already landed on `main` to keep the primary work focused and easier to review. Also it reduces work having to handle constant merge conflicts.

This is more of a proposal and I'm happy to make changes. Changes are super easy to make with this smaller set of changes. I included some questions and areas for feedback above. Let me know what you think!

sfshaza2 (commented):
@parlough, the way I read this, it appears that you are coalescing the Flutter SDK page, the breaking changes pages, and the release notes pages under "/release". Is that correct? If so, I LOVE IT.

parlough:
Yes, alongside introducing a new top-level **Releases** entry in the sidenav for that content.

InMatrix:
> The primary change here where I'm looking for the most feedback is on the new top-level subnav category of Releases to make these docs more discoverable. How does this location seem? Should it be below Reference instead? Should it be a child of Reference?

I kind of like what Android Studio does. They've got "what's new" first and then "get started" second in their left nav. Similarly, we have this page https://docs.flutter.dev/whats-new but it's not in the left nav. Maybe we can add a category named "Stay up to date" immediately below "Get started." This category would include this "what's new" page, the pages in your "Releases" category, and a few other relevant pages, such as https://docs.flutter.dev/development/tools/sdk/upgrading. 

What do you think?

parlough:
That sounds pretty good to me! Let me try it out and see what it ends up feeling like.

parlough:
I've proposed those changes in https://github.com/flutter/website/pull/8516. Hoping for specific feedback on the order and names there. Thanks!

parlough:
Superseded by https://github.com/flutter/website/pull/8516

PR #8493, opened 2023-04-05
[Proposal] Link to breaking changes & migrations from release notes index

parlough:
We already collect release blogs and notes on the "Release notes" page already, but not the associated breaking changes/migration docs, which are potentially the most impactful.


domesticmouse (approved):

sfshaza2 (approved):
Great idea!

PR #8508, opened 2023-04-06
Add a new `make` command for easier staging

parlough:
- Removes prexisting built files to avoid outdated generated page
- Uses the currently running dev container (or starts one) to avoid having to rerun setup which takes a long time. Doesn't run the link checker
- Utilizes a new `_config_stage.yml` to set `{{site.url}}`` to empty, so that links use relative links and don't point to localhost or docs.flutter.dev, but rather the appropriate staged site on Firebase
- Builds the site with that URL, then stages the site to the version specified with `firebase use` or to the project specified with `FIREBASE_ALIAS`

For example the following will stage the site with functional links to https://parlough-docs-flutter-dev.web.app/:

```terminal
make stage-local FIREBASE_ALIAS=parlough-docs-flutter-dev
```


parlough:
Feel free to try this out with your own Firebase instance. I created this branch on `flutter/website`, so you should just be able to pull it down!

sfshaza2 (approved):
lgtm. I can't wait!

PR #8519, opened 2023-04-07
Add Provenance Column

sealesj:
Provenance is now being uploaded for all platforms for Flutter releases, so we would like to make this link more readily available to users on each release. This change offers a link for each release to the associated provenance. 

b/277230427

## Presubmit checklist


parlough:
Thanks @sealesj! Could you update the PR against `main`? The `_os` file recently moved and changed names to https://github.com/flutter/website/blob/main/src/release/_release_os.md.

I have a feeling users may not be familiar with the term "Provenance". I wonder if we can describe it somewhere on the page or perhaps link the header to a definition like SLSA's provenance page? https://slsa.dev/provenance. What do you think?

sealesj:
@parlough That's a great point and thank you for the comment! I will update with a link to more details on what provenance is and also get this updated with the restructuring from `main`

sealesj:
@parlough  How does this update look? I've added the link to the SLSA website describing provenance as well as simplified the link to view to display as "View" Screenshot 2023-04-10 at 3 25 46 PM

src/release/_release_os.md
```diff
@@ -12,8 +12,8 @@ Select from the following scrollable list:
 
 <div class="scrollable-table">
   <table id="downloads-{{id}}-{{channel}}" class="table table-striped">
-  <thead><tr><th>Flutter version</th><th>Architecture</th><th>Ref</th><th class="date">Release Date</th><th>Dart version</th></tr></thead>
-  <tr class="loading"><td colspan="5">Loading...</td></tr>
+  <thead><tr><th>Flutter version</th><th>Architecture</th><th>Ref</th><th class="date">Release Date</th><th>Dart version</th><th><a href="https://slsa.dev/provenance">Provenance</a></th></tr></thead>
```
parlough, line 15: However, I'm now looking at the file this downloads, and I'm worried there may still be some confusion for the user. It doesn't seem so useful on its own especially since the scheme on slsa.dev won't be seen in the plaintext representation of the file. Do we or slsa have some documentation on how to use this file to run verification? Perhaps we can link there in the meantime as that seems more useful.

I'm not familiar with this at all, so sorry about the back and forth!

parlough (commented):
Thanks for making those changes. I do think it looks good and it's nice to provide the link to some resource!

src/release/_release_os.md
sfshaza2: I agree with @parlough . I also think we should add a sentence or two describing what provenance means in the context of a release.  I also don't like having multiple links with the same descriptive text. But I'm not sure how this should be fixed.

sealesj: Would something like this be a good approach to those issues? I've moved the link to the SLSA provenance website into a descriptive sentence, and then labeled the provenance file by the flutter version
Screenshot 2023-04-11 at 12 27 22 PM

sealesj: @sfshaza2 I took at shot at placing some provenance inspection steps here, but would you have a more preferred placement/formatting of these instructions?
3DiYFm2ubpeD4mo

sfshaza2:
@sealesj and I had a brief meeting and I made some suggestions. I am expecting an update soon.  Thanks, @sealesj!

sfshaza2 (commented):
This PR doesn't yet reflect the changes that Jesse and I talked about.

sealesj:
> This PR doesn't yet reflect the changes that Jesse and I talked about.

I've updated the PR to use new verbiage and note formatting 
Screenshot 2023-04-14 at 10 26 43 AM

src/release/archive.md
```diff
@@ -19,6 +19,17 @@ Check out [Flutter’s channels][] for details.
 For details about what's new in the major Flutter releases,
 check out the [release notes][] page.
 
+
+{{site.alert.secondary}}
+  A note on provenance: <a href="https://slsa.dev/provenance">provenance</a>
```
parlough, line 24:
```suggestion
  A note on provenance: [provenance](https://slsa.dev/provenance)
```

src/release/archive.md
```diff
@@ -19,6 +19,17 @@ Check out [Flutter’s channels][] for details.
...
+  describes how software artifacts are built.
+  If you would like view provenance in a more readable format, run the
+  following command using the provenance file URL from a release:
+  ```
```
parlough, line 28:
````suggestion

  ```terminal
````

parlough (approved):
Thanks for those updates explaining what the file is and how to use it!

The archive code looks good to me. Wait for a final approval on the text by Shams :)

sfshaza2 (approved):
@sealesj, once you apply some small changes, lgtm. Thanks!

PR #8542, opened 2023-04-12
Update contextual-survey-metadata.json

jayoung-lee:
More placeholders and (tentative) details added for the upcoming contextual survey. 

The feature is not complete yet. Once it's shipped, we might consider migrating this over to the GCS bucket.


sfshaza2 (approved):
lgtm

PR #8550, opened 2023-04-13
Add max-width of 100% default to images in primary content

parlough:
This matches what is used in dart-lang/site-www.

### DIFF
--- a/src/_sass/components/_content.scss
+++ b/src/_sass/components/_content.scss
@@ -35,6 +35,10 @@
 
   .container {
     max-width: $site-content-max-width;
+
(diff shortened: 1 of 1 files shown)


sfshaza2 (commented):
What if the image is less than the width already? We don't want to force a larger image.

parlough:
@sfshaza2 Yep! This PR just sets the default maximum to 100% so they won't expand outside of the width of the body as the page gets less wide (For example on mobile). They will still prefer their default width.

sfshaza2 (approved):
lgtm!

PR #8551, opened 2023-04-14
Increase consistency between sidenav styles and increase density

parlough:
This PR better standardizes styles between the two sidenavs, fixes some padding issues, properly indicates active entries, and increases density. This can land now but will be even more important as we add new top-level entries with future IA work.

- Increase consistency between the mobile (thin) and normal (wide) sidenav.
  - Fix font and text styles on mobile sidenav to match the normal side
  - Highlight active top-level entry on both, not just mobile sidenav
- Prevent sidenav text overlapping footer on scroll
- Slightly reduce vertical padding between top-level entries to improve density

**Staged:** https://parlough-style-flutter-dev.web.app/


sfshaza2 (approved):
lgtm

PR #8557, opened 2023-04-14
Update to Ruby 3.2.2 and update gems

parlough:
[Ruby 3.2.2](https://www.ruby-lang.org/en/news/2023/03/30/ruby-3-2-2-released/) includes a few security fixes.


sfshaza2 (approved):
lgtm

Issue #8321, opened 2023-02-28
Misleading instruction about downloading source from GitHub


InMatrix:
### Page URL

https://docs.flutter.dev/get-started/install/windows/

### Page source

https://github.com/flutter/website/tree/main/src/get-started/install/windows.md

### Describe the problem

> If you don’t want to install a fixed version of the installation bundle, you can skip steps 1 and 2. Instead, get the source code from the [Flutter repo](https://github.com/flutter/flutter) on GitHub, and change branches or tags as needed. 

This instruction might lead to misunderstanding that once you download an installation bundle, you're stuck with a particular version or channel of Flutter, which is not true. 

This instruction seems to be redundant, since there's a link to https://docs.flutter.dev/development/tools/sdk/releases above it. 

### Expected fix

Consider dropping this paragraph.

PR #8562, opened 2023-04-16
[Docs] [Install] [Windows] remove confusing instruction

piedcipher:

Removes following paragraph & the `git clone` command:

> If you don’t want to install a fixed version of the installation bundle, you can skip steps 1 and 2. Instead, get the source code from the Flutter repo on GitHub, and change branches or tags as needed.

Because, (👇 explanation from the issue description)

> This instruction might lead to misunderstanding that once you download an installation bundle, you're stuck with a particular version or channel of Flutter, which is not true.
>
> This instruction seems to be redundant, since there's a link to https://docs.flutter.dev/development/tools/sdk/releases above it.

_Issues fixed by this PR (if any):_ Fixes: #8321

## Presubmit checklist

### DIFF
--- a/src/get-started/install/_get-sdk-win.md
+++ b/src/get-started/install/_get-sdk-win.md
@@ -21,15 +21,6 @@
   `C:\Program Files\` that requires elevated privileges.
 {{site.alert.end}}
 
-If you don't want to install a fixed version of the installation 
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):

sfshaza2 (approved):
lgtm

PR #8581, opened 2023-04-19
Minor standardization and cleanup in `archive.js`

parlough:
To make future review of new and updated code easier, this PR slightly cleans up and standardizes `archive.js`:

- Use `const` and `let` instead of `var`
- Use template strings where it makes sense
- Standardize to single quotes (to be consistent across the file and with our Dart style)
- Standardize function syntax
- Removes some dead/unused code and comments
- Other minor cleanup

**Staged site to verify functionality:**
https://parlough-docs-flutter-dev.web.app/release/archive

### DIFF
--- a/src/assets/js/archive.js
+++ b/src/assets/js/archive.js
@@ -70,7 +70,7 @@ let macOSArm64ArchiveFilename = '';
 
 // Listen for the macOS arm64 download link to be clicked and update
 // the example unzip command with correct arm64 filename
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):
Is it worth introducing `prettier` or other JavaScript linter?

parlough:
Not a large priority since we don't change the JS much, but I've opened an issue (https://github.com/dart-lang/site-www/issues/4771) to track that possibility.

PR #8584, opened 2023-04-19
Add jq Link

sealesj:
Based on some feedback, it seems like it could be helpful to add a short blurb about the `jq` command and where it could be found so that the user can more easily run the command on the provenance 

## Presubmit checklist


sfshaza2 (approved):
lgtm. We don't use "may", which is considered a legal term.  https://developers.google.com/style/word-list/#may

PR #8588, opened 2023-04-19
Document ClipboardData constructor breaking change

loic-sharma:
Document the breaking change introduced by https://github.com/flutter/flutter/pull/122446.

I will create a dart fix to automate this migration.

Part of: https://github.com/flutter/flutter/issues/121976

## Presubmit checklist


src/release/breaking-changes/clipboard-data-required.md
```diff
@@ -0,0 +1,62 @@
...
+
+The [`ClipboardData constructor`][]'s `text` argument is no longer nullable.
+Code that provides `null` to the `text` argument must be migrated to provide
+an empty string `''`.
+
+## Context
+
+In preparation for supporting multiple clipboard data variants, the
+`ClipboardData` constructor now requires that at least one data variant is
+provided.
```
loic-sharma, line 16: I didn't mention this, but platforms were inconsistent in how they handled `null`. Some would reset the clipboard, some set the clipboard to the text `"null"`, others would crash. This was an unintended consequence of platforms assuming that at least one data variant is provided. We made the behavior consistent by disallowing this case.

sfshaza2: How about something like this:

"Previously, platforms were inconsistent in how they handled `null`. Behavior is now consistent across platforms. If you are interested in the low level details, see PR xxx."

I can imagine that some would want more info.

sfshaza2 (approved):
This lgtm. I won't merge this in case you want to add a note about the inconsistent behavior, but this looks good otherwise!

src/release/breaking-changes/clipboard-data-required.md
loic-sharma: Added!

PR #8602, opened 2023-04-20
Update DevTools docs.

kenzieschmoll:
Update DevTools images for the latest build, and fix a couple things with performance and cpu profiler docs.


sfshaza2 (approved):
lgtm

PR #8621, opened 2023-04-24
Losslessly optimize images

parlough:
There have been a lot of new images on the site recently, so this PR losslessly optimizes them to reduce site build and load time.


sfshaza2 (approved):
LOVE THIS!  lgtm

PR #8627, opened 2023-04-25
Replace Webby mention with I/O in banner

parlough:
The Webby voting has ended. This PR removes the Webby mention and reintroduces the I/O call to action.

Screenshot of banner


domesticmouse (approved):

sfshaza2 (approved):
lgtm

PR #8682, opened 2023-05-12
Update macos deployment instructions to bump minimum supported version to 10.14

dnfield:
@stuartmorgan tells me this was bumped. @cbracken fyi

### DIFF
--- a/src/deployment/macos.md
+++ b/src/deployment/macos.md
@@ -94,7 +94,7 @@ In the **Identity** section:
 In the **Deployment info** section:
 
 `Deployment Target`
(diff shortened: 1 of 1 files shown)


atsansone (approved):

parlough (approved):
Thanks for updating this.

Matches what we changed to in https://docs.flutter.dev/reference/supported-platforms

stuartmorgan (approved):
LGTM.

More authoritative than my say-so is https://docs.flutter.dev/reference/supported-platforms

cbracken (commented):
Thanks - yes I bumped the minimum in Dec.

PR #8688, opened 2023-05-15
adding missing createArgsCodec override in doc of platform-views - iOS

GJJ2019:

_Issues fixed by this PR (if any):
- adding missing createArgsCodec override in doc of platform-views - ios

## Presubmit checklist


atsansone:
@domesticmouse : Could you check this code? Thanks!

mossmana:
Adding @stuartmorgan, who can help find a reviewer for this. I'm probably not the best contact for iOS related code.

mossmana:
Also, @GJJ2019 would it be possible to link to the associated issue on this PR?

GJJ2019:
> Also, @GJJ2019 would it be possible to link to the associated issue on this PR?

I'm not sure there are issues for this but just codec doc was missing so i created the pr. Shall I file issue as well ?

stuartmorgan:
That's an optional part of the protocol, and it's only used if there are view creation arguments, which there aren't in this example, so the premise that the code is missing is not correct. It's not there because it would be dead code. Closing since the example is correct—and less verbose—as-is.

(Also, the Obj-C code in this PR is not at all correct; it's in the wrong part of the object definition, and on the wrong object.)

GJJ2019:
but if i try `creationParamsCodec: const StandardMessageCodec(),` in uikitview then arg returns null for ios so createargscodec is must 

see this https://docs.flutter.dev/platform-integration/android/platform-views
https://docs.flutter.dev/platform-integration/ios/platform-views

stuartmorgan:
> then arg returns null for ios so createargscodec is must

Please re-read my comment above about this example.

GJJ2019:
ok i understand now thx

PR #8692, opened 2023-05-15
More detailed descriptions of the channels

Hixie:
This PR adds details to the channel descriptions in the "Upgrading Flutter" page, providing additional information to help users in finding the channel (release build) that suits their needs.

This supersedes https://github.com/flutter/website/pull/8689


src/release/upgrade.md
```diff
@@ -81,14 +135,9 @@ the Dart [`pub outdated` documentation]({{site.dart-site}}/tools/pub/cmd/pub-out
...
 
-## Keeping informed
-
-We publish breaking change announcements to the
-[Flutter announcements mailing list][flutter-announce].
-Aside from subscribing to receive announcements,
-we'd love to hear from you!
-
 [Flutter SDK archive]: {{site.url}}/release/archive
-[release channels]: {{site.repo.flutter}}/wiki/Flutter-build-release-channels
```
parlough, line 92: Looks like this link definition is still reference on line 7 - https://github.com/flutter/website/blob/a2df8ed07998abedacca1f8bfc8296ac5b482382/src/release/upgrade.md?plain=1#L7

Perhaps with this extra information, the link can be dropped?

Hixie: oops. yeah i am trying to remove links to the wiki from this page. The goal is for the website to be our developer documentation and the wiki the contributor documentation.

Hixie: fixed

parlough: I like that transition. Thanks!

parlough:
Is this ready for review or should it still be a draft?

Hixie:
Yup, not sure why it got marked draft, sorry!

src/release/upgrade.md
```diff
@@ -19,22 +19,67 @@ $ flutter upgrade
...
-  The `dev` channel was retired as of Flutter 2.8.
-{{site.alert.end}}
+### The **stable** channel
+
+We recommend the **stable** channel
+for new users
+and for production app releases.
+It is updated roughly every three months,
+and may additionally receive occasional hot fixes
+for high-severity or high-impact issues.
```
atsansone, line 50: **nits:** Write in active voice per [GSG](https://developers.google.com/style/voice). Limit adverbs as they make Google sound unsure.
```suggestion
We recommend the **stable** channel for new users
and for production app releases.
The team updates this channel about every three months.
The channel may receive occasional hot fixes
for high-severity or high-impact issues.
```

src/release/upgrade.md
```diff
@@ -19,22 +19,67 @@ $ flutter upgrade
...
+
+The Flutter team's plugins and packages are continually tested against the latest **stable** release.
```
atsansone, line 52: **issue:** Write in active voice per [GSG](https://developers.google.com/style/voice).
```suggestion
The Flutter team tests its plugins and packages against the latest **stable** release.
```

src/release/upgrade.md
```diff
@@ -19,22 +19,67 @@ $ flutter upgrade
...
+This channel has passed all our public testing,
+has been verified against test suites for Google products that use Flutter,
+and has been vetted against [contributed private test suites][test registry].
+The **beta** channel receives regular hot fixes
+to address newly discovered important issues.
+
+The **beta** channel is essentially the same as the **stable** channel
+but updated monthly instead of quarterly.
+Indeed, when the **stable** channel is updated,
+it is updated to the latest **beta** release.
```
atsansone, line 69: **issues:**
- Write in active voice per [GSG](https://developers.google.com/style/voice).
- Avoid equivocation unless needed.
```suggestion
The **beta** channel has the latest stable release.
This channel includes the most recent version of Flutter
that passed the following tests:

- All our public testing.
- Test suites for Google products that use Flutter.
- All [contributed private test suites][test registry].

The **beta** channel receives regular hot fixes
to address discovered important issues.

The **beta** channel works like the **stable** channel
but updated each month instead of each quarter.
When the **stable** channel updated,
the update comes from the latest **beta** release.
```

src/release/upgrade.md
```diff
@@ -81,14 +135,9 @@ the Dart [`pub outdated` documentation]({{site.dart-site}}/tools/pub/cmd/pub-out
...
-[Flutter announcements mailing list][flutter-announce].
-Aside from subscribing to receive announcements,
-we'd love to hear from you!
-
 [Flutter SDK archive]: {{site.url}}/release/archive
-[release channels]: {{site.repo.flutter}}/wiki/Flutter-build-release-channels
 [flutter-announce]: {{site.groups}}/forum/#!forum/flutter-announce
 [pubspec.yaml]: {{site.dart-site}}/tools/pub/pubspec
+[test registry]: https://github.com/flutter/tests
+[contributing to Flutter]: https://github.com/flutter/flutter/blob/master/CONTRIBUTING.md
```
atsansone, line 142: **issue:** Change to match prior change.
```suggestion
[Flutter contributors]: https://github.com/flutter/flutter/blob/master/CONTRIBUTING.md
```

src/release/upgrade.md
```diff
@@ -19,22 +19,67 @@ $ flutter upgrade
...
-We recommend using the **{{site.sdk.channel}}** channel
-unless you need a more recent release.
+We currently have one other channel, **master**.
+This is the channel used by people [contributing to Flutter][].
+
+This channel is not as thoroughly tested as the **beta** and **stable** channels.
+
+We do not recommend using this channel as it more likely to contain serious regressions.
+
+The latest documentation for the **master** branch is at: <https://master-api.flutter.dev>
```
atsansone, line 80: **issues:** Write in active voice per [GSG](https://developers.google.com/style/voice).
```suggestion
We have one other channel, **master**.
[Flutter contributors][] use this channel.

We test this channel less than the **beta** and **stable** channels.

We do not recommend using this channel as it might contain serious regressions.

To learn about the latest API documentation for the **master** branch,
check out <https://master-api.flutter.dev>
```

src/release/upgrade.md
```diff
@@ -73,6 +118,15 @@ use the `upgrade` command:
 $ flutter pub upgrade
 ```
 
+To update to the _latest possible version_ of
+all the dependencies listed in the `pubspec.yaml` file,
+and automatically update the constraints in that file,
+use the `upgrade` command with `--major-versions`:
```
atsansone, line 124: **issue:** Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading).

```suggestion
Flutter lets you update to the _latest possible version_ of
all dependencies in your `pubspec.yaml` file.
To make this update and update the constraints in that file,
run the `upgrade` command with `--major-versions`:
```

atsansone (changes requested):
@parlough : A few changes, then I can review again.

parlough:
I'm assuming you meant to tag @Hixie, not me :)

src/release/upgrade.md
Hixie: We're not testing them, the computers test them. The idea is that the CI is using this branch to test the plugins. But I don't want to imply it's the only branch being used.

src/release/upgrade.md
Hixie: The problem with this way of phrasing it is that it implies that there's no other tests that it passed, and I don't want to make that implication. I'm also a bit dubious about saying a channel "includes" something (the channel is just a branch and so has at any one time just one version of Flutter). Finally my concern with saying it "works like" the stable channel is that it missed the key point, which is that it really is the same as the beta channel (literally, the commit from the beta channel goes to the stable channel). Equivocation is needed there though because there are technical subtleties that mean they're not actually identical.

I'm not sure which parts are passive voice here, could you elaborate on that?

src/release/upgrade.md
Hixie: I don't think "Flutter contributors" is a good link text to use for linking to the contributing guide. The guide doesn't list the people contributing. It explains how to contribute to Flutter.

I don't think "To learn about the latest API documentation" makes sense, after all, that link doesn't teach you anything about the API documentation, it _is_ the API documentation.

Hixie: i'm also a bit reluctant to say "we" test the channels, given that there are a number of different processes for testing that aren't all "we" (for one, most of the testing is automated).

src/release/upgrade.md
Hixie: "Flutter lets you" implies that we could stop you, which seems like a weird implication since this is just a text file.

Hixie:
I applied the changes I agreed with and left comments for the others. :-)

atsansone (commented):
@Hixie : OK, I got a little overzealous on the passive voice, but my intention was to remove ambiguity. I hope I've answered your questions and provided better suggestions.

src/release/upgrade.md
Hixie: gotcha. Changed to "The continuous integration for the Flutter team's plugins and packages includes testing against the latest stable release."

src/release/upgrade.md
Hixie: > I ask "who makes the update?"

Ah, I see. I use the passive voice specifically to avoid saying who or what is performing something when I think being explicit about it will just make things more complicated in ways that are not relevant to the discussion. :-)

The answer to "who makes the update" is a whole complicated topic. It's partially automated, we're moving to a two-agent system, it's specifically not _anyone_ on the team... I'm not really sure how to succinctly say who it is who's doing it! :-)

> How is the beta channel different?

I mean, to the extent that it's different, this is what this whole document is trying to explain. What we're trying to do here is convey the idea that the only relevant difference between beta and stable is the update cadence. There are, however, lots of minor differences that are tricky to explain without getting into the weeds (I tried to explain the API docs difference and the plugin testing difference by mentioning those explicitly, but there's lots of other minor differences that are more nuanced).

"The **beta** channel contains most of the same code as the **stable** channel, save features to be released in the next quarter." isn't any more accurate (I would say it's more misleading actually).

> Is it different in that includes some patches and hotfixes?

Both beta and stable branches get cherry-picks.

> does it include new features to be released in the next stable release?

We're trying to express that "beta" branch releases are stable releases, just like releases on the "stable" branch. In fact the "stable" branch release is just a copy of a previous "beta" branch release. Thus, by definition if a feature ends up in a "stable" branch release it would have previously been in a "beta" branch release.

src/release/upgrade.md
Hixie: That would break the symmetry with the previous paragraph, which I think would harm readability, but I've made a similar change to reduce the sentence length; I put "This also automatically update the constraints in the `pubspec.yaml` file." after the command and removed the relevant clause from the paragraph in question.

Hixie:
Thanks for the explanations, PTAL!

atsansone (approved):

johnpryan (approved):

PR #8750, opened 2023-05-25
Update jquery to 3.7.0 and js-cookie to 3.0.5

parlough:
We've tested these on dart.dev for a few weeks with no issues.


sfshaza2 (approved):
lgtm

PR #8791, opened 2023-05-31
Break-out navigation from top-level UI category in sidenav

parlough:
Also adjusts to task-orientated titles.

This is a minimal change to try it out.


sfshaza2 (approved):
lgtm

PR #8794, opened 2023-05-31
Fix duplicated content  in Material 2 widget catalog

parlough:
If you visit https://docs.flutter.dev/ui/widgets/material2, you'll see all content on the page is duplicated and the second frontmatter is present as well.

This also changes the short-title to "Material 2" so it has a slightly different breadcrumb from the main Material 3 catalog.


sfshaza2 (approved):
Thanks for catching this, @parlough! lgtm

atsansone (approved):

Issue #8756, opened 2023-05-25
Correct code example that returns error on 'Adding a Flutter screen to an iOS app' page


eKaramihov:
### Page URL

https://docs.flutter.dev/add-to-app/ios/add-flutter-screen/

### Page source

https://github.com/flutter/website/tree/main/src/add-to-app/ios/add-flutter-screen.md

### Describe the problem

The example below to run a flutter engine with entry route cause:

FlutterDefaultDartEntrypoint causes Xcode error: 
```terminal
Undefined symbol: _FlutterDefaultDartEntrypoint
```

Example:
```dart
let flutterEngine = FlutterEngine()
// FlutterDefaultDartEntrypoint is the same as nil, which will run main().
engine.run(
  withEntrypoint: FlutterDefaultDartEntrypoint, initialRoute: "/onboarding")
```

leighajarett:
Yes this seems to be an issue I was able to replicate. @hellohuanlin it seems to work with specifying "main" as the entrypoint -- do you think that's okay to use in the example instead?

hellohuanlin:
Im not familiar with this, but just from the comment, `FlutterDefaultDartEntrypoint` seems to be just a placeholder, if i read it right. Can you try if passing `nil` works?

hellohuanlin:
@eKaramihov did you accidentally used `_FlutterDefaultDartEntrypoint` rather than `FlutterDefaultDartEntrypoint` in your code?

leighajarett:
I tried both @hellohuanlin  - "main" worked so I think it's fine to leave

PR #8796, opened 2023-05-31
Change the default entry point to a string for Swift

leighajarett:
_Issues fixed by this PR (if any):_ https://github.com/flutter/website/issues/8756

## Presubmit checklist

### DIFF
--- a/src/add-to-app/ios/add-flutter-screen.md
+++ b/src/add-to-app/ios/add-flutter-screen.md
@@ -690,7 +690,7 @@ FlutterViewController.
 let flutterEngine = FlutterEngine()
 // FlutterDefaultDartEntrypoint is the same as nil, which will run main().
 engine.run(
-  withEntrypoint: FlutterDefaultDartEntrypoint, initialRoute: "/onboarding")
+  withEntrypoint: "main", initialRoute: "/onboarding")
 ```
 {% sample Objective-C %}
 <!--code-excerpt "Creating engine" title-->


sfshaza2 (approved):
lgtm

src/add-to-app/ios/add-flutter-screen.md
```diff
@@ -690,7 +690,7 @@ FlutterViewController.
 let flutterEngine = FlutterEngine()
 // FlutterDefaultDartEntrypoint is the same as nil, which will run main().
 engine.run(
-  withEntrypoint: FlutterDefaultDartEntrypoint, initialRoute: "/onboarding")
+  withEntrypoint: "main", initialRoute: "/onboarding")
```
hellohuanlin, line 693: Im not familiar with this, but just from the comment, FlutterDefaultDartEntrypoint seems to be just a placeholder, if i read it right. Can you try if passing nil works?

hellohuanlin: may also wanna remove the comment above this line

hellohuanlin: actually, i may be terribly wrong - `FlutterDefaultDartEntrypoint` is an actual string that should just work: 

```
If this is FlutterDefaultDartEntrypoint (or nil); this will default to `main()`. 
```

Not sure why `FlutterDefaultDartEntrypoint` fails to work here...

PR #8801, opened 2023-06-01
Update videos.md

DemX86:
Update the title of the Maximilian Schwarzmüller video course and move it up because now it is the most actual course you can find — it was completely re-recorded from the ground in May 2023


parlough (approved):

parlough:
Thanks for updating this!

PR #8805, opened 2023-06-01
Update to latest fonts and font APIs

parlough:
At the same times, let's the fonts use `swap` to allow a first paint without waiting for the fonts.


sfshaza2 (approved):
lgtm

PR #8807, opened 2023-06-01
Simplify 'Samples & tutorials' top-level category

parlough:
The Flutter Gallery shows a lot of the relevant code within the app and has links to the source, so I think linking to the app should be sufficient and reduce unnecessary choices.

As for the "Samples apps on GitHub", it leads to a standalone site, so the "on GitHub" is a little confusing and a bit unnecessary. Especially since not all of the demos are actually hosted on the flutter/samples repo.

### DIFF
--- a/src/_data/sidenav.yml
+++ b/src/_data/sidenav.yml
@@ -47,11 +47,9 @@
 
 - title: Samples & tutorials
   children:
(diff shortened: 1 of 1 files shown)


atsansone (approved):
LGTM. Makes sense.

PR #8811, opened 2023-06-01
Move windows platform migrations alongside other migration docs

parlough:


loic-sharma (approved):
Thanks for fixing this, I'll make sure to put future migrations in the right location! 😅

parlough:
Thanks for taking a look :)

Yeah the staging has been great, I'm glad other's are appreciating it too! It does require the branch to be on `flutter/website` but everyone in flutters hackers should benefit and not have to worry about staging themselves :)

sfshaza2 (approved):
lgtm

PR #8816, opened 2023-06-01
Add categorical headers to Resources in sidenav

parlough:
The resources category was getting overwhelmingly large and hard to visually parse, but dropdowns will hide a bunch of the content.

This aims to follow a similar design to devsite and others which have non-dropdown headers. Open to feedback on the design, layout, etc. Happy to drop this altogether as well if it seems like the wrong direction.


atsansone (approved):
@parlough : I like it, but I have to ask: Why not subheads? It would curb the content overload. I'm still willing to let it pass, though. We can always revisit it.

sfshaza2 (approved):
lgtm

parlough:
> @parlough : I like it, but I have to ask: Why not subheads? It would curb the content overload. I'm still willing to let it pass, though. We can always revisit it.

I see "Resources" as a section that you might not know what you're looking for when you open. Having headers helps with the content overload while still not hiding things too deep. 

I'm thinking it will be helpful in quite a few scenarios and see this as a good place to try it out. Similar functionality is used extensively on the Firebase docs and I quite like how it turns out.

sfshaza2:
@parlough, can we land this?

PR #8840, opened 2023-06-06
Add missing code in native iOS views of Flutter Platform Views

nvbien2000:
According to issue #28124 (https://github.com/flutter/flutter/issues/28124) and Stack Overflow question 76414194 (https://stackoverflow.com/questions/76414194), a part of Swift/Objective-C code is missing that led to nil arguments which are passed from Flutter to native. Thanks for @CameronStuartSmith with Swift code & @irohitb with Objective-C code.


leighajarett:
@cyanglaz are you able to take a look at this?

src/platform-integration/ios/platform-views.md
```diff
@@ -242,6 +246,10 @@ and the platform view provides a reference to the
                              binaryMessenger:_messenger];
 }
 
+- (NSObject<FlutterMessageCodec>*)createArgsCodec {
```
cyanglaz, line 249: This protocol method is optional. Implementing is necessary when `createWithFrame`'s `arguments` parameter is not nil. See: https://github.com/flutter/engine/blob/main/shell/platform/darwin/ios/framework/Headers/FlutterPlatformViews.h#L45-L49

We should add a comment here to explain it to avoid making people believe this method is required. How about:
// Implementing this method is only necessary when the `arguments` in `createWithFrame` is not nil.

src/platform-integration/ios/platform-views.md
```diff
@@ -112,6 +112,10 @@ class FLNativeViewFactory: NSObject, FlutterPlatformViewFactory {
             arguments: args,
             binaryMessenger: messenger)
     }
+
+    public func createArgsCodec() -> FlutterMessageCodec & NSObjectProtocol {
```
cyanglaz, line 116: Ditto with my comment on the Objective-C version.

cyanglaz (commented):
Thank you for adding this! We just need to add a little comment on the method implementation to explain that when it is necessary.

src/platform-integration/ios/platform-views.md
nvbien2000: Dear @cyanglaz , I added comment as your request.

src/platform-integration/ios/platform-views.md
nvbien2000: Dear @cyanglaz, I added comment as your request.

nvbien2000 (commented):
add comment for function createArgsCodec as Expert's request

atsansone (approved):

atsansone:
@cyanglaz : Do you approve this PR?

cyanglaz (approved):
LGTM! Thanks again!

sfshaza2 (approved):
lgtm, as per @cyanglaz. Thanks!

PR #8847, opened 2023-06-06
Fix RiverPod passage

### LINKED ISSUE #8507: Fix passage about Riverpod in 'State management' page
### Page URL

https://docs.flutter.dev/development/data-and-backend/state-mgmt/options#riverpod

### Page source

_No response_

### Describe the problem

In https://github.com/flutter/website/pull/8348 recommending `Provider` was removed. However the description for Riverpod says `another good choice`. 

Since there is no `first` good choice anymore the word `another` doesn't make sense. 

### Expected fix

Personally I would list Riverpod as the new recommended approach, given it is the evolution of `Provider` which was recommended. And potentially even remove `Provider` from the list of state management approaches since it's not _really_ a state management solution without additional work, since it's just Providing data. And having it on the list _might_ confuse people.

### Additional context

_No response_

atsansone:
Fixes #8507

### DIFF
--- a/src/data-and-backend/state-mgmt/options.md
+++ b/src/data-and-backend/state-mgmt/options.md
@@ -42,9 +42,9 @@ Things to review before selecting an approach.
 
 ## Riverpod
 
-Riverpod, another good choice, is
-similar to Provider and is compile-safe and testable.
-Riverpod doesn't have a dependency on the Flutter SDK.
+Riverpod works in a similar fashion to Provider.
+It is also compile-safe and testable.
+Riverpod doesn't depend on the Flutter SDK.
 
 * [Riverpod][] homepage
 * [Getting started with Riverpod][]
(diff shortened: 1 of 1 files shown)


src/data-and-backend/state-mgmt/options.md
```diff
@@ -42,9 +42,9 @@ Things to review before selecting an approach.
 
 ## Riverpod
 
-Riverpod, another good choice, is
-similar to Provider and is compile-safe and testable.
-Riverpod doesn't have a dependency on the Flutter SDK.
+Riverpod works in a similar fashion to Provider.
+It is also compile-safe and testable.
+Riverpod doesn't depend on the Flutter SDK.
```
sfshaza2, line 47: Rewriting passive voice.

```suggestion
Riverpod, which works in a similar fashion to Provider,' is also compile-safe and testable.
Riverpod doesn't depend on the Flutter SDK.
```

sfshaza2 (approved):
Other than my one comment, lgtm

Issue #8800, opened 2023-06-01
[PAGE ISSUE]: 'Internationalizing Flutter apps'


ozbonus:
### Page URL

https://docs.flutter.dev/accessibility-and-localization/internationalization/

### Page source

https://github.com/flutter/website/tree/main/src/accessibility-and-localization/internationalization.md

### Describe the problem

In the section *Adding your own localized messages*, step 6 tells the use to run `flutter gen-l10n` or alternatively... `flutter gen-l10n` to generate files "without running the app", which doesn't make sense because neither option involves running the app. Here's the full text of the step:

> Now, run flutter `gen-l10n` so that codegen takes place. You should see generated files in `${FLUTTER_PROJECT}/.dart_tool/flutter_gen/gen_l10n`. Altenatively, you can also run `flutter gen-l10n` to generate the same files without running the app.

### Expected fix

I propose this text instead:

> Now, run `flutter run` and codegen will take place automatically. You should see generated files in `${FLUTTER_PROJECT}/.dart_tool/flutter_gen/gen_l10n`. Altenatively, you can also run `flutter gen-l10n` to generate the same files without running the app.

I believe this is closer to what the author intended to write.

### Additional context

The original text implied that codegen of localization files will take place automatically when building and running an app, and indeed that is what happens when following all of the steps up to this point and running `flutter run`.

PR #8864, opened 2023-06-09
Cleaning up internationalization page

sfshaza2:
Fixed https://github.com/flutter/website/issues/8800

But I also did a general pass through the page and cleaned it up.

### DIFF
--- a/src/accessibility-and-localization/internationalization.md
+++ b/src/accessibility-and-localization/internationalization.md
@@ -15,48 +15,44 @@ description: How to internationalize your Flutter app.
   * How to define the locales an app supports.
 {{site.alert.end}}
 
-If your app might be deployed to users who speak another language then
-you'll need to internationalize it. That means you need to write
-the app in a way that makes it possible to localize values like text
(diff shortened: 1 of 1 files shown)


src/accessibility-and-localization/internationalization.md
```diff
@@ -189,24 +190,26 @@ Widget build(BuildContext context) {
...
+To add localized text to your application,
+complete the following instructions:
 
-1. Add the `intl` package as a dependency,
-   using the version pinned by `flutter_localizations` with `any`:
+1. Add the `intl` package as a dependency:
 
    ```terminal
-   $ flutter pub add intl:any
+   $ flutter pub add intl
```
parlough, line 208: I think the `:any` portion was useful here as it is a dependency pinned by flutter/`flutter_localizations`, and apps should just pick up the dependency from there.

parlough (approved):
Nice cleanup! One comment:

parlough:
Thanks, looks good to me!

PR #8875, opened 2023-06-12
Correct MainActivity.kt with plugin-safe code

### LINKED ISSUE #7900: [PAGE ISSUE]: 'Hosting native Android views in your Flutter app with Platform Views'
### Page URL

https://docs.flutter.dev/development/platform-integration/android/platform-views/

### Page source

https://github.com/flutter/website/tree/main/src/development/platform-integration/android/platform-views.md

### Describe the problem

I would highly recommend to make a small improvement in one example. 
In the section [On the platform side](https://docs.flutter.dev/development/platform-integration/android/platform-views#on-the-platform-side) there is example of **MainActivity.kt** which overrides only one method **configureFlutterEngine**, but this method doesn't call **super()**, so if you use any plugins, they won't be registered. 
It leads to an exception, which does not explains that issue appeared because of missed **super()** call.
`Unhandled Exception: PlatformException(error, java.lang.IllegalStateException: Trying to create a platform view of unregistered type [plugin name]`

### Expected fix

```
package dev.flutter.example

import io.flutter.embedding.android.FlutterActivity
import io.flutter.embedding.engine.FlutterEngine

class MainActivity : FlutterActivity() {
    override fun configureFlutterEngine(flutterEngine: FlutterEngine) {
        super.configureFlutterEngine(flutterEngine)
        flutterEngine
                .platformViewsController
                .registry
                .registerViewFactory("<platform-view-type>", NativeViewFactory())
    }
}
```

### Additional context

_No response_

atsansone:
Fixes #7900


bparrishMines (approved):
LGTM

parlough (approved):

PR #8878, opened 2023-06-12
Fix highlighting of navigation entries in sidenav

parlough:
Since the navigation docs are within `/ui`, they were causing the earlier UI section to be indicated as active incorrectly in the sidenav, rather than the navigation docs.

**Before:**

Example of UI being highlighted instead of navigation


sfshaza2 (approved):
lgtm

PR #8879, opened 2023-06-12
Remove unnecessary version requirement specification

parlough:
There is no value in specifying this version requirement, as 1.12 is a long time ago.


sfshaza2 (approved):
lgtm

PR #8881, opened 2023-06-12
Update example created pubspec for latest changes

parlough:
Changes to support Dart 3 and includes `flutter_lints` which is included by the default `flutter create` command.


sfshaza2 (approved):
lgtm

Issue #4018, opened 2020-04-30
Code excerpt header boxes regression


xster:
Our specified header box titles are now missing

was:
<img width="751" alt="78847846-e1f98b00-79c4-11ea-9892-c56d8a3d8146" src="https://user-images.githubusercontent.com/156888/80749730-beb18f80-8adb-11ea-9045-a816f6c3879e.png">

now:
<img width="749" alt="78847860-eaea5c80-79c4-11ea-8e51-269531edbd73" src="https://user-images.githubusercontent.com/156888/80749743-c6713400-8adb-11ea-8d8b-f81da5fa6a9d.png">

Seems to have been removed by https://github.com/flutter/website/pull/3773. The headers used to describe the files to edit are very useful in project setup type instructions where the user has to edit multiple files.

micpap25:
@sfshaza2 I can help out with this. Do you know off the top of your head what file the code excerpt headers are controlled by? I can hunt around the files if you don't.

aadarshadhakalg:
I have #5026 pull reqquest on this.

sfshaza2:
@khanhnwin, is this still a valid issue or can we close?

parlough:
As far as I can tell, this functionality is still missing for snippets without code excerpts, since it was tied to that tool.

sfshaza2:
Ah. Do we care?

parlough:
I’d like to track this, but not enough for it to be p1, as the information can be portrayed other ways. Demoting to p3.

PR #8887, opened 2023-06-13
Reintroduce titles to code snippets without excerpts

parlough:
These were all commented out because at some point the code excerpt tool was changed breaking them. This reenables all usages by specifying a specific title instead of a source file so the excerpt tool won't fail because it can't find them.

In review, please check out a few locations to verify the filename properly shows up above the code snippet :)  

Closes https://github.com/flutter/website/issues/4018


atsansone (approved):
Verified on add to app pages. Looks good! Would this affect other pages as well, or only those that had a title?

parlough:
This only affects the broken ones that were trying to explicitly add a filename title.

If other excerpts want a title they can add one.

sfshaza2 (approved):
lgtm

PR #8896, opened 2023-06-14
Replace reference to `_dartobservatory._tcp` with `_dartVmService._tcp`

bkonyi:
Fixes https://github.com/flutter/flutter/issues/126838

### DIFF
--- a/src/add-to-app/ios/project-setup.md
+++ b/src/add-to-app/ios/project-setup.md
@@ -423,7 +423,7 @@ Make a copy of it called **Info-Release.plist** and add it to your Xcode project
 
 <li markdown="1">
 In **Info-Debug.plist** _only_ add the key `NSBonjourServices`
-and set the value to an array with the string `_dartobservatory._tcp`.
+and set the value to an array with the string `_dartVmService._tcp`.
 Note Xcode will display this as "Bonjour services".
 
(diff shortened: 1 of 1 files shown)


parlough:
@vashworth I don't really have the context for this, but is this something that users will need to update? Should this have a small migration guide for the next Flutter release so users are aware?

vashworth:
> @vashworth I don't really have the context for this, but is this something that users will need to update? Should this have a small migration guide for the next Flutter release so users are aware?

Once `_dartobservatory._tcp` is deprecated, users will need to update to use `_dartVmService._tcp`. Also, users on versions prior to 3.10.0, would still need to use `_dartobservatory._tcp`.

I'm not sure what we generally do in these situations. A migration guide could be the way to go. Alternatively, we could have the documentation say to add both `_dartobservatory._tcp` and `_dartVmService._tcp`. There's no harm in putting both in their plist.

parlough:
If `_dartVmService._tcp` is already supported in 3.10, I don't think we need to include `_dartobservatory._tcp` as well. 

However, I do think a really short migration guide with some extra context under the next release would be super useful - https://docs.flutter.dev/release/breaking-changes

parlough (approved):
Thanks for this, looks good to me.

Consider following up with a migration guide :)

sfshaza2 (approved):
lgtm

PR #8905, opened 2023-06-16
Add redirect for old c-interop article

parlough:
I've seen a few instances of this old page still being linked to, but not having a destination. This PR introduces a redirect for it.


sfshaza2 (approved):
lgtm

PR #8819, opened 2023-06-02
bad-smells corection

tasospal:
--cookie-notice.html

1. Use semantic HTML elements: Consider using more semantic HTML elements to improve the structure and accessibility of the code. In this case, you can use the <article> element instead of <section> to represent the cookie notice.
2. Add appropriate text alternatives to links: It's good practice to provide text alternatives using the title attribute for links that open in a new window (target="_blank"). This helps users who may use assistive technologies to understand the purpose of the link.
3. Consider adding labels to form elements: If the "Okay" button is intended to perform an action, such as accepting the cookie notice, you may want to consider wrapping it in a <form> element and adding a <label> element to provide a text description for the button. This improves accessibility and allows users to interact with the button using the keyboard.
4. Use CSS classes for styling: In the code you provided, there are CSS class names (container, btn, btn-primary) that suggest the use of CSS for styling. Ensure that these classes have associated CSS styles defined in a separate stylesheet to keep the HTML code clean and separate from presentational concerns.

--footer.html

1. Use consistent naming conventions: In the code, the CSS class names use both hyphens (site-footer__wrapper, site-footer__logo) and underscores (site-footer__link-list). It's better to use a consistent naming convention throughout the code, such as either hyphens or underscores.
2. Provide descriptive alt text for images: The  tag has an alt attribute, but the value "Flutter Logo" is not very descriptive. Ideally, the alt attribute should provide alternative text that accurately describes the content and purpose of the image for users who cannot see it.
3. Open external links in a new window: For external links, such as the link to the Flutter community in Chinese (https://flutter.cn), it's a good practice to add the target="_blank" attribute to open them in a new browser tab or window. This provides a better user experience and prevents users from navigating away from your site.
4. Ensure clear and concise link text: The link text "We stand in solidarity with the Black community. Black Lives Matter." in the last <li> element of the link list is quite long and may not be suitable as link text. Consider providing a more concise and focused description or creating a separate section for such important announcements.
5. Improve accessibility of the licenses paragraph: The <p> element with the class "licenses" contains important licensing information. To improve accessibility, consider providing additional visual cues or using proper heading levels to indicate the significance of the content.

--header.html

1. Remove unnecessary data attributes: The <div class="site-header__sheet-bg"> element has data attributes that duplicate the attributes of the parent <div> element. Since they have the same values and serve the same purpose, you can remove the duplicate data attributes to simplify the code.
[...]

PR #8946, opened 2023-06-25
Revert HTML changes from "Update template HTML with a11y fixes (#8819)"

parlough:
Reverts the HTML/template changes from #8819.

It introduced some issues, primarily not being able to close the side menu on mobile/small screens. 

There are a few minor formatting and semantic issues as well, such as incorrect line breaking in the cookie notice, some less direct text in the footer, multiple `<main>` elements on the 404 page, and not following our style in a few cases.

To fix the issues quickly, this PR reverts the changes. Improvements from the original PR can be revisited as follow-up, and should be done isolated, rather than in one big PR.

This PR does not revert the JS changes from the PR as there are conflicts due to other recent changes made and they seem to be working fine.

### DIFF
--- a/src/404.html
+++ b/src/404.html
@@ -7,25 +7,19 @@ permalink: /404.html
 ---
 
 <div class="not-found__wrapper">
-  <header>
-    <h1>404</h1>
-    <p>Sorry, we couldn’t find that page…</p>
-  </header>
-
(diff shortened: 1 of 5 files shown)


sfshaza2 (approved):
Wow. lgtm

Issue #4878, opened 2020-10-14
Fix duplication of setting PATH on 'macOS install' page


MuhammadimYusoff:
Page URL: https://docs.flutter.dev/get-started/install/macos
Page source: https://github.com/flutter/website/tree/master/src/docs/get-started/install/macos.md

Description of issue:

Guys please, at this section:
https://flutter.dev/docs/get-started/install/macos#update-your-path

Number/Step 3, why do I need to include "/flutter/bin"?
FUCKING WORKING AFTER I HAVE REMOVE IT! REMOVE IT!

Please, edit this page, I have been in HELL figuring out what the fuck is wrong with the path you guys given!

Regards.

sfshaza2:
@MuhammadimYusoff, can you provide your actual path? And what platform/version are you using? These instructions work for me, so I would like to narrow it down.

atsansone:
This results from having already set the PATH in Step 3 of https://docs.flutter.dev/get-started/install/macos#get-sdk. As that was already set, this section addition caused an issue. We will refactor this to fix the issue.

sfshaza2:
Actually, the instructions are correct. However, I've updated the text to (hopefully) make it more clear. The first time, it tells you how to update the path for the current terminal session. But when you open a new terminal session, the path reverts to the previous value. The second set of instructions (in Update your path), explain how to permanently update the path.

atsansone:
Step 3 could use `echo $PATH` with a result after it to make it clear what the user should see. This would be consistent with GAR requirements.

Something like:

```terminal
$ export PATH="$PATH:`pwd`/flutter/bin"
$ echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/home/user/development/flutter/bin:/home/user/development/dart-sdk/bin:/snap/bin
```

PR #8951, opened 2023-06-26
Clarifying the update your path instructions

sfshaza2:
Fixes https://github.com/flutter/website/issues/4878

While I was at it, I removed all non-essential uses of "may" with either "might" or "could", as was appropriate. This is in accordance with the GDSG.


atsansone (changes requested):
Please separate this into two PRs. I had a really difficult time seeing the change you made against the original issue. Thanks!

sfshaza2:
@atsansone, that would be an enormous amount of work. Perhaps focus on the install files, which are where the actual work is.

atsansone:
Not too much work. Just tested this.

```terminal
gh pr checkout 8951                                           // Requires GitHub CLI
git checkout -b fix-might upstream/set-path                   // Make new branch from PR
git reset --soft HEAD^                                        // Return all files to stage
git restore --staged src/get-started/install/_path-mac.md     // Remove file from stage
git restore --staged src/get-started/install/_get-sdk-mac.md  // Remove file from stage
git checkout -- .                                             // Clear unstaged files
git commit -m "Update may to might"                           // Create new commit
git push origin fix-might                                     // Push new commit
git checkout set-path                                         // Checkout the PR again
git reset                                                     // Unstage all files in the PR
git add src/get-started/install/_*mac.md                      // Stage the macOS files
git checkout -- .                                             // Remove the other unstaged files
git commit --amend                                            // Create new commit on original PR
git push origin set-path --force-with-lease                   // Push new commit to original PR
```

atsansone (approved):
Let's land it. If we need to revisit, we can.

sfshaza2:
Thanks, Tony. I'll try not to do that to you again. It's definitely part of my OCD.  :D

PR #8969, opened 2023-07-01
Remove `flutter format` mention now that it is no longer functional

parlough:
Remove `flutter format` mention now that it is no longer functional and is set to be removed in the next Flutter release.


atsansone (approved):

sfshaza2 (approved):
lgtm

Issue #8983, opened 2023-07-05
Replace usages of Font Awesome with Material Icons


parlough:
### Describe the problem

We currently use a mix of Font Awesome and Material Icons causing visual inconsistency. Besides that, the way we use Font Awesome results in a huge JS file slowing down page load.

### Expected fix

Replace usages of Font Awesome with material icons.

### Additional context

dart.dev finalized equivalent work in https://github.com/dart-lang/site-www/commit/c95e61b96ad27bd04de01fb913a416fb5e3ea702

ulisseshen:
Hello @parlough how I can help you with this? I find this kind of work exciting.

parlough:
I appreciate the offer @ulisseshen, but the only remaining place to update is the [Get started: Install](https://docs.flutter.dev/get-started/install) page and @atsansone has the assets necessary. So it'll probably be up to him.

PR #8995, opened 2023-07-08
Remove most uses of font awesome

parlough:
Contributes to https://github.com/flutter/website/issues/8983

The logo SVGs are the same used in the dart.dev footer.

The only remaining usages of Font Awesome after this is on the [Get started: Install page](https://docs.flutter.dev/get-started/install), which will need more work to find suitable replacements for.


atsansone (approved):
@parlough : Good work. Let me know if you want or need help on the Getting Started section. The performance score on Lighthouse went up slightly. Would recommend that you run your own Lighthouse tests on this and the current page to see where and how performance improved.

parlough:
Thanks for taking a look!

I don't expect scores to go up much until I fully remove Font Awesome alongside the Get Started changes. I'll make sure to measure loading differences before and after that work though.

PR #9021, opened 2023-07-11
Added China notice to all download related pages

### LINKED ISSUE #3594: Add another warning in archive download pages to go to the chinese mirror when in china
There's a china warning in https://flutter.dev/docs/get-started/install

but in https://flutter.dev/docs/get-started/install/windows and https://docs.flutter.dev/release/archive where the actual archive SDK zips are linked, we should also add another warning to redirect to https://flutter.cn/docs/development/tools/sdk/releases where the zips are in faster mirrors

atsansone:
Fixes #3594

Staged views

image

image


sfshaza2 (approved):
lgtm

PR #9039, opened 2023-07-12
Suggest cc'ing ian@hixie.ch for now

Hixie:
Our forwarding is broken due to Google Groups being more aggressive about spam control. While we sort this out, we want to make sure we don't lose reports.

## Presubmit checklist


sfshaza2 (approved):
lgtm

Issue #8985, opened 2023-07-05
Migrate styles to SASS module system


parlough:
### Describe the problem

`@import` is deprecated in favor of `@use` in Dart Sass (which we are now using) as part of their new [module system](https://sass-lang.com/blog/the-module-system-is-launched/).

### Expected fix

Migrate to `@use` instead of `@import`.

### Additional context

dart.dev did so in https://github.com/dart-lang/site-www/pull/4974

parlough:
Finishing this will make a bunch of follow-up style changes and improvements much easier to implement and review, so increasing this to P1.

Issue #8030, opened 2023-01-02
☂ Migrate to Bootstrap 5 or remove it


parlough:
### Describe the problem

Bootstrap 5 is the current release Bootstrap, replacing Bootstrap 4. We use it heavily across the site and we want to make sure we stay up to date. This will also allow us to eventually drop Jquery since Bootstrap 5 no longer uses it. Beyond that, this will also make a dark mode slightly easier with its (not yet released) [color mode functionality](https://getbootstrap.com/docs/5.3/customize/color-modes/).

### Expected fix

We should migrate away from Bootstrap 4 to 5. 

### Additional context

There may be some incompatibilities due to Bootstrap now using the Dart sass compiler which supports some newer features. We may have to configure Jekyll to use [sass-embedded](https://rubygems.org/gems/sass-embedded) instead somehow, which is implemented with Dart sass :)

PR #9058, opened 2023-07-17
Migrate styles to SASS modules

parlough:
`@import` is being deprecated in favor of `@use` in Dart Sass (which we are now using). Read more within Sass's [`@import` documentation](https://sass-lang.com/documentation/at-rules/import).

Contributes to https://github.com/flutter/website/issues/8030 since it will make that migration easier, with all bootstrap usages namespaced. Also enables better code analysis and auto-completion in IDEs.

Fixes https://github.com/flutter/website/issues/8985


atsansone (approved):
@parlough : This makes sense. I'll re-read the @use docs, but I get where this is going. Good job.

sfshaza2 (approved):
lgtm

PR #9059, opened 2023-07-17
Cleanup and adjust install help links

parlough:
- Don't use relative links (which break when staging and on some translated versions of the site)
- Don't use inline styles
- Use material icons instead of Font Awesome (https://github.com/flutter/website/issues/8983)

**Staged example:** https://flutter-docs-prod--pr9059-misc-cleanup-install-jrqqvfer.web.app/get-started/install/windows#get-the-flutter-sdk

### DIFF
--- a/src/_sass/base/_base.scss
+++ b/src/_sass/base/_base.scss
@@ -226,6 +226,24 @@ dd {
   }
 }
 
+.install-help {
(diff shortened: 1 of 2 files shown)


sfshaza2 (approved):
lgtm

PR #9060, opened 2023-07-17
Open Provenance information in new tab

parlough:
This opens provenance links in the archive table in a new tab to not navigate users away from the site.


sfshaza2 (approved):
lgtm

Issue #1778, opened 2018-11-27
Validate `permalink` fields in sidenav.yml file as part of a build process


chalin:
Followup to https://github.com/flutter/website/pull/1777#issuecomment-442132142.

There is a quick and easy implementation that we could do in Liquid, but that will slow down the build process (since the sidenav is rebuilt for each page). An out-of-band check would be more efficient.

Issue #9013, opened 2023-07-11
Incorporate high-quality cookbook recipes into sidenav when appropriate


parlough:
- [x] #9073
- [x] Incorporate maintenance cookbooks in the sidenav (https://github.com/flutter/website/pull/9125)
- [x] Incorporate navigation cookbooks in the sidenav (https://github.com/flutter/website/pull/9109)
- [x] Incorporate networking cookbooks in the sidenav (https://github.com/flutter/website/pull/9061, https://github.com/flutter/website/pull/9127)
- [x] Incorporate persistence cookbooks in the sidenav (https://github.com/flutter/website/pull/9128)
- [x] Incorporate testing cookbooks in the sidenav (https://github.com/flutter/website/pull/9125)

---

This should mostly until https://github.com/flutter/website/issues/9011 is completed:

- [x] #9140 (https://github.com/flutter/website/pull/9147)
- [x] #9141 (https://github.com/flutter/website/pull/9153)
- [x] #9142 (https://github.com/flutter/website/pull/9147, https://github.com/flutter/website/pull/9148)
- [x] #9143 (https://github.com/flutter/website/pull/9150)
- [x] #9144 (https://github.com/flutter/website/pull/9148)
- [x] Incorporate images cookbooks in the sidenav (https://github.com/flutter/website/pull/9130)
- [x] #9145
- [x] Incorporate plugins cookbooks in the sidenav (https://github.com/flutter/website/pull/9130)

PR #9061, opened 2023-07-17
Incorporate networking cookbooks in sidenav with new tooling

parlough:
Introduces new tooling for sidenav active entry calculation (which entry is open and should be highlighted). Previously it was one pass, so only found the first entry which begins with the same path. So cookbook entries found `/cookbook` under Samples and tutorials. The new functionality generates a sort of tree structure which finds the closest path but allows fallback to the next closest. So `/cookbook/networking/fetch-data` will prefer that, then `/cookbook/networking`, then `/cookbook`. This process is a little more complicated, but determining activenav data for each page is much faster, scaling with the path rather than how many sidenav entries there are.

As a demo and first use of this functionality, add some of the networking cookbooks under **Data & backend > Networking** in the sidenav. It also adjusts their prev -> next order to better match a potential learning workflow. 

Also implements some validation of the `sidenav.yml` file and regenerates all pages when the sidenav is updated.

---

Closes https://github.com/flutter/website/issues/9073, contributing to https://github.com/flutter/website/issues/9013
Closes https://github.com/flutter/website/issues/1778 by adding validation to permalink entries in the sidenav data


parlough:
I'd still like to add to and adjust the comments to clarify the behavior of the new tooling better, but this is ready for initial review :)

sfshaza2 (commented):
I love this! But, maybe remove the "next" button on the last recipe? Clicking that took me to the JSON recipe, which is the next recipe, but the sidenav was pretty lost.

sfshaza2 (approved):
I see you removed the next and prev links, so lgtm!

InMatrix:
This change looks great to me. Thanks for making it work.

johnpryan (approved):
LGTM!

Issue #8889, opened 2023-06-13
Migrate our js files away from using jquery


parlough:
### Describe the problem

We'd eventually like to stop using jquery, especially since Bootstrap 5 ([which we hope to migrate to](https://github.com/flutter/website/issues/8030)) no longer requires it. 

### Expected fix

Migrate all code in https://github.com/flutter/website/tree/main/src/assets/js away from using jquery.

parlough:
If anyone wants to help with this, please keep your pull requests simple. Migrate a few related functions, or if a file is simple, just one file. Thanks so much!

ulisseshen:
It may not be an easy job, but I appreciate it very much and I can try to help with that. However, I have no previous experience collaborating with open source projects.

parlough:
We always appreciate help! I'm happy to help with any issues you have as well.

The repository's [README](https://github.com/flutter/website/tree/main#flutter-website) has some information on contributing and setting up the site. 

Usually contributing looks like some mix of [forking the repository](https://docs.github.com/en/get-started/quickstart/fork-a-repo), making and testing your changes in a [local branch](https://gist.github.com/Chaser324/ce0505fbed06b947d962#doing-your-work), then pushing them to GitHub and opening a [pull request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request). From there one of us will review it and provide feedback in a back and forth process.

As for this particular issue, perhaps a good place to start would be a single function (or even a part of one) in the [`archive.js`](https://github.com/flutter/website/blob/main/src/assets/js/archive.js) folder. It is used on the [Flutter SDK archive page](https://docs.flutter.dev/release/archive) for populating the archive tables. There are a lot of good documents online about migrating away from jquery to normal JavaScript.

If you'd like to work on that, go for it, and let us know if you have any questions about anything or something doesn't work.

ulisseshen:
I already hosted this project locally, I have no problem with it. A good starting point could be this [archive.js](https://github.com/flutter/website/blob/main/src/assets/js/archive.js) file you mentioned.

My questions now are:
How can I relate my changes to this topic?
Will Jquery's ajax calls be replaced by the fetch api?

PS: Good initiative. As awesome as it is, the Jquery era has long since passed.

parlough:
@ulisseshen 

> How can I relate my changes to this topic?

[...]

PR #9063, opened 2023-07-18
Finished migrate archive.js file to pure js

ulisseshen:

Related with #8889 and #9022 

## Presubmit checklist


parlough:
Thanks for finishing migrating this file. I'm excited to take a look!

Can you handle resolving the conflicts with the `main` branch?

ulisseshen:
> Thanks for finishing migrating this file. I'm excited to take a look!
> 
> Can you handle resolving the conflicts with the `main` branch?

I will settle this now. I didn't mind when I saw this conflict notice because I'm new to contributions, and I thought it wouldn't affect.

src/assets/js/archive.js
```diff
@@ -122,66 +125,93 @@ function createTableCell(content, dataClass) {
...
+  });
 
   // Update inlined filenames in <code> element text nodes:
-  const fileNamePrefix = 'flutter_';
-  const code = $(`code:contains("${fileNamePrefix}")`);
-  const textNode = $(code).contents().filter(function () {
-    return this.nodeType === 3 && this.textContent.includes(fileNamePrefix);
+  const filteredElements = filterCodeElements();
+  filteredElements.forEach(function (node) {
+    const newText = node.textContent.replace(new RegExp(`^(.*?)\\b${FILE_NAME_PREFIX}\\w+_v.*`), `$1${archiveFilename}`);
```
parlough, line 206: This regex doesn't seem to be working to replace the filenames on the install pages. I think because using `textContent` now combines it all and the part to be replaced is on the second line. Adding the [multiline flag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions#advanced_searching_with_flags) (`m`) should get it working.

```suggestion
    const newText = node.textContent.replace(new RegExp(`^(.*?)\\b${FILE_NAME_PREFIX}\\w+_v.*`, 'm'), `$1${archiveFilename}`);
```

src/assets/js/archive.js
```diff
@@ -122,66 +125,93 @@ function createTableCell(content, dataClass) {
...
-  $(textNode).replaceWith(`unzip ~/Downloads/${macOSArm64ArchiveFilename}`);
-});
 
-/*
-releases: A list of Flutter releases
-base_url: link for sdk download link such as storage.googleapis.com...
-os: macos, windows, or linux
-[optional] arch: Only specify if there's additional architecture, such as arm64
-*/
+/**
```
parlough, line 167: I really appreciate you adding these JSDocs!

src/assets/js/archive.js
```diff
@@ -122,66 +125,93 @@ function createTableCell(content, dataClass) {
...
-  const textNode = $(code).contents().filter(function () {
-    return this.nodeType === 3 && this.textContent.includes(fileNamePrefix);
+(() => {
+  document.querySelector('.download-latest-link-macos-arm64')
+    .addEventListener('click', function () {
+      // Update inlined filenames in <code> element text nodes with arm64 filename:
+      const filteredElements = filterCodeElements();
+
+      filteredElements.forEach(function (node) {
+        node.textContent = `unzip ~/Downloads/${macOSArm64ArchiveFilename}`;
```
parlough, line 146: This seems to replace the entire code block, when we only want to replace the filename. Perhaps we can reuse some of the regex logic from below setting the filename the first time.

**Before clicking:**

Code block with all text

**After clicking (with this change):**

Code block with all text replaced

**After clicking (expected):**

Code block on live docs.flutter.dev

Let me know if you need any help or pointers fixing this one.

src/assets/js/archive.js
```diff
@@ -244,16 +278,28 @@ function getProvenanceLink(os, release, date, channel) {
 
 
 // Send requests to render the tables.
-$(function () {
-  if ($('#sdk-archives').length) {
+(() => {
```
parlough, line 281: Nice idea to maintain the previous behavior with [IIFEs](https://developer.mozilla.org/en-US/docs/Glossary/IIFE).

parlough (changes requested):
Thanks again for working on this! Overall really clean migrations. 🥳 

When testing I found a few small issues. Let me know if I made a mistake or you want to discuss any of them further.

src/assets/js/archive.js
ulisseshen: I found the reason for this and I will work on this tomorrow. Tnks

src/assets/js/archive.js
```diff
@@ -122,66 +125,93 @@ function createTableCell(content, dataClass) {
...
 // Listen for the macOS arm64 download link to be clicked and update
 // the example unzip command with correct arm64 filename
-$('.download-latest-link-macos-arm64').click(function () {
-  // Update inlined filenames in <code> element text nodes with arm64 filename:
-  const fileNamePrefix = 'flutter_';
-  const code = $(`code:contains("${fileNamePrefix}")`);
-  const textNode = $(code).contents().filter(function () {
-    return this.nodeType === 3 && this.textContent.includes(fileNamePrefix);
+(() => {
+  document.querySelector('.download-latest-link-macos-arm64')
```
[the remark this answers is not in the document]
ulisseshen: Already solved!

src/assets/js/archive.js
```diff
@@ -122,66 +125,93 @@ function createTableCell(content, dataClass) {
 }
 
 function updateTableFailed(os) {
-  const tab = $(`#tab-os-${os}`);
-  tab.find('.loading').text('Failed to load releases. Refresh page to try again.');
+  const tab = document.getElementById(`tab-os-${os}`);
+  const loadingElements = tab.querySelectorAll('.loading');
```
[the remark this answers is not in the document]
ulisseshen: Done

src/assets/js/archive.js
```diff
@@ -122,66 +125,98 @@ function createTableCell(content, dataClass) {
...
+      return node.nodeType === Node.TEXT_NODE && node.textContent.includes(FILE_NAME_PREFIX);
+    });
+  });
+
+  const code = filteredElements[0];
+  const textNode = [...code.childNodes].filter(function (node) {
+    return node.nodeType === Node.TEXT_NODE && node.textContent.includes(FILE_NAME_PREFIX);
+  })[0];
+  return textNode;
+}
```
parlough, line 171: Your new solution looks pretty good and is clever, however it causes a failure if one of the pages with a download button doesn't have a `code` element with the filename, like the [Windows install page](https://docs.flutter.dev/get-started/install/windows/) doesn't. Beyond that, it will only support replacing one instance of the filename. 

How about we take a simpler approach that replaces all instances, and uses a modified regex that supports replacing both the placerholder text and existing filenames. Something like this:

```js
const filenameReplacement = new RegExp(`^(.*?)\\b${FILE_NAME_PREFIX}\\w+_v?[X|0-9]+\\..*`, 'm');

/**
 * Replaces the placeholder text or the old filename in code blocks
 * with the specified {@link archiveFilename}.
 * 
 * @param archiveFilename The new filename to replace the
 *   old one in code blocks with
 */
function replaceFilenameInCodeElements(archiveFilename) {
  const codeElements = document.querySelectorAll('code');

  codeElements.forEach((e) => {
    e.childNodes.forEach((node) => {
      if (node.nodeType === Node.TEXT_NODE &&
          node.textContent.includes(FILE_NAME_PREFIX)) {
        const text = node.textContent;
        node.textContent = text.replace(
            filenameReplacement,
            `$1${archiveFilename}`
        );
      }
    });
  });
}
```

Then we would only need to call this method in the cases that use it.

parlough (changes requested):
Thanks so much for making those adjustments. Mostly looks good!

I noticed one remaining issue though, and left a potential solution.

src/assets/js/archive.js
ulisseshen: My doubt is where I will call this method. I will make some tests

src/assets/js/archive.js
```diff
@@ -122,66 +125,98 @@ function createTableCell(content, dataClass) {
...
-  const textNode = $(code).contents().filter(function () {
-    return this.nodeType === 3 && this.textContent.includes(fileNamePrefix);
-  });
-  const text = $(textNode).text();
-  const newText = text.replace(new RegExp(`^(.*?)\\b${fileNamePrefix}\\w+_v.*`), `$1${archiveFilename}`);
-  $(textNode).replaceWith(newText);
+  const textNode = filterTextNodeInCodeElement();
+  const text = textNode.textContent;
+  const newText = text.replace(new RegExp(`^(.*?)\\b${FILE_NAME_PREFIX}\\w+_v.*`, 'm'), `$1${archiveFilename}`);
+  textNode.textContent = newText;
```
parlough, line 213: @ulisseshen Here's one place you would consider calling the suggested method, replacing this code and passing in `archiveFilename`. The other would be for the macos arm64 code on L146-149.

This will allow the function to handle the finding and replacing, making this area a bit cleaner.

ulisseshen: I solved this with your solution.

ulisseshen:
changes are made! I was unable to assing this to review.

parlough (approved):
Thank you so much for all of your work on this! The changes look and work great.

ulisseshen:
I feel much happy for contribute on this project, in this migrate task in particular. 

I'll continue work on this. @parlough you can help me guided me in which another file I can work on.

parlough:
Glad to hear that. Thanks for your continued help!

I haven't taken a close look at the code, but consider checking out [`main.js`](https://github.com/flutter/website/blob/main/src/assets/js/main.js) and converting some of it. There might be some code that interacts with bootstrap that is a challenge to convert, so feel free to skip those if needed. Always feel free to reach out with questions too!

PR #9088, opened 2023-07-21
Change the example code to match the code description

DemX86:
The code description says:
> Put the text in a `Container`

but the code itself uses a `Padding` widget


sfshaza2 (commented):
This looks fine to me but, @DemX86, you have to run tools/refresh-code/excerpts. That's why the build is breaking.

DemX86:
I am sorry but I have failed to run `tool/refresh-code-excerpts.sh` script — it doesn't work on my WSL installation.
I have updated the code example `examples/layout/lakes/step4/lib/main.dart` manually with second commit but with no luck for test passing.
Can anyone help me and run the script on your machine?

parlough:
I'm super sorry about the issues with WSL and windows in general. I hope to look further into the compatibility issues once I get a chance. I appreciate all of the fixes and help you provide despite those challenges though.

I'd be happy to regenerate the snippets for you. Looks like we'll need to adjust a few more source examples for other steps first though to be consistent: [step5](https://github.com/flutter/website/blob/main/examples/layout/lakes/step5/lib/main.dart) and [step6](https://github.com/flutter/website/blob/main/examples/layout/lakes/step6/lib/main.dart).

Would you like to update those source `.dart` files or would you prefer I do?

Thanks again for your help!

DemX86:
@parlough Thanks for help! I didn’t guess that it was necessary to fix the example code of the rest steps too

domesticmouse:
In this case I feel like the text is correct, using a container for padding is inefficient. Maybe updating the text to agree with the code would be a better fix.

DemX86:
> In this case I feel like the text is correct, using a container for padding is inefficient. Maybe updating the text to agree with the code would be a better fix.

This option also exists.

I think that way: the article introduces to reader only the `Container` widget — it's already used prior Step 4 and has a mention about in the text — the tutorial has no word about the `Padding` widget in the description. So I decided that whole article uses a `Container` widget to control the padding and margin in the layout.

domesticmouse:
What I meant to say is that the text is incorrect. 

But yes, this is a major rewrite to correct. I'm happy to see this PR landed so the code and text agree, but we do need to take an action item to update it to reflect the fact that the code, as written, will probably generate linter warnings for the ineffeciency of using a container when a padding widget does the job much more efficiently =)

domesticmouse (approved):

parlough (approved):
@domesticmouse Is right and we should consider modifying the text to introduce `Padding` in that section, but matching the text with the code seems good to do for now. I opened https://github.com/flutter/website/issues/9118 to track that improvement.

Thanks again for this and fixing those build failures!

Issue #9098, opened 2023-07-22
Find a better home for 'Background processes' page


parlough:
### Page URL

https://docs.flutter.dev/packages-and-plugins/background-processes/

### Page source

https://github.com/flutter/website/tree/main/src/packages-and-plugins/background-processes.md

### Describe the problem

The page doesn't belong under "Packages & plugins".

parlough:
@sfshaza2 I'm thinking to keep this page out of the sidenav for now. It just links to other resources and I think it introduces some confusion since it intermixes two topics. That of isolates and that of persistent background processes (mostly on mobile operating systems).

What do you think about keeping it out of the sidenav until we improve the page? Tracked in https://github.com/flutter/website/issues/1986

atsansone:
@parlough : Maybe **Data and Backend** would be more consistent? Many of these articles cover asynchronous activity.

PR #9104, opened 2023-07-23
Reorganize 'Packages & plugins' section

parlough:
This does temporarily remove "Background processes" from the sidenav, but I'm tracking reincorporating it in https://github.com/flutter/website/issues/9098.


sfshaza2 (approved):
You're on fire, sir!

PR #9106, opened 2023-07-24
Temporarily remove cookbook recipe prev->next links

parlough:
After reviewing them, many of the orders don't necessarily make sense. Beyond that, the cookbook recipes are most useful when learning a specific thing rather than a sequential order of experiences, hence the idea of them being standalone recipes.

Anyway, maintaining them will make the next phase of Information Architecture work much more challenging to implement and review with extra changes and consideration surrounding them.

This PR removes them, with the option to revisit when the IA work is done, so we can more thoughtfully approach a prev->next order.


sfshaza2 (approved):
woah! lgtm

PR #9107, opened 2023-07-24
Rename tools category and locate with other reference docs

parlough:
I decided to rename the category to "Tools & editors", but let me know if you think just "Tools" or "Tooling" is better :) 

Closes https://github.com/flutter/website/issues/9108

### DIFF
--- a/src/_data/sidenav.yml
+++ b/src/_data/sidenav.yml
@@ -373,7 +373,9 @@
     - title: Loading sequence and performance
       permalink: /add-to-app/performance
 
-- title: Tools & features
+- divider
+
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
I guess I'm ok with that, though editors are tools. One good thing is that it makes it easier for the brain to see it, with the longer text. I can live with it.

parlough:
> I guess I'm ok with that, though editors are tools. One good thing is that it makes it easier for the brain to see it, with the longer text. I can live with it.

The longer text helped me too. Let's try it out then since it's super easy to change that portion back :)

PR #9111, opened 2023-07-24
"dismissForMinutes" instead of Days

jayoung-lee:
`dismissForDays` --> `dismissForMinutes`


jayoung-lee:
Note: We'll move this file to `flutter/uxr` repo soon!

PR #9121, opened 2023-07-25
Enable new text wrap balance for page headers

parlough:
A [new option](https://developer.chrome.com/blog/css-text-wrap-balance/) (mostly for headers) to automatically introduce line breaks in text under 4 lines when it makes sense. It only works in Chromium-based browsers for now.

**After:**
Example header after text-wrap: balance change

**Before:**
Example header before text-wrap: balance change

**I recommend** visiting a [few pages](https://flutter-docs-prod--pr9121-misc-enable-new-text-owy06646.web.app/) and seeing how you feel about the results. This is a rather subjective change.


sfshaza2 (approved):
lgtm

PR #9126, opened 2023-07-25
Reduce margin below page headers

parlough:
Slightly reduces the margin below h1 headers and breadcrumbs to get to the content faster. This makes it not as small as dart.dev's, but still smaller than we have now.


sfshaza2 (approved):
lgtm

PR #9154, opened 2023-07-27
Propose another default install path for Windows

AlexV525:
There are many file access exception issues with Flutter that are caused by the inaccessible path for the Flutter installation on Windows. The PR changes the default path which should be available for users most of the time.

## Presubmit checklist


AlexV525:
image

atsansone (approved):

PR #9164, opened 2023-07-28
Add /ui exception for mini-toc generation

parlough:
This allows index pages linked to from the breadcrumbs to be generated properly, such as the text index page. They were empty since the generation didn't account for `/ui` having multiple entries in the sidenav.

**Before:** https://docs.flutter.dev/ui/design/text

**After:** https://flutter-docs-prod--pr9164-fix-9156-yw9b52uw.web.app/ui/design/text

### DIFF
--- a/src/ui/design/index.md
+++ b/src/ui/design/index.md
@@ -1,6 +1,6 @@
 ---
 layout: toc
-title: Design
+title: Design & theming
 description: Content covering designing Flutter apps.
 sitemap: false
 ---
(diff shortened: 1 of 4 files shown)


src/ui/design/index.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: toc
-title: Design
+title: Design & theming
```
atsansone, line 3: **nit:** This would be better phrased as `Design & Themes` if design was intended as a noun. I would still recommend that.
```suggestion
title: Design & themes
```

atsansone (approved):
LGTM with one nit.

src/ui/design/index.md
parlough: Design is intended as a verb here. It's not about the designs of your app, but about building that.

Issue #1950, opened 2018-12-08
Add Top button for easily reaching the top page


mshwf:
Page URL: https://docs.flutter.dev/get-started/flutter-for/xamarin-forms-devs
Page source: https://github.com/flutter/website/tree/main/src/get-started/flutter-for/xamarin-forms-devs.md

Description:
Please add Top button, it's extremely important for long web pages, when reading from cell phones. Instead of running my fingers on the screen.

chalin:
@galeyang @InMatrix - I've been increasingly testing the site from mobile as well, and I've felt this need too, especially because several site pages are very long. Any feedback concerning this?

As you might recall, I had implemented (via #1271) a "scroll to top" button based on the webdev site implementation, but we decided against it because it wasn't in the original design (cf. #1272).

IIRC Daniel had suggested that if we choose implement such a feature, it should be a floating FAB in the lower right corner.

cc @maguinis @kwalrath

InMatrix:
Sounds reasonable to me. It would be mostly for Android phone users though, since on iOS you could just tap the top of the screen. It would be nice if we can avoid showing this button on Mobile Safari.

galeyang:
Agreed with this issue. The proposal of floating button in lower right corner sounds good because we won't able to see fixed table of content on a narrow screen, like mobile.

sfshaza2:
I like this idea, though I'm not sure @chalin has time for this before he leaves.

InMatrix:
@galeyang and I are assigned to the issue, but I think fixing this issue requires a developer. I will unassigned Gale and I for now. Hope we will have resource to add this simple feature on mobile soon.

parlough:
As part of the new mobile version of the table of contents ("On this page") implemented in https://github.com/flutter/website/issues/11750, you can now quickly navigate to the top of the page (or any other header) with a button in the opened dialog. 

<img width="397" alt="Mobile TOC opened in layout tutorial" src="https://github.com/user-attachments/assets/2d56d1a5-1abb-4fe5-a117-8d1b23c4a3f9" />

Thanks for the feedback and for all the discussion here.

PR #9167, opened 2023-07-29
Begin to migrate `main.js` from JQuery to pure js

ulisseshen:

Related with https://github.com/flutter/website/issues/8889

Initialy I just migrated the ` adjustToc` function and made some comments in the code, so you can help me to understand the usage of the code in question.

## Presubmit checklist


sfshaza2:
@craiglabenz

craiglabenz:
The website still loads and builds for me, but I don't know what behavior this code was/is powering, so I can't visually confirm that everything still works as before.

Could I bug you to add screenshots, @ulisseshen?

src/assets/js/main.js
```diff
@@ -79,20 +79,55 @@ function scrollSidebarIntoView() {
...
+function adjustTOC() {
 
   const tocId = '#site-toc--side';
 
   const tocHeader = document.querySelector(tocId + ' header');
 
   if (tocHeader) {
-    tocHeader.addEventListener('click', function (_) {
-      $('html, body').animate({scrollTop: 0}, 'fast');
+    //Shouldn't it be a "to top" button?
```
ulisseshen, line 98: This kind of code `scrollToTop` is actioned on click of TOC header

In the below imagem, you will need to click on the header, or `Contents` to do this work.
image

Shoud it be a ^ `to top` button?

ulisseshen (commented):
I made a review and I put a brief print.

craiglabenz (approved):
Manually confirmed this still works (now that I know what it does in the first place). LGTM

src/assets/js/main.js
parlough: You are right it's not intuitive and it won't help on mobile where the TOC isn't visible lower down in mobile layouts. The ideal would be a floating button in the bottom right once users have scrolled down a little bit. We're tracking improved functionality in https://github.com/flutter/website/issues/1950 which would be nice to have if you are interested in working on that.

For now, you can remove these comments since we're tracking the improvement in that issue :)

src/assets/js/main.js
```diff
@@ -85,20 +85,55 @@ function scrollSidebarIntoView() {
   }
 }
 
-function adjustToc() {
-  // Adjustments to the jekyll-toc TOC.
+/**
+ * Adjusts the behavior of the table of contents (TOC) on the page.
+ * 
+ * This function enables a "scrollspy" feature on the TOC, where the active link in the TOC is updated
+ * based on the currently visible section in the page.
```
parlough, line 92: To shorten this a bit:

```suggestion
 * This function enables a "scrollspy" feature on the TOC, 
 * where the active link in the TOC is updated
 * based on the currently visible section in the page.
```

src/assets/js/main.js
```diff
@@ -85,20 +85,55 @@ function scrollSidebarIntoView() {
...
+ * 
+ * Enables a "back to top" button in the TOC header.
+ * returns {void}
```
parlough, line 95: Thanks for these JS docs! I think it's fine if we don't specify void:

```suggestion
```

src/assets/js/main.js
```diff
@@ -85,20 +85,55 @@ function scrollSidebarIntoView() {
...
+ */
+function adjustTOC() {
```
parlough, line 98:
```suggestion
```

src/assets/js/main.js
```diff
@@ -85,20 +85,55 @@ function scrollSidebarIntoView() {
...
+    tocHeader.addEventListener('click', (_) => {
+      _scrollToTop();
     });
   }
 
-  $('body').scrollspy({ offset: 100, target: tocId });
+  //Appear not to be used anywhere
+  //'class' active dont make any effect on toc
+  const tocElement = document.querySelector(tocId);
+  if(tocElement){
```
parlough, line 114:
```suggestion
  if (tocElement){
```

src/assets/js/main.js
```diff
@@ -85,20 +85,55 @@ function scrollSidebarIntoView() {
...
+    window.addEventListener('scroll', _spyOnScroll);
+  }
+
+  function _scrollToTop() {
+    const distanceBetweenTop = document.documentElement.scrollTop || document.body.scrollTop;
+    if (distanceBetweenTop > 0) {
+      window.requestAnimationFrame(_scrollToTop);
+      const divisorToMakeAnimationFast = 4;
+      window.scrollTo(0, distanceBetweenTop - distanceBetweenTop / divisorToMakeAnimationFast);
+    }
```
parlough, line 124: Cool, this is a clever solution!

However, since the functionality is pretty obscure, what do you think using the browser's built-in smooth scrolling functionality? That way the code is easier to understand.

[According to MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollTo) something like the following should work:

```js
window.scrollTo({
  top: 0,
  behavior: 'smooth'
});
```

If you prefer your current solution, I'm happy with that too :)

src/assets/js/main.js
```diff
@@ -85,20 +85,55 @@ function scrollSidebarIntoView() {
...
+    //Shouldn't it be a "to top" button? 
+    //Its doesn't make sense, it's not intutive.
+    tocHeader.addEventListener('click', (_) => {
+      _scrollToTop();
     });
   }
 
-  $('body').scrollspy({ offset: 100, target: tocId });
+  //Appear not to be used anywhere
+  //'class' active dont make any effect on toc
```
parlough, line 112: On the live site, such as on https://docs.flutter.dev/ui/layout, it highlights the currently visible header on the page in grey instead of the default blue within the TOC. 

The scrollspy functionality to do this comes from bootstrap and I think for now it's fine to keep using it through jquery since our TOC is built with that in mind. Perhaps just add a comment explaining this.

We have to keep jquery until we migrate away from Bootstrap 4 anyway (https://github.com/flutter/website/issues/8030), so we can migrate this usage once we do that.

parlough (changes requested):
Thanks for the PR and sorry for the delay. Some comments and a suggestion :)

src/assets/js/main.js
ulisseshen: I don´t used behavior smooth, because I traiedmantain the same velocity that Jquery providing, but I believe that the default behavior is more clean and make the same job in a half of velocity, what is not significantly importantly. I'll apply your suggestion.

src/assets/js/main.js
ulisseshen: I can help with the issue #1950, taking advantage that I'm already work on that in this moment.

PS: I'll remove comments, and in the future I'll make them in a review.. looks more profissional =D

src/assets/js/main.js
```diff
@@ -85,20 +85,49 @@ function scrollSidebarIntoView() {
...
+    tocHeader.addEventListener('click', (_) => {
+      _scrollToTop();
     });
   }
 
-  $('body').scrollspy({ offset: 100, target: tocId });
+  const tocElement = document.querySelector(tocId);
+  if (tocElement) {
+    window.addEventListener('scroll', _spyOnScroll);
+  }
```
parlough, line 112: I think my last comment may have got lost, but I don't think this change works as intended. It looks like it will add the `active` class to the entire TOC when scrolled past 100px. The intended functionality is to set an individual element as active when its corresponding header (`<h1>`, `<h2`, etc.) is in view.

Since this functionality is from bootstrap through jquery, I think it's fine to keep using it for now: `$('body').scrollspy({ offset: 100, target: tocId });` rather than trying to rewrite that functionality.

We can worry about migrating it if/when we move away from Bootstrap 4 (https://github.com/flutter/website/issues/8030).

parlough (changes requested):
Thanks for those adjustments! The scrollspy still needs to be adjusted.

parlough (approved):
I tested it and it all works well. Thank you for continuing this work and documenting everything. I really appreciate the help :)

PR #9174, opened 2023-08-01
Reduce stage expiration time

parlough:
There's a limit on the amount of channels active at one time which we passed twice over the past week. 7 days should prevent this from happening while still being sufficient for review time. Staging can be manually retriggered in the GitHub actions UI if a link expired already or by merging in `main`.


sfshaza2 (approved):
lgtm

PR #9192, opened 2023-08-04
Add a brief mention of the Flutter Terminal Embedder

parlough:
While we explore what else we can do with this sample, linking it as a resource seems like a win.


sfshaza2 (approved):
lgtm

PR #9193, opened 2023-08-04
Fix broken API doc links in old release notes

parlough:
The old links were 404ing on the marketing page.


sfshaza2 (approved):
Good catch, @parlough!

PR #9194, opened 2023-08-05
Improve refresh code excerpts when checking codes

AlexV525:
Detecting code excerpts during checking codes would not tell what codes are refreshed during the check. This PR adds `git diff` result to tell the difference.

This was inspired during our sync work and implemented in https://github.com/cfug/flutter.cn/pull/1341.

## Presubmit checklist


AlexV525:
@parlough WDYT about this one? We can also update the code_excerpt_updater tool to output changes.

parlough:
Sorry that I haven't taken a look at this yet. I'm working through a long list of PRs.

Without taking a look, adding the support into the tool itself sounds more flexible, but using `git diff` might be a good solution for now. Thanks for working on this and I'll make sure to take a look as soon as I can :D

AlexV525:
ping @parlough

parlough:
I don't want to land this until https://github.com/flutter/website/pull/9416 or a similar fix for `dart` and `flutter` being inaccessible in the container lands, but I think it may be better to just add the functionality to the code excerpter. It has all the information it needs, just needs a flag and to print them out.

AlexV525:
Sounds fair. I'll see if I can manage to make a request this week for the tool.

AlexV525:
Rethink about the diff, there is no proper way (for Dart officials) to diff changes between string content in Dart. So it might be better to use `git diff` as a solution.

tool/check-code.sh
```diff
@@ -42,9 +42,14 @@ flutter --version
...
   echo "=> Refreshing code excerpts..."
   (
+    commitHash=$(git reflog -n1 --format='%h')
+    git add .
+    git commit -m "Temporarily commit code excerpt changes"
     set -x
     tool/refresh-code-excerpts.sh --keep-dart-tool
+    git reset --soft $commitHash
   ) || (
+    git --no-pager diff
```
parlough, line 52: Won't this only show the diff if there is an error? I believe if all excerpts were successfully refreshed, it won't result an error, so the diff won't be shown.

Maybe this could be moved directly below the `git reset`? It's probably fine if it always runs.

tool/check-code.sh
```diff
@@ -42,9 +42,14 @@ flutter --version
 if [[ $REFRESH ]]; then
   echo "=> Refreshing code excerpts..."
   (
+    commitHash=$(git reflog -n1 --format='%h')
```
parlough, line 45: Could you put this new behavior under a new flag, like `--refresh-diff` and then specify that in `test.sh`? So that those running this locally can opt in to the behavior.

parlough (commented):
Sorry about the delay @AlexV525!

You're right that using `git diff` will be much easier and provide better details.

One suggestion and a question, otherwise looks good to me.

tool/check-code.sh
AlexV525: Should we opt-in? IMO it should at least opt-out. This is helpful whether we test locally or remotely.

tool/check-code.sh
AlexV525: > Won't this only show the diff if there is an error? I believe if all excerpts were successfully refreshed, it won't result an error, so the diff won't be shown.

IIRC, from actual tests, refreshes will throw exceptions if there are codes been refreshed and return `0` if no codes have been refreshed.

If "successfully refreshed" means there are some changes during the refresh and the process ends properly, that is a misunderstanding.

tool/check-code.sh
parlough: That's fine too. It would just be nice to have some way to disable it.

tool/check-code.sh
parlough: I tried it out and the diff doesn't seem to run when excerpts were refreshed. I guess that's because calling refresh script isn't the last executed command in the subshell.

AlexV525: I took another look of your suggestion, if all excerpts were refreshed, why the diff is required? The diff flow is only meant to help users to identify not refreshed files.

AlexV525:
Guys, I don't really like merging from the main branch too often. 😄 It'll somehow affect my acknowledgment of my branchs' status.

domesticmouse:
> Guys, I don't really like merging from the main branch too often. 😄 It'll somehow affect my acknowledgment of my branchs' status. It can be outdated, but without conflicts.

Keeping PRs up to date enables easier reviewing as the delta is more accurate

parlough (approved):
I think let's try this out then and see in what situations it helps. We can always expand when/where the diff functionality runs in the future.

Thanks for this Alex and again sorry about the wait.

PR #9200, opened 2023-08-07
Standardize extra anchor targets

parlough:
In HTML4 both `id` and `name` were allowed, but technically only `id` is allowed for anchors now.


sfshaza2 (approved):
Interesting! I didn't know that and I primarily used "name". lgtm

Issue #6929, opened 2022-03-16
Add README file to Implicit Animations Examples directory


micpap25:
### Describe the problem

Some of the examples folder for implicit animations is not utilized.
Folders such as opacity0/9/10, and container0/9/10 do not seem to be used in the code.

### Expected fix

If these files have a use, I believe the folder would benefit from a README explaining their purpose.
If not, these should be deleted.

### Additional context

https://github.com/flutter/website/tree/main/examples/animation/implicit

PR #9201, opened 2023-08-07
Fix and cleanup implicit animations codelab

parlough:
Essentially this PR is focused on improving the clarity of the steps which are diff based, matching starting and ending runnable samples to the steps, removing unnecessary code, and completing some cleanup of the examples and the surrounding text.

**Detailed list of changes:**

- Sets up code excerpts for the starting and complete example code which were mismatched with the actual steps/instructions. Then updates them to match.
- Adds missing type specifiers and simplifies variable declarations.
- Adjusts trailing commas and formatting in opacity sample for easier to read diffs.
- Removes extra folders/steps that are not currently used.
- Adds missing type specifiers and simplifies variable declarations.
- Moves away from inline ignores since the diff tooling doesn't remove them, making the diffs more confusing and the line numbers inconsistent between steps. Ignoring in the analysis options file allows for consistent line numbers.
- Changes text of button from "change" to "Change" to match instructions.
- Fixes a few instructions in the codelab's text.
- Removes the embedded survey.
- Adds a README to the example directory explaining its setup.

**Staged:** https://flutter-docs-prod--pr9201-fix-implicit-animati-24cnhdzx.web.app/codelabs/implicit-animations

Closes https://github.com/flutter/website/issues/6929


examples/animation/implicit/container1/lib/main.dart
```diff
@@ -1,7 +1,3 @@
-// Copyright 2019 the Dart project authors. All rights reserved.
```
domesticmouse, line 1: I'm curious why this has been deleted. We need all code to have copyright headers.

domesticmouse (approved):

examples/animation/implicit/container1/lib/main.dart
parlough: It causes a little confusion with the line numbers with our current diff setup. I can add them back since it's not a major problem.

PR #9205, opened 2023-08-07
feat: Add hover styles to TOC entries

### LINKED ISSUE #9016: Add hover styles to TOC entries
### Describe the problem

On hover, there is no indication which element in the TOC you are interacting with.

### Expected fix

Add hover styles, similar to how we did for the main sidenav.

### Additional context

_No response_

ulisseshen:
Was added hover style to TOC entries 
Closes #9016 
## Presubmit checklist


src/_sass/components/_toc.scss
```diff
@@ -4,6 +4,15 @@
   display: none;
   right: 0;
 
+
+  @mixin nav-link-not-active-hover {
+    &:not(.disabled) {
```
parlough, line 9: I'm not sure the TOC entries can be disabled, at least I can't find any area where they can be. Do you know of any? If not, we can likely remove the `&:not(.disabled) {` portion.

src/_sass/components/_toc.scss
```diff
@@ -4,6 +4,15 @@
   display: none;
   right: 0;
 
+
+  @mixin nav-link-not-active-hover {
```
parlough, line 8: Since this doesn't have a special case for active, perhaps just `nav-link-hover`?

parlough (commented):
Thanks so much for this accessibility improvement!

A question and a suggestion:

src/_sass/components/_toc.scss
ulisseshen: You is correctly, I only copy and paste from another place.

src/_sass/components/_toc.scss
ulisseshen: This name is more clean.

parlough (approved):
Nice 🥳 ! Thanks so much :D

Issue #9203, opened 2023-08-07
Feedback on the new information architecture


InMatrix:
The updated left navigation looks great! I took a closer look at the current implementation and I'd like to share the following feedback for discussion:

- Consider renaming “Development” to “Beyond UI” or “Further Development”, since the UI section is also about development.

- Moving “Accessibility & localization” into the “User interface” section, since both deal with UIs.

About the Design & theming category under User interface:

- It needs an introductory doc, but we can add this later. 
- In the long run, this section should probably covers major design systems Flutter supports (Material Design, Cupertino, etc), how to work with themes, styling text, and painting custom graphics. 
- Moving “Use themes to share colors and font styles” one level up to “Design & theming”, since the concept is more general than Material Design.
- Moving “Add a drawer to a screen” and “Add tabs to your app” to Navigation & routing, since these two docs are about creating UIs for navigation.
- Moving “Display a snackbar” to “Interactivity”
- After these changes, “Material design” can points to the “Overview” doc, since there’s no other doc under it. The new menu hierarchy within “Design & theming” will look like this:
  - Material Design overview
  - Use themes to share styles
  - \> Text
  - \> Custom graphics

Cc: @parlough @sfshaza2 @stef-luo

parlough:
Thanks for the suggestions! I also wasn't sure about "Development", and I particularly like "Beyond UI". Moving "Accessibility & localization" makes sense as well.

For the various "Theming & design" issues, I agree with most of them. It is the section that has helped us identify the most holes in our documentation. I've started creating individual issues but need to create a few more.

I was hesitant to put some of those docs outside of "Material Design" due to some of their strong ties to material/`MaterialApp`, but especially with some of these gaps, I think moving them out makes sense.

I'll prepare a few PRs and keep working on opening issues for those holes. Thanks again!

InMatrix:
Thanks @parlough! 

Cc: @leighajarett in case she has thoughts on the "Design & theming" section. There probably is a good follow-up project for us to figure out what docs should be there in the fullness of time.

PR #9206, opened 2023-08-07
Adjust name of "Development" category to "Beyond UI"

parlough:
Using the "Beyond UI" suggestion from https://github.com/flutter/website/issues/9203. Open to other ideas though and it's easy to change in the future since it's just a label :)


sfshaza2 (approved):
I like it!

PR #9207, opened 2023-08-07
Fix Ecosystem tab to match marketing site

parlough:
We recently added Monetization under "Development", so it's no longer necessary here.


sfshaza2 (approved):
lgtm

Issue #9218, opened 2023-08-08
[PAGE ISSUE]: 'Navigate to a new screen and back'


imagochan:
### Page URL

https://docs.flutter.dev/cookbook/navigation/navigation-basics/

### Page source

https://github.com/flutter/website/tree/main/src/cookbook/navigation/navigation-basics.md

### Describe the problem

The dart code inside the interactive exampel section is wrong. As in it´s just HTML instead of real code.

### Expected fix

The interactive example code should be checked and replaced by the correct code intended for this section.

parlough:
Thanks for opening an issue! If you hard refresh the page, the embedded DartPads should be working now.

PR #9219, opened 2023-08-08
Fix DartPads failing to load due to main.js logic running too early

parlough:
Closes https://github.com/flutter/website/issues/9218

### DIFF
--- a/src/assets/js/archive.js
+++ b/src/assets/js/archive.js
@@ -283,8 +283,7 @@ function getProvenanceLink(os, release, date, channel) {
 
 
 // Send requests to render the tables.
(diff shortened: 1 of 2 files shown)


atsansone (approved):

parlough:
\cc @ulisseshen I didn't notice this on reviewing https://github.com/flutter/website/pull/9167, but we don't want the included logic to run until after the DOM is verified to have been finished loading.

ulisseshen:
This make sense now. it really should be `document.addEventListener("DOMContentLoaded",(){})`. I'll keep it in mind or anothers migrations.

PR #9225, opened 2023-08-09
Removing the Happy Paths recommendations

sfshaza2:
Removing Happy Paths.


firebase.json
```diff
@@ -69,6 +69,7 @@
       { "source": "/networking", "destination": "/data-and-backend/networking", "type": 301 },
       { "source": "/overview", "destination": "/resources/architectural-overview", "type": 301 },
       { "source": "/platform-channels", "destination": "/platform-integration/platform-channels", "type": 301 },
+      { "source": "/packages-and-plugins/happy-paths*", "destination": "/packages-and-plugins/favorites", "type": 301 },
```
parlough, line 72: I don't think this will cover the recommended page. Can just separate them so its easy to understand.

```suggestion
      { "source": "/packages-and-plugins/happy-paths", "destination": "/packages-and-plugins/favorites", "type": 301 },
       { "source": "/packages-and-plugins/happy-paths/recommended", "destination": "/packages-and-plugins/favorites", "type": 301 },
```

parlough (approved):
Thanks for this cleanup. One fix:

Issue #9231, opened 2023-08-10
Take down the Flutter I/O banner


atsansone:
The Flutter I/O banner can be taken down now. Please remove at your earliest convenience.

PR #9256, opened 2023-08-16
Add a new banner for Flutter 3.13

parlough:
I thought this would be fun, but let me know if you prefer something simpler.

Closes https://github.com/flutter/website/issues/9231


atsansone (approved):

parlough:
Thanks for taking a look @atsansone!

PR #9269, opened 2023-08-18
[Windows] Document the show window migration

loic-sharma:
Flutter 3.7 and 3.10 has a race condition that can result in the window not being shown (see https://github.com/flutter/flutter/issues/119415). Projects need to be migrated to fix this bug.

Most Windows projects will be migrated automatically by running `flutter windows build` or `flutter run` unless they've done significant changes to their `windows/runner/flutter_window.cpp` file.

This documents how to migrate a Windows project manually. Note that this fix depends on an API that was added in Flutter 3.13. In other words, projects that are still using Flutter 3.7 or 3.10 should not do this migration.

_Issues fixed by this PR (if any):_ https://github.com/flutter/flutter/issues/127695

image

## Presubmit checklist

### DIFF
--- a/src/release/breaking-changes/index.md
+++ b/src/release/breaking-changes/index.md
@@ -51,6 +51,7 @@ release, and listed in alphabetical order:
 * [Removed `ignoringSemantics`][] properties
 * [Deprecated `RouteInformation.location`][] and its related APIs
 * [Updated EditableText scroll into view behavior][]
+* [Migrate a Windows project to ensure the window is shown][]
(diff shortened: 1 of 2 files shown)


parlough (approved):
Thanks so much for staying on top of this and writing this! I really appreciate it :D

As a reference for next time, consider adding a migration guide once the necessary change lands in beta. These guidelines can still be helpful for that set of users. If something changes before stable is released, we can always readjust :)

PR #9276, opened 2023-08-21
Updated Xcode license flag

### LINKED ISSUE #9239: Xcode license agreement command double dash problem
### Page URL

https://docs.flutter.dev/get-started/install/macos#install-xcode

### Page source

_No response_

### Describe the problem

```sh
sudo xcodebuild --license
```

This command doesn't work with new Xcode versions.

### Expected fix

`--license` option must be a single dash.

```sh
sudo xcodebuild -license
```

### Additional context

My Xcode version:

```sh
xcodebuild -version
```

> Xcode 14.3.1
> Build version 14E300c

atsansone:
Fixes #9239

### DIFF
--- a/src/get-started/install/_ios-setup.md
+++ b/src/get-started/install/_ios-setup.md
@@ -30,8 +30,6 @@ To develop Flutter apps for iOS, you need a Mac with Xcode installed.
        $ sudo xcodebuild -license
        ```
 
-       The single dash for the `license` option was introduced in Xcode 14.
-
 Versions older than the latest stable version might still work,
 but are not recommended for Flutter development.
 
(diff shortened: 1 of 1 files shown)


src/get-started/install/_ios-setup.md
```diff
@@ -19,11 +19,18 @@ To develop Flutter apps for iOS, you need a Mac with Xcode installed.
...
+    {: type="a"}
+    1. Open Xcode and confirm.
+
+    1. Open the Terminal and run:
+
+       ```terminal
+       $ sudo xcodebuild -license
+       ```
+
+       The single dash for the `license` option was introduced in Xcode 14.
```
parlough, line 33: If you're going to say something was changed, you should generally explain what worked before. Otherwise, it will just cause extra reading or confusion on what to do before.

In this case, since the instructions say to install the latest version of xcode and that's all Apple makes easy to do, I think it should be fine to not include any note.

```suggestion
```

parlough: Actually, I just looked through some related posts online and it seems the single dash syntax has (always?) been the supported syntax and wasn't just introduced in Xcode 14. So this note is definitely fine to remove.

parlough (approved):
Thanks for fixing this!

One comment/fix:

Issue #9246, opened 2023-08-14
Update creating Linux Snap application to include `snap install`


RickPoleshuck:
### Page URL

https://docs.flutter.dev/deployment/linux

### Describe the problem

Following these instructions on an Ubuntu 22.04 Desktop had a couple of problems.

1. The instructions should suggest trying it out locally with $ sudo snap install <pgrm.snap> --devmode before pushing to snap store.
2. The program runs with a large black border without a usable title bar after installation with --devmode. This border is only on the linux desktop with snapcraft running. It works correctly on another linux desktop.
 

### Expected fix

1. Add $ sudo snap install <pgrm.snap> --devmode  to the documentation.
2. Since it works correctly on another linux desktop, maybe only the documentation needs to be changed.

### Additional context

You can use the txtr_dsk project within https://github.com/RickPoleshuck/txtr to duplicate this problem.

kenvandine:
I would recommend using core22 instead of core18.

RickPoleshuck:
> I would recommend using core22 instead of core18.

"Extension 'flutter-stable' does not exist" when specifying core22

kenvandine:
The usage is a little different for core22.  You can see https://snapcraft.io/docs/flutter-plugin

RickPoleshuck:
> The usage is a little different for core22. You can see https://snapcraft.io/docs/flutter-plugin
core22 never finishes building on my ubuntu 22.04. It gets stuck at "Creating new instance from image 'core22' from remote 'craft-com.ubuntu.cloud-buildd' " 
Snapcraft and flutter seems fragile. My best guess is that core22 requires lxd and there is a problem with lxd on my version of linux. When I say 'my best guess', I mean a confidence of about 20% :-)

RickPoleshuck:
> > The usage is a little different for core22. You can see https://snapcraft.io/docs/flutter-plugin
> > core22 never finishes building on my ubuntu 22.04. It gets stuck at "Creating new instance from image 'core22' from remote 'craft-com.ubuntu.cloud-buildd' "
> > Snapcraft and flutter seems fragile. My best guess is that core22 requires lxd and there is a problem with lxd on my version of linux. When I say 'my best guess', I mean a confidence of about 20% :-)

I am able to build the snap a brand new ubuntu virtual machine. I have configured lxd exactly the same way on both machines. Should I do a pull request to adjust the documentation to use core22 and mention testing with --devmode?

PR #9282, opened 2023-08-23
update linux snapcraft instructions to use core22

RickPoleshuck:
The instructions for creating a Linux snap have been adjusted to use core22. Multipass does not work for core22 and that alternative for building has been removed. Also, a short note about how to test before publishing was added.

Fixes https://github.com/flutter/website/issues/9246

## Presubmit checklist


parlough:
Thanks for the pull request!

\cc @kenvandine Could you take a look at this change or redirect to someone else who could? I'm not familiar with snaps. I'd appreciate it.

src/deployment/linux.md
```diff
@@ -33,30 +32,8 @@ Use the following instructions to set up your build environment.
 At the command line, run the following:
 
 ```terminal
-$ sudo snap install snapcraft --classic
+$ sudo snap install snapcraft --classics
```
kenvandine, line 35: This should actually be "--classic".

src/deployment/linux.md
```diff
@@ -342,10 +316,18 @@ To use the LXD container backend:
...
-## Publish
+Once the `snapcraft.yaml` file is complete,
+
+## Test the snap
 
 Once the snap is built, you'll have a `<name>.snap` file
-in your root project directory. You can now publish the snap.
+in your root project directory. 
+
+$ sudo snap install ./super-cool-app_0.1.0_amd64.snap --devmode
```
kenvandine, line 326: It's better to test without devmode.  It's fine to try devmode early in development, but always test without devmode before publishing.

RickPoleshuck:
If we recommend testing without devmode, we need to add instructions
for adding a signature to the snap.

kenvandine:
> If we recommend testing without devmode, we need to add instructions for adding a signature to the snap.
> […](#)
> On Wed, 2023-08-23 at 18:46 -0700, Ken VanDine wrote: @kenvandine commented on this pull request. In src/deployment/linux.md: > Once the snap is built, you'll have a `<name>.snap` file -in your root project directory. You can now publish the snap. +in your root project directory. + +$ sudo snap install ./super-cool-app_0.1.0_amd64.snap --devmode It's better to test without devmode. It's fine to try devmode early in development, but always test without devmode before publishing. — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you authored the thread.Message ID: ***@***.***>

Ah, I see why you suggested devmode.  You need to add --dangerous for a locally built snap.  Devmode drops all confinement, so it's not a good test of functionality.  So for the PR just replace devmode with dangerous.

RickPoleshuck:
Thank you. I really don't like snaps choice of words. devmode seems
safer than dangerous. :-)

kenvandine (changes requested):
A couple minor changes, then it'll be good to go.

src/deployment/linux.md
```diff
@@ -305,7 +277,7 @@ Place the .desktop file in your Flutter project
 under `<project root>/snap/gui/super-cool-app.desktop`.
 
 **Notice**: icon and .desktop file name must be the 
-same as your app name in yaml file!
+same as your app name in yaml file! Also, your icon file must be world readable.
```
kenvandine, line 280: It's also probably not worth calling out the permissions of the icon file, that has to be a pretty extreme corner case to try to use an icon without read permissions.

kenvandine (changes requested):

src/deployment/linux.md
RickPoleshuck: I was annoyed that snapcraft didn't just change the permissions on the icon file during the build process. It might be an edge case, but I hit the edge.

kenvandine: > I was annoyed that snapcraft didn't just change the permissions on the icon file during the build process. It might be an edge case, but I hit the edge.

Yeah, but the same could be true for lots of the files, like the desktop file and other assets.  I don't think it's worth calling all of those out, the usual mode for files created would work.

cbracken:
Ah apologies @RickPoleshuck just noticed that you replied already to @kenvandine. I pushed an update without noticing that you were already back to this PR. I'll leave it to the two of you to work out the rest!

src/deployment/linux.md
RickPoleshuck: @kenvandine , in my case, I have my umask set to 0007, a holdover from when my Unix development computer was shared by many people. The icon file was the only issue that was an issue for me. But, I agree that it is an edge case and am okay either way. Only very paranoid or very old linux programmers set their umask to 0007. I am not very paranoid. :-)

kenvandine (approved):
Looks good, thanks!

src/deployment/linux.md
```diff
@@ -342,10 +314,18 @@ To use the LXD container backend:
 $ snapcraft --use-lxd
 ```
 
-## Publish
+Once the `snapcraft.yaml` file is complete,
```
domesticmouse, line 317: It feels odd to have a title as a continuation of a sentence. Can we turn this into a complete sentence?

domesticmouse (approved):
PTAL @atsansone

src/deployment/linux.md
cbracken: Or just delete this line altogether, perhaps? It's really just leading into the next two sections, and their titles provide pretty much all the reader needs to know.

kenvandine: I agree it would be better to remove the line.

src/deployment/linux.md
```diff
@@ -342,10 +314,18 @@ To use the LXD container backend:
...
+
```
cbracken, line 318:
```suggestion
```

sfshaza2 (approved):
lgtm

PR #9302, opened 2023-08-25
Replaced flutter pub run with dart run

### LINKED ISSUE #9301: Change all instances of `flutter pub run` with `dart run`
`flutter pub run` is deprecated.

```terminal
➜  my_app flutter pub run
Deprecated. Use `dart run` instead.
Must specify an executable to run.

Usage: dart pub run <executable> [arguments...]
-h, --help                              Print this usage information.
    --[no-]enable-asserts               Enable assert statements.
    --enable-experiment=<experiment>    Runs the executable in a VM with the given experiments enabled.
                                        (Will disable snapshotting, resulting in slower startup).
-C, --directory=<dir>                   Run this in the directory <dir>.

Run "dart help" to see global options.
See https://dart.dev/tools/pub/cmd/pub-run for detailed documentation.
```

atsansone:
Fixes #9301

### DIFF
--- a/src/cookbook/testing/unit/mocking.md
+++ b/src/cookbook/testing/unit/mocking.md
@@ -126,7 +126,7 @@ void main() {
 Next, generate the mocks running the following command:
 
 ```terminal
-$ flutter pub run build_runner build
+$ dart run build_runner build
 ```
 
 ## 4. Write a test for each condition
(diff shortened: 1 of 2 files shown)


parlough (approved):
Thanks! I was hesitant to make this change earlier because the error messages with a messed up setup are pretty confusing still, but I'm exploring ways to improve those in pub.

atsansone:
Understood, but the message that `flutter pub run` displays makes it clear that it's deprecated. We can be consistent with the command at least.

parlough:
Oh sorry, I meant when running `dart run` instead. If there is a Flutter dependency, it currently gives obscure failure messages if your Dart SDK does not match the Flutter SDK

Issue #8272, opened 2023-02-16
[Makefile] Impossible to setup on Windows


AlexV525:
### Describe the problem

https://github.com/flutter/website/blob/5494bf1ac41f10fe0fb236c66677c1b5a5c8c3dc/Makefile#L172-L173

Since most of the commands start with `clean`, there is no way to run `rm` on Windows command lines.

```dart
C:\ProgramData\chocolatey\bin\make.exe -f X:\Coding\Garages\GitHub\flutter-website\Makefile -C X:\Coding\Garages\GitHub\flutter-website build
process_begin: CreateProcess(NULL, git rev-parse --short HEAD, ...) failed.
X:\Coding\Garages\GitHub\flutter-website\Makefile:12: pipe: No error
make: Entering directory 'X:/Coding/Garages/GitHub/flutter-website'
make clean
make[1]: Entering directory 'X:/Coding/Garages/GitHub/flutter-website'
rm -rf _site .jekyll* src/.jekyll* *.log tmp .dart_tool
make[1]: Leaving directory 'X:/Coding/Garages/GitHub/flutter-website'
process_begin: CreateProcess(NULL, git rev-parse --short HEAD, ...) failed.
Makefile:12: pipe: No error
process_begin: CreateProcess(NULL, rm -rf _site .jekyll* src/.jekyll* *.log tmp
.dart_tool, ...) failed.
make (e=2): 系统找不到指定的文件。
make[1]: *** [Makefile:153: clean] Error 2
make: *** [X:\Coding\Garages\GitHub\flutter-website\Makefile:135: build] Error 2
make: Leaving directory 'X:/Coding/Garages/GitHub/flutter-website'

Process finished with exit code 2

```

AlexV525:
Also, copying Flutter from the host will cause platform issues that unable to start any Flutter commands.

atsansone:
@AlexV525 : Your second issue is a Windows Git issue.

Check out this [StackOverflow post](https://stackoverflow.com/questions/29045140/env-bash-r-no-such-file-or-directory) that explains the issue.

TL;DR: Git translated UNIX line breaks to Windows line breaks on `git clone`. `bash` has no idea what to do with those line breaks.

```terminal
$ git config --global core.autocrlf input
```

That ensured that all _writes_ to git and GitHub use LF only. 

```terminal
$ git config --global core.autocrlf false
```

This ensures that all _reads_ to git and GitHub use LF only. 

[...]

Issue #6443, opened 2021-11-05
Fix website local builds on Windows


skytomo221:
### Describe the problem

When I set up docker image on Windows Power Shell
`make up` fails by jekyll

<details>
<summary>logs</summary>

```console
> make up
docker-compose up site
Recreating website_site_1 ... done
Attaching to website_site_1
site_1  | bundle exec jekyll serve \
site_1  |       --host 0.0.0.0 \
site_1  |       --port 4002 \
site_1  |       --config _config.yml,_config_dev.yml \
site_1  |       --livereload \
site_1  |       --incremental \
site_1  |       --trace
site_1  | Configuration file: /app/_config.yml
site_1  | Configuration file: /app/_config_dev.yml
site_1  | bundler: failed to load command: jekyll (/usr/local/bundle/bin/jekyll)
site_1  | /usr/local/bundle/gems/jekyll-4.2.0/lib/jekyll/external.rb:60:in `require': /app/src/_plugins/breadcrumb.rb:1: unknown regexp option - t (SyntaxError)
site_1  | ../../site-shared/src/_plugins/breadcru...
site_1  |      ^~~~~
site_1  |       from /usr/local/bundle/gems/jekyll-4.2.0/lib/jekyll/external.rb:60:in `block in require_with_graceful_fail'
site_1  |       from /usr/local/bundle/gems/jekyll-4.2.0/lib/jekyll/external.rb:57:in `each'
site_1  |       from /usr/local/bundle/gems/jekyll-4.2.0/lib/jekyll/external.rb:57:in `require_with_graceful_fail'
site_1  |       from /usr/local/bundle/gems/jekyll-4.2.0/lib/jekyll/plugin_manager.rb:89:in `block in require_plugin_files'
site_1  |       from /usr/local/bundle/gems/jekyll-4.2.0/lib/jekyll/plugin_manager.rb:87:in `each'
site_1  |       from /usr/local/bundle/gems/jekyll-4.2.0/lib/jekyll/plugin_manager.rb:87:in `require_plugin_files'
site_1  |       from /usr/local/bundle/gems/jekyll-4.2.0/lib/jekyll/plugin_manager.rb:21:in `conscientious_require'
site_1  |       from /usr/local/bundle/gems/jekyll-4.2.0/lib/jekyll/site.rb:131:in `setup'
site_1  |       from /usr/local/bundle/gems/jekyll-4.2.0/lib/jekyll/site.rb:36:in `initialize'
[...]

Issue #6201, opened 2021-08-26
`make setup` fails with `/usr/bin/env: ‘bash\r’: No such file or directory`


TahaTesser:
### Describe the problem

When I set up docker image on Windows
`make setup` fails by 14/16

<details> 
<summary>logs</summary> 
	
```console
Taha@Taha-PC MINGW64 ~/Code/cloned/website (master)
$ make setup
make clean
make[1]: Entering directory 'C:/Users/Taha/Code/cloned/website'
rm -rf _site .jekyll* src/.jekyll* *.log tmp example.g .dart_tool
make[1]: Leaving directory 'C:/Users/Taha/Code/cloned/website'
docker-compose build site
#1 [internal] load build definition from Dockerfile
#1 sha256:f561a8fef6cfe35823face857bde797a1c76b658b443e9a5e51c415e341effaa
#1 transferring dockerfile: 32B done
#1 DONE 0.0s

#2 [internal] load .dockerignore
#2 sha256:c72d804e4519ff661ddd75b049e656be79bd43eec3dfe302228373b4b39b99ec
#2 transferring context: 34B done
#2 DONE 0.0s

#3 [internal] load metadata for docker.io/library/ruby:3-buster
#3 sha256:df66394f91f845e96404cd0f06a808649ad0177e61733aadd5eff962a00ff40c
#3 DONE 0.7s

#4 [dev  1/16] FROM docker.io/library/ruby:3-buster@sha256:a24fb0a5dbbc2a2d1d87ca47ae3eeb643e9427da4a7849541aa0a0f2308860d1
#4 sha256:b342b20b19a5bc2536e1a5e0626480811f405d94d1c612f4ba98fc114893a587
#4 DONE 0.0s

#9 [internal] load build context
#9 sha256:e1fb1ff79debb986bfd813370f9ffc6d910279c0105814502da22139a8f8bb6f
#9 transferring context: 1.71MB 2.6s done
#9 DONE 2.7s

#5 [dev  2/16] RUN curl -sL https://deb.nodesource.com/setup_15.x -o node_setup.sh &&       bash node_setup.sh 1> /dev/null
#5 sha256:ab5bdd1089285e5a5d29d04522279a4b5aca7b0cfb131a063bbc3c1c454cd55e
#5 CACHED

#8 [dev  5/16] RUN gem install bundler
#8 sha256:84e9c48ade5f0d6a31d994076ec7f10afcce96929923779c57846bbc37d9416e
#8 CACHED

[...]

PR #9307, opened 2023-08-26
🔥 Remove `flutter` as a submodule

AlexV525:
The PR removes the `flutter` repo as a submodule to avoid conflicts when building the website on different platforms. It was first introduced in #1257 but with no explanation so we cannot track back what's the purpose, but removing this should not cause difficulties in maintaining the website.

### Additional info

- The build process is slower than the previous one because the checkout step needs an extra 3~5 minutes, depending on the network condition. See [now](https://github.com/flutter/website/actions/runs/5985413443/job/16237673762?pr=9307) and [previous](https://github.com/flutter/website/actions/runs/5983398408/job/16237673076?pr=9298) checks.

_Issues fixed (partially) by this PR (if any):_
- https://github.com/flutter/website/issues/6201
- https://github.com/flutter/website/issues/6443
- https://github.com/flutter/website/issues/8272

## Presubmit checklist


AlexV525:
If we need `flutter` for locally builds, we might consider add it to the `.gitignore` too, and maybe with some more setup scripts.

AlexV525:
With the change, I was able to run `make test` locally on my Windows setup. But still not able to run `make up` because some symbolic link issues between platforms.

parlough:
Thanks for working on this @AlexV525!

I haven't taken a close look yet, but I'll give it some thought tonight :) 

Currently we benefit from dependabot updating the submodule for each stable Flutter release. This allows us to explicitly move to a new stable version of Flutter, rather than the build using whatever the stable branch happens to be at, potentially breaking the build with updates. In practice, this isn't usually an issue since we are prepare for new stable builds in advance.

AlexV525:
> This allows us to explicitly move to a new stable version of Flutter, rather than the build using whatever the stable branch happens to be at, potentially breaking the build with updates.

It can be done if we wrote the current ref we're using into configurations and use some dependabot mechanism to upgrade as I imagine.

parlough (approved):
Sorry about the delay on reviewing this. Thank you!

PR #9319, opened 2023-08-28
Move widget catalog to under UI in sidenav

parlough:
Fixes https://github.com/flutter/website/issues/9263

### DIFF
--- a/src/_data/sidenav.yml
+++ b/src/_data/sidenav.yml
@@ -65,7 +65,8 @@
 - header: User interface
 - title: Introduction
   permalink: /ui
-      
+- title: Discover widgets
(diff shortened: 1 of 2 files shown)


atsansone (commented):
@parlough , the problem with an active title is that all of others are not. This lacks parallel structure. I don't think "Widget Catalog" is bad. It's just generic.

parlough:
> @parlough , the problem with an active title is that all of others are not. This lacks parallel structure. I don't think "Widget Catalog" is bad. It's just generic.

Great point! I've switched back to "Widget catalog" :)

atsansone (approved):

PR #9330, opened 2023-08-29
"Add Cursor Pointer for Search Icon on Website"

### LINKED ISSUE #9329: "Add Cursor Pointer for Search Icon on Website"
### Describe the problem

I noticed that the cursor pointer is missing when hovering over the search icon on our website. This issue makes it difficult for users to identify that the search icon is clickable

### Expected fix

To fix this problem, we should add the CSS class that makes elements appear as clickable links or buttons when users hover over them. This will provide visual feedback to users that the search icon is interactive.

### Additional context

This issue affects the user experience on our website, especially when users attempt to search for content. By adding the cursor pointer, we can enhance usability and guide users to engage with the search functionality more effectively.

Hittitech:
This issue affects the user experience on our website, especially when users attempt to search for content. By adding the cursor pointer, we can enhance usability and guide users to engage with the search functionality more effectively.
https://docs.flutter.dev/

Fixes #9329 "Add Cursor Pointer for Search Icon on Website"

## Presubmit checklist


parlough (changes requested):
Thanks for the pull request!

However, I don't think this solution will work. `cursor-pointer` is a class provided by some CSS frameworks, such as Tailwind, but we don't currently have any equivalent class.

We'd likely need to add `cursor: pointer` to the appropriate styles in the SCSS files. 

I'm not sure with our current setup for the search icon if we can get this to work though, as if we add `cursor: pointer` to the `::before` pseudo element where the icon is added, it won't work since it's within the bounds of the `input` which needs be `cursor: text`. 

Any ideas? If not, we would have to adjust the search field to use a separate element for the search icon rather than a `::before` pseudo-element.

Thanks again! Let us know what you think :)

Hittitech:
> Thanks for the pull request!
> 
> However, I don't think this solution will work. `cursor-pointer` is a class provided by some CSS frameworks, such as Tailwind, but we don't currently have any equivalent class.
> 
> We'd likely need to add `cursor: pointer` to the appropriate styles in the SCSS files.
> 
> I'm not sure with our current setup for the search icon if we can get this to work though, as if we add `cursor: pointer` to the `::before` pseudo element where the icon is added, it won't work since it's within the bounds of the `input` which needs be `cursor: text`.
> 
> Any ideas? If not, we would have to adjust the search field to use a separate element for the search icon rather than a `::before` pseudo-element.
> 
> Thanks again! Let us know what you think :)


Thank you for replying . And yes you are correct here I realized that this will not work but as you mentioned .
we can add cursor pointer to the input form which will work out but the expanded input box will have `cursor as pointer` over it
secondly even   after entering the search query users expect that `clicking on the search icon ` will work to search but its also not working we have to use enter from keyboard to search .
I would suggest having search icon  separately and having an onclick action to the search icon/button which not only solves current issue but also this one .
Another thing I would like to mention is that  documentation website so we should have this searching feature in a place where its easily accessible and noticeable to the users like:
we can have a input box already active and expanded anywhere in the header or just below the header where ever its more accessible to the users and easily they can directly enter their search key and click on icon to search ,
We can also provide search suggestions with the input when the user is typing,

These things seems will provide better user experience as I believe most users uses documentation while learning or implementing specific feature in their project and search feature  is the mostly used part of documentation.
these all seems better to me than current one
However you all knows better . let me know what you thinks and how can I contribute more towards your idea :)

atsansone:
This appears to duplicate https://github.com/flutter/website/pull/9354. If you want to comment on that PR, it was tested to work with minor issues.

PR #9390, opened 2023-09-14
Update android setup device requirements

parlough:
To match the "Supported" column on [Supported deployment platforms](https://docs.flutter.dev/reference/supported-platforms). We could technically use the "Best effort" versions but Google play services is dropping support for those and 19/20 devices count for <= 0.5% of all active Android devices. This prevents any potential confusion and is more future proof.


sfshaza2 (approved):
lgtm

atsansone (approved):

PR #9393, opened 2023-09-14
Fix links to `dart` tool documentation

parlough:
The existing link was redirecting to the `dart run` page, but the text on these pages is referring to the entire `dart` tool.


sfshaza2 (approved):
lgtm

PR #9397, opened 2023-09-14
Announce a breaking change of added `dispose()`.

polina-c:
Discussion: https://docs.google.com/document/d/1ec5djPP14b1u93O9fHrJJf3gbONay1nV4sBV2-XXOIw/edit?resourcekey=0-CFEu4HWAW8yP0knhoVfKxQ

This announce is added as already released, because we are adding `dispose` already long time and some of the changes may be already released: https://github.com/flutter/devtools/issues/3951

Tech writers, please do not merge this before tech sign off. @goderbauer , can you review please?

### DIFF
--- a/src/release/breaking-changes/dispose.md
+++ b/src/release/breaking-changes/dispose.md
@@ -1,12 +1,12 @@
 ---
-title: Added missed `dispose()` for some disposable objects in Flutter
+title: Added missing `dispose()` for some disposable objects in Flutter
 description: >
   'dispose()' might fail because of double disposal.
 ---
 
 ## Summary
 
-Missed 'dispose()' are added for some disposable objects.
+Missing 'dispose()' are added for some disposable objects.
 If some other code also invokes 'dispose()' for the objects,
 and the object is protected from double disposal,
 the second 'dispose()' fails with the following error message:
--- a/src/release/breaking-changes/dispose.md
+++ b/src/release/breaking-changes/dispose.md
@@ -1,71 +1,17 @@
 ---
-title: Replace with title of breaking change
+title: Added missed `dispose()` for some disposable objects in Flutter
 description: >
-  Brief description similar to the "context" section below. 
-  Text should break at 80 chars or less.
+  'dispose()' may fail because of usage after disposal.
 ---
-
-{% comment %}
-  PLEASE READ THESE GENERAL INSTRUCTIONS:
-  * All lines of text should be 80 chars OR LESS.
-    The writers strongly prefer semantic line breaks:
-    https://github.com/dart-lang/site-shared/blob/main/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks
-  * DON'T SUBMIT a PR weeks and weeks in advance.
-    Doing this causes it to get stanky in the website
-    repo and usually develops conflicts in the index file.
-    Ideally, submit a PR once you have confirmed
-    info on the version number where the breaking
-    change landed.
-  * One of the most important things to fill out 
-    in this template is the *Timeline* section.
-    I won't approve/merge the PR until the "landed in"
-    release info is provided. For example:
-    `Landed in version: 1.21.0-5.0.pre<br>`.
-    Do NOT list the PR in this section. Also, don't
-    fill in the "stable" release info unless it's
-    already in a published stable release.
-    After a stable release, I go through and confirm
-    that updates have made it to stable and I then
-    update the breaking change and the index file.
-  * The text in this page should be backwards looking,
-    so write about previous behavior in past tense,
-    not future tense. People are reading this months
-    from now when the change is likely in the stable
-    release, not today. Don't say "in a month" or
-    talk about your plan to do something next week.
-    Assume you've done it, and that they're looking
-    back to figure out how to migrate their code.
-  * Use sentence case for headings and titles.
-    (`## Migration guide`, NOT `Migration Guide`)
-  * DON'T use the abbreviation `i.e.` or `e.g.`.
-    Use "for example" or "such as", and similar.
-  * For links, use the macros where possible.
-    See the examples at the end of this template,
-    but don't use "github.com" or "api.flutter.dev" or
-    "pub.dev" in your URLs. Use the {{site.github}},
-    {{site.api}}, or {{site.pub}} macros.
-  * AVOID "will" when possible, in other words,
-    stay in the present tense. For example:
-    Bad: "When encountering an xxx value,
-          the code will throw an exception."
-    Good: "When encountering an xxx value,
-           the code throws an exception."
-    Good use of "will": "In release 2.0, the xxx API
-          will be deprecated."
-  * Finally, delete the comment tags and text from the
-    final PR.
-{% endcomment %}
 
 ## Summary
 
 {% comment %}
-  A brief (one- to three-line) summary that gives
-  context as to what changed so that someone can
-  find it when browsing through an index of
-  breaking changes, ideally using keywords from
-  the symptoms you would see if you had not yet
-  migrated (for example, the text from probable
-  error messages).
+  Missed 'dispose()' are added for some disposable objects. If some other code
+  also invoked 'dispose()' for these objects, and the class is protected from double
+  disposal, the second 'dispose()' will fail with error message:
+
+  Once you have called dispose() on a <class name>, it can no longer be used.  
 {% endcomment %}
 
 ## Background
--- a/src/release/breaking-changes/dispose.md
+++ b/src/release/breaking-changes/dispose.md
@@ -6,9 +6,10 @@ description: >
 
 ## Summary
 
-Missed 'dispose()' are added for some disposable objects. If some other code
-also invokes 'dispose()' for the objects, and the object is protected from double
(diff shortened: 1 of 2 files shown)


src/release/breaking-changes/dispose.md
```diff
@@ -0,0 +1,51 @@
...
+title: Added missed `dispose()` for some disposable objects in Flutter
+description: >
+  'dispose()' may fail because of double disposal.
+---
+
+## Summary
+
+Missed 'dispose()' are added for some disposable objects. If some other code
+also invokes 'dispose()' for the objects, and the object is protected from double
+disposal, the second 'dispose()' will fail with error message:
```
atsansone, line 11: **chore (if-minor):** Use [semantic breaks](https://sembr.org/).
```suggestion
Missed 'dispose()' are added for some disposable objects.
If some other code also invokes 'dispose()' for the objects,
and the object is protected from double disposal,
the second 'dispose()' fails with the following error message:
```

src/release/breaking-changes/dispose.md
```diff
@@ -0,0 +1,51 @@
...
+
+`Once you have called dispose() on a <class name>, it can no longer be used.` 
+
+## Background
+
+There is a rule that owner of an object should dispose it.
+
+This rule was broken in some places in Flutter Framework: owners were not disposing the disposable objects.
+The issue was fixed by adding `dispose()`. The added `dispose()` can cause failures in debug mode,
+if there is another `dispose()`, and the object is protected from double disposal.
```
atsansone, line 21: **chore (if-minor):** Use [semantic breaks](https://sembr.org/). Condition, then instruction per [GSG](https://developers.google.com/style/sentence-structure). Write in active voice per [GSG](https://developers.google.com/style/voice).
```suggestion
This rule was broken in some places in Flutter Framework:
owners were not disposing the disposable objects.
To fix the issue, `dispose()` was added.
If there is another `dispose()`, and the object is protected from double disposal,
the added `dispose()` can cause failures in debug mode.
```

src/release/breaking-changes/dispose.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Migration guide
+
+If you got error like `Once you have called dispose() on a <class name>, it can no longer be used.`,
+and the error is raised for your code, 
+update the code to call `dispose()' only in cases when your code created the object.
```
atsansone, line 27: **issue:** Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading).
````suggestion
If the following error was raised for your code, 
update the code to call `dispose()' only in cases when your code created the object.

```nocode
Once you have called dispose() on a <class name>, it can no longer be used.
```
````

**Review by atsansone — APPROVED** [review]
@polina-c : A few suggestions which you can accept or not. LGTM % your review of those suggestions.

File: src/release/breaking-changes/dispose.md [hunk 4]
```diff
@@ -0,0 +1,51 @@
+---
+title: Added missed `dispose()` for some disposable objects in Flutter
+description: >
+  'dispose()' may fail because of double disposal.
```
sfshaza2, line 4:
```suggestion
  'dispose()' might fail because of double disposal.
```

src/release/breaking-changes/dispose.md
sfshaza2: Please fix the grammar:
Missed 'dispose()' are added for some disposable objects.
If some other code also invokes 'dispose()' for the objects,

=>
Missing calls to `dispose()` were added for some disposable objects.
If `dispose()` is called elsewhere on the same object

src/release/breaking-changes/dispose.md
```diff
@@ -0,0 +1,51 @@
...
+
+Missed 'dispose()' are added for some disposable objects. If some other code
+also invokes 'dispose()' for the objects, and the object is protected from double
+disposal, the second 'dispose()' will fail with error message:
+
+`Once you have called dispose() on a <class name>, it can no longer be used.` 
+
+## Background
+
+There is a rule that owner of an object should dispose it.
```
sfshaza2, line 17: Is this a rule or a convention? If a convention:

```suggestion
The convention is that the owner of an object should dispose of it.
```

src/release/breaking-changes/dispose.md
sfshaza2: Please fix the grammar:

=>
This convention was broken in some places:
owners were not disposing the disposable objects.
The issue was fixed by adding a call to `dispose()`.
However, if the object is protected from double disposal,
this can cause failures when running in debug mode
and `dispose()` is called elsewhere on the object.

src/release/breaking-changes/dispose.md
sfshaza2: Please fix the grammar:
If you encounter the following error, then remove
any additional calls in your code that dispose of the object:

```nocode
Once you have called dispose() on a <class name>, it can no longer be used.
```

src/release/breaking-changes/dispose.md
```diff
@@ -0,0 +1,51 @@
...
+if (xIsCreatedByMe) {
+  x.dispose();
+}
+```
+
+If the error is for Flutter code, this means wrong `dispose()` happened first, and the error was raised
+for correct `dispose()` that happened second.
+
+You can locate the wrong `dispose()` by temporary adding `print(StackTrace.current)`
+to the body of the failed method `dispose`.
```
sfshaza2, line 47: It seems like you don't really need this first sentence?
If so, delete. That second sentence is super useful.

```suggestion
If the error occurs in Flutter code, `dispose()` was
called incorrectly the first time.

You can locate an incorrect call to `dispose()` by
temporary calling `print(StackTrace.current)`
in the body of the failed `dispose` method.
```

src/release/breaking-changes/index.md
```diff
@@ -51,6 +50,7 @@ release, and listed in alphabetical order:
 
 ### Released in Flutter 3.13
 
+* [Add missed `dispose()` for some disposable objects in Flutter][]
```
sfshaza2, line 53:
```suggestion
* [Add missing calls to `dispose()` for some disposable objects in Flutter][]
```

src/release/breaking-changes/index.md
```diff
@@ -60,6 +60,7 @@ release, and listed in alphabetical order:
 * [Migrate a Windows project to ensure the window is shown][]
 * [Updated `Checkbox.fillColor` behavior][]
 
+[Add missed `dispose()` for some disposable objects in Flutter]: {{site.url}}/release/breaking-changes/dispose
```
sfshaza2, line 63:
```suggestion
[Add missing calls to `dispose()` for some disposable objects in Flutter]: {{site.url}}/release/breaking-changes/dispose
```

sfshaza2 (commented):
A few grammatical improvements.

src/release/breaking-changes/dispose.md
polina-c: Shams, in most cases, the code should not be removed, but condition should be added. So, it should be fixed, not removed.

src/release/breaking-changes/dispose.md
polina-c: print(StackTrace.current) makes sense if the error is for Flutter code, otherwise it is not needed.
I updated text to made it clearer.

src/release/breaking-changes/dispose.md
```diff
@@ -0,0 +1,60 @@
+---
+title: Added missing `dispose()` for some disposable objects in Flutter
+description: >
+  'dispose()' might fail because of double disposal.
+---
+
+## Summary
+
+Missing 'dispose()' are added for some disposable objects.
```
kenzieschmoll, line 9: nit: Missing calls to 'dispose()'

src/release/breaking-changes/index.md
```diff
@@ -38,7 +38,6 @@ release, and listed in alphabetical order:
 * [Deprecated just-in-time navigation pop APIs for Android Predictive Back][]
 * [Deprecated `Paint.enableDithering`][]
 * [Updated default text styles for menus][]
-* [Windows build path changed to add the target architecture][]
```
kenzieschmoll, line 41: was this intentionally deleted?

kenzieschmoll (approved):
LGTM, but @goderbauer should review before merge

src/release/breaking-changes/index.md
polina-c: returned. thanks

src/release/breaking-changes/dispose.md
```diff
@@ -0,0 +1,60 @@
+---
+title: Added missing `dispose()` for some disposable objects in Flutter
+description: >
+  'dispose()' might fail because of double disposal.
+---
+
+## Summary
+
+Missing calls to 'dispose()' are added for some disposable objects.
```
goderbauer, line 9: For added context it would be nice to list somewhere in this doc examples of where we added previously missing disposes, e.g. "The FooController owned by the BarWidget was missing a dispose" or something like that.

goderbauer (approved):
LGTM

sfshaza2 (approved):
lgtm

PR #9416, opened 2023-09-19
Simplify test setup and fix inaccessible Dart/Flutter in container

parlough:
https://github.com/flutter/website/pull/9307 caused `dart` and `flutter` to not be accessible when running inside the container. This PR attaches the cloned Flutter repository as a volume to allow it to persist and be accessible.

To be compatible with these changes, also modifies the `test.sh` script to be simpler, not allowing switching branches separately from the container. When it's ran from GitHub actions, a new container is created anyway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,8 +21,6 @@ WORKDIR /app
 
 
 # ============== INSTALL FLUTTER ==============
-# NOTE that this will fail if you have not cloned the repo with --recurse-submodules
-# or run `git submodule update --init --recursive` after cloning.
 FROM base AS flutter
 
 COPY ./site-shared ./site-shared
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -32,9 +30,10 @@ ARG FLUTTER_BUILD_BRANCH=stable
 ENV FLUTTER_BUILD_BRANCH=$FLUTTER_BUILD_BRANCH
 ENV FLUTTER_ROOT=flutter
 ENV FLUTTER_BIN=flutter/bin
-ENV PATH="/app/flutter/bin:$PATH"
+ENV PATH="/flutter/bin:$PATH"
```
AlexV525, line 33: I used to remove this, and found out some commands won't be able to find the `flutter`. We should find out how can we rely on `FLUTTER_ROOT`.

parlough: I also had that issue but I think specifying it as a volume in this PR fixed those cases for me. 

Could you try out these changes?

sfshaza2 (approved):
lgtm

atsansone (approved):

PR #9426, opened 2023-09-19
Updated IntelliJ versions on editor page

### LINKED ISSUE #6934: Change "Set up an editor" to recommend "latest stable" version
### Page URL

https://docs.flutter.dev/get-started/editor/

### Page source

https://github.com/flutter/website/tree/main/src/get-started/editor.md

### Describe the problem

As recommended by @sfshaza2 in #6860 
The "Set up an editor" page currently lists specific versions of editors which can become outdated if the page is left untouched for too long.

### Expected fix

Update the page to recommend the "latest stable" version of Android Studio / VSCode rather than a specific version which could become outdated.
We can use this issue to discuss how this might be implemented.

### Additional context

_No response_

atsansone:
Fixes #6934

### DIFF
--- a/src/get-started/editor.md
+++ b/src/get-started/editor.md
@@ -107,11 +107,11 @@ follow Microsoft's instructions for the relevant platform:
 Android Studio and IntelliJ IDEA offer a complete,
 IDE experience once you install the Flutter plugin.
 
-To install one of the following IDEs, follow their instructions:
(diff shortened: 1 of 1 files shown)


parlough (approved):
That's simpler! I've almost always used the EAP and mostly have never had problems, so this seems fine.

PR #9454, opened 2023-09-25
Fixed typo in universal links page

### LINKED ISSUE #9407: Change section 3, step 1 in 'Set up universal links for iOS' to use different copy for app ID
### Page URL

https://docs.flutter.dev/cookbook/navigation/set-up-universal-links/

### Page source

https://github.com/flutter/website/tree/main/src/cookbook/navigation/set-up-universal-links.md

### Describe the problem

in Section 3. Hosting apple-app-site-association file #apple-app-site-association

The page says
`1. Set the appID value to your Flutter application ID.`

The term "Flutter application ID" is incorrect.  "Flutter application ID" is an Android term, and it's equivalent to the iOS "bundle ID". However appID should not be set to either the application ID or the bundle ID, it should be set to the constructed value as described in the preceding section.

### Expected fix

Change the term "Flutter application ID" to "\<team id\>.\<bundle id\>"

### Additional context

_No response_

atsansone:
Fixes #9407

### DIFF
--- a/src/cookbook/navigation/set-up-universal-links.md
+++ b/src/cookbook/navigation/set-up-universal-links.md
@@ -12,13 +12,12 @@ Deep linking is a mechanism for launching an app with a URI. This URI
 contains scheme, host, and path, and opens the app to a specific
 screen.
 
-A _universal link_ is a type of deep link that uses `http` or `https` and
-is exclusive to Apple devices.
+A _universal link_ is a type of deep link that uses `http` or `https`
+and is exclusive to Apple devices.
 
-
-Setting up universal links requires one to own a web domain. Otherwise, consider
-using [Firebase Hosting][] or [GitHub Pages][] as a temporary solution.
-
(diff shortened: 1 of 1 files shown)


src/cookbook/navigation/set-up-universal-links.md
```diff
@@ -140,21 +138,21 @@ The hosted file should have the following content:
 }
 ```
 
-1. Set the `appID` value to your Flutter application ID.
+1. Set the `appID` value to `<team id>.<bundle id>`.
```
parlough, line 141: I think this is fine, but maybe we could just link to and reference the above [App ID](https://docs.flutter.dev/cookbook/navigation/set-up-universal-links/#app-id) header since it gives more explicit details and avoids repeating ourselves.

parlough (approved):
Thanks for the change and for the cleanup!

One suggestion to consider though:

PR #9459, opened 2023-09-27
Update integration test to use find by key

DemX86:
We don't need these keys since we find the widgets by tooltip in the integration test below, not by the keys:
`final Finder fab = find.byTooltip('Increment');`


domesticmouse:
Should we instead update the sample test to use find by key? That seems like a good use case to document.

DemX86:
I have replaced `find.byTooltip` with `find.byKey` method in tests as @domesticmouse suggested.
I have leaved the `Key('counter')` for `Text` widget removed since tests don't use it anyway.

DemX86:
Sorry but I again have difficulties with code excerpts: I tried to update them manually but with no luck — there are still two code excerpts somewhere and I have no idea where to find them.

Script `refresh-code-excerpts.sh` still doesn't work on my WSL:
```
$ sh refresh-code-excerpts.sh
refresh-code-excerpts.sh: 3: set: Illegal option -o pipefail
$ bash refresh-code-excerpts.sh
/usr/bin/env: ‘bash\r’: No such file or directory
```
And I have no option to run it on another machine with clean Linux distributive.

domesticmouse:
Looks like the code isn't compiling straight?

```

  error • The argument type 'String' can't be assigned to the parameter type 'Key' • integration_test/counter_test.dart:20:35 • argument_type_not_assignable
```

DemX86:
I made a mistake with usage of the `find.byKey()` method — I have passed the string as an argument instead of a `Key` object. Sorry for that — editing the code from the GitHub web interface without an IDE was not a great idea.

I fixed that issue but there is still:

> 1 out of 894 fragments needed updating

If you point me to that file with last code excerpt I will fix that too.

DemX86:
Thanks for helping, Brett!

domesticmouse:
@parlough looks like I don't have cloud build permissions?

parlough:
> @parlough looks like I don't have cloud build permissions?

No one has really told me about that work, but I think(?) the cloud build setup is still being worked on. I don't have permission either. For now, it should be fine to ignore the check. \cc @drewroengoogle

parlough (approved):
Thanks for this and for your patience with the excerpts. Looks good to me :) 

Will wait for approval by @domesticmouse too though.

drewroengoogle:
Hi! Sorry about the confusion with the cloud build check. This can be ignored, it is something I am working on (https://github.com/flutter/website/pull/9468) that I accidentally triggered for all presubmit PRs. It has been turned off!

PR #9460, opened 2023-09-27
Delete src/f/flutter-survey-metadata.json

jayoung-lee:
The file is now at 

https://github.com/flutter/uxr/blob/master/surveys/flutter-survey-metadata.json


atsansone (approved):
@jayoung-lee : If the intent is to move this file to UXR, LGTM!

PR #9461, opened 2023-09-27
Delete src/f/contextual-survey-metadata.json

jayoung-lee:
Similar to https://github.com/flutter/website/pull/9460, this file is now at 

https://github.com/flutter/uxr/blob/master/surveys/contextual-survey-metadata.json


atsansone (approved):
@jayoung-lee : If the intent is to move this file to UXR, LGTM!

PR #9475, opened 2023-10-03
Add missed semicolons

DemX86:


atsansone (approved):

atsansone (changes requested):
This edit needs to be made in the Dart file referenced in the code_excerpt tag. If that file and this excerpt don't match, the build throws a blocking error.

DemX86:
@atsansone
The code excerpt for this example was fine already—the problem was with the `replace` regular expression which intents to strip the `return` keyword and the trailing semicolon from the excerpt code.

parlough (approved):
Thanks!

atsansone (approved):

PR #9508, opened 2023-10-12
Better organize testing sidenav contents and fix url

### LINKED ISSUE #9434: Review link order in side menu for the testing pages
### Page URL

https://docs.flutter.dev/cookbook/testing/integration/introduction/

### Page source

https://github.com/flutter/website/tree/main/src/cookbook/testing/integration/introduction.md

### Describe the problem

The problem is in the left side menu (Beyond UI --> Testing & debugging --> Testing)

1. According to the testing overview, I think the order should be Unit testing, Widget testing then Integration testing
2. Links under Integration testing point to the wrong pages. "Introduction" points to "Integration testing" and "Write and run an integration test" points to "Introduction". They should be swapped

### Expected fix

That part of the menu should look like this:

- Beyond UI
  - Testing & debugging
    - Testing
      - Unit testing
        - ...
      - Widget testing
        - ...
      - Integration testing
        - Introduction (links to https://docs.flutter.dev/cookbook/testing/integration/introduction)
        - Write and run an integration test (links to https://docs.flutter.dev/testing/integration-tests)
        - ...

### Additional context

_No response_

No0b-master:
- Fixed the SIdeNav of the testing overview by adding the Widget Testing, Unit testing, and Integration testing inside the testing overview tree 

- Fixed the permalinks of Integration testing children by swapping the links of Introduction and integration test (Previously they are wrong)

Fixes #9434 

## Presubmit checklist

PR_proof


No0b-master:
@parlough i did make a new PR which only focuses on the side nav changes

atsansone:
@No0b-master : This change may run contrary to the recent restructuring of content in the left nav. The goal there was to not nest too many items under too many levels.

@sfshaza2 : Can you provide more insight here?

parlough:
> @parlough i did make a new PR which only focuses on the side nav changes

Thank you for doing that! This will be much easier to review :)

src/_data/sidenav.yml
```diff
@@ -403,31 +403,32 @@
...
-          permalink: /cookbook/testing/integration/introduction
-        - title: Profile an integration test
-          permalink: /cookbook/testing/integration/profiling
-    - title: Unit testing
-      children:
-        - title: Introduction
-          permalink: /cookbook/testing/unit/introduction
-        - title: Mock dependencies
-          permalink: /cookbook/testing/unit/mocking
+        - title: Widget testing
```
parlough, line 407: I am curious why you nested these three sections under "Testing overview"? Could you explain more on what you're trying to achieve?

The "Testing overview" page serves as sort of an introduction to everything under the bold **Testing** header and before the **Debugging** header. Also, if it's given children, it won't be openable from the sidenav.

src/_data/sidenav.yml
```diff
@@ -403,31 +403,32 @@
...
+              permalink: /cookbook/testing/widget/scrolling
+            - title: Simulate user interaction
+              permalink: /cookbook/testing/widget/tap-drag
+        - title: Integration testing
+          children:
+            - title: Introduction
+              permalink: /cookbook/testing/integration/introduction
+              match-page-url-exactly: true
+            - title: Write and run an integration test
+              permalink: /testing/integration-tests
```
parlough, line 423: Thanks for swapping these! Taking a closer look however, I remember now that there is still a lot of overlap between these two pages.

What do you think about removing the "/testing/integration-tests" one from the sidenav for now in this pull request, then a separate issue can be made for updating and consolidating these pages?

```suggestion
            - title: Introduction
              permalink: /cookbook/testing/integration/introduction
              match-page-url-exactly: true
```

src/_data/sidenav.yml
```diff
@@ -403,31 +403,32 @@
...
+            - title: Profile an integration test
+              permalink: /cookbook/testing/integration/profiling
+        - title: Unit testing
+          children:
+            - title: Introduction
+              permalink: /cookbook/testing/unit/introduction
+            - title: Mock dependencies
+              permalink: /cookbook/testing/unit/mocking
```
parlough, line 431: I see the [original issue](https://github.com/flutter/website/issues/9434) suggestd moving "Unit testing" first, up before "Integration testing". I am thinking that makes sense and it is consistent with what is on the [overview page](https://docs.flutter.dev/testing/overview). What do you think?

parlough (commented):
Thanks so much for being patient on the review and getting this new PR up and running :)

I have a few comments and questions for you to consider. Let me know what you think or if you need any help! Thanks again.

_Sidenote: I also really appreciate the thorough PR description and screenshot. That's a great practice to have._

src/_data/sidenav.yml
No0b-master: I also thought what was the point of putting Unit testing, Widget testing, and integration testing inside the testing overview,
but the issue states that these tests should be under a subsection so I think what we should do is put  Unit testing, Widget testing, and integration testing inside the Testing section as their children, In this case the testing overview will come as the child of Testing subsection, Let me know what you think

src/_data/sidenav.yml
No0b-master: I think this should remain in this PR because the issue made #9434, was made to capture this problem also, so making another issue with the same description will create confusion

src/_data/sidenav.yml
No0b-master: Yeah this can be done, I'll commit this change

No0b-master: @parlough I've changes the order and committed the changes , please have a look

src/_data/sidenav.yml
droidbg: @No0b-master I guess the issue does not stated to put it under a subsection, first point was about order in which it is displayed in nav bar. 

`1. According to the testing overview, I think the order should be Unit testing, Widget testing then Integration testing`

parlough: Yeah I think it was meaning below the **Testing** subheader, not necessarily as children. We added that to avoid too much nesting/dropdowns which makes content a bit hard to find. Also this change currently results in **Testing overview** being unable.

At least for this PR, would you mind not nesting these as children? Thanks so much!

sfshaza2:
Yes, thanks, @No0b-master, for providing a screenshot of your changes! I agree with Parker that when we re-organized the sidenav recently, that we were trying to avoid deep nesting of topics. Adding the testing sections as children doesn't really add clarity.

atsansone:
@sfshaza2 : Should we close this, then?

No0b-master:
@parlough I have removed the nest under testing, and it  looks like this now 

Screenshot 2023-11-10 at 8 28 47 AM

parlough (approved):
Looks good to me! 

Thanks for working on this and those adjustments! I appreciate your help and patience :D

PR #9680, opened 2023-10-20
Standardize example analysis and formatting

parlough:
Sets up an `example_utils` package at the root of `/examples` which all nested examples now depend on and use as the source for their analysis. Uses `package:flutter_lints` v3 and enables a few new lints, then fixes resulting code where necessary and completes some other nearby cleanup while there.


parlough:
**@parlough's blog:**

_My ability to type lessens everyday..._

examples/example_utils/pubspec.yaml
```diff
@@ -0,0 +1,9 @@
+name: example_utils
```
domesticmouse, line 1: I'm guessing you'll need to add this project to `.github/dependabot.yml` for the next time `flutter_lints` gets updated?

domesticmouse (approved):

domesticmouse:
> **@parlough's blog:**
> 
> _My ability to type lessens everyday..._

You and me both ;-)

examples/example_utils/pubspec.yaml
parlough: Ah thanks for reminding me! I got distracted by fixing diagnostics that I forgot. Done now!

PR #9706, opened 2023-10-26
 Add `flutter pub get` to "Internationalizing Flutter apps" that also generates the translation files

AlexanderThiele:
calling `flutter pub get` also generates the translation files so that `flutter run` is not the only solution to generate the files.

## Presubmit checklist

### DIFF
--- a/src/ui/accessibility-and-internationalization/internationalization.md
+++ b/src/ui/accessibility-and-internationalization/internationalization.md
@@ -262,7 +262,7 @@ complete the following instructions:
    }
    ```
 
-6. Now, run `flutter run` and codegen takes place automatically.
+6. Now, run `flutter pub get` or `flutter run` and codegen takes place automatically.
    You should find generated files in
    `${FLUTTER_PROJECT}/.dart_tool/flutter_gen/gen_l10n`.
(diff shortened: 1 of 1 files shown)


parlough (commented):
Thanks so much for the pull request!

I'm curious what version of Flutter you're running, because I think `flutter pub get` doesn't consistently generate localizations on the current stable release (https://github.com/flutter/flutter/commit/312ef541156c0b0c8553626678e37e551b99383a).

If that's the case, we might want to hold off on documenting this until the next stable release to avoid any potential confusion. What do you think?

AlexanderThiele:
Hey @parlough,

currently working with: Channel stable, 3.13.8, on macOS 14.0 23A344 darwin-arm64.

I just added localizations to my app and noticed, that `flutter pub get` is enough to create the `.dart_tool` folder including its `flutter_gen` files.

also important: You have to add `generate: true` as described in step 2 above the change to make it work with `flutter pub get`.
```yaml
flutter:
  generate: true
```

parlough (approved):
Thanks for confirming the version. Sounds good to me then!

Thanks for thinking to add this and for the PR :)

PR #9709, opened 2023-10-26
Add go link for "File System Access from DevTools" design doc

kenzieschmoll:
Adds go link for doc go/file-system-access-from-devtools


parlough (approved):
Thank you!

PR #9715, opened 2023-10-27
Fix find command for Linux install

### LINKED ISSUE #9714: Update command in the 'Linux Installation Guide' page
### Page URL

https://docs.flutter.dev/get-started/install/linux/

### Page source

https://github.com/flutter/website/tree/main/src/get-started/install/linux.md

### Describe the problem

The command

`find / -type d -name "flutter/bin" 2>/dev/null`

returns the warning bellow so it will never find the path:

`find: warning: ‘-name’ matches against basenames only, but the given pattern contains a directory separator (‘/’), thus the expression will evaluate to false all the time.  Did you mean ‘-wholename’?`

### Expected fix

The following command should work correctly:

`find / -type d -wholename "*/flutter/bin" 2>/dev/null`

### Additional context

My machine is Debian 11

atsansone:
Fixes #9714

### DIFF
--- a/src/get-started/install/_path-linux-chromeos.md
+++ b/src/get-started/install/_path-linux-chromeos.md
@@ -26,17 +26,17 @@ To add Flutter to your `PATH` for _any_ terminal session,
 follow these steps:
 
 1. Find your Flutter SDK installation path.
-   
+
     ```terminal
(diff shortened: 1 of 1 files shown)


parlough (approved):
Thanks!

PR #9735, opened 2023-10-31
Feature 1

Braumski:
delet stuff


parlough (changes requested):
Hi @Braumski, thanks for the pull request.

However, it's not clear what you're trying to achieve with the PR, as it removes files we rely on. Could you provide some details?

I'm going to close this as we can't review it without the CLA signed [as outlined in the above comment](https://github.com/flutter/website/pull/9735#issuecomment-1787107327) and I don't expect we'll want to delete all of these files.

If you need help contributing or have something specific you'd like to work on, let us know and we'd be happy to help. Thanks!

PR #9737, opened 2023-10-31
Add info on group test function

### LINKED ISSUE #5761: Explain, link `group` function on 'An introduction to unit testing' page
Page URL: https://docs.flutter.dev/cookbook/testing/unit/introduction
Page source: https://github.com/flutter/website/tree/master/src/docs/cookbook/testing/unit/introduction.md

Description of issue:
It's not really clear (at least for me) what is the use case of group keyword.
In the example, Counter is being initiated for each test so how those test are related if they don't share the same Counter instance?
I don't understand the benefit of grouping the tests by reading this document.

Also, What I would like to find on this page (and eventually found by digging the source code) is the setUp and tearDown functions.

As most Android users will have experience with those functions it worth mentioning them.
Hope it makes sense :)

atsansone:
Fleshes out the `group` function for `flutter_test`.

Fixes #5761


examples/cookbook/testing/unit/counter_app/test/group.dart
```diff
@@ -1,25 +1,21 @@
-import 'package:counter_app/counter.dart';
-import 'package:test/test.dart';
+import "package:counter_app/counter.dart";
```
domesticmouse, line 1: I'm curious about this change. I thought we had prefer single quotes lint rule in place. If not, we should add it.

examples/cookbook/testing/unit/counter_app/lib/counter.dart
```diff
@@ -1,7 +1,5 @@
 class Counter {
   int value = 0;
-
```
domesticmouse, line 3: Why are we deleting these blank lines? I believe these add to readability in this instance.

examples/cookbook/testing/unit/counter_app/test/counter_test.dart
```diff
@@ -5,9 +5,7 @@ import 'package:test/test.dart';
 void main() {
   test('Counter value should be incremented', () {
     final counter = Counter();
-
```
domesticmouse, line 8: Ditto

domesticmouse (changes requested):

examples/cookbook/testing/unit/counter_app/test/group.dart
parlough: We do have the lint added, CI caught it.

atsansone:
```suggestion
import 'package:counter_app/counter.dart';
```

examples/cookbook/testing/unit/counter_app/test/group.dart
```diff
@@ -1,25 +1,21 @@
 import 'package:counter_app/counter.dart';
-import 'package:test/test.dart';
+import "package:test/test.dart";
```
sfshaza2, line 2: No, we use single quotes for import. Please change.

sfshaza2 (approved):
I agree with @domesticmouse about the empty lines. They aid in readability. Also, we use single quotes for imports, not double. Please fix.
I am approving because I don't need to review again, but please at least fix the quotes.

sfshaza2:
Interesting. I didn't realize that the double quotes on an import would break the build. Good to know.

domesticmouse:
> Interesting. I didn't realize that the double quotes on an import would break the build. Good to know.

Linters gunna lint!

examples/cookbook/testing/unit/counter_app/lib/counter.dart
atsansone: I removed them because this did not seem to be consistent with other examples. I'll revert this change.

atsansone:
@domesticmouse : Back to you. The extra lines, though, appear to be inserted somewhat randomly and inconsistent with other pages. We can review that issue at another time.

domesticmouse (approved):

Issue #9743, opened 2023-11-01
Refactor Toolkits as TL section


atsansone:

parlough:
Fixed by https://github.com/flutter/website/pull/9744 :)

PR #9744, opened 2023-11-02
Introduce a new top-level sidenav category for toolkit content

parlough:
This work is primarily to introduce a new top-level sidenav category for toolkits so we can more prominently link to the corresponding cookbook recipes.

However, the two toolkits we have are not the only unique goals/app categories we should cover and I doubt anyone would look under "Toolkits" for this kind of content. So I am proposing a new top-level entry called "Build by category". To showcase what could live here, I also added Monetization and Maps. While we don't have content on the docs website for these areas yet we have great content elsewhere that's currently not as discoverable from the site. Perhaps this will give it a chance to shine!

Note that this doesn't change the file structure, at least not yet, so we can easily make adjustments by removing or adding links.

**Looking for feedback:**

- Does the name make sense to you?
  - Would something like "Use cases", "Toolkits", or something else be better?
- Does the location in the sidenav make sense?
  - Should it instead be at the bottom of "Beyond UI", under "Resources", next to "Samples", etc?
- Does the initial content I added seem worthwhile, despite much of it being external?
  - Would it be better to just have the toolkits?
  - What other subcategories make sense to add that we have content for?
  - Am I missing any high quality content to link to?
- How are the names of the individual entries?

**Staged:** https://flutter-docs-prod--pr9744-feature-build-by-cat-pbqz0954.web.app/resources/games-toolkit

_To find the changes on the staged site, scroll down in the left sidenav. The new category is currently directly under "Beyond UI"._

Fixes https://github.com/flutter/website/issues/9743

### DIFF
--- a/src/_data/sidenav.yml
+++ b/src/_data/sidenav.yml
@@ -254,6 +254,29 @@
 - divider
 - header: Beyond UI
 
+- title: Build by category
+  children:
+    - title: Games
+      children:
+        - title: Create a casual game
+          permalink: /resources/games-toolkit
+        - title: Build a game with Flame
+          permalink: https://codelabs.developers.google.com/codelabs/flutter-flame-game
+    - title: News
+      children:
+        - title: Build a news app
+          permalink: /resources/news-toolkit
+    - title: Monetization
+      children:
+        - title: Overview
+          permalink: https://flutter.dev/monetization
+        - title: Add in-app purchases
+          permalink: https://codelabs.developers.google.com/codelabs/flutter-in-app-purchases#0
+    - title: Maps & navigation
+      children:
+        - title: Add maps to your app
+          permalink: https://developers.google.com/maps/flutter-package
+
 - title: Data & backend
   permalink: /data-and-backend
   children:
@@ -620,11 +643,6 @@
       permalink: https://github.com/flutter/flutter/blob/main/CONTRIBUTING.md
     - title: Design documents
       permalink: /resources/design-docs
-    - header: Toolkits
-    - title: Casual Games Toolkit
-      permalink: /resources/games-toolkit
-    - title: Flutter News Toolkit
-      permalink: /resources/news-toolkit
 
 - title: Reference
   permalink: /reference
--- a/src/_data/sidenav.yml
+++ b/src/_data/sidenav.yml
@@ -61,6 +61,29 @@
     - title: Codelabs
       permalink: /codelabs
 
+- title: App solutions
+  children:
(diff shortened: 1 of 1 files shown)


parlough:
\cc @zoeyfan and @InMatrix - I'd appreciate your feedback and ideas here as well! Thanks 😄

atsansone (approved):
@parlough : This is what I was looking for. @zoeyfan , your thoughts?

zoeyfan:
Thank you so much for starting this work. Bringing it to top level definitely has improved its visibility. Meanwhile, can I make some suggestions in terms of the location and category name? 

1) The name "Build by category" feels a bit too general. I wonder if we can call it "Industry solutions". Coincidentally, I found GCP also has the same naming for industry specific solution. E.g. https://cloud.google.com/solutions#industry-solutions

2. Monetization doesn't feel like its a part of industry solutions. Can monetization become a standalone category under "Beyond UI"?

3. Another minor nitpick: technically speaking, the industry solution (vertical toolkit) still has a lot of UI components, does it fit under Beyond UI? I also understand that it probably doesn't fit into the "User Interface" section. Maybe industry solutions can be directly below "Samples and codelabs", so we have 4 things in the first section?

parlough:
Thanks for the feedback @zoeyfan!

In response to your suggestions and after discussing further with @sfshaza2 I've made some changes:

> I wonder if we can call it "Industry solutions"

Shams and I agreed this isn't what developers that use the docs will look for, especially for games. We came up with "App solutions" and found it was a good balance. What do you think?

> Can monetization become a standalone category under "Beyond UI"?

For now we'd like to avoid doing this as we don't have sufficient docs for monetization yet to be its own top-level category. Do you think it's fine to keep "Monetization" under "App solutions" for now? We can always easily pull it out if we introduce more content.

> Maybe industry solutions can be directly below "Samples and codelabs", so we have 4 things in the first section?

This makes sense, especially for the toolkits content! I've moved the section there.

<br>

Thanks again for the feedback. Let me know what you think and we can get this landed :)

zoeyfan:
@parlough Thank you for replying. 
* I am okay with "App solutions" for now, and happy to revisit it again. 
* Regarding "monetizations", there are actually a few other resources we can link: 
1) New recipe we are currently building: https://docs.flutter.dev/cookbook/games/google-mobile-ads
2) https://codelabs.developers.google.com/codelabs/admob-ads-in-flutter?hl=en#0
Given we now have 3 resources, would that change your mind about making it a standalone category?

PR #9752, opened 2023-11-02
Refactor the cloudbuild template

drewroengoogle:
NOTE: Given the current status of cloud build logs not publishing to github reliably, this will not be enabled after this PR. The intention of this PR is to clean up the current script to make it more maintainable.

Refactors the current cloudbuild template for staging
* Moves the cloudbuild template to the `cloud_build` directory
* Adds a custom docker image that can be built and deployed to gcr.io regularly, which includes `gh-cli` and `firebase-tools`. This abstracts a lot of the "magic" of installing these tools into a docker image, and also saves the time of having to reinstall them on the docker image every single time this staging job runs.
* Refactors the current template by moving the logic to scripts, and then call the scripts from the cloudbuild template. The scripts themselves have also been organized for reability.


drewroengoogle:
@parlough This PR is based on our conversation on some potential improvements for maintainability. Let me know what you think, and if there are any ways to make this better!

atsansone (approved):
LGTM. Question: Is it worth building an image with Packer to eliminate other install needs (maybe replaced with update calls)?

drewroengoogle:
> LGTM. Question: Is it worth building an image with Packer to eliminate other install needs (maybe replaced with update calls)?

Would you be able to explain a bit more for my understanding? I don't have direct experience with Packer but my understanding is that it is best for creating images across multiple platforms, which I don't believe is something we do in the current checks.

cloud_build/staging_container/Dockerfile
parlough, line 1: I I like that this is in the repository now, definitely less magic, which is great! Thank you.

However, I don't fully understand the "can be built and deployed to gcr.io regularly" portion. Is this rebuilt when it changes or does someone have to manually run the build through the cloud console?

Either way, could you document this inline with a `README.md` in this directory? If there's an internal reference/doc as well, consider linking to that as well with "If you're a Googler, learn more at ...".

cloud_build/scripts/stage_site_and_comment_on_github.sh
parlough, line 1: I'm glad you were able to pull this out into a separate file. It's much easier to read and review! 💯

cloud_build/scripts/stage_site_and_comment_on_github.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -e
+
+deploy_to_firebase_staging_channel () {
+    echo "Deploying website to a staging channel on firebase..."
+
+    FIREBASE_DEPLOY_RESPONSE=$(firebase hosting:channel:deploy --expires 7d pr$PR_NUMBER-$HEAD_BRANCH --project=$PROJECT_ID)
+    echo "$FIREBASE_DEPLOY_RESPONSE"
+    FIREBASE_STAGING_URL=$(grep -Eo "https://$PROJECT_ID--[a-zA-Z0-9./?=_%:-]*" <<< "$FIREBASE_DEPLOY_RESPONSE")
```
parlough, line 9: This is much easier to understand for me, thank you!

Could you add a comment above the line explaining what it's doing though. The grep and regex might confuse others in the future.

parlough (approved):
This is all amazing @drewroengoogle, thanks so much!

I have a few asks related to documentation, but don't let them prevent you from merging this.

sfshaza2 (approved):
lgtm. Thanks, @drewroengoogle!

atsansone:
> > LGTM. Question: Is it worth building an image with Packer to eliminate other install needs (maybe replaced with update calls)?
> 
> Would you be able to explain a bit more for my understanding? I don't have direct experience with Packer but my understanding is that it is best for creating images across multiple platforms, which I don't believe is something we do in the current checks.

You could use Packer (https://www.packer.io/) to build a Docker image with everything already installed. That way, your script could run bundle update, etc. to be ready to build. Not having to install all of the base packages and extensions should save a bit of time.

cloud_build/scripts/stage_site_and_comment_on_github.sh
drewroengoogle: Added!

cloud_build/staging_container/Dockerfile
drewroengoogle: For now I put this Dockerfile in this repository because I'm not sure of a better place to put it, but it can be used in dart.dev eventually as well, since it has the tools needed to deploy to firebase and comment. It currently is manually triggered, but once there is a more solid place to put this, we can set it up to automatically build and push to artifact registry when any changes related to the Dockerfile are committed. I've added a README to explain this for now!

drewroengoogle:
> > > LGTM. Question: Is it worth building an image with Packer to eliminate other install needs (maybe replaced with update calls)?
> > 
> > 
> > Would you be able to explain a bit more for my understanding? I don't have direct experience with Packer but my understanding is that it is best for creating images across multiple platforms, which I don't believe is something we do in the current checks.
> 
> You could use Packer (https://www.packer.io/) to build a Docker image with everything already installed. That way, your script could run bundle update, etc. to be ready to build. Not having to install all of the base packages and extensions should save a bit of time.

That might be a worthwhile future improvement. This PR contains a Dockerfile which installs some of the base requirements for the website staging, so those no longer need to be installed each time.

parlough (approved):
Thanks so much for the extra documentation!

cloud_build/staging_container/Dockerfile
parlough: Sounds great to me, thanks!

PR #9757, opened 2023-11-03
Add wildcard

filiph:
The current wording makes it look as if there are literally `.entitlements` files (i.e. hidden files, such as `.gitignore`).

But in fact, they are files like "DebugProfile.entitlements" and "Release.entitlements". I think the wildcard makes it clearer what is meant.

## Presubmit checklist

### DIFF
--- a/src/data-and-backend/networking.md
+++ b/src/data-and-backend/networking.md
@@ -15,7 +15,7 @@ Some platforms require additional steps, as detailed below.
 ### Android
 
 Android apps must [declare their use of the internet][declare] in the Android
-manifest (`AndroidManifest.xml `):
+manifest (`AndroidManifest.xml`):
 
 ```
 <manifest xmlns:android...>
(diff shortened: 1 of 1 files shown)


src/data-and-backend/networking.md
```diff
@@ -27,7 +27,7 @@ manifest (`AndroidManifest.xml `):
 
 ### macOS
 
-macOS apps must allow network access in the relevant `.entitlements` files. 
+macOS apps must allow network access in the relevant `*.entitlements` files.
```
domesticmouse, line 30: Should we be more specific with actual file names or link to more detail elsewhere?

domesticmouse (approved):

src/data-and-backend/networking.md
filiph: There's already a link to more details below. I think this is okay. I just threw me as I was looking for an `.entitlements` file in a hurry.

parlough (approved):
Makes sense, thanks!

PR #9764, opened 2023-11-05
Increase contrast of secondary aside

parlough:
Accomplishes this by slightly lightening the background color. Helps especially for the blueish code text, but also slightly for normal text as well.


domesticmouse (approved):

sfshaza2 (approved):
lgtm

PR #9776, opened 2023-11-09
Ads tweaks

filiph:
Leftover tweaks to the new ads codelab.

I merged https://github.com/flutter/website/pull/9730 before adding the changes suggested by @parlough. The ones I forgot are tweaks such as two instances of linking classes to their API docs.

The one change that needs explanation is removing `unawaited` from the initialization code after it was suggested and added by @parlough. Here's the commit description:

> I belatedly realized that `unawaited()` requires an additional import. For simplicity, I’m going to omit `unawaited` for this recipe. I’m going to assume that people who know `await_futures` (because they turned it on), know how to deal with the lint warning.

It's not ideal, but in my opinion, it's better than complicating the recipe with yet another small step that just adds `import 'dart:async';` — which is something that is irrelevant to ads.
 

## Presubmit checklist

### DIFF
--- a/src/cookbook/plugins/google-mobile-ads.md
+++ b/src/cookbook/plugins/google-mobile-ads.md
@@ -137,19 +137,17 @@ $ flutter pub add google_mobile_ads
 
 ## 4. Initialize the Mobile Ads SDK
 
-You need to initialize the Mobile Ads SDK before loading ads.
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):

domesticmouse:
PTAL @atsansone

parlough (approved):
Makes sense to me. Thanks!

PR #9786, opened 2023-11-12
Add `/go/macros-for-leak-tracker` doc redirect

polina-c:


parlough (changes requested):
Thanks for adding this! Looks like the doc isn't available publicly yet though. Would you mind sharing it so anyone with the link can add suggestions?

polina-c:
> Thanks for adding this! Looks like the doc isn't available publicly yet though. Would you mind sharing it so anyone with the link can add suggestions?

Done. Thanks for checking.

parlough (approved):
Thank you for updating that!

PR #9804, opened 2023-11-14
Update Flutter supported-platforms for Flutter 3.16

stef-luo:
Update the Flutter supported platform versions for Flutter 3.16. 

**Staged:** https://flutter-docs-prod--pr9804-stef-luo-patch-1-c1mp1znq.web.app/reference/supported-platforms


src/reference/supported-platforms.md
```diff
@@ -40,10 +40,10 @@ following table:
 |Platform version|Supported|Best effort|Unsupported|
 |----------------|---------|-----------|-----------|
 | Android SDK    |21-34|19-20|18-|
-| iOS            |16|11-15|10-, arm7v 32-bit|
+| iOS            |16|11-15, 17|10-, arm7v 32-bit|
```
domesticmouse, line 43: iOS 17 should be supported, not best effort.

src/reference/supported-platforms.md
```diff
@@ -40,10 +40,10 @@ following table:
...
 | Linux Debian   |10-12|9-|any 32-bit|
-| Linux Ubuntu   |18.04 LTS|20.04-23.04|any 32-bit|
-| macOS          |Ventura (13+)|Mojave (10.14) to Monterey (12)|High Sierra (10.13-) |
+| Linux Ubuntu   |20.04 LTS|20.04-23.04|any 32-bit|
```
domesticmouse, line 45: This is confusing, 20.04 is both Supported and Best Effort, while 18.04 is now neither supported, nor best effort, or even unsupported.

src/reference/supported-platforms.md
```diff
@@ -40,10 +40,10 @@ following table:
...
+| macOS          |Ventura (13+)|Mojave (10.14) to Monterey (12), Sonoma (14)|High Sierra (10.13-) |
```
domesticmouse, line 46: Sonoma should be supported, not best effort.

parlough: This came from Chris. I think it is generally about the CI testing we have done for the versions.

src/reference/supported-platforms.md
parlough: Same here, came from the iOS team, I think because the release had minimal CI testing on iOS 17.

src/reference/supported-platforms.md
parlough: Oops, the best effort was supposed to be updated to `20.10-23.04`.

src/reference/supported-platforms.md
domesticmouse: Eeek, that needs to be fixed pronto. I'll hassle Chris.

src/reference/supported-platforms.md
domesticmouse: iOS 17 is apparently already installed on a 1/3rd of devices. https://telemetrydeck.com/blog/ios-market-share-10-23/

src/reference/supported-platforms.md
parlough: As for 18.04, it has reached its end of its standard LTS support. Do you think it's worth adding to the `Unsupported` column or do you think it can be gleamed from not being in the first two columns?

src/reference/supported-platforms.md
parlough: Yeah I definitely see why you see it as important to have it there. Developers would definitely like support to be signaled.

We just currently define **Supported** as "Google-tested platforms that are automatically tested on every commit by continuous integration testing."

Since 3.16 was cut before iOS 17 release, I'm just assuming it had limited automatic testing on the platform version. I'm not too familiar with the Flutter testing setup nor these support levels. Maybe we can pull some others in who are.

Do you think "Best effort" is too off putting? Maybe we can consider different names for the levels of support.

sfshaza2:
@domesticmouse, @parlough, @stef-luo, thanks for being on top of this!

PR #9807, opened 2023-11-14
Move 3.16 migration guides to 3.16 in index

parlough:


sfshaza2 (approved):
lgtm. Thanks for cleaning up breaking changes!

PR #9812, opened 2023-11-15
Move linkcheck workflow out of Docker-based build

parlough:
The Cloud Build staging workflow was flaky, most likely due to the linkcheck workflow.

To maintain logs, we also run the linkcheck workflow on PRs, so we can safely remove it from staging.

To accomplish this, this PR separates link checking from the primary site build so we can run just the build during staging and run the linkcheck after build.

This has the added benefit of reducing both staging time and general setup time of the container.


domesticmouse (approved):

.github/workflows/build.yml
```diff
@@ -60,9 +82,13 @@ jobs:
...
           node-version: ${{ env.NODE_VERSION }}
       - name: Install Firebase CLI
-        run: npm install -g firebase-tools@12.7.0
-      - name: Deploy to Firebase hosting
-        run: make deploy
+        run: npm install -g firebase-tools@12.9.1
+      - uses: dart-lang/setup-dart@b64355ae6ca0b5d484f0106a033dd1388965d06d
+        with:
+          sdk: stable
+      - run: tool/check-links.sh
```
drewroengoogle, line 89: For my understanding, we want to check the links both when the PR is opened and staging is triggered, and when it is merged?

drewroengoogle (approved):

.github/workflows/build.yml
parlough: That's what has been done traditionally. It's not super essential but it can help occasionally since we don't use a merge queue currently.

PR #9845, opened 2023-11-22
Fix list

filiph:
Looks like the last commit removed space between the numeral and the text, so it's no longer recognized as a list.

See https://docs.flutter.dev/cookbook/games/firestore-multiplayer, the first list ("High tick rate" vs "Low tick rate").


sfshaza2 (approved):
lgtm. Sorry about that. <ahem>

filiph:
No problem at all. 

FWIW, I think the build seems to be failing on an unrelated test. No hurry to merge here.

domesticmouse (approved):

parlough (approved):
Thank you!

PR #9864, opened 2023-11-28
Add sections to Flavors page about flavor-conditional asset bundling

andrewkolos:
This PR introduces additional documentation for a new feature, the conditional bundling of assets based on app flavor, which will be introduced by https://github.com/flutter/flutter/pull/132985.

I've decided to incorporate the documentation into the flavors page, as I believe this is where this new feature would be most relevant. I do not suspect many users would start using the flavors feature specifically for this new feature.

I've also tried to keep this documentation simple, explaining how to use the feature entirely through an example (similar to how the [Adding assets and images](https://docs.flutter.dev/ui/assets/assets-and-images) works).

## Presubmit checklist

### DIFF
--- a/src/deployment/flavors.md
+++ b/src/deployment/flavors.md
@@ -244,8 +244,7 @@ If you aren't familiar with how to add assets to your app, see
 
 If you have assets that are only used in a specific flavor in your app, you can
 configure them to only be bundled into your app when building for that flavor.
-You can utilize this to prevent your app bundle sizes from being bloated by
-unused assets.
+This prevents your app bundle size from being bloated by unused assets.
 
 Here is an example:
 
--- a/src/deployment/flavors.md
+++ b/src/deployment/flavors.md
@@ -242,7 +242,7 @@ check out the integration test samples in the [Flutter repo][].
 If you aren't familiar with how to add assets to your app, see
 [Adding assets and images][].
 
-If you have assets that are only used in specific flavor in your app, you can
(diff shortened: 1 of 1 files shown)


gnprice:
> I've decided to incorporate the documentation into the flavors page, as I believe this is where this new feature would be most relevant. I do not suspect many users would start using the flavors feature specifically for this new feature.

This seems reasonable to me.

I think it would be important, though, to include at least a mention of this feature at https://docs.flutter.dev/ui/assets/assets-and-images#specifying-assets . That's the place that otherwise serves to document what the syntax is for the `flutter: assets: …` part of the pubspec.yaml file, so it's valuable for that to cover (either directly or indirectly) all the features that can appear there.

In particular, consider the case of a user who's maintaining a project they didn't themself create from scratch, and they're trying to understand how the assets configuration works. Their starting point is "how do assets in a Flutter pubspec work", not "how do I use flavors", so they'll find that page with its "Specifying assets" section. If the pubspec file they're looking at has this `path:`/`flavors:` syntax, and they don't see anything resembling that there, then they'd be frustrated that the docs are incomplete.

sfshaza2 (approved):
I made a few tiny tweaks, but this looks good, @andrewkolos! Can you also add the link in Assets & images to here? I'll go ahead an approve so you can land it when ready.

atsansone:
@andrewkolos : This looks good. If you take it out of draft and update to the base branch, you can merge.

PR #9880, opened 2023-11-30
Remove games cookbook index redirect

parlough:
We now have an index to place at this location with links to the recipes, so we don't need the redirect anymore.


sfshaza2 (approved):
Woot!! lgtm!

PR #9914, opened 2023-12-11
Add `title` attributes to all `iframe` videos per WCAG guidelines

### LINKED ISSUE #9528: Add titles to embedded iframes (YouTube)
Add `title` attribute to all `iframe` tags throughout the site. These most commonly appear with YouTube video embeds. The iframes without this attribute violates WCAG [H64](https://www.w3.org/TR/WCAG20-TECHS/H64).

**Error:** Iframe element requires a non-empty title attribute that identifies the frame.
[htmlcs: WCAG2AA.Principle2.Guideline2_4.2_4_1.H64.1](https://squizlabs.github.io/HTML_CodeSniffer/Standards/WCAG2/#sc2_4_1)

**Example:**
```html
<iframe style="max-width: 100%; width: 100%; height: 230px;" src="https://youtube.com/embed/xWV71C2kp38" frameborder="0" allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowfullscreen=""></iframe>
```

**Location:**
(selector "html > body > div:nth-child(6) > div > main > div > div:nth-child(13) > div:nth-child(1) > div > iframe")

atsansone:

Add `title` attributes to all `iframe` videos per WCAG guidelines. Also adds lazy loading and standardizes other attributes as well.

Fixes #9528 

## Presubmit checklist


domesticmouse:
CI is cranky:

```console
check_link_references: Invalid link references found!

codelabs/index.html:
[null-safe][]

packages-and-plugins/using-packages/index.html:
[<em>caret syntax</em>][]
```

domesticmouse (approved):

src/add-to-app/android/project-setup.md
```diff
@@ -59,6 +58,7 @@ android {
 <div class="tab-pane active" id="with-android-studio" role="tabpanel" aria-labelledby="with-android-studio-tab" markdown="1">
 
 ### Integrate with Android Studio
+
```
parlough, line 61: These `{:.no_toc}` entries can't have the space before them, as otherwise the class won't apply to the header. Mind reverting these changes?

src/_includes/docs/yt_shims.liquid
parlough, line 1: Is there a reason for these to be an included liquid file over a global site variable in the `_config.yml`?

parlough (commented):
Thanks for working on this @atsansone! One issue and a question:

src/add-to-app/android/project-setup.md
atsansone: My Markdown linter is doing this on save and I can't turn it off, but I did fix it from making this change. I'll revise.

src/_includes/docs/yt_shims.liquid
atsansone: I understand this, but I had two reasons:

1. It's not needed for the majority of files (~50 of ~475).
2. Shorter replacements would be more helpful. Otherwise, I'm making a declaration at the start to `assign` shorter replacement for each file.

src/add-to-app/android/project-setup.md
parlough: With the 11ty work, these will be on the same line as the header, so hopefully the linter can handle that.

I turned off fixing lints of save for mine, as it's too disruptive and not necessarily all correct.

src/_includes/docs/yt_shims.liquid
parlough: I raise this question as it's not really convertible between different static site generators and templating engines, and the goal is to move away from `include`s eventually due to their problematic rules.

I think it's fine for now since these shouldn't have to change much in the short term, but note I'll likely have to convert this to a different, more standard implementation and syntax with the migration.

Issue #6424, opened 2021-10-27
Add point about initializing localization objects on 'Internationalizing Flutter apps' page


akiross:
### Page URL

https://docs.flutter.dev/accessibility-and-localization/internationalization

### Page source

https://github.com/flutter/website/tree/main/src/docs/development/accessibility-and-localization/internationalization.md

### Describe the problem

In the page, step 8, it is mentioned that

> 8. Use AppLocalizations anywhere in your app. Here, the translated message is used in a Text widget.

but I believe this is a bit misleading, because "anywhere" led me to naively think that this code was valid:

```dart
import 'package:flutter_gen/gen_l10n/app_localizations.dart';

void main() => runApp(MyWidget());

class MyWidget extends StatelessWidget {
    @override
    Widget build(BuildContext context) {
        return MaterialApp(
            localizationsDelegates: AppLocalizations.localizationsDelegates,
            supportedLocales: AppLocalizations.supportedLocales,
            home: Text(AppLocalizations.of(context)!.helloWorld));
    }
```

But it does not, because the localization objects are not yet initialized when the `AppLocalizations.of(context)!.helloWorld` is evaluated, so things crash with a null exception. It's simple to understand that `AppLocalizations.of(context)` is null in this instance, but it's not obvious that it is because it is uninitialized.

### Expected fix

I believe the documentation could be a little more clear and contextually mention (in point 7 or 8) that the material app has to actually be started to initialize the `AppLocalizations` and that using `AppLocalizations.of(context)!.helloWorld` won't work before that.

In retrospective the error is clear, but a minor fix could help avoiding this problem at all.

sfshaza2:
That is an excellent point, @akiross! I have updated the text.

PR #9916, opened 2023-12-12
Clarifying when AppLocalizations can be initialized.

sfshaza2:
Fixes https://github.com/flutter/website/issues/6424


domesticmouse (approved):

parlough (approved):
Thanks!

I updated the broken link reference as well.

Issue #8324, opened 2023-03-01
Replace `CachedNetworkImage` with `Image.network` on 'Work with cached images' page


bryanoltman:
### Page URL

https://docs.flutter.dev/cookbook/images/cached-images/

### Page source

https://github.com/flutter/website/tree/main/src/cookbook/images/cached-images.md

### Describe the problem

This page recommends the CachedNetworkImage package, which I believe to be unnecessary now that the `Image.network` constructor exists and "All network images are cached regardless of HTTP headers".

### Expected fix

Recommend the built-in `Image.network` constructor instead of a third-party package.

atsansone:
Escalated to P1 as it was found repeatedly over 4 years. This should get extra attention.

domesticmouse:
Proposed actions:
1) Update https://github.com/flutter/website/blob/main/src/cookbook/images/network-image.md to remove the `Image.network` doesn't cache
2) Remove https://github.com/flutter/website/tree/main/src/cookbook/images/cached-images.md

sg?

domesticmouse:
#9923 is blocked until we get confirmation that `Image.network` caches retrieved images on all six platforms.

PR #9923, opened 2023-12-13
Drop `cached_network_image` package

domesticmouse:
Drop references to `cached_network_image` package because the [`Image.network`](https://api.flutter.dev/flutter/widgets/Image/Image.network.html) named constructor caches images.

Closes https://github.com/flutter/website/issues/8324


src/cookbook/images/cached-images.md
parlough, line 1: Can you remove this from the index and add a redirect to the updated network recipe?

domesticmouse: Can you point me in the right direction for redirects?

domesticmouse: nvm, found it

src/cookbook/images/network-image.md
```diff
@@ -31,15 +31,13 @@ Image.network(
...
 
-## Placeholders and caching
+## Image fade in with placeholders
 
 The default `Image.network` constructor doesn't handle more advanced
-functionality, such as fading images in after loading, or caching images
-to the device after they're downloaded. To accomplish these tasks, see
-the following recipes:
+functionality, such as fading images in after loading. To accomplish this task, see
+the following recipe:
```
parlough, line 38: Since this isn't a list anymore, you can just inline the mention.

```suggestion
functionality, such as fading images in after loading.
To accomplish this task,
check out [Fade in images with a placeholder][].
```

parlough (commented):
Thanks for working on this @domesticmouse!

Two comments and a question:

Is there somewhere on the remaining page that we can mention the widget's caching capability or do you think that's not necessary?

src/cookbook/images/cached-images.md
parlough: Ah sorry I walked away and forgot to submit my review. Glad you did!

sfshaza2:
Woah, woah, woah! I've been talking (or trying to) talk with both the web and the framework teams about this issue. Yes, the Image.network constructor now caches the image, HOWEVER, it's not clear if this also works for web. Do you know, @domesticmouse?

sfshaza2:
Here's the content of the email I sent @kevmoo on Dec 7th, which has also been sent to the web and framework teams.

From: shams
Thu, Dec 7, 12:36 PM (6 days ago)
to Kevin

We have two cookbook recipes that kinda sorta don't play well together. One recommends using `cached_network_image` for caching images. The other suggests the `Image.network` constructor, which I gather was more recently added.

The thing is, it's not entirely clear what is supported on the web. I've been looking at both and trying to come up with a list of pros and cons, so we can make sensible recommendations.

Can you help elucidate?

**Image.network**
pros:
- automatically caches image (web??)
- image constructors allow you to specify image decode size to save space when caching
- supports several types of image formats and downloading mechanisms
- supports animated GIFs
- provides loading progress

cons:
- on web, specifying a decode size is only supported when using a CanvasRenderer; not supported for an HTMLRenderer

**cached_network_image package**
pros:
- supports "placeholder" images
- supports fading images in as they're loading

cons:
- it's a 3p package...

-s-

atsansone (approved):

sfshaza2:
I've converted this to a draft as I don't want this landed until this is resolved.

johnpryan:
On the web, [images use the cache provided by the browser](https://stackoverflow.com/questions/61186689/caching-image-on-flutter-web-is-it-required). So I think that the behavior is different regardless of whether you use Image.network or cached_network_image (cached_network_image explicitly uses an in-memory cache on the web, probably because explicit caching isn't necessary for most apps).

I haven't validated this behavior myself, but this would be a good thing to do as part of our web docs.

sfshaza2:
OK, I'm going to accept this. Landing.  :)

sfshaza2 (approved):
lgtm

PR #9935, opened 2023-12-15
Remove DevTools survey metadata json and add redirect.

kenzieschmoll:
The survey metadata file was copied to the `flutter/uxr` repo in https://github.com/flutter/uxr/pull/127.


parlough (approved):
Cool, thanks! Looks like the new version is already being uploaded to the storage bucket, so feel free to merge when ready :)

PR #9987, opened 2024-01-03
Release notes for DevTools 2.31.0

polina-c:


src/tools/devtools/release-notes/images-2.31.0/link-to-doc.png
domesticmouse, line 1: I feel like we might want something that looks a bit more polished than this hand drawn arrow.

src/tools/devtools/release-notes/images-2.31.0/hover-for-dropdown.png
domesticmouse, line 1: This screenshot appears cluttered and confusing. Would increasing the amount of context in the image make it easier to understand?

polina-c: Two notes:
1. We save screen size, because the area of release notes in DevTools is a vertical side panel.
2. Explanation should help with understanding.

I improved wording for this item.

+ @CoderDake (Dan) who created this feature 

Do you think we can create better image?

Another question. Would it be better for both us and users to add this information as text for item, instead of T-Short sizes. Like: "CPU Sampling rate: 4 per second"

Dan will be back next week. So may be lets merge the PR this way in order to may be follow up later.

src/tools/devtools/release-notes/images-2.31.0/link-to-doc.png
polina-c: From my perspective hand drawn arrow catches attention much better than anything more polished. 

Created alternative option:

Screenshot 2024-01-03 at 11 25 08 AM

Feel free to create something yourself, based on the image here: https://github.com/flutter/devtools/pull/6825

polina-c: Let me know if you like new image more or want to suggest something else.

domesticmouse (approved):

parlough (approved):
Thanks for adding this! Sounds good to land this now so the link is there.

PR #10001, opened 2024-01-06
Update iOS minimum version documentation to 12.0

jmagman:
Move iOS 11 to the unsupported tier.  

**This should not be merged until https://github.com/flutter/flutter/issues/140923 is cherry-picked to stable and released.**

Part of https://github.com/flutter/flutter/issues/140474

### DIFF
--- a/src/add-to-app/ios/project-setup.md
+++ b/src/add-to-app/ios/project-setup.md
@@ -35,7 +35,7 @@ For an example using SwiftUI, see the iOS directory in [News Feed App][].
 Your development environment must meet the
 [macOS system requirements for Flutter][]
 with [Xcode installed][].
-Flutter supports iOS 11 and later.
+Flutter supports iOS 12 and later.
 Additionally, you will need [CocoaPods][]
 version 1.10 or later.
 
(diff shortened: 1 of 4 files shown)


parlough (approved):
Thanks for being on top of updating all of these references!

If this is going to be landing on the stable channel soon, I'd actually consider merging this sooner rather than waiting for that. That way users can have up to date expectations  and potentially avoid the issue.

I think there's more chance for confusion or harm with a lower minimum than a higher one.

Up to you though! :)

jmagman:
> If this is going to be landing on the stable channel soon, I'd actually consider merging this sooner rather than waiting for that. That way users can have up to date expectations and potentially avoid the issue.

I think I'd rather wait and submit this as simultaneously as possible with the hotfix since the supported versions page reflects stable:
> As of the current stable release, support for deploying Flutter apps is shown in the following table:

Until that merges, users would have to manually edit some files to change their minimum from the default (11) to iOS 12.

jmagman:
Okay this is ready, merging on @parlough's LGTM.  Thanks!

parlough:
Thank you! 🚀

Issue #4130, opened 2020-05-25
'Layouts in Flutter' page issue


CivanDogan:
Page URL: https://flutter.dev/docs/development/ui/layout/
Page source: https://github.com/flutter/website/tree/master/src/docs/development/ui/layout/index.md

Found a typo? You can fix it yourself by going to the page source and clicking the pencil icon. Or finish creating this issue.

Description of issue: This tutorial is awsome but probably outdated. Sizing Widgets (https://flutter.dev/docs/development/ui/layout/) is not working as in the example. Setting flex :2 not makes the images get bigger vertically .  Thank you 

![image](https://user-images.githubusercontent.com/34371430/82844501-6ce90300-9ee9-11ea-8d6b-d72686d83ee1.png)

CivanDogan:
It's because of the Image.asset because Expanded take up space as in tutorial.
![image](https://user-images.githubusercontent.com/34371430/82845413-21385880-9eed-11ea-970e-579152828ac9.png)

Issue #2719, opened 2019-06-01
Update screenshot for align widget example on 'Layouts in Flutter' page


fodil-a:
Page URL: https://docs.flutter.dev/ui/layout#aligning-widgets
Page source: https://github.com/flutter/website/tree/master/src/docs/development/ui/layout/index.md

Found a typo? You can fix it yourself by going to the page source and clicking the pencil icon. Or finish creating this issue.

Description of issue: the image of `row_column` example for Column align is wrong  
It shows:
![image](https://user-images.githubusercontent.com/21066051/58754103-0b56aa00-84ca-11e9-9ecb-781877f47164.png)

Instead of showing something like:
![image](https://user-images.githubusercontent.com/21066051/58754115-26291e80-84ca-11e9-8d66-859d8132c2bb.png)

The errors are:
1. The images are not the same as in the repository [here](https://github.com/flutter/website/tree/master/examples/layout/row_column/images)
2. The image does not show the specificity of MainAxisAlignment.spaceEvenly, which is to space evenly better, BEFORE and AFTER.

PR #10007, opened 2024-01-08
FWE

ericwindmill:
This branch is a place where we can have a conversation about how FWE V1 content will be presented. The content of this page is less important (right now) than the UI.


src/get-started/learn-more.md
```diff
@@ -1,54 +1,163 @@
...
+with Flutter widgets. In this section, you’ll learn how to build simple and
+complex layouts in Flutter with some of the most common layout widgets in
+Flutter. You’ll also learn about how to use DevTools to understand how your apps
+are being laid out, and you’ll learn how to debug one of the most common Flutter
+errors, which is related to the most common Flutter layout widgets.
+
+### Introduction to layouts
+
+The following tutorial will introduce you to layouts in Flutter, and give you an
+opportunity to work with the most common layout widgets.
```
ericwindmill, line 21: Row and Column should be called out explicitly here, since they're so important and it otherwise might seem like I forgot about them.

src/get-started/learn-more.md
```diff
@@ -1,54 +1,163 @@
...
+
+1. Tutorial: [Layouts in Flutter][]
```
ericwindmill, line 23: [As Khanh pointed out in the original doc] We should have an icon of some sort that says "Take Action!" next to resources that allow/encourage the user to write code.

site-shared
ericwindmill, line 1: Didn't mean to commit this

src/get-started/learn-more.md
```diff
@@ -1,54 +1,163 @@
...
+Flutter. You’ll also learn about how to use DevTools to understand how your apps
+are being laid out, and you’ll learn how to debug one of the most common Flutter
+errors, which is related to the most common Flutter layout widgets.
+
+### Introduction to layouts
+
+The following tutorial will introduce you to layouts in Flutter, and give you an
+opportunity to work with the most common layout widgets.
+
+1. <i class="material-icons" aria-hidden="true">bolt</i> Take action! - Tutorial: [Layouts in Flutter][]
```
ericwindmill, line 23: (As Khanh pointed out in the original doc) We should have an icon of some sort that says "Take Action!" next to resources that allow/encourage the user to write code. 

This will obviously need to be styled and ideally made into a Liquid variable a la {{site.alert...

src/get-started/learn-more.md
```diff
@@ -1,54 +1,163 @@
...
+how Flutter’s layout algorithm works.
+
+The following resources will teach you what constraints are, some widgets that
+allow you to work with constraints, and how to deal with “unbounded constraints”
+type errors.
+
+1. Article: [Understanding Constraints][]
+2. Video: [Expanded - Flutter Widget of the Week][]
+3. Video: [Flexible - Flutter Widget of the Week][]
+4. Video: [Intrinsic widgets - Decoding Flutter][]
```
ericwindmill, line 37: I'm not sure if we should embed videos in V1, because the videos will so dominate the page. I imagine people would just scroll right past links and be drawn right into the videos

src/get-started/learn-more.md
```diff
@@ -1,54 +1,163 @@
...
-[Flutter for SwiftUI developers]: {{site.url}}/get-started/flutter-for/swiftui-devs
-[Flutter for UIKit developers]: {{site.url}}/get-started/flutter-for/uikit-devs
-[Flutter for React Native developers]: {{site.url}}/get-started/flutter-for/react-native-devs
-[Flutter samples]: https://flutter.github.io/samples
-[Flutter for web developers]: {{site.url}}/get-started/flutter-for/web-devs
-[Flutter for Xamarin.Forms developers]: {{site.url}}/get-started/flutter-for/xamarin-forms-devs
-[Introduction to Flutter's UI]: {{site.url}}/ui
-[mailing list]: mailto:{{site.email}}
-[Udacity online Flutter training]: https://www.udacity.com/course/build-native-mobile-apps-with-flutter--ud905
+## Layout
```
ericwindmill, line 9: When we start adding other topics, should this be one long page or multiple pages? 

Pros of multiple pages (imo):

- Future proofing, assuming we plan on adding a lot to this over time.
- Easier to navigate for folks who don't want to sit down for 20 hours straight.
- Easier to maintain

Cons of multiple pages:

- We'll end up with a new sub-sidenav that looks very similar to the current Sidenav.

ericwindmill: Alternatively, we can do something like Android did with expansion panels:
https://developer.android.com/courses/android-basics-compose/course

(I have _serious_ misgivings about the content of that page and I would really like to avoid using it is a guide, but it does have nice UI)

site-shared
ericwindmill, line 1: Didn't mean to commit these submodules

src/get-started/learn-more.md
sfshaza2: Definitely don't embed videos in this page.

atsansone (commented):
@ericwindmill : This duplicates a good bit of the [Layout intro page](https://docs.flutter.dev/ui/layout). That page has two other issues (#4130 and #2719) that need resolution, plus it duplicates a lot of content that is better explained in the Layout tutorial that I just finished updating. It may be worth a discussion before more work on this.

johnpryan:
@atsansone This isn't meant to be a replacement for these docs. (It actually links directly to the Layout intro page and Layout tutorial already.)

src/get-started/learn-more.md
ericwindmill: Doing even an "MVP" implementation of UI similar to the Android course page is relatively significant amount of work. I highly recommend we come to an agreement on what the page should look like before spending time on this.

RedBrogdon: I have been imagining a "learn more" page with links to individual topic pages. Doesn't mean we have to do it that way, but it could help with analytics about which topics get traffic, and probably make it easier for mobile readers.

src/get-started/learn-more.md
RedBrogdon: @sfshaza2, you use icons next to links sometimes, right? Is there a set/font that you pull from and we could reuse here?

src/get-started/learn-more.md
RedBrogdon: +1 for not embedding directly, since it could increase memory demand and load times. One alternative would be to create a small playlist for each topic, and link to individual videos within the playlist. That would make it easier for people to get from one to another as they follow along, and folks might get some bonus use out of the playlists if they just stumble onto them while browsing YT.

src/get-started/learn-more.md
```diff
@@ -1,54 +1,163 @@
...
+### Constraints
+
+Understanding “constraints” in Flutter are an important part of understanding
+how Flutter’s layout algorithm works.
+
+The following resources will teach you what constraints are, some widgets that
+allow you to work with constraints, and how to deal with “unbounded constraints”
+type errors.
+
+1. Article: [Understanding Constraints][]
```
RedBrogdon, line 34: Style question: should these be ordered lists, or would bullets work just as well (e.g. "all things being equal we recommend going top-down on these four things, but it's not required")?

src/get-started/learn-more.md
```diff
@@ -1,54 +1,163 @@
...
+to as the “unbounded constraints” error. If there was only one type error you
+should be prepared to confront before you start building Flutter apps, it’d be
+this type.
+
+1. Article: [DevTools - Widget Inspector][]
+2. Video: [Unbounded height and width - Decoding Flutter][]
+
+### Advanced Content
+
+- [2D Scrolling][]
```
RedBrogdon, line 118: Question: would we expect (as a general pattern) to have at least a sentence or two under every `###` heading introducing whatever items are there?

ericwindmill: I envisioned (and briefly talked to Khanh about) having an Advanced Content and Reference section at the end of topics, both of which have optional content.  I'm not sure what I'd write under these headings, but I'm open to ideas. 

On the current docs.flutter.dev/ui/layout page, there's a "Other resources" section with the description: "The following resources might help when writing layout code." I'm going to start by copying that.

ericwindmill: I'm also happy to just skip the "advanced content" altogether

src/get-started/learn-more.md
ericwindmill: FWIW, I'm using `material-symbols`, which I think is the icon font that the rest of the site uses. (I think it's the only icon font loaded into the site)

ericwindmill:
I imagine we'll want to change the URL from /fwe/

domesticmouse (approved):

ericwindmill:
@sfshaza2  I updated the symbol to be this, which I think what we have used elsewhere.
Screenshot 2024-02-07 at 11 42 12 AM

src/get-started/fwe/layout.md
```diff
@@ -0,0 +1,161 @@
...
+
+* Video: [LayoutBuilder - Flutter Widget of the Week][]
+* Video: [MediaQuery - Flutter Widget of the Week][]
+* Video: [Building platform adaptive apps][]
+
+<i class="material-symbols" aria-hidden="true">label</i> Tutorial: [Adaptive Apps codelab][]
+
+### DevTools and debugging for layout
+
+Flutter has a robust suite of DevTools that will help you work with any number
```
sfshaza2, line 103:
```suggestion
Flutter has a robust suite of DevTools that help you work with any number
```

src/get-started/fwe/layout.md
```diff
@@ -0,0 +1,161 @@
...
+of aspects of Flutter development. The “Widget Inspector” tool is particularly
+useful when building Layouts (and working with widgets in general).
```
sfshaza2, line 105:
```suggestion
useful when building layouts (and working with widgets in general).
```

src/get-started/fwe/layout.md
```diff
@@ -0,0 +1,161 @@
...
+
+Additionally, perhaps the most common error you’ll run into while building a
+Flutter application is due to incorrectly using Layout widgets, and is referred
```
sfshaza2, line 108:
```suggestion
Flutter application is due to incorrectly using layout widgets, and is referred
```

src/get-started/fwe/layout.md
```diff
@@ -0,0 +1,161 @@
...
+
+<i class="material-symbols" aria-hidden="true">label</i> Tutorial: [Layouts in Flutter][]
+
+### Constraints
+
+Understanding “constraints” in Flutter are an important part of understanding
+how Flutter’s layout algorithm works.
+
+The following resources will teach you what constraints are, some widgets that
+allow you to work with constraints, and how to deal with “unbounded constraints”
```
sfshaza2, line 30: I'm wondering. Does these resources actually deal with unbounded restraints? Because you have that covered in the last bit of this page.

src/get-started/fwe/layout.md
```diff
@@ -0,0 +1,161 @@
...
+### DevTools and debugging for layout
+
+Flutter has a robust suite of DevTools that will help you work with any number
+of aspects of Flutter development. The “Widget Inspector” tool is particularly
+useful when building Layouts (and working with widgets in general).
+
+Additionally, perhaps the most common error you’ll run into while building a
+Flutter application is due to incorrectly using Layout widgets, and is referred
+to as the “unbounded constraints” error. If there was only one type error you
+should be prepared to confront before you start building Flutter apps, it’d be
```
sfshaza2, line 110:
```suggestion
should be prepared to confront before you start building Flutter apps, it would be
```

sfshaza2 (approved):
LGTM! I love love love that Dash icon.  :D

PR #10024, opened 2024-01-10
Migration guide for deprecations removed after 3.16 release

Piinks:

This adds the migration guide for the deprecations that have been removed since the 3.16 release.

Fixes https://github.com/flutter/flutter/issues/139243

## Presubmit checklist

### DIFF
--- a/src/release/breaking-changes/3-16-deprecations.md
+++ b/src/release/breaking-changes/3-16-deprecations.md
@@ -101,9 +101,9 @@ Relevant PRs:
 * Removed in [#139267][]
 
 [`TextButton`]: {{site.api}}/flutter/material/TextButton-class.html
-[`Chip`]: {{site.api}}/flutter/material/ElevatedButton-class.html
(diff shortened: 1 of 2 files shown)


src/release/breaking-changes/3-16-deprecations.md
```diff
@@ -0,0 +1,276 @@
+---
+title: Deprecated API removed after v3.16
+description: >-
+After reaching end of life, the following deprecated APIs
+were removed from Flutter.
```
parlough, line 5: I think this is breaking the build since YAML expects indentation here:

```suggestion
  After reaching end of life, the following deprecated APIs
  were removed from Flutter.
```

parlough (approved):
Thanks for this! One fix then looks good to me :)

src/release/breaking-changes/3-16-deprecations.md
Piinks: Ah thank you! A bit of copy pasta. 🍝 
I feel like I always hit this when I write these because the indents don't follow.

PR #9736, opened 2023-10-31
Refactor macOS Getting Started

atsansone:
Total refactor of macOS install.

Fixes #2461
Fixes #2529
Fixes #4995
Fixes #5374
Fixes #7475
Fixes #7647
Fixes #9520

PR #10031, opened 2024-01-11
Adds PATH for macOS

### LINKED ISSUE #9351: Change PATH configuration line on 'macOS install' page
### Page URL

https://docs.flutter.dev/get-started/install/macos/

### Page source

https://github.com/flutter/website/tree/main/src/get-started/install/macos.md

### Describe the problem

The use of pwd in the PATH configuration line (export PATH="$PATH:pwd/flutter/bin") is incorrect and unnecessary. The correct way to add a directory to the PATH is to specify the absolute path to that directory directly. The pwd command, which stands for "print working directory," is used to display the current directory, but it's not needed in the PATH configuration.

### Expected fix

 Rename to: export PATH="$PATH:[PATH_OF_FLUTTER_GIT_DIRECTORY]/bin" as it is under "Update your path" also step 3, using which did not bring any confusion

### Additional context

Step: "**3. Add the flutter tool to your path:**"  is under section **Get the Flutter SDK**

### LINKED ISSUE #10015: Add VS Code extension mention to the 'Start building Flutter native desktop apps on macOS' page
### Page URL

https://docs.flutter.dev/get-started/install/macos/desktop/

### Page source

https://github.com/flutter/website/tree/main/src/get-started/install/macos/desktop.md

### Describe the problem

The "Use VS Code to install" steps should mention the need to install the Flutter extension.

### Expected fix

_No response_

### Additional context

_No response_

### LINKED ISSUE #10036: Install page issues for developing on iOS
### Page URL

https://docs.flutter.dev/get-started/install/macos/mobile-ios?tab=vscode

### Page source

_No response_

### Describe the problem

I ran into a few issues with the install page when setting up my new computer. 

First, users need to install the VSCode extension before being able to navigate to the start a new flutter project - I don't think that was clear from this document. Next, it says "Git for Windows" even though this is the macOS page

### Expected fix

_No response_

### Additional context

_No response_

### LINKED ISSUE #10040: macOS hardware requirements mention x86-64 cpu cores?
### Page URL

### Page source

https://github.com/flutter/website/tree/main/src/get-started/install/macos/mobile-ios.md

### Describe the problem

The hardware requirements on this page mention "x86_64" CPU cores. It's not clear to me why the x86_64 specifier is there, as it could lead users to thinking ARM Mac laptops aren't supported, or just confuse them in general.

I've also never heard of this core count requirement. Where did these numbers come from?

### Expected fix

I guess drop the x86_64 specifier? Or just the core count requirement completely.

### Additional context

_No response_

atsansone:
Adds PATH instructions for macOS.
Fixes #9351
Fixes #10015
Fixes #10040 
Fixes #10036


atsansone:
@johnpryan : Can you review this as DRE on call this week? The main change is in the **Download and Install** tab of [the macOS install page](https://flutter-docs-prod--pr10031-fix-9351-7bq7dgok.web.app/get-started/install/macos/mobile-ios?tab=download#install-the-flutter-sdk). The other change is in the **Operating system** section of the [Software requirements](https://flutter-docs-prod--pr10031-fix-9351-7bq7dgok.web.app/get-started/install/macos/mobile-ios?tab=download#software-requirements) section. Thanks!

src/_includes/docs/install/reqs/macos/set-path.md
```diff
@@ -0,0 +1,39 @@
+{% assign terminal=include.terminal %}
+{% assign target = include.target %}
+
+### Add Flutter to your `PATH`
+{:.no_toc}
+
+To run Flutter commands in {{terminal}},
+You need to add Flutter to the `PATH` environment variable.
+THis section covers the permanent addition of the path to the
```
johnpryan, line 9:
```suggestion
This section covers the permanent addition of the path to the
```

src/_includes/docs/install/compiler/xcode.md
```diff
@@ -98,7 +92,26 @@ To [install and set up CocoaPods][cocoapods], run the following commands:
...
+    ```terminal
+    $ cat <<EOF >> ~/.zshenv
+    # Configure CocoaPods
+
+    export GEM_HOME=\$HOME/.gem
+    if [[ \$PATH != *"\$GEM_HOME"* ]]
+    then
+        export PATH=\$GEM_HOME/bin:\$PATH
+    fi
+    EOF
```
johnpryan, line 106: We don't need to dictate how the user should add a directory to their path in their `.zshenv` file. I would prefer give instructions that align more with the [Cocoapods Getting Started guide](https://guides.cocoapods.org/using/getting-started.html) like this:

```
export GEM_HOME=$HOME/.gem
export PATH=$GEM_HOME/bin:$PATH
```

src/_includes/docs/install/compiler/xcode.md
```diff
@@ -98,7 +92,26 @@ To [install and set up CocoaPods][cocoapods], run the following commands:
    {{prompt1}} brew install cocoapods
    ```
 
-Using Homebrew reduces potential issues with chipsets and install permissions.
```
johnpryan, line 101: @jmagman Why do we recommend Homebrew, but the [CocoaPods Getting Started guide](https://guides.cocoapods.org/using/getting-started.html) says to use `gem install` with the system Ruby installation? 

@atsansone It would be great if we can lean more on the CocoaPods documentation here. At the very least, we should try to match what they recommend.

src/_includes/docs/install/compiler/xcode.md
```diff
@@ -98,7 +92,26 @@ To [install and set up CocoaPods][cocoapods], run the following commands:
...
+    fi
+    EOF
+    ```
+
+1. Load the changes to your current open terminal sessions
+   using the following command.
+
+    ```terminal
+    $ source ~/.zshenv
+    ```
```
johnpryan, line 114: I think we should just leave this out, we could just say "Restart your terminal".

src/_includes/docs/install/reqs/macos/set-path.md
```diff
@@ -0,0 +1,39 @@
...
+Flutter to your `PATH` environment variable.
```
sfshaza2, line 10:
```suggestion
you need to add Flutter to the `PATH` environment variable.
You can add it manually each time you launch a terminal 
but it's easier, as shown here,
to add it to your environment so that it's always available.
```

src/_includes/docs/install/reqs/macos/set-path.md
```diff
@@ -0,0 +1,39 @@
...
+    ```terminal
+    $ cat <<EOF >> ~/.zshenv
+    # Configure Flutter
+
+    export FLUTTER_HOME=\$HOME/Projects/flutter
+    if [[ \$PATH != *"\$FLUTTER_HOME"* ]]
+    then
+        export PATH=\$FLUTTER_HOME/bin:\$PATH
+    fi
+    EOF
```
johnpryan, line 28: Same comment as above, this code is overcomplicated.

src/_includes/docs/install/reqs/macos/zsh-config.md
```diff
@@ -0,0 +1,51 @@
...
+
+<details markdown="1">
+<summary><strong>To verify your shell configuration, expand this section</strong></summary>
+
+Like most UNIX-like operating system, macOS can support multiple shells,
+like `bash`, `zsh`, and `sh`.
+
+As of the October 2019 release of macOS Catalina (macOS 10.15),
+Z shell or `zsh` is the default shell for macOS.
+Apple made this change to due to licensing changes for `bash`.
```
johnpryan, line 11: Take this line out, we don't need it.
```suggestion
```

src/_includes/docs/install/reqs/macos/zsh-config.md
```diff
@@ -0,0 +1,51 @@
...
+
+#### Check and set `zsh` as default
+{:.no_toc}
+
+1. To verify `zsh` has been installed and set as
+   the default macOS shell, run the following commands.
+
+    ```terminal
+    $ which zsh; dscl . -read ~/ UserShell
```
johnpryan, line 20: This can be simplified to:

```suggestion
$ echo $0
/bin/zsh
```

src/_includes/docs/install/reqs/macos/zsh-config.md
```diff
@@ -0,0 +1,51 @@
...
+    ```
+
+    {{terminal}} should print the following as its response.
+
+    ```terminal
+    /bin/zsh
+    UserShell: /bin/zsh
+    ```
+
```
johnpryan, line 29: Take these lines out as per comment above:

```suggestion
```

sfshaza2 (commented):
Clarified some text.

src/_includes/docs/install/reqs/macos/zsh-config.md
```diff
@@ -0,0 +1,51 @@
...
+1. If you need to install `zsh`, run the following command.
+
+    ```terminal
+    $ brew install zsh
+    ```
+
+   To check if your install succeeded,
+   run the command in the previous step after installing `zsh`.
+
+1. If you need to change your default shell, run the following command.
```
johnpryan, line 39: We should explain what the command does before we tell them to run it:

```suggestion
1. To change your shell to Zsh, run the following command.
```

src/_includes/docs/install/reqs/macos/zsh-config.md
```diff
@@ -0,0 +1,51 @@
...
+
+    ```terminal
+    $ chsh -s `which zsh`
+    ```
+
+    This command changes your shell to where you installed `zsh`.
```
johnpryan, line 45: Take this line out (see above comment)

```suggestion
```

src/_includes/docs/install/reqs/macos/zsh-config.md
```diff
@@ -0,0 +1,51 @@
+{% assign target = include.target %}
+
+<details markdown="1">
+<summary><strong>To verify your shell configuration, expand this section</strong></summary>
+
+Like most UNIX-like operating system, macOS can support multiple shells,
```
johnpryan, line 6: This is the type of doc that tends to get out of date, and I don't think we need this page. We could probably just link to Apple's [Change the default shell in Terminal on Mac](https://support.apple.com/zh-sg/guide/terminal/trml113/mac) instead.

src/_includes/docs/install/flutter/download.md
```diff
@@ -84,6 +90,9 @@
...
 
-{% if os == 'Windows' %}
-{% include docs/install/reqs/windows/set-path.md terminal=terminal %}
-{% endif %}
+{% case os %}
+{% when 'Windows' %}
+{% include docs/install/reqs/windows/set-path.md terminal=terminal target=target %}
+{% when 'macOS' %}
+{% include docs/install/reqs/macos/set-path.md terminal=terminal target=target %}
+{% endcase %}
```
johnpryan, line 98: I don't know if it's just me but this structure is hard to review with all of these includes. I've noticed that the size of the `_includes` directory is increasing and I'm not sure that's a great trend for maintainability overall. If there are places where we can just add the docs in-line I think that would be better in the long term. 

Includes are great for reusable snippets of instructions but might have diminishing returns if we use them too much.

src/_includes/docs/install/flutter/vscode.md
```diff
@@ -128,3 +133,5 @@
    1. Restart VS Code.
 
 [development tools prerequisites]: #development-tools
+[Visual Studio Code]: https://code.visualstudio.com/docs/setup/mac
+[Flutter extension for VS Code]: https://marketplace.visualstudio.com/items?itemName=Dart-Code.flutter
```
johnpryan, line 137: Unused link?

src/_includes/docs/install/reqs/macos/base.md
```diff
@@ -31,6 +35,13 @@ software packages.
 
 Flutter supports macOS {{site.devmin.macos}} or later.
 
+If you run a supported macOS version,
+Z shell (`zsh`) is your default login and interactive shell.
```
johnpryan, line 39: I think it's more correct to call it [Zsh](https://www.zsh.org/).

```suggestion
Zsh (`zsh`) is your default login and interactive shell.
```

src/_includes/docs/install/reqs/macos/set-path.md
```diff
@@ -0,0 +1,39 @@
...
+### Add Flutter to your `PATH`
+{:.no_toc}
+
+To run Flutter commands in {{terminal}},
+You need to add Flutter to the `PATH` environment variable.
+THis section covers the permanent addition of the path to the
+Flutter to your `PATH` environment variable.
+This guide presumes that you are using the latest default shell, `zsh`.
+This section adds Flutter to the `PATH` using the `.zshenv` file.
+Z shell uses this file only for environment variables and loads its
```
johnpryan, line 13:
```suggestion
Zsh uses this file only for environment variables and loads its
```

src/_includes/docs/install/reqs/macos/zsh-config.md
```diff
@@ -0,0 +1,51 @@
...
+like `bash`, `zsh`, and `sh`.
+
+As of the October 2019 release of macOS Catalina (macOS 10.15),
+Z shell or `zsh` is the default shell for macOS.
```
johnpryan, line 10:
```suggestion
Zsh or `zsh` is the default shell for macOS.
```

johnpryan (changes requested):

src/_includes/docs/install/compiler/xcode.md
atsansone: The extra code around this was to keep the PATH sane if the user runs `source` more than once.

src/_includes/docs/install/compiler/xcode.md
atsansone: The reason I included Homebrew is that it was included in a previous version of this doc for the reasons stated in the Software requirements: Homebrew handles chipset and permissions issues for you. I would love to avoid: 

1. Sending users to too many other sites.
2. Needing to cover off more edge cases when something doesn't work.

src/_includes/docs/install/compiler/xcode.md
atsansone: If you're like me, having to restart 5-6 terminals would be... unpleasant. This command is a shorter way to do this.

src/_includes/docs/install/reqs/macos/zsh-config.md
atsansone: Can do. I waivered on this, though, because the _why_ of it kept creeping into my mind.

src/_includes/docs/install/reqs/macos/zsh-config.md
atsansone: No, it can't. The second part tells you what your default shell is. The `echo $0` only tells you what shell is running in your current terminal.

src/_includes/docs/install/reqs/macos/zsh-config.md
atsansone: I might need to tweak that slightly. This step is optional.

src/_includes/docs/install/reqs/macos/zsh-config.md
atsansone: This section wouldn't get out of date unless Apple changes the default shell (again). It's also hidden unless the reader needs to see it. That's because, well, they shouldn't unless they're running macOS 10.14 or earlier.

src/_includes/docs/install/flutter/download.md
atsansone: It can be on first pass, but future updates may be easier. Most of these types of pages get a lot of reuse as most (~75%) of the content is the same for all 4 development platforms.

src/_includes/docs/install/reqs/macos/set-path.md
atsansone: I'd like to keep this more opinionated and avoid writing about another method that they'll find they need to change later. That's what made the original macOS install page so cumbersome.

src/_includes/docs/install/compiler/xcode.md
johnpryan: I still think this is unnecessary code to make the user copy-paste. It's also not very readable. I know I spent a while staring at it before I understood what it was doing. But if you ad just said "Add this to your path in `.zshenv` I would have been able to do it with whichever editor I normally use, and add it in a place that makes sense, perhaps with a comment. We don't need to hand-hold the user through this.

src/_includes/docs/install/compiler/xcode.md
johnpryan: Why? Don't you just need to close the terminal window and open a new one?

src/_includes/docs/install/flutter/download.md
johnpryan: It's _really_ hard for me to follow what's going on. I have to jump around from file to file rather than just read it in-line.

src/_includes/docs/install/compiler/xcode.md
jmagman: @johnpryan I wouldn't have suggested Homebrew, I wasn't a reviewer on https://github.com/flutter/website/pull/9736.  

https://docs.flutter.dev/get-started/install/macos/mobile-ios#install-cocoapods currently sounds like Homebrew is a requirement to do Flutter iOS development it just isn't.  cc @christopherfujino 

I don't use Homebrew to manage Ruby or CocoaPods.  I use the built in Ruby version to macOS and I follow CocoaPod's own installation instructions.  For the purpose of this doc I would personally just point them to https://guides.cocoapods.org/using/getting-started.html#installation and call it a day instead of trying to replicate dependency's installation instructions.

src/_includes/docs/install/reqs/macos/set-path.md
johnpryan: It's hard to read, I would just describe the code they need to add to their shell initialization (whether it's Zsh, Bash, or something else)

src/_includes/docs/install/compiler/xcode.md
parlough: I'm also not a fan of Homebrew and skip any steps involving it. Leaning on Cocoapods own installation instructions and the options there seems good 👍

christopherfujino: I got lost searching through the git history, but TL;DR we used to recommend brew install cocoapods, but this caused pain for our users, and is NOT the official way to install this, so we standardized on `gem install cocoapods` across the tool and website as our official recommendation.

christopherfujino: > For the purpose of this doc I would personally just point them to https://guides.cocoapods.org/using/getting-started.html#installation and call it a day instead of trying to replicate dependency's installation instructions.

This is aligned with what the tool tells users to do if we detect it's either missing or broken: https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/macos/cocoapods.dart#L50

src/_includes/docs/install/reqs/macos/zsh-config.md
johnpryan: I think it's safe to assume the user is running their default shell.

src/_includes/docs/install/reqs/macos/zsh-config.md
johnpryan: In general this is not a doc that we need to maintain. In some cases it might be better to have no doc page at all, especially in this case anything we put here will already be covered by the official Apple / Zsh documentation. Need to be careful we don't increase our maintenance costs unnecessarily.

src/_includes/docs/install/reqs/macos/set-path.md
atsansone: That's the thing: It should be `zsh` across the board at this version of the OS. I don't want to assume a particular level of knowledge or familiarity with all areas involved.

src/_includes/docs/install/reqs/macos/zsh-config.md
atsansone: We can't make too many assumptions. This is one that I think would be important to explain.

johnpryan: I don't think average Flutter developer is concerned about this honestly, and this isn't even documenting something about Flutter. I would vote that we keep our docs as general as possible since not every developer even uses Zsh, and they don't need to use Zsh to use Flutter, Cocoapods, or anything else.

src/_includes/docs/install/compiler/xcode.md
atsansone: Most any explanation I've consulted doesn't recommend the default macOS install of Ruby. We used Homebrew elsewhere, so it seemed better to have one way to install everything instead of one way for this app, another for this other app, etc. Think as a new user: keep it simple.

We refer to Homebrew in about as many pages in the public docs: around 1-2 times each.

Looking at this as a new user, I'd like to recommend simple, consistent ways to handle installs wherever possible. Homebrew offers that option.

Further, the CocoaPods install docs themselves are out of date (referring to `bash_profile`, for example). 

Although, @christopherfujino , your point about what exists in the code makes sense and may need a future revisit.

atsansone: @jmagman : Understood that CocoaPods is not required to build iOS apps for Flutter. It's required to build Flutter plugins for iOS apps, right? We could make that note, but not in a Getting Started. I would then need to add an additional section or page later to cover that off. What say you?

src/_includes/docs/install/compiler/xcode.md
atsansone: I run [Tmux](https://github.com/tmux/tmux/wiki) and have multiple tabs open in my Terminal app.

src/_includes/docs/install/flutter/vscode.md
atsansone: Fixed!

src/_includes/docs/install/compiler/xcode.md
jmagman: CocoaPods is required as soon as you add a plugin.  I believe in the docs that used to be called out.

src/_includes/docs/install/reqs/macos/set-path.md
jmagman: I disagree, in this doc it's so opinionated I find it difficult to understand what is required to use Flutter and what's just recommended.  Like why `.zshenv` and not `.zshrc`?  Why would Flutter need me to update `GEM_HOME`?  Have we now messed up the environment for someone who already has Ruby installed?  Why are we in the business of telling people to update that env?

atsansone: The `zsh` docs say that environment variables should be set in `.zshenv` because that file is loaded regardless of it being a login or an interactive shell. The other files are not always loaded. 

This is also what opinionated means in this case. We're not offering options, we're providing a path to completion. When this is published, we may get issues filed against this and we can address any raised concerns then.

jmagman: I guess I'm disagreeing that this doc should be opinionated.  I suggest we put this page back to:
1. Install Xcode (link to Xcode in the Mac App Store, or Apple docs)
2. Install Cocoapods (link to CocoaPods install page)
3. Put Flutter on your PATH (link to how to put things on your PATH, or maybe a blog post or something like "here's an example of how to do that!"

etc.

As someone who has had to close many open issues and add workarounds to the Flutter tool related to installation and environment issues resulting from our _own_ docs, I'd really prefer not to be in the business of keeping the dependency installation docs up-to-date, or to confuse experienced developers who know how to add executables to their PATH, or make it seem like we have dependencies that we don't (Homebrew).

Once it's installed and built, the `flutter` tool itself has more up-to-date hints about how to get your environment set up (you have Xcode installed, but you need to accept the license, etc)

atsansone:
@johnpryan , @sfshaza2 , @christopherfujino , @jmagman : Thank you all for your input on this PR. I believe we've had some constructive conversations that will produce a better docs page. I will make the updates to remove Homebrew, simplify PATH instructions, simplify CocoaPods install instructions, and clarify per-session vs. permanent PATH changes in my next commit. This is an FYI and the next commit will be a little later today.

Again, my thanks for your time and assistance!

src/_includes/docs/install/compiler/xcode.md
atsansone: It still is. ;)

src/_includes/docs/install/reqs/macos/zsh-config.md
atsansone: They need to know which shell they're running to properly install all of these apps because of where they need to set the permanent PATH. That's why I'm calling this out: They should use the macOS default, `zsh`, and the default file for setting environment variables, `.zshenv`.

src/_includes/docs/install/compiler/xcode.md
atsansone: I changed this to opening a text editor and editing the `~/.zshenv` file.

src/_includes/docs/install/reqs/macos/base.md
```diff
@@ -1,10 +1,22 @@
...
 
 ## System requirements
 
 To install and run Flutter,
-your {{os}} environment must meet these requirements:
+your {{os}} environment must meet the following hardware
+and software requirements.
+
+All of the commands in this guide have been tested on a MacBook Pro M1
+running macOS 14.2.1.
```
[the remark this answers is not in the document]
atsansone: I would rather not. It heads off certain questions.

src/_includes/docs/install/compiler/xcode.md
```diff
@@ -71,34 +71,50 @@ With Xcode, you can run Flutter apps on an iOS device or on the simulator.
...
+   # Configure CocoaPods
+
+   # Set the Ruby gem directory to your .gem directory.
+   export GEM_HOME=$HOME/.gem
+
+   # Check if Flutter is in your PATH, and, if not, add it. 
+   if [[ $PATH != *"$GEM_HOME"* ]]
+   then
+       export PATH=$GEM_HOME/bin:$PATH
+   fi
```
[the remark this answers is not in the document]
atsansone: I understand your point, but don't agree with it. I stand by the change.

src/_includes/docs/install/reqs/macos/set-path.md
```diff
@@ -0,0 +1,52 @@
...
+   # Configure Flutter
+
+   # Set the Flutter home directory to your Flutter install directory.
+   export FLUTTER_HOME=$HOME/development/flutter
+
+   # Check if Flutter is in your PATH, and, if not, add it. 
+   if [[ $PATH != *"$FLUTTER_HOME"* ]]
+   then
+       export PATH=$FLUTTER_HOME/bin:$PATH
+   fi
```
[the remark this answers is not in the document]
atsansone: I understand your point, but don't agree with it. I stand by the change.

src/_includes/docs/install/reqs/macos/set-path.md
christopherfujino: drive by comment, @atsansone you're right, users will copy paste shell snippets to their config. also, many tools auto append. this leads to users having very long, incomprehensible config files and obscure runtime shell environments that we are not easy for my team to debug.

src/_includes/docs/install/reqs/macos/zsh-config.md
atsansone: I understand your point. I am willing to make any updates should they come. I stand by the change.

src/_includes/docs/install/reqs/macos/zsh-config.md
atsansone: I understand your point. I believe this can answer questions that may be raised later. I stand by the change. If the community submits issues to the contrary, we can revisit then.

src/_includes/docs/install/reqs/macos/zsh-config.md
atsansone: Same comment as line 20.

johnpryan (approved):

src/_includes/docs/install/compiler/xcode.md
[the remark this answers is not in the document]
sfshaza2: I agree with John here. In all of my Unix life of 40+ years, I have NEVER defined the GEM_HOME environment variable. Please simplify.

src/_includes/docs/install/compiler/xcode.md
```diff
@@ -71,34 +71,33 @@ With Xcode, you can run Flutter apps on an iOS device or on the simulator.
 
 {% endif %}
 
-### Install CocoaPods
+### Install CocoaPods (Optional)
```
domesticmouse, line 74: Optional feels wrong. If required?

domesticmouse (approved):

src/_includes/docs/install/compiler/xcode.md
sfshaza2: If someone is advanced enough to use Tmux, they should be advanced enough to know how to handle this scenario

sfshaza2:
By the way, regarding "presumes" vs "assumes":
- "Assumes" is much more consistent with our docs
- It's easier for non-native English speakers to understand

PR #10038, opened 2024-01-12
Adds uninstall page

### LINKED ISSUE #5695: Include instructions to uninstall Flutter
Some individuals wanted to uninstall flutter and its folders completely. 
We should include instructions to delete the folders of the flutter installation and the .pub_cache folders 

related to flutter/flutter#30014 and flutter/flutter#29278

atsansone:
Add uninstall page. Fixes #5695

### DIFF
--- a/src/get-started/uninstall/index.md
+++ b/src/get-started/uninstall/index.md
@@ -151,3 +151,9 @@ remove the `.pub-cache` directory from your home directory.
 {% endfor -%}
 </div>
 {% comment %} End: Tab panes. {% endcomment -%}
+
+## Reinstall Flutter
+
+You can [reinstall Flutter](/get-started/install) at any time.
+If you removed the configuration directories,
+reinstalling Flutter restores them to default settings.
--- a/src/get-started/uninstall/index.md
+++ b/src/get-started/uninstall/index.md
@@ -147,6 +147,10 @@ remove the `.pub-cache` directory from your home directory.
 {{rm-pub-dir | strip}}
 ```
 
+{% if os=='Windows' -%}
+{% include docs/install/reqs/{{os | downcase}}/unset-path.md terminal=terminal -%}
+{% endif %}
+
 </div>
 
 {% endfor -%}
--- a/src/get-started/uninstall/index.md
+++ b/src/get-started/uninstall/index.md
@@ -25,33 +25,58 @@ delete the directories that store Flutter and its configuration files.
 
 {% case os %}
 {% when 'Windows' -%}
-   {% assign dirinstall='C:\dev\' %}
-   {% assign dirconfig='%USERPROFILE%\' %}
-   {% assign unzip='Extract-Archive' %}
-   {% assign path='C:\dev\' %}
-   {% assign prompt-start='C:\>' %}
-   {% assign prompt=path | append: '>' %}
-   {% assign terminal='PowerShell' %}
-   {% assign rm = 'Remove-Item -Recurse -Force -Path ' %}
-   {% capture rm-sdk %}Remove-Item -Recurse -Force -Path '{{dirinstall}}flutter'{% endcapture %}
-{% when "macOS" -%}
-   {% assign dirinstall='~/development' %}
-   {% assign dirconfig='~/' %}
-   {% assign path='~/development/' %}
-   {% assign terminal='the Terminal' %}
-   {% assign prompt='$' %}
-   {% assign prompt-start=prompt %}
-   {% assign rm = 'rm -rf ' %}
-   {% assign rm-sdk = 'rm -rf ' | append: dirinstall | append: '/flutter' %}
+{% assign dirinstall='C:\dev\' %}
+{% assign localappdata='%LOCALAPPDATA%\' %}
+{% assign appdata='%APPDATA%\' %}
+{% assign unzip='Extract-Archive' %}
+{% assign path='C:\dev\' %}
+{% assign prompt-start='C:\>' %}
+{% assign prompt=path | append: '>' %}
+{% assign terminal='PowerShell' %}
+{% assign rm = 'Remove-Item -Recurse -Force -Path ' %}
+{% capture rm-sdk %}Remove-Item -Recurse -Force -Path '{{dirinstall}}flutter'{% endcapture %}
+{% capture dart-files %}
+{{localappdata}}.dartServer
+{{appdata}}.dart
+{{appdata}}.dart-tool
+{% endcapture %}
+{% capture rm-dart-files %}
+{{prompt}} {{rm}} {{localappdata}}.dartServer,{{appdata}}.dart,{{appdata}}.dart-tool
+{% endcapture %}
+{% capture flutter-files %}{{appdata}}.flutter-devtools{% endcapture %}
+{% capture rm-flutter-files %}
+{{prompt}} {{rm}} {{flutter-files}}
+{% endcapture %}
+{% capture rm-pub-dir %}
+{{prompt}} {{rm}} {{localappdata}}Pub\Cache
+{% endcapture %}
 {% else -%}
-   {% assign dirinstall='~/development' %}
-   {% assign dirconfig='~/' %}
-   {% assign path='~/development/' %}
-   {% assign terminal='a shell' %}
-   {% assign prompt='$' %}
-   {% assign prompt-start=prompt %}
-   {% assign rm = 'rm -rf ' %}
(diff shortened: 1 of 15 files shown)


src/get-started/uninstall/index.md
```diff
@@ -0,0 +1,153 @@
...
+{% capture rm-flutter-files %}
+{{prompt}} {{rm}} {{flutter-files}}
+{% endcapture %}
+{% capture rm-pub-dir %}
+{{prompt}} {{rm}} {{localappdata}}Pub\Cache
+{% endcapture %}
+{% else -%}
+{% assign dirinstall='~/development' %}
+{% assign dirconfig='~/' %}
+{% assign path='~/development/' %}
```
johnpryan, line 56:
```suggestion
{% assign path='~/path/to/' %}
```

src/get-started/uninstall/index.md
```diff
@@ -0,0 +1,153 @@
...
+
+```terminal
+{{rm-pub-dir | strip}}
+```
+
+</div>
+
+{% endfor -%}
+</div>
+{% comment %} End: Tab panes. {% endcomment -%}
```
johnpryan, line 153: This section should live on dart.dev https://github.com/dart-lang/site-www/issues/3187

johnpryan (approved):

sfshaza2 (commented):
Overall, this page looks good and I like the use of tabs. (But are they visible enough? My eyeballs slid over them on my first read of the staged version.)

- Several times this PR uses the phrase "This guides presumes..." That struct me a bit odd, so I looked into using _presumes_ vs _assumes_. The Google style guide doesn't seem to mention it (at least I couldn't find it), but many sources consider "presume", at least in some instances, rude. I looked at our website, and we use assumes 45 times, and presumes 3 times. I personally prefer "assumes" over "presumes", as the latter seems a bit off-putting, but this is a nit, as it's more of an issue of which is easier for non-native English speakers? If we _do_ choose one over the other, we should list that in go/dart-terms
- Can you add a friendly, "Come back anytime" sort of note at the end?
- I assume you've had this tech reviewed, as I know Unix (so, by extension, macOS, Linux, and Chrome), but not Windows, so I have no idea about those commands.
- This is a nit with the site rendering, not this PR specifically, but at first, I couldn't see the asterisks in the terminal commands. So, instead of seeing "rm -rf .dart*" I saw "rm -rf .dart". The copy button worked fine, of course, and I confirmed it by tanking my .dart files <oops>, but I wish that the text was more readable for the visually impaired (like me)
- @parlough, can you weigh in on the location of this file in the site nav? I have sort of mixed feelings about it, as it's under "Getting Started", but I'm not sure if there's a better place.

src/get-started/uninstall/index.md
atsansone: **comment:** This path would be consistent with the install path in the Install Guides. I could change it there, too. Can we mark this as an issue to resolve separately?

src/get-started/uninstall/index.md
atsansone: **comment:** It can't just live there. The location of the configuration files is the same for both, but a complete removal should include all steps here.

atsansone:
@sfshaza2 : Good points. 

1. I had to check the presume vs. assume issue. [Merriam-Webster clarifies in this way](https://www.merriam-webster.com/grammar/assume-vs-presume#:~:text='Presume'%20is%20the%20word%20to,to%20use%20is%20'assume'.):

> _Although presume and assume both mean "to take something as true," "presume" implies more confidence or evidence backed reasoning. An "assumption" suggests there is little evidence supporting your guess. Think carefully before using them interchangeably or you may lose some meaning._

2. Good point. I'll do that.
3. I tested all of these commands myself on my laptop and my Windows box.
4. I know, right? Same with semi-colons, which are used to put two commands on the same line. Maybe we can fix that.
5. I'm not thrilled with it, but I didn't want to be accused of burying it, either!

src/get-started/uninstall/index.md
johnpryan: Can fix fix this as part of this PR? We shouldn't be specifying `/development/` as the directory you install the SDK in. I install my in `~/code/sdks` so if I followed these instructions I would be doing the wrong thing.

src/get-started/uninstall/index.md
johnpryan: > It can't just live there. 
Yes it can and it should. We don't want duplicate docs between the sites.

src/get-started/uninstall/index.md
sfshaza2: Tony, please ignore my emoji's on your comment, but I see now way to delete them. I meant to thumbs up John's comment.

johnpryan: (You can click the emojis again to delete them)

src/get-started/uninstall/index.md
sfshaza2: Good point, @johnpryan. We do want to avoid duplicatoin, so docs.flutter.dev should talk about removing Flutter. Then, if you want to remove Dart, go to dart.dev.

sfshaza2:
Wow, those asterisks are MUCH more readable! thx

parlough:
> @parlough, can you weigh in on the location of this file in the site nav? I have sort of mixed feelings about it, as it's under "Getting Started", but I'm not sure if there's a better place.

> I'm not thrilled with it, but I didn't want to be accused of burying it, either!

I personally probably would have not added it to the sidenav at all, and instead link to it from some relevant pages. I can't imagine any user ever looking at the site sidenav to figure out how to uninstall Flutter.

If you think it deserves a spot in the sidenav though, I agree that it shouldn't be under "Get Started". I think another fine place would be under "Stay up to date" next to "Update" and "SDK archive" as that's the most relevant content we have.

atsansone:
@sfshaza2 , @parlough : Moved uninstall to under What's New on the left nav.

src/get-started/uninstall/index.md
atsansone: This is not about deleting Dart per se. It's about deleting Flutter. If you delete Flutter, you delete Dart at the same time, not at some other time. It doesn't make sense to send the user elsewhere to delete a config directory for Dart which is part of their complete _Flutter_ install.

johnpryan:
I agree with @parlough that that this link doesn't need to be in the sidenav, unless we have data that suggests a lot of visitors to our docs site are trying to find the uninstall link. Can we add a link to this page from the [get-started/install](https://docs.flutter.dev/get-started/install)? That would probably be the first place most users would look if they want to change things about their installation.

atsansone:
@johnpryan @sfshaza2 @parlough : In reading the original request, linking from Install and/or Update Flutter should be fine. This page needs information on how to remove Flutter from the PATH as well.

sfshaza2 (commented):
So, @atsansone, why the empty https://github.com/flutter/website/pull/10038/files#diff-04964e3043b84ee2b1bbf612391351b1a20b8a156f1629d1c5d6a1d71ca21ee9 file?

I'm mostly ok with this PR, but I am not ok with having instructions for removing Dart on this site, but nothing on the Dart site. That portion should be moved out of this PR and into a PR for dart.dev. And then one can point to the other.

atsansone:
It's the unset path instructions for macOS. I will populate it when the other PR is merged.

There's already an issue on the Dart site that I'll do after this one.

domesticmouse (approved):

sfshaza2 (commented):
It occurs to me that there's some nuance here that needs to be covered.
If they've ONLY installed Flutter (and not Dart), then the Dart installation is underneath Flutter. So removing the Flutter directory ALSO removes the Dart directory. (They still need to remove it from their path, however.)

If they've installed Dart AS WELL as Flutter, then there are TWO installations of Dart. I assume that the PATH refers to the Dart that is inside the Flutter install. This should really be explained somewhere.

atsansone:
> It occurs to me that there's some nuance here that needs to be covered. If they've ONLY installed Flutter (and not Dart), then the Dart installation is **underneath** Flutter. So removing the Flutter directory ALSO removes the Dart directory. (They still need to remove it from their path, however.)
> 
> If they've installed Dart AS WELL as Flutter, then there are TWO installations of Dart. I assume that the PATH refers to the Dart that is inside the Flutter install. This should really be explained somewhere.

The instructions cover removing the _configuration files_ for Dart, which Flutter creates once installed. These instructions don't cover removing Dart itself. 

image

sfshaza2 (approved):
lgtm

PR #10039, opened 2024-01-12
Explain disposables and memory risky objects.

polina-c:


polina-c:
@dnfield  and @kenzieschmoll , could you tech review the change?

domesticmouse (approved):
PTAL @atsansone

src/tools/devtools/memory.md
```diff
@@ -46,6 +46,17 @@ is no longer used (see [Dart garbage collection][]).
 
 [Dart garbage collection]: {{site.medium}}/flutter/flutter-dont-fear-the-garbage-collector-d69b3ff1ca30
 
+### Object types
+
+#### Disposable object
+
+A Dart object, that has method `dispose()`, which, by de-facto convention, is expected to be invoked,
+when the object is not needed any more, to avoid memory leaks.
```
kenzieschmoll, line 54: wording/ punctuation nit: "A Dart object that has a method `dispose()`, which, by convention, is expected to be invoked when the object is not needed anymore to avoid memory leaks."

src/tools/devtools/memory.md
```diff
@@ -46,6 +46,17 @@ is no longer used (see [Dart garbage collection][]).
...
+
+#### Memory-risky object
+
+A Dart object, that, if not disposed or disposed but not GCed, may cause a memory leak.
```
kenzieschmoll, line 58: nit: remove comma after object

src/tools/devtools/memory.md
sfshaza2, line 54:
```suggestion
A Dart object that has a `dispose()` method is, by convention, expected to be invoked
when the object isn't needed anymore to avoid memory leaks.
```

src/tools/devtools/memory.md
sfshaza2, line 58: Can we even remove "might"? Doesn't it for sure cause a memory leak?

```suggestion
A Dart object that is not disposed, or is disposed but not GCed, might cause a memory leak.
```

sfshaza2 (commented):
Tightening up the language a bit.

src/tools/devtools/memory.md
polina-c: Used different rephrasing to keep meaning.

src/tools/devtools/memory.md
polina-c: Nope. It causes memory leak if only not disposed or not GCed.
It is ok, if it is disposed and GCed.

src/tools/devtools/memory.md
```diff
@@ -46,6 +46,17 @@ is no longer used (see [Dart garbage collection][]).
 
 [Dart garbage collection]: {{site.medium}}/flutter/flutter-dont-fear-the-garbage-collector-d69b3ff1ca30
 
+### Object types
+
+#### Disposable object
+
+A Dart object, that has method `dispose()`. The method, by convention, is expected to be invoked
+when the object isn't needed anymore to avoid memory leaks.
```
sfshaza2, line 54:
```suggestion
A disposable object is any Dart object that defines a `dispose()` method.
To avoid memory leaks, Flutter expects that `dispose` is invoked when the
object isn't needed anymore.
```

src/tools/devtools/memory.md
```diff
@@ -46,6 +46,17 @@ is no longer used (see [Dart garbage collection][]).
...
+
+#### Memory-risky object
+
+A Dart object that, if not disposed or disposed but not GCed, might cause a memory leak.
```
sfshaza2, line 58:
```suggestion
A Dart object if not disposed or disposed but not GCed, might cause a memory leak.
```

sfshaza2 (approved):
Polina and I revised over GVC. LGTM

PR #10072, opened 2024-01-21
Remove external icons

parlough:
These are [no longer necessary](https://developers.google.com/style/whats-new#november-8,-2023) in favor of other link guidelines.


domesticmouse (approved):

atsansone (approved):
We may need to check our verbiage around links, but this makes sense.

PR #10099, opened 2024-01-29
Generate an index file for DevTools to fetch release notes from

parlough:
The file would be located at `https://docs.flutter.dev/f/devtools-releases.json`. The generated file currently looks like:

```json
{
  "latest": "2.31.0",
  "releases": {
    "2.31.0": "/tools/devtools/release-notes/release-notes-2.31.0-src.md",
    "2.30.0": "/tools/devtools/release-notes/release-notes-2.30.0-src.md",
    "2.29.0": "/tools/devtools/release-notes/release-notes-2.29.0-src.md",
    "2.28.5": "/tools/devtools/release-notes/release-notes-2.28.5-src.md",
    // ... Remaining entries
  }
}
```

I included the full path as it makes it easier for us to migrate their location in the future, but if you think it's not necessary, let me know. Then I can just include a list rather than a map.

I'm open to suggestions on the file location, JSON format, and/or information contained :)

---

Contributes to https://github.com/flutter/devtools/issues/7114

_As follow-up once migrated away from Jekyll, I will also use this file to generate the page files._


domesticmouse (approved):

atsansone:
@parlough : Why not `/data` so Liquid can use it for loops, etc.?

parlough:
@atsansone The source yaml file is stored in `_data` and used to generate the release notes index as well as this JSON file.

kenzieschmoll (approved):

sfshaza2 (approved):
lgtm

PR #10133, opened 2024-02-11
Update supported versions and platforms for 3.19

parlough:
**Staged:** https://flutter-docs-prod--pr10133-release-3-19-supported-vers-fwawg0o4.web.app/reference/supported-platforms

Supported deployment platform table changes:
- Reorders platforms by category (mobile, desktop, web)
- Uses longer wording instead of `-` and `+` that users have reported confusion of
- Switches 17 to be marked "Supported" on iOS
- Removes arm32 mention from iOS unsupported column, since no arm32 devices support iOS 12.
- Adjust Windows "Unsupported" category to focus on any 32-bit as 32-bit releases still exist for W10
- Move Debian "Supported" to `11, 12` from `10-12` and `10` to "Best effort" since `10` LTS will reach EOL soon, and `9` has.
- Increase Ubuntu "Best effort" from `20.10-23.04` to `22.04-23.10` so we don't include a bunch of unsupported interim releases.
- Adjust Ubuntu unsupported to mention non-LTS before 22.04 as we shouldn't try to support very old interim releases.

I'd love to include Windows 11 in "Supported", but it looks like AMD64 Windows might still be tested only on W10.


sfshaza2 (approved):
lgtm!

PR #10140, opened 2024-02-13
Update Windows supported platforms: Updates Windows 7 and 8 to unsupported

itsjustkevin:
## Description
Updates Windows Supported Deployment Platforms, moving Windows 7 and 8 to the unsupported tier. 

## Issue addressed
- https://github.com/flutter/flutter/issues/140830

## Presubmit checklist

### DIFF
--- a/src/reference/supported-platforms.md
+++ b/src/reference/supported-platforms.md
@@ -19,7 +19,7 @@ which apps built with Flutter might be deployed:
    but are only tested on an ad-hoc basis.
 1. **Unsupported**<br>
    Platforms that we don't test or support.
-   
+
 {% comment %}
 **IMPORTANT NOTE**:
 When changing the minimum support version of macOS, Windows, or Linux,
@@ -48,7 +48,7 @@ following table:
 | web - Firefox  |latest 2 releases|99+| |
 | web - Safari   |latest 2 releases|14+| |
 | web - Edge     |latest 2 releases|96+| |
-| Windows        |10|7, 8, and 11|Vista-, any 32-bit|
+| Windows        |10|11|Vista-, 7, 8, any 32-bit|
 {:.table.table-striped.table-centered-after-first}
 </div>
 


parlough:
I thought 3.19 still supports Windows 7 and 8, then 3.21 will drop them?

domesticmouse (approved):

domesticmouse:
PTAL @sfshaza2

atsansone (approved):
LGTM. Thanks for the update!

parlough (changes requested):
Thanks all for taking a look and approving.

However: Requesting changes to prevent accidental merging so my [above question](https://github.com/flutter/website/pull/10140#issuecomment-1941999247) can be clarified:

> I thought 3.19 still supports Windows 7 and 8, then the next beta and 3.22 will drop them?
> 
> While I'd like to introduce a UI for indicating beta platform support, this table currently is for the stable channel.

itsjustkevin:
> @itsjustkevin @cbracken I thought 3.19 still supports Windows 7 and 8, then the next beta and 3.22 will drop them?
> 
> While I'd like to introduce a UI for indicating beta platform support, this table currently is for the `stable` channel.

The decision came down that this is the last stable release that will support those platforms.

parlough:
By this, do you mean 3.19? If so, we should keep it on here, and then remove it closer to 3.22.

We'll also be publishing a breaking change guide for this on the [breaking change index](https://docs.flutter.dev/release/breaking-changes).

itsjustkevin:
> > The decision came down that this is the last stable release that will support those platforms.
> 
> By "this", do you mean 3.19? That's what Chris mentioned on the linked bug. If so, we should keep it on here for now, and then remove it closer to 3.22.
> 
> We'll also be publishing a breaking change guide for this on the [breaking change index](https://docs.flutter.dev/release/breaking-changes).

Yeah, I see what you mean @parlough.  The timing is a bit confusing here, do we publish after the release, during the next release I am not entirely sure how to effectively communicate this.

parlough:
Generally we communicate changes like this through the https://docs.flutter.dev/release/breaking-changes index page and the Flutter announce Google group: https://groups.google.com/g/flutter-announce. Those mentions can indicate that 3.19 will be the last release with support for Windows 7 and 8, link to relevant material, what developers need to do, etc.

I plan to expand this supported deployment platforms to have a way to indicate what platforms beta supports as well, and I'll make sure to account for this.

itsjustkevin:
100% fair and thanks for this @parlough.  

On the announcement message, I already have one cooked up, just getting it reviewed and set to send.

parlough:
Sounds great, thanks so much Kevin!

I'm going to close this change, but I'll make sure the supported platforms get updated for 3.19 (and again for 3.22), then account for this change where possible.

PR #10154, opened 2024-02-15
Remove Set up Editor link from Install page

### LINKED ISSUE #10109: Can navigate to orphaned "Set up an editor" page from Get started flow
The page can be navigated to with the "Set up an editor" next button on https://docs.flutter.dev/get-started/install, but then it leads to a page not in the sidenav.

Should this page still exist in the flow? If not, should its relevant content be moved to https://docs.flutter.dev/tools/vs-code and https://docs.flutter.dev/tools/android-studio?

atsansone:
Removes next link to Set up Editor Fixes #10109


parlough (approved):
Thanks for following up on this @atsansone! I think this will make more sense now.

PR #10155, opened 2024-02-15
Fixed macOS unzip command

### LINKED ISSUE #10117: Fix unzip command for macOS install page
### Page URL

https://docs.flutter.dev/get-started/install/macos/mobile-android?tab=download#download

### Page source

https://github.com/flutter/website/blob/main/src/_includes/docs/install/flutter/download.md

### Describe the problem

The command
```terminal
$ unzip ~/development/flutter_sdk_v1.0.0.zip ~/development/
```
provided under

> 3. Extract the zip file into the directory you want to store the Flutter SDK.

returns an error and does not work
```terminal
Archive: ~/development/flutter_macos_arm64_3.16.9-stable.zip
caution: filename not matched:  ~/development
```

### Expected fix

To unzip an archive with the `unzip` command in a Unix-like terminal (aka, macOS), we don't specify the target directory as a direct argument. Instead, we use the `-d` option followed by the path to the directory where we want to extract the files.
If the target directory doesn't exist, unzip will attempt to create it.

So the value of `{{uz}}` under

> 3. Extract the zip file into the directory you want to store the Flutter SDK.

must be:
```terminal
$ unzip ~/development/flutter_sdk_v1.0.0.zip -d ~/development/
```

Also, we might want to consider removing this redundant step

> 2. Create a folder where you can install Flutter.

### Additional context

I've tried fixing the issue myself but I'm unable to locate where these template variables (like `{{uz}}`) are defined or how they are managed within the project.

atsansone:
Fixed macOS unzip command. Fixes #10117


sfshaza2 (commented):
So, we want to unzip to a different directory?

PR #10170, opened 2024-02-20
Clean up yt_shims before migration

parlough:
This aligns closer with other similar variables on the site and will make the 11ty migration easier to review. It also avoids the mistake of accidentally not including the shims.


atsansone (approved):
LGTM. Sounds great.

sfshaza2 (approved):
lgtm

Issue #4446, opened 2020-08-05
Preselect OS that we're running on for install page


timsneath:
Page URL: https://flutter.dev/docs/get-started/install/
Page source: https://github.com/flutter/website/tree/master/src/docs/get-started/install/index.md

Description of issue:
Right now, if you click on "Get Started" from the title bar, you get to this page:
https://flutter.dev/docs/get-started/install

I like the visual strength, but this adds another click to the process compared to the old experience. Studies show that every click has a significant impact in funnel completion. We should be able to detect the OS of the user agent and start there, even if we also offer the other options as an easy click. 

See for example how streamlined [VSCode](https://code.visualstudio.com/) have made their experience:
![image](https://user-images.githubusercontent.com/2319867/89435343-97311b00-d6f9-11ea-9a8e-7af6c5fc7ff3.png)
(and obviously, you see a different button if you visit that page from Windows or Linux).

sfshaza2:
This flow hasn't changed recently, so I am a bit confused by your comment, "this adds another click to the process compared to the old experience".  This change would require some infrastructure expertise.

timsneath:
Oh, you're right! It's not one more click after all. The page has [changed in the last couple of months](https://web.archive.org/web/20200616041720/https://flutter.dev/docs/get-started/install), but only for the better. 

Nevertheless, I still think the main point remains -- we can remove an unnecessary click by getting the user straight to install instructions for their current machine, and give them the opportunity to pick a different platform if they need to.

sfshaza2:
Oh, I don't disagree with that.  :). I was just confused by the assertion that it had changed. Yes, the last change was from Patrice (our former infrastructure guy) who improved the Linux option.

I'm definitely leaving this open, but we don't have infrastructure support at the moment.

atsansone:
@parlough : FYI per prior discussions.

\cc: @InMatrix

PR #10171, opened 2024-02-20
Emphasize current OS card on install page

parlough:
I'm not sure we want to pre-navigate to the target page, as then there's the question of how to get back if that's not what they want. What this PR does instead is highlight the current OS's card and auto focuses it on site load. So attention is drawn to the likely correct card and keyboard users can just click enter. I also added some aria labels to make it more clear what the cards are navigating to.

Closes https://github.com/flutter/website/issues/4446

### DIFF
--- a/src/_sass/base/_base.scss
+++ b/src/_sass/base/_base.scss
@@ -38,7 +38,6 @@ dd {
 .card-footer.card-footer--transparent {
   border: none;
   background-color: transparent;
-
 }
 
 .card-footer.card-footer--links {
--- a/src/_sass/vendor/_bootstrap.scss
+++ b/src/_sass/vendor/_bootstrap.scss
@@ -116,6 +116,10 @@ a.card {
   font-family: $site-font-family-alt;
   font-size: bootstrap.$font-size-lg;
   margin-bottom: bs-spacer(2);
+
+  .card-highlight & {
+    font-weight: 700;
+  }
 }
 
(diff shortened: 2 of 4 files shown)


loic-sharma:
Could this be done without moving focus, perhaps through just styling? Moving the initial focus can be confusing when using a screen reader, even with the added labels.

atsansone:
@parlough : This is a nice add, but the issue asks to steer the reader directly to the necessary page. A macOS user would be redirected to https://docs.flutter.dev/get-started/install/macos for example.

If that's not possible, I would agree with @loic-sharma : Maybe just change the style to the aligned OS rather than the focus because that does alter the a11y of the page.

parlough:
> The issue asks to steer the reader directly to the necessary page. A macOS user would be redirected to https://docs.flutter.dev/get-started/install/macos for example.

This is possible, but then some UI element or text would need to be added near the top of that page to help misdirected users back. The extra clutter doesn't seem worth it to me, but I'd be happy to explore ideas people have.

> Moving the initial focus can be confusing when using a screen reader, even with the added labels.

Yeah I was worried about this. Thanks for pushing me away from it. I'll remove the autofocus, which can likely be better handled by a standardized, hidden "Jump to main content" button anyway.

parlough:
Closing this for now. I’ll revisit this with a less intrusive experience soon. Thanks for the feedback!

PR #10176, opened 2024-02-21
Remove and clean up old top-level event and competition pages

parlough:
Removes the old `/apprentice-giveaway`, `/adoptawidget`, `/clock`, and `/create` top-level marketing pages, their custom styles, images, etc. None have been used or referenced for multiple years now. This PR does however introduce some relevant redirect destinations for those paths.

This will help the site and its styles to be be easier to maintain, slightly improve site build times, and ease the migration to 11ty.


firebase.json
```diff
@@ -206,7 +206,7 @@
       { "source": "/whats-new", "destination": "/release/whats-new", "type": 301 },
       { "source": "/testing/oem-debuggers", "destination": "/testing/native-debugging", "type": 301 },
 
-      { "source": "/adoptawidget", "destination": "https://github.com/flutter/flutter/blob/master/CONTRIBUTING.md", "type": 301 },
+      { "source": "/adoptawidget", "destination": "https://github.com/flutter/flutter/blob/main/CONTRIBUTING.md", "type": 301 },
```
domesticmouse, line 209: As far as I can tell `master` is still the default channel on github.

domesticmouse (approved):

sfshaza2 (approved):
Yeah, I am always loathe to delete that stuff, as I tend to reference it when we offer similar challenges. However, that stuff is now primarily on the marketing channel. LGTM

parlough:
> Yeah, I am always loathe to delete that stuff, as I tend to reference it when we offer similar challenges. However, that stuff is now primarily on the marketing channel. LGTM

Yeah I've hesitated to do it, but it felt right to do it during this larger change. Especially since some include content that might be confusing or misleading now that they are over.

A few have pages have results blog posts which I redirected to here, which is nice. I also made sure the web archive has an up-to-date archive of each page before opening the PR, and of course the git history is there if we ever want to revive them (or a shorter history) in some form :)

sfshaza2:
Thanks, Parker! Landing.

firebase.json
parlough: Yeah it is, but `main` also works and that's the intended end state, so I make sure to switch to it where possible. Helps us avoid making that transition any more challenging.

Let me know if you think it's important to change it back though!

goderbauer:
@sfshaza2 @parlough Two images (`reflectly-hero-600px.png` and `dart-diagram-small.png`) removed in this PR are referenced from Flutter's README, see the broken embeds of "Reflectly hero image" and "Dart diagram" at https://github.com/flutter/flutter/blob/master/README.md.

Could those images be restored or could you fix up the README if they shouldn't be used anymore there?

parlough:
Ah thanks for calling that out! I opened https://github.com/flutter/website/pull/10240 to get the links working again and will work on a longer term solution and update once I get a chance.

I did check if any Dart and Flutter repos were using the images, but I checked any using the hosted versions on the site, I didn't consider the raw GitHub links being used. Sorry about that!

goderbauer:
No worries and thank you for the quick fix!

PR #10193, opened 2024-02-22
Fixed chipsets to supported platforms

### LINKED ISSUE #10192: Update chipset architectures for Flutter
### Page URL

https://docs.flutter.dev/reference/supported-platforms

### Page source

_No response_

### Describe the problem

The chipset list needs to be updated for Android, macOS, Linux, and Windows.

Per @reidbaker : Android supports x64, Arm32, and Arm64 chipsets plus the x86 via emulator.
Per @cbracken : macOS supports x64, Arm64.

### Expected fix

Update the chipset table.

### Additional context

_No response_

- [X] I will try and fix this problem on docs.flutter.dev.

atsansone:
Fixed hardware architectures to supported platforms. Fixes #10192


reidbaker:
I was just informed that there is a way to make x86 32bit flutter work. Documentation is here https://github.com/flutter/flutter/wiki/JIT-Release-Modes the performance is not great and this is an "at your own risk" level of support. 

No changes requested to the support documentation.

src/_data/chipsets.yml
```diff
@@ -9,11 +9,11 @@
   x64: '<span class="material-symbols" style="color: #158477" aria-label="The Flutter SDK supports the x64 architecture on Linux" role="img">verified</span>'
   arm32: '<span class="material-symbols" style="color: #D43324" aria-label="The Flutter SDK does not support the Arm32 architecture on Linux" role="img">dangerous</span>'
   arm64: '<span class="material-symbols" style="color: #158477" aria-label="The Flutter SDK supports the Arm64 architecture on Linux" role="img">verified</span> '
-  risc: '<span class="material-symbols" style="color: #D43324" aria-label="The Flutter SDK does not support the RV64GC architecture on Linux" role="img">dangerous</span>'
+  risc: '<span class="material-symbols" style="color: #DADCE0" aria-label="No version of the Flutter SDK exists for the RV64GC architecture on Linux" role="img">do_not_disturb_on</span>'
```
cbracken, line 12: Question: why is this phrased differently for Linux than for macOS and Windows, given all three only support x64 and arm64?

cbracken (approved):
lgtm for desktop modulo one question.

sfshaza2 (commented):
Once Chris and Parker are happy, we can land this.

src/_data/chipsets.yml
atsansone: All target platforms don't support RV64GC. It is the same language for all platforms ([Windows](https://github.com/flutter/website/pull/10193/files/759a42c219f07baebbfc46a6459730cec4c5f665#diff-275e58b217f338e06bd26b903fb57382eb3d972beef3a73583adcae657b9297cR6), [macOS](https://github.com/flutter/website/pull/10193/files/759a42c219f07baebbfc46a6459730cec4c5f665#diff-275e58b217f338e06bd26b903fb57382eb3d972beef3a73583adcae657b9297cR18), [Linux](https://github.com/flutter/website/pull/10193/files/759a42c219f07baebbfc46a6459730cec4c5f665#diff-275e58b217f338e06bd26b903fb57382eb3d972beef3a73583adcae657b9297cR12), [iOS](https://github.com/flutter/website/pull/10193/files/759a42c219f07baebbfc46a6459730cec4c5f665#diff-275e58b217f338e06bd26b903fb57382eb3d972beef3a73583adcae657b9297cR24), and [Android](https://github.com/flutter/website/pull/10193/files/759a42c219f07baebbfc46a6459730cec4c5f665#diff-275e58b217f338e06bd26b903fb57382eb3d972beef3a73583adcae657b9297cR30)).

dnfield (changes requested):
I think the table is confusing and should be removed.

I'm confused about the difference between "unsupported on all channels" and "no version exists." Are there actually arm32 Windows devices out there that people would like us to support but we're not? There definitely are arm32 iOS devices out there somewhere that we no longer support (but you could download an old version of Flutter and make it kind of work). Why is the Windows one red and the iOS one grey? Ditto on macOS and probably Linux. I'm not aware of any RISCV 64 platforms that are currently widely available for end users, and in fact Flutter/Dart is relatively close to being able to support it (we're mainly not because there's not demand yet). Why is that red when other things are grey? IA32 architectures are exceedingly rare at this point as well.


I think the table would make more sense as prose for each platform. On most platforms, it doesn't even make sense to bring up x86 at this point because the devices are just pretty rare and old - we could mention some of the unique cases more easily this way without creating confusion about why we'd support x86 in emulation on Android but not iOS (where it's never existed). Ditto on RISCV, where maybe it will exist soon but isn't quite there yet.

cbracken:
Perhaps these could be inlined in brackets beside the platform names in the main chart, which either way, I think should probably go at the top since that's the main thing people come to this page looking for.

Perhaps something like:
Windows (x64, arm64)
macOS (x64, arm64)
Debian Linux (x64, arm64)
...

atsansone:
@cbracken : I like this idea. The original PR was to make this table mirror its Dart equivalent, but this could work, too.

@dnfield : Thoughts?

dnfield:
I think the new chart looks good. I mentioned offline, browsers should just have `N/A`, and iOS should not have `x64`

dnfield (approved):

dnfield:
LGTM, thanks!

PR #10174, opened 2024-02-20
Added chipsets to supported platforms

atsansone:
Added hardware architectures to supported platforms. 

Fixes #10136
Fixes #6713

PR #10197, opened 2024-02-23
Revert "Added chipsets to supported platforms"

zanderso:
Reverts flutter/website#10174

The update has incorrect information, and the folks who would need to correct it are OOO, so I'm sending this revert to at least go back to the old state until the changes can be reviewed by the most knowledgable team members.


atsansone:
A correction was reviewed in #10193. That will update this page.

Issue #9450, opened 2023-09-25
Review and update docker command in website setup


nithinkodadhavadee:
### Describe the problem

I was trying to build this site locally and faced issues with the command `make setup`. I am new to docker so I could not understand the error that I got.

this is the error I got:
```
touch .env
make clean
make[1]: Entering directory '/home/somanna/Scripts/flutter/website'
rm -rf _site .jekyll* src/.jekyll* *.log tmp .dart_tool
make[1]: Leaving directory '/home/somanna/Scripts/flutter/website'
docker-compose build --no-cache site
make: docker-compose: No such file or directory
make: *** [Makefile:62: setup] Error 127

```

### Expected fix

I wonder if `docker-compose` can be replaced with `docker compose` as the earlier one is no longer available. I have the necessary changes, so I guess I can create a pull request. 

### Additional context

after a million Google searches later I stumbled upon this [post from starkoverflow](https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose) which says that docker-compose has been deprecated so I am supposed to use `docker compose` instead. I made a few changes to the makefile, removing dashes from `docker-compose` to make it `docker compose`.  Now I can successfully build the site and use the docs.

Issue #10203, opened 2024-02-26
Remove reliance on Docker and migrate away from Jekyll


parlough:

sfshaza2:
Where are we on this, @parlough? Given that it's a P1 bug, I'd like to get this addressed and closed.

parlough:
Working on fixing the final discrepancies in https://github.com/flutter/website/pull/10336 :)

PR #10216, opened 2024-02-28
Remove most Docker usage

parlough:
This PR results in most of the site infrastructure no longer being functional when developing locally. This is a temporary step in the transition away from Docker and Jekyll to make the final migration and its change review easier. You'll need to temporarily rely on the testing and staging on GitHub. Some testing can be ran locally with `dart run flutter_site <command>` if you have the latest Flutter installation.

This PR also pulls out the code excerpt check, as it's independent of the Dart channel and testing. This avoids unnecessary work and enables quicker visibility into failure reasons on CI :)

Please let me know if you have any concerns with this landing or need help with something after it lands. Thank you!

Closes https://github.com/flutter/website/issues/9450


.dockerignore
```diff
@@ -2,8 +2,7 @@
 .gitignore
 AUTHORS
 CONTRIBUTING.md
-docker-compose.yml
 LICENSE
 node_modules/
 README.md
-flutter/
+./flutter/
```
domesticmouse, line 8: I believe this should be `/flutter/`

domesticmouse (approved):

.dockerignore
parlough: Good point. Thanks!

Revisiting this, we don't even need the entry anymore anyway since we don't clone Flutter in the build process anymore.

PR #10226, opened 2024-02-29
Remove usages of `site.url` variable

parlough:
We don't rely on this, nor have we ever. It has caused confusion and inconsistencies, ending up being one of the most common issues when a new contributor opens a PR.

The downstream Chinese-translated version of the docs site does however rely on this variable being used. Dropping it will result in a temporary regression for them that I think they shouldn't account for by staying on an older commit until https://github.com/flutter/website/issues/10203 is solved. As another solution would end up being temporary and likely not worth the effort.


sfshaza2 (approved):
Thanks for letting me know about this ahead of the PR! I will be glad to get rid of this macro!  lgtm

parlough:
@AmosHuKe @AlexV525 👋 Just wanted to make you aware that I'll be dropping the `{{site.url}}` usage from the site to make some upcoming changes easier to review and land. If this causes issues for flutter.cn, I recommend holding off on updating to a commit with these changes. Then once my followup work lands, we can explore a better solution that works for you compared to always requiring the `{{site.url}}`.

As always, thanks for all the work you do, and sorry about the inconvenience. I'll make sure to keep you updated! Let me know if you have any concerns though.

AlexV525:
@parlough We shall have discussions about this when we are about to upgrade, soon or later. Thanks for the heads up!

PR #10267, opened 2024-03-11
Only create TOC entries for h2 and h3

parlough:
Limits TOC entries to h2 and h3 as it's usually not desired for h4 to be included anyway, as can be seen by the removal of `no_toc` in many h4 cases here. In the future, we can configure this per page as necessary.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -126,7 +126,7 @@ plugins:
 
 toc:
   min_level: 2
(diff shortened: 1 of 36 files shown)


sfshaza2 (approved):
Thank you! This really bugs me when I see it. lgtm

PR #10268, opened 2024-03-11
Remove use of include_cached plugin

parlough:
Contributes to https://github.com/flutter/website/issues/10203 as we'll only use normal Liquid includes and render syntax after the migration.

The usage of `include_cached` was simply a performance optimization, so this doesn't affect rendering results.


sfshaza2 (approved):
lgtm

Issue #10290, opened 2024-03-14
Bitrise link are broken, route doesn´t exists


hansobvius:
### Page URL

https://docs.flutter.dev/deployment/cd/

### Describe the problem

![image](https://github.com/flutter/website/assets/21185950/26c5bd8e-df3b-4f20-93f4-aea1fcc0209f)

PR #10297, opened 2024-03-16
Fix links to Bitrise Flutter docs

parlough:
They also don't have a link destination that automatically picks a language, so they have to include the `/en/`.

Fixes https://github.com/flutter/website/issues/10290


sfshaza2 (approved):
lgtm

PR #10301, opened 2024-03-16
Remove transitional 'GPU thread' mentions

parlough:
It has been around 3+ years since the transition was made, so we can probably remove these transitional mentions now. I've kept a modified form of one mention as it still seemed useful.

\cc @filiph


filiph:
This LGTM, thanks @parlough! I'm glad the term "GPU" is still around, for SEO.

PR #10322, opened 2024-03-21
Minimize Windows Android install list to only Android Studio

CaseyHillers:

https://docs.flutter.dev/get-started/install/windows/mobile?tab=virtual shows all the install options for Windows devices. There's seems to be a lot unnecessary steps added, and this seems to be it falling back to not having the target set correctly.

I would expect this list to be [Android Studio] instead of [Visual Studio, Android Studio, Google Chrome]

N/A

## Presubmit checklist

### DIFF
--- a/src/get-started/install/windows/mobile.md
+++ b/src/get-started/install/windows/mobile.md
@@ -2,7 +2,7 @@
 title: Start building Flutter Android apps on Windows
 description: Configure your system to develop Flutter mobile apps on Windows.
 short-title: Make Android apps
-target: Android
+target: mobile
 config: WindowsAndroid
 devos: Windows
 next:


atsansone:
@CaseyHillers : That value can't be changed without altering the includes and how variables are passed. Happy to discuss.

CaseyHillers:
Do you have any pointers or docs I can take a look at to learn more about the template language? I was looking at the target tags, and the included one seems to be looking for "mobile" but that didn't seem to work here.

atsansone:
@CaseyHillers : I'm fixing this as a part of https://github.com/flutter/website/pull/10393. The language is called Liquid, made by Shopify.

PR #10329, opened 2024-03-21
Added help for Flutter web

### LINKED ISSUE #9792: Add remedy to 'Web enable for existing app' error on  'Building a web application with Flutter' page
### Page URL

https://docs.flutter.dev/platform-integration/web/building/

### Page source

https://github.com/flutter/website/tree/main/src/platform-integration/web/building.md

### Describe the problem

In the last section 'Add web support to an existing app' there is no guidance on what to do if after the stated command the response is 'The web is currently not supported on your local environment.'

### Expected fix

After the stated command add the following:
If you get the message 'The web is currently not supported on your local environment.' then run the following command:
$ flutter config --enable-web
Then run the above flutter create command again.

### Additional context

I copied a project from one Ubuntu development computer to another. The editor in both computers is Visual Studio Code. Flutter version is:

Flutter 3.13.2 • channel stable • https://github.com/flutter/flutter.git
Framework • revision ff5b5b5fa6 (3 months ago) • 2023-08-24 08:12:28 -0500
Engine • revision b20183e040
Tools • Dart 3.1.0 • DevTools 2.25.0

Although the chrome device was in the device list in the original project, it was missing in the project copied to the new development computer. When I ran the 'flutter create --platforms web . command I received the following response:

You can find general documentation for Flutter at: https://docs.flutter.dev/
Detailed API documentation is available at: https://api.flutter.dev/
If you prefer video documentation, consider: https://www.youtube.com/c/flutterdev

In order to run your application, type:

  $ cd .
  $ flutter run

Your application code is in ./lib/main.dart.

**The web is currently not supported on your local environment.**
For more details, see: https://flutter.dev/docs/get-started/web

I ran the command: flutter config --enable-web and then ran the flutter create command again to fix the problem.

I believe it would save time for those not familiar with flutter config.

atsansone:
Added help for Flutter web. Fixes #9792

### DIFF
--- a/src/platform-integration/web/building.md
+++ b/src/platform-integration/web/building.md
@@ -124,6 +124,12 @@ Run the following command to generate a release build:
 $ flutter build web
 ```
 
+If you receive a `not supported` error, run the following command:
+
+```terminal
+$ flutter config --enable-web
+```
+
 A release build uses [dart2js][]
 (instead of the [development compiler][])
 to produce a single JavaScript file `main.dart.js`.
@@ -137,7 +143,7 @@ You can also include `--web-renderer html`  or `--web-renderer canvaskit` to
 select between the HTML or CanvasKit renderers, respectively. For more
 information, see [Web renderers][].
 
-For more information, see
+To learn more, see
 [Build and release a web app][].
 
 ## Add web support to an existing app
@@ -149,6 +155,12 @@ from your project's top-level directory:
 
 ```terminal
 $ flutter create --platforms web .
+```
+
+If you receive a `not supported` error, run the following command:
+
+```terminal
+$ flutter config --enable-web
 ```
 
 [Build and release a web app]: /deployment/web


src/platform-integration/web/building.md
```diff
@@ -124,6 +124,12 @@ Run the following command to generate a release build:
 $ flutter build web
 ```
 
+If you receive a `not supported` error, run the following command:
+
+```terminal
+$ flutter config --enable-web
```
domesticmouse, line 130: I'm a little confused on this one. My understanding is that web is enabled by default, and the only time that web will be disabled is when a user has explicitly run `flutter config --no-enable-web` at some point in the past for some reason.

Thoughts @kevmoo ?

kevmoo: I'm not familiar w/ this aspect of the CLI *at all*

domesticmouse: In which case, let's land it as is. For most people this will be a null op. I'd suggest that maybe we think seriously about ripping this functionality out of the CLI, however.

domesticmouse (approved):

Issue #10263, opened 2024-03-09
Remove the graf about CanvasKit performance on 'Displaying images on the web' page


bahaa-kallas:
### Page URL

https://docs.flutter.dev/platform-integration/web/web-images/

### Page source

https://github.com/flutter/website/tree/main/src/platform-integration/web/web-images.md

### Describe the problem

> [As of today](https://github.com/flutter/website/blob/main/src/platform-integration/web/%7B%7Bsite.repo.flutter%7D%7D/issues/71884), using too many HTML elements with the CanvasKit renderer might hurt performance. If images interleave non-image content Flutter needs to create extra WebGL contexts between the <img> elements. If your application needs to display a lot of images on the same screen all at once, consider using the HTML renderer instead of CanvasKit.

This section references a bug ticket that was marked as completed. Therefore the note "probably" no longer a valid. 

### Expected fix

Assuming that the performance issue was fixed and the note no longer applies. Removing the entire section would make sense. 
But if the note is still relevant then The linked issue in the note must be removed because its "allegedly" fixed (Based on the closing comment) otherwise it will lead to some confusion for the reader. 

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

PR #10330, opened 2024-03-24
Clean up web images layout, links, and formatting

parlough:
- Removes mention of old CanvasKit limitation
  - Fixes https://github.com/flutter/website/issues/10263
- Don't have h1 headers inline, as there shouldn't be more than one
  - Adjust other headers for that level change
- Use non-English MDN links to enable navigation to user's preferred language
- Include link definitions near usage

### DIFF
--- a/src/platform-integration/web/web-images.md
+++ b/src/platform-integration/web/web-images.md
@@ -1,37 +1,41 @@
 ---
-title: Displaying images on the web
+title: Display images on the web
 short-title: Web images
 description: Learn how to load and display images on the web.
 ---
 
-The web supports the standard [`Image`][1] widget to display images.
+The web supports the standard [`Image`][] widget to display images.
 However, because web browsers are built to run untrusted code safely,
 there are certain limitations in what you can do with images compared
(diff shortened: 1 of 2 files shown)


src/platform-integration/web/web-images.md
```diff
@@ -1,37 +1,41 @@
...
 This section summarizes the technologies available
 across Flutter and the web,
 on which the solutions below are based on.
 
-## Images in Flutter
+### Images in Flutter
 
-Flutter offers the [`Image`][1] widget as well as the low-level
-[`dart:ui/Image`][11] class for rendering images.
+Flutter offers the [`Image`] widget as well as the low-level
```
domesticmouse, line 23: Is this a valid link format in MD?

domesticmouse (approved):

src/platform-integration/web/web-images.md
parlough: Oops, thanks for calling that out.

It is, but we prefer to use the full `[text][]` format for consistency with `[text][link definition]` syntax.

Issue #10271, opened 2024-03-12
VS Code links to docs pages that seem to no longer resolve


mit-mit:
As part of the getting stated install flow, our VS Code plugins link to instructions in our docs for setting path using Urls like `https://docs.flutter.dev/get-started/install/windows#update-your-path`:
https://github.com/Dart-Code/Dart-Code/blob/master/src/shared/constants.ts#L96

Did these Urls change recently? It looks like they no longer resolve. If so, we should put redirects in place.

atsansone:
Redirects:

- https://docs.flutter.dev/get-started/install/windows/desktop?tab=download#update-your-windows-path-variable
- https://docs.flutter.dev/get-started/install/macos/desktop?tab=download#add-flutter-to-your-path

Linux will change after #10302 merges.

mit-mit:
@atsansone was this resolved?

atsansone:
@mit-mit : I may need to create a hidden page that covers this. I checked with @kenzieschmoll about changing this in future DevTool versions.

mit-mit:
We still need older releases to work though. Can you just add redirects for the old Urls?

atsansone:
Unfortunately, no. The Firebase JSON redirects can't point from URL fragment to URL fragment. (See https://github.com/angular/angular/issues/45861.)

So, this redirect can't be made:
```json
{ "source": "/get-started/install/:platform#update-your-path",
"destination":
/get-started/install/:platform/desktop?tab=download#add-flutter-to-your-path", "type": 301 },
```

This is why I'm thinking about a hidden page. That way I can link to a whole page, rather than a URL fragment.

sfshaza2:
Oh, about the hidden page idea. Would it require documenting how to set the path for ALL platforms and ALL shells? If so, that seems pretty icky

sfshaza2:
Actually, I think aa better solution is to create a "go" link, such as "docs.flutter.dev/go/set-path" and that link can be changed without any issues, I think. Once again, this requires changing the VSCode plugin, as well as the site, but it feels much cleaner than the hidden page.

kenzieschmoll:
CC @DanTup

parlough:
> We still need older releases to work though. Can you just add redirects for the old Urls?

[...]

PR #10357, opened 2024-04-03
Add more client-side fragment redirects for old get started pages

parlough:
This adds redirects from the `/get-started/install/<platform>` pages if a fragment (`#link-target-fragment`) from the old page format is specified. These JS based redirects are necessary as fragments aren't sent to the server, so Firebase redirects won't work.

We likely won't maintain these indefinitely, but a lot of links referencing these still exist in old versions of tools and the internet, so we can keep them for the foreseeable future.

Fixes https://github.com/flutter/website/issues/10271


domesticmouse (approved):

sfshaza2 (approved):
Thanks so much, @parlough! A very workable solution. In future, I think it's best if we set up go links for tooling links.

PR #10362, opened 2024-04-04
Add missing note about building Android plugin

stuartmorgan:
Every other platform's section had the note about needing to build at least once before opening the native project, but it was missing for Android; this adds it.

It also updates the iOS and macOS docs to add the `--config-only` flag, which is an option that was added after these docs were originally written, and makes the step much faster. (Windows and Linux are unchanged since that option does not yet exist for those platforms).

## Presubmit checklist


sfshaza2 (approved):
lgtm.  Thx, @stuartmorgan!

PR #10373, opened 2024-04-05
Fix code blocks on layout index page

parlough:
Standard Markdown requires an extra new line after HTML before Markdown markup.


domesticmouse (approved):

sfshaza2 (approved):
lgtm

PR #10374, opened 2024-04-05
Don't use definition list and list syntax at same time

parlough:


domesticmouse (approved):

sfshaza2 (approved):
ahhh. I like it sometimes. lgtm

Issue #10364, opened 2024-04-04
Evaluate order of Kotlin/Java snippets


parlough:
Evaluate each location which has a Kotlin and Java snippet to see what order makes sense. Consider preferring the Kotlin snippet for most scenarios, potentially besides add-to-app.

PR #10375, opened 2024-04-05
Reorder duplicated Kotlin and Java snippets

parlough:
After discussing with others, it seems to make sense to place Kotlin snippets before Java. It's been the Flutter tool's default for many years and is now Android's default as well. The one exception is in the add-to-app docs, where Java might be more common for old apps, but this PR also puts Kotlin first there as well for consistency.

Closes https://github.com/flutter/website/issues/10364


domesticmouse (approved):

sfshaza2 (approved):
lgtm

PR #10383, opened 2024-04-08
Speed up code excerpt updater

parlough:
Slightly reduce runtime of code excerpt updater by doing less unneeded work :) 

- Only run on `src/content` directory. All code excerpt instructions are within that directory currently.
- Don't add any of the global replacements, they currently are relied on by any of the excerpts on the site.


domesticmouse (approved):

sfshaza2 (approved):
lgtm. woot!!

Issue #10202, opened 2024-02-26
Update DevTools install references to start or use


parlough:
DevTools used to be installed from pub, but that's no longer necessary as it has been bundled with the Dart SDK (and therefore Flutter) for a few years now. We should go through and update all mentions of installation to instead discuss or focus on usage.

PR #10390, opened 2024-04-08
Doing a pass to make sure that DevTools is correctly used

sfshaza2:
Fixes https://github.com/flutter/website/issues/10202

### DIFF
--- a/src/content/tools/devtools/app-size.md
+++ b/src/content/tools/devtools/app-size.md
@@ -1,5 +1,5 @@
 ---
-title: Using the app size tool
+title: Use the app size tool
 description: Learn how to use the DevTools app size tool.
 ---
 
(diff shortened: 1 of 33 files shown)


domesticmouse (approved):

atsansone:
Looks good, but may want to check with @mit about naming. I think I remember being told to just use "Dart DevTools" without Flutter in the name at all.

sfshaza2:
@mit-mit, the last time we discussed this, we were told to call it either Flutter DevTools or Dart DevTools, or both. @atsansone says that the advice is now to call them Dart DevTools. Please clarify.

src/content/tools/devtools/app-size.md
```diff
@@ -267,7 +267,7 @@ the Wonderous App using DevTools, check out the
 [App Size Tool tutorial][app-size-tutorial]. Various strategies
 to reduce an app's size are also discussed.
 
-[Use the treemap]: #using-the-treemap
+[Use the treemap]: #use-the-treemap
```
domesticmouse, line 270: Won't this change break incoming references?

sfshaza2: If there are any, yup.

sfshaza2:
@mit-mit, I'm going to land this. But, when you return, let me know if you want to chat about it.

Issue #9527, opened 2023-10-16
Add alt text


atsansone:
Add `alt` attribute to all images on the site. This violates WCAG [H37](https://www.w3.org/TR/WCAG20-TECHS/H37).

Refer to [the W3C’s alt Decision Tree](https://www.w3.org/WAI/tutorials/images/decision-tree/) and [Axess Lab’s Ultimate Guide to Alt Texts](https://axesslab.com/alt-texts/) for guidance.

Issue #6487, opened 2021-11-18
Improve image format/sizing and caching


guidezpl:
### Describe the problem

The only major performance issue highlighted by Lighthouse regards images. Estimated savings for serving images in next-gen formats, efficiently encoding images, and properly sizing them are ~1.9s

Assets also all have a 1h cache policy, which seems quite short for static content.

### Expected fix
- [ ] use a better image format
- [ ] optimize/resize images
- [ ] increase cache policy for assets

danagbemava-nc:
Screenshots of Lighthouse report for context.

<details>
<summary>screenshots</summary>

![Screenshot 2021-11-19 at 10 57 23](https://user-images.githubusercontent.com/88313112/142611911-08d8d9fd-c113-4ccd-b3c6-5a70c23e6318.png)


![Screenshot 2021-11-19 at 10 58 14](https://user-images.githubusercontent.com/88313112/142611907-0af8485c-acc3-43bd-ab75-8541dc53f4d1.png)



</details>

guidezpl:
@parlough I believe I saw PRs from you that fixed this?

PR #10391, opened 2024-04-09
Optimize images as part of production builds

parlough:
Use 11ty's first-party [image transformation plugin](https://www.11ty.dev/docs/plugins/image/) to automatically optimize images, convert them to `png`, `webp`, and `avif`, and then transform the site HTML to use the [`<picture>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture) to select the best one for the user. This only runs in production deploy builds to reduce serving time. Also applies lazy loading and async decoding to all images, even when serving without optimizations. The transform also adds a hash of the image, allowing us to expand caching for images in Firebase hosting.

To account for the transformed HTML structure, some minor HTML and CSS changes were needed as well.

Overall, this reduces page load time on pages that use images, and reduces unnecessary downloads that were due to a relatively short cache time for images. On a page with a few images, I saw lighthouse perf score increases from 5-15 points.

Closes https://github.com/flutter/website/issues/6487
Closes https://github.com/flutter/website/issues/9527 as this plugin also validates `alt` is always specified


sfshaza2:
Parker, can we either land this or close? I really hate stinky PRs that hang around for weeks and weeks.

parlough:
@sfshaza2 Yep, sorry! I was waiting for the dart.dev equivalent to land to work out the kinks there first. It landed today with https://github.com/dart-lang/site-www/commit/bcbff3ea28c25a6d2fb44b2e5e3b761c704a4a7f.

I'll work on getting this updated and ready for review sometime this week.

domesticmouse (approved):

Issue #10389, opened 2024-04-08
Review use of "devtools" in Getting Started


sfshaza2:
### Page URL

https://docs.flutter.dev/platform-integration/android/install-android/install-android-from-linux?tab=virtual

### Page source

/src/content/platform-integration/android/install-android/install-android-from-linux.md

### Describe the problem

This issue applies to this page and many other pages. Recently some pages were updated/added that don't use the terms "DevTools" or "web" correctly.

First, it's either Flutter DevTools, Dart DevTools (both refer to the same suite of tools), or Chrome DevTools. So far as I know, none of these DevTools run on any other browsers besides Chrome, so "web DevTools" is incorrect.

Also, check the Google word list. Unless it's the start of a sentence, it's "web", not "Web".

Finally, "devtools" is not a word. We are talking about two specific sets of tooling. We don't want to weaken the meaning of that term by devolving into a generic short word that could apply to any tooling that is developer related.

Also, regarding the pages that are titled as follows: "Add Android devtools for Flutter from Linux start". I don't even know what this title means (nor any of those pages titled similarly). You don't install DevTools, you get it when you download Flutter/Dart. So you launch it or open it. You do install the Flutter/Dart plugins in your IDE, and maybe you meant that but what is "from a Linux start"? (Another experienced Flutter user told me that they don't understand these titles either).

Also, what are "Android devtools"?

cc @kenzieschmoll 

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

parlough:
There's still cases of this on the site.

PR #10402, opened 2024-04-11
More fixes to adding a platform pages

parlough:
- "devtools" is a term that generally refers to the Dart/Flutter DevTools app, we shouldn't conflate it with the various apps/software/tools developers using when building for/developing on/debugging against a platform. This PR continues and expand on changes began in https://github.com/flutter/website/commit/8c0003e94988362179fd8532e201439425ed5ecd to move away from these usages of "devtools".
- The [web](https://docs.flutter.dev/platform-integration/web/install-web) and [android](https://docs.flutter.dev/platform-integration/android/install-android) pages had some rendering issues due to whitespace causing rendering differences. This PR addresses those.
- This PR also standardize styles and Liquid usage between the equivalent pages for different platforms.

Feel free to push any changes you'd like to this branch. I haven't had a chance to understand what content is covered or to put too much thought into the names. I put the page names as "Add X as a target platform ..." for now to be consistent with https://github.com/flutter/website/commit/8c0003e94988362179fd8532e201439425ed5ecd, but it seems the content might actually be "Set up your development environment for targeting X platform with Flutter"?

Closes https://github.com/flutter/website/issues/10389

@atsansone @sfshaza2 I've having trouble wrapping my head around these pages, or perhaps just their structure. It's also not clear to me what developers who didn't or don't follow the current version of the getting started instructions should do. I think it might be beneficial if us three could chat through this whole flow together. For now though, I kept this individual PR focused on fixing the rendering issues and replacing the "devtools" mentions.

### DIFF
--- a/src/_includes/docs/install/next-steps.md
+++ b/src/_includes/docs/install/next-steps.md
@@ -34,15 +34,15 @@ consult the following resources.
 {%- endcase %}
 {%- case config.add-xcode %}
 {%- when 'Y' %}
-* [Add macOS compliation tools](/platform-integration/macos/install-macos/install-macos-from-{{target}})
+* [Add macOS compilation tools](/platform-integration/macos/install-macos/install-macos-from-{{target}})
 {%- endcase %}
 {%- case config.add-linux-tools %}
 {%- when 'Y' %}
-* [Add Linux compiliation tools](/platform-integration/linux/install-linux/install-linux-from-{{target}})
+* [Add Linux compilation tools](/platform-integration/linux/install-linux/install-linux-from-{{target}})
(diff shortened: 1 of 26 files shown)


atsansone (approved):
LGTM. Thanks!

atsansone:
@parlough : We can chat about this more, but I hope some of what we discussed today helped.

sfshaza2:
Thanks so much for diving into this, @parlough. I agree that some of these docs had become quite confusing to me. I see that you and @atsansone talked, but let me know if we need further discussion.

sfshaza2 (approved):
lgtm

Issue #10409, opened 2024-04-12
[PAGE ISSUE]: 'Start building Flutter iOS apps on macOS'


loic-sharma:
### Page URL

https://docs.flutter.dev/get-started/install/macos/mobile-ios/

### Page source

https://github.com/flutter/website/tree/main/src/content/get-started/install/macos/mobile-ios.md

### Describe the problem

This section is incomplete: https://docs.flutter.dev/get-started/install/macos/mobile-ios?tab=physical#start-developing-ios-on-apps-with-flutter

> Congratulations! Now that you have installed all prerequisites and the Flutter SDK, you should be able to start developing Flutter apps for iOS on .

### Expected fix

```diff
- Congratulations! Now that you have installed all prerequisites and the Flutter SDK, you should be able to start developing Flutter apps for iOS on .
+ Congratulations! Now that you have installed all prerequisites and the Flutter SDK, you should be able to start developing Flutter apps for iOS on macOS.
```

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

PR #10417, opened 2024-04-14
Update install includes to newer render syntax

parlough:
Issues with variable scope with LiquidJS includes were causing values to be incorrect or missing. The `include` syntax we are using is only supported to support migration from Jekyll anyway. Update to the [newer `render` tag](https://liquidjs.com/tags/render.html) which introduces an isolated scope for each `render` that all data must be explicitly passed in to.

Fixes https://github.com/flutter/website/issues/10409


parlough:
Closing in favor of https://github.com/flutter/website/pull/10475

PR #10421, opened 2024-04-15
corepack-install

jarrodcolburn:
- **update devcontainer to run `corepack install`**
I wasn't running 'corepack install' before when spinning up container. I think I'm supposed to.


sfshaza2 (approved):
This sounds goo to me. Can you confirm, @parlough?

.devcontainer/devcontainer.json
```diff
@@ -13,7 +13,7 @@
 		"PATH": "${containerEnv:PATH}:${containerEnv:FLUTTER_ROOT}/bin"
 	},
 	"onCreateCommand": "git clone --branch=stable https://github.com/flutter/flutter.git ${containerEnv:FLUTTER_ROOT} && ${containerEnv:FLUTTER_ROOT}/bin/flutter doctor",
-	"postCreateCommand": "git submodule update --init --recursive && flutter pub get && pnpm install",
+	"postCreateCommand": "git submodule update --init --recursive && flutter pub get && corepack install && pnpm install",
```
domesticmouse, line 16: Going from https://www.npmjs.com/package/corepack I'm guessing this is meant to be `corepack enable`

sfshaza2: Is that true, @jarrodcolburn? Should it be `corepack enable`, instead of `install`?

jarrodcolburn: Here's my belief...
* `corepack` is already enabled in the container 
* `corepack install` needs to run in the project's root

domesticmouse (approved):
LGTM

PR #10428, opened 2024-04-17
Revert "Add a link to the Flutter for web developers guide"

domesticmouse:
Reverts flutter/website#10422

The content at `/get-started/install` is not really an introduction to Flutter for web developers. It's how to get started using Flutter for the web.

I'd argue we probably need to think about what content we would give a developer coming from the web to Flutter, but "install these tools" is probably not it.

### DIFF
--- a/src/content/index.md
+++ b/src/content/index.md
@@ -32,13 +32,12 @@ Once you've gone through [Get started][],
 including [Write your first Flutter app][],
 here are some next steps.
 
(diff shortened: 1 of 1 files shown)


sfshaza2:
You're right, Brett. Sorry

sfshaza2 (approved):
lgtm. I was too preoccupied with the capitalization of "web".

Issue #10458, opened 2024-04-26
The deploy docs for Android still use the deprecated `Version` suffix for compileSdk / targetSdk / minSdk


navaronbracke:
### Page URL

https://docs.flutter.dev/deployment/android#review-the-gradle-build-configuration

### Describe the problem

The `compileSdkVersion` / `minSdkVersion` / `targetSdkVersion` properties have been deprecated and replaced by 
`compileSdk` / `minSdk` / `targetSdk`.

### Expected fix

The documentation should be updated to reflect the new property names, and perhaps include a reference to the old names for people that still need to migrate to the new names.

The `flutter.compileSdkVersion` / `flutter.minSdkVersion` / `flutter.targetSdkVersion` references should remain unchanged, as this is part of the Flutter tool.

### I would like to fix this problem.

- [X] I will try and fix this problem on docs.flutter.dev.

sfshaza2:
Fixed in https://github.com/flutter/website/pull/10459. Thanks, @navaronbracke!

PR #10459, opened 2024-04-26
Update references to compile / min target sdk property

navaronbracke:
This updates the Android docs to reference to the new property names of the deprecated `minSdkVersion` / `targetSdkVersion` / `compileSdkVersion`

_Issues fixed by this PR (if any):_ https://github.com/flutter/website/issues/10458

## Presubmit checklist

### DIFF
--- a/src/content/deployment/android.md
+++ b/src/content/deployment/android.md
@@ -323,11 +323,11 @@ to verify that the values are correct.
 `applicationId`
 : Specify the final, unique [application ID][].
   
-`minSdkVersion`
+`minSdk`
 : Specify the [minimum API level][] on which you designed the app to run.
   Defaults to `flutter.minSdkVersion`.
 
-`targetSdkVersion`
+`targetSdk`
 : Specify the target API level on which you designed the app to run.
   Defaults to `flutter.targetSdkVersion`.
   
@@ -349,7 +349,7 @@ to verify that the values are correct.
 
 #### Under the `android` block
   
-`compileSdkVersion`
+`compileSdk`
 : Specify the API level Gradle should use to compile your app.
   Defaults to `flutter.compileSdkVersion`.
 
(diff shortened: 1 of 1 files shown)


src/content/deployment/android.md
```diff
@@ -323,11 +323,11 @@ to verify that the values are correct.
 `applicationId`
 : Specify the final, unique [application ID][].
   
-`minSdkVersion`
+`minSdk`
```
navaronbracke, line 326: @bartekpacia I was contemplating if we should tell people that these are replacements for the deprecated properties with similar names. Android Studio at least gives you a warning if you use the old names, so it should be obvious?

src/content/deployment/android.md
```diff
@@ -323,11 +323,11 @@ to verify that the values are correct.
...
 : Specify the [minimum API level][] on which you designed the app to run.
   Defaults to `flutter.minSdkVersion`.
```
navaronbracke, line 328: @bartekpacia Should we also do a similar change in `flutter_tools`, to keep it consistent? We could make a backwards compatible change, although I'm not sure if it is worth to even rename in the tool.

src/content/deployment/android.md
bartekpacia: Yes, I think it's a good idea.

src/content/deployment/android.md
bartekpacia: I don't have a strong opinion on it tbh. Seems very minor.

bartekpacia (approved):
lgtm

src/content/deployment/android.md
navaronbracke: I added a `Note` block, like we do for other things on this page. I didn't build a staged version of the docs yet, so I don't know if that looks alright.

navaronbracke:
This is the new output (thanks for running this in a staging environment!)

image

sfshaza2 (approved):
Thanks, @navaronbracke and @bartekpacia!  lgtm

PR #10475, opened 2024-04-30
Avoid include shadowing issues in getting started docs

parlough:
Fixes https://github.com/flutter/website/issues/10409

Issues with variable scope with LiquidJS includes were causing values in parent scopes to be reassigned and later incorrect. We will eventually move away from the deprecated `include` syntax in favor of `render`, but this fixes the issue for now by avoiding unnecessary variable shadowing.


sfshaza2 (approved):
I leave the breaking build to you, but lgtm

atsansone (approved):
I see the utility, but I'm glad a better solution will be forthcoming.

PR #10481, opened 2024-05-01
Fix formatting error

atsansone:
Corrects small formatting error


sfshaza2 (approved):
I don't see a bug #, but ok. lgtm

PR #10507, opened 2024-05-04
Break up examples/development to match newer IA

parlough:


domesticmouse:
CI is cranky

```
Because tools depends on example_utils from path which doesn't exist (could not find package example_utils at "../../example_utils"), version solving failed.
```

domesticmouse (approved):
LGTM once CI is fixed

PR #10508, opened 2024-05-04
Continue migration from `include` to `render`

parlough:
The `include` syntax we're using is supported as a fallback for migration, and [`render`](https://liquidjs.com/tags/render.html) is the appropriate replacement tag. Its included content doesn't inherit its parents scope, helping prevent mistakes and potentially improving incremental rebuilds.

Also remove some unused includes.


sfshaza2 (approved):
Wow... "render". lgtm

PR #10524, opened 2024-05-07
Prevent a few icons from being mistakenly read by screen readers

parlough:
- In the sidenav, set an empty alt content for screen readers for the arrows.
- In the inline TOC, hide the icons themselves from screen readers and add explanatory titles to the surrounding spans, providing more details to all users.


atsansone:
@parlough : Works as expected. I'll notify the Core a11y team.

PR #10585, opened 2024-05-13
Update delete-data.md

### LINKED ISSUE #10577: [PAGE ISSUE]: 'Delete data on the internet'
### Page URL

https://docs.flutter.dev/cookbook/networking/delete-data/

### Page source

https://github.com/flutter/website/tree/main/src/content/cookbook/networking/delete-data.md

### Describe the problem

After clicking the "Delete Data" button, the following error occurs:

image

This is because the `Album.fromJson` factory method throws an `FormatException` exception while parsing an empty json map.

### Expected fix

_No response_

### Additional context

_No response_

johnpryan:
Fix "Delete data on the internet" cookbook sample.

The Album.fromJson method doesn't handle an empty map, which is returned from the jsonplaceholder API when the DELETE method is called. 

Fixes #10577

### DIFF
--- a/examples/cookbook/networking/delete_data/lib/main.dart
+++ b/examples/cookbook/networking/delete_data/lib/main.dart
@@ -31,11 +31,11 @@ Future<Album> deleteAlbum(String id) async {
 
   if (response.statusCode == 200) {
     // If the server did return a 200 OK response,
-    // then parse the JSON. After deleting,
+    // then return an empty Album. After deleting,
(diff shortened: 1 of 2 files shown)


parlough (approved):
Thanks! But you'll need to update the source snippets as well.

domesticmouse (approved):
LGTM once @parlough's concerns are addressed

sfshaza2 (approved):
lgtm, once the build succeeds.

PR #10590, opened 2024-05-14
Update the banner for I/O 2024 livestream

sfshaza2:
as is says


domesticmouse (approved):

parlough (approved):
Thanks!

I also updated the watch text to be the link.

PR #10593, opened 2024-05-14
[wasm] clarify that JS is still complied, even with the --wasm flag

kevmoo:


src/content/platform-integration/web/wasm.md
```diff
@@ -93,7 +93,14 @@ flutter build web --wasm
 ```
 
 The command produces output into the `build/web` directory relative to the
-package root.
+package root, just like `flutter build web`.
+
+:::note
+Even with the `--wasm` flag, `flutter build web` will still compile the
+application with JavaScript. If WasmGC support is not detected at runtime,
```
parlough, line 100:
```suggestion
application to JavaScript. If WasmGC support is not detected at runtime,
```

parlough (approved):
One adjustment :)

domesticmouse (approved):

PR #10600, opened 2024-05-15
fix: replace an incorrect link

percival888:
replace an incorrect link


parlough (approved):
Nice catch, thanks for fixing this!

Issue #10613, opened 2024-05-16
Fix CLI examples to not select prompts


sethladd:
### Page URL

https://docs.flutter.dev/get-started/install/macos/mobile-ios/

### Page source

https://github.com/flutter/website/tree/main/src/content/get-started/install/macos/mobile-ios.md

### Describe the problem

I copied the entire line from one of the instruction codeblocks, and pasted into my terminal. When I executed what I pasted in, I got "command not found".

Specifically `$ sudo softwareupdate --install-rosetta --agree-to-license` is what was copied and then pasted.

Even more specifically: When you triple click on the text, or you just blindly select all the text, you get the prompt `$`:

![Screenshot 2024-05-16 at 9 29 38 AM](https://github.com/flutter/website/assets/5479/551d6554-ffd6-478e-b584-84353d2b27a9)

And when you paste the text into your terminal, you also paste in the `$` which of course is itself not a command. Hence the error when you try to run it.

This is a minor inconvenience for users moving quickly (as I was) and just copying and pasting what the webpage tells me to :)

### Expected fix

What would be awesome is if someone triple clicks in the text, or blinding selects all the text, and they do NOT copy the prompt `$`.

Like this:

![Screenshot 2024-05-16 at 9 29 46 AM](https://github.com/flutter/website/assets/5479/64f97b27-e1ea-4c7d-8c05-b42f014d8844)

### I would like to fix this problem.

- [X] I will try and fix this problem on docs.flutter.dev.

sethladd:
Here's how we can fix this easily, across the site. Add a tiny bit of CSS like this:

```
.code-block-body:before {
    color: #bdc1c6;
    content: "$";
    padding-right: 8px;
}
```

or something very close to that. I'll have to experiment, but the key thing is `:before`

sethladd:
Before I go and remove all the literal `$` from the codeblocks, and apply CSS like the above, I'm curious what @parlough thinks.

atsansone:
@sethladd : Interesting. The copy button in the upper right of each code block will allow you to copy without the $. Maybe we need to make that more obvious?

sethladd:
Dunno, I'm not trained in my muscle memory to even look for that copy button. I see text I'm supposed to copy, and I use my mouse to copy it :)

atsansone:
Part of the issue with this specific fix is that it doesn't account for Windows command prompt (`C:>`) or Powershell (`PS`) prompts. There might be additional changes for which we have yet to account.

[...]

PR #10637, opened 2024-05-21
Replace terminal marker with CSS-based marker

parlough:
Fixes https://github.com/flutter/website/issues/10613 for the common case of a `$` terminal marker for the `console` language. It looks at each line's contents to determine if the line starts with a `$`, then removes that and adds a CSS class to add it instead.

### DIFF
--- a/src/_11ty/plugins/highlight.js
+++ b/src/_11ty/plugins/highlight.js
@@ -122,7 +122,7 @@ function _highlight(
     ? _parseNumbersAndRanges(highlightLines)
     : null;
 
-  const noHighlight = 'noHighlight' in attributes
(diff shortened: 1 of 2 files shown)


atsansone:
@parlough : Could you not also create the PowerShell case? Their prompt is `PS C:\>`. (Yes, I know that it changes to the current path, but that might be more than we could manage?) The `cmd` prompt is just `C:\>`, but I believe we're encouraging use of PowerShell in all cases.

sfshaza2 (approved):
I don't want to use a PowerShell prompt in our examples. lgtm

domesticmouse (approved):

PR #10649, opened 2024-05-22
Fix Xcode install

### LINKED ISSUE #10615: Normalize Xcode and CocoaPods install on 'Start building Flutter iOS apps on macOS' page
### Page URL

### Page source

https://github.com/flutter/website/tree/main/src/content/get-started/install/macos/mobile-ios.md

### Describe the problem

Following the install process, it appears we don't explicitly tell someone to install Xcode inline at the right point.

Actual:

Under "Development Tools" I saw a summary of what I assumed is what we will do by following these steps. Including installing Xcode. I glanced over this section because it didn't sound like I needed to do this _right now_.

Especially because the "Development Tools" section talks about installing CocoaPods, but later on in the article, we explicitly call out a step to install CocoaPods.

So, when you get to `$ sudo sh -c 'xcode-select -s /Applications/Xcode.app/Contents/Developer && xcodebuild -runFirstLaunch'` , that will fail because you didn't install XCode (again, because it was kind of hidden/implied above).

### Expected fix

Expected:

Under "Configure iOS development", the first sub-header should be "Install Xcode", before "Configure XCode".

This matches the later explicit "Install CocoaPods" instructions later.

### Additional context

_No response_

atsansone:
Fixes #10615
Fixes https://github.com/flutter/website/issues/10650

### DIFF
--- a/src/_includes/docs/install/flutter-doctor.md
+++ b/src/_includes/docs/install/flutter-doctor.md
@@ -8,7 +8,7 @@
 {% case include.devos %}
 {% when 'Windows' -%}
    {% assign terminal='PowerShell' %}
-   {% assign prompt='C:\>' %}
+   {% assign prompt='PS C:\>' %}
 {% when "macOS" -%}
    {% assign terminal='your Terminal' %}
    {% assign prompt='$' %}
--- a/src/_includes/docs/install/compiler/xcode.md
+++ b/src/_includes/docs/install/compiler/xcode.md
@@ -3,20 +3,29 @@
 
 {% assign prompt1='$' %}
 
-### Configure Xcode
+### Install and configure Xcode
 
 {% if include.attempt=="first" %}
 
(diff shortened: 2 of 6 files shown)


src/_includes/docs/install/flutter-doctor.md
```diff
@@ -8,7 +8,7 @@
 {% case include.devos %}
 {% when 'Windows' -%}
    {% assign terminal='PowerShell' %}
-   {% assign prompt='C:\>' %}
+   {% assign prompt='PS C:\>' %}
```
domesticmouse, line 11: Why powershell?

domesticmouse (approved):

src/_includes/docs/install/flutter-doctor.md
atsansone: The instructions themselves always said you needed to install / use PowerShell, so I'm just making the prompting consistent.

PR #10688, opened 2024-06-01
Update release note links

erkro1:


sfshaza2 (commented):
Thanks so much for this PR, @erkro1! But @Piinks beat you to it!

PR #10706, opened 2024-06-04
Fixes PATH variance, filename replacement script

### LINKED ISSUE #10662: Resolve difference between install and PATH instructions for Linux
### Page URL

https://docs.flutter.dev/get-started/install/linux/desktop

### Page source

_No response_

### Describe the problem

["Download then install flutter - Step 3"](https://docs.flutter.dev/get-started/install/linux/desktop?tab=download#download-then-install-flutter)
> Extract the file into the directory you want to store the Flutter SDK.
> `tar -xf ~/Downloads/flutter_linux_3.22.1-stable.tar.xz -C /usr/bin/`
> When finished, the Flutter SDK should be in the /usr/bin/flutter directory.

The next step:
["Add Flutter to your PATH"](https://docs.flutter.dev/get-started/install/linux/desktop?tab=download#add-flutter-to-your-path)
> To run Flutter commands in a shell, add Flutter to the PATH environment variable.
> ...
> To add Flutter to your PATH, expand the entry for your default shell, then choose the command.
> `echo 'export PATH="$HOME/development/flutter/bin:$PATH"' >> ~/.bash_profile`

### Expected fix

Update the instructions to use the same installation location.

### Additional context

_No response_

- [X] I will try and fix this problem on docs.flutter.dev.

atsansone:
Fixes #10662

### DIFF
--- a/src/_data/shells.yml
+++ b/src/_data/shells.yml
@@ -1,14 +1,14 @@
 - name: bash
-  set-path: echo 'export PATH="$HOME/development/flutter/bin:$PATH"' >> ~/.bash_profile
+  set-path: echo 'export PATH="/usr/bin/flutter/bin:$PATH"' >> ~/.bash_profile
 - name: zsh
-  set-path: echo 'export PATH="$HOME/development/flutter/bin:$PATH"' >> ~/.zshenv
+  set-path: echo 'export PATH="/usr/bin/flutter/bin:$PATH"' >> ~/.zshenv
 - name: fish
-  set-path: fish_add_path -g -p $HOME/development/flutter/bin
+  set-path: fish_add_path -g -p /usr/bin/flutter/bin
 - name: csh
-  set-path: echo 'setenv PATH "$HOME/development/flutter/bin:$PATH"' >> ~/.cshrc
+  set-path: echo 'setenv PATH "/usr/bin/flutter/bin:$PATH"' >> ~/.cshrc
 - name: tcsh
-  set-path: echo 'setenv PATH "$HOME/development/flutter/bin:$PATH"' >> ~/.tcshrc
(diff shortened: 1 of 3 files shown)


domesticmouse (approved):

src/_data/shells.yml
```diff
@@ -1,14 +1,14 @@
 - name: bash
```
sfshaza2, line 1: Might it be that Flutter is NOT in /usr/bin?

src/_includes/docs/install/flutter/download.md
```diff
@@ -8,17 +8,17 @@
    {% assign path='C:\\user\\{username}\\dev' %}
    {% assign flutter-path='C:\\user\\{username}\\dev\\flutter' %}
    {% assign terminal='PowerShell' %}
-   {% assign prompt='C:>' %}
+   {% assign prompt='PS C:\\>' %}
```
sfshaza2, line 11: I thought we weren't using :PS C"?

sfshaza2 (commented):
Not a full tech review, just questions.

src/_includes/docs/install/flutter/download.md
atsansone: We encourage PowerShell everywhere, and the PowerShell prompt is `PS C:\\>`. I noticed it on my PC the other day and wanted to be consistent / correct.

src/_data/shells.yml
atsansone: True, but this is consistent with the previous advice given on the page, which is to consider adding `flutter` to `/usr/bin`. If they follow inline, this should be OK.

sfshaza2 (approved):
lgtm

Issue #10713, opened 2024-06-06
Add new step to 'Create flavors of a Flutter app' page [iOS]


qwadrox:
### Page URL

https://docs.flutter.dev/deployment/flavors/

### Page source

https://github.com/flutter/website/tree/main/src/content/deployment/flavors.md

### Describe the problem

**Description:**
Following the documentation for renaming schemes in the Podfile, an error occurs when building the iOS app. The issue arises after running `flutter clean/pub get` and `pod install`. The build fails with an error indicating it is unable to load contents of a file list. Manually modifying the `ios/Flutter/Debug.xcconfig` file resolves the issue, suggesting a potential missing step in the documentation.

**Steps to Reproduce:**
1. Rename the schemes in the `Podfile` as follows:
    ```ruby
    project 'Runner', {
      'Debug-free' => :debug,
      'Profile-free' => :release,
      'Release-free' => :release,
      'Debug-prod' => :debug,
      'Profile-prod' => :release,
      'Release-prod' => :release,
    }
    ```
2. Run `flutter clean` `flutter pub get`..
3. Run `pod install`.
4. Attempt to build the iOS app.

**Expected Result:**
The iOS app should build successfully without errors.

**Actual Result:**
The build fails with the following error:
```bash
Failed to build iOS app
Error (Xcode): Unable to load contents of file list: '/Target Support Files/Pods-Runner/Pods-Runner-frameworks-Release-prod-input-files.xcfilelist'
```

**Workaround:**
Modifying the `ios/Flutter/Debug.xcconfig` file manually resolves the issue:
```diff
- #include? "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug.xcconfig"
+ #include? "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug-free.xcconfig"
+ #include? "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug-prod.xcconfig"
#include "Generated.xcconfig"
```

**Environment:**
- Flutter version: 3.22.1
- Xcode version: 15.4
- Podfile configuration: as in the sample project

**Additional Information:**
- Reproducible sample project: [new_flavor_test_app](https://github.com/qwadrox/new_flavor_test_app)
- The error appears before the commit where the manual modification was applied.

### Expected fix

It seems that there might be a missing step in the documentation regarding the modification of the `xcconfig` files after renaming schemes in the `Podfile`. This needs to be confirmed and updated in the documentation to prevent similar issues in the future.

[...]

PR #10716, opened 2024-06-06
WIP Updating the iOS/macOS flavors doc

sfshaza2:
Fixes https://github.com/flutter/website/issues/10713,
for setting up flavors in iOS and macOS.

cc @qwadrox @jmagman

### DIFF
--- a/src/content/deployment/flavors.md
+++ b/src/content/deployment/flavors.md
@@ -116,7 +116,7 @@ build configurations already created.
 </li>
 </ol>
 
-### Manually update podfile
+### Manually update Podfile
 
 After running `flutter clean`, `flutter pub`, and `pod install`,
 you receive a build error indicating that it is unable to load
--- a/src/content/deployment/flavors.md
+++ b/src/content/deployment/flavors.md
@@ -187,8 +187,8 @@ value to `$(PRODUCT_NAME)`.
 Now you have set up your flavor by making a `free` scheme
 in Xcode and setting the build configurations for that scheme.
 
-For more information,
-skip to the [Launching your app flavors][]
+For more information, skip to the
+[Launch your app flavors](#launch-your-app-flavors)
 section at the end of this document.
 
 ### Update plugin configurations
@@ -348,7 +348,6 @@ For packages that support creating flavors, check out the following:
 * [`flutter_flavor`][]
 * [`flutter_flavorizr`][]
 
-[Launching your app flavors]: /deployment/flavors/#launching-your-app-flavors
 [Flutter repo]: {{site.repo.flutter}}/blob/master/dev/integration_tests/flavors/lib/main.dart
 [iOS]: {{site.repo.flutter}}/tree/master/dev/integration_tests/flavors/ios
 [macOS]: {{site.repo.flutter}}/tree/master/dev/integration_tests/flavors/macos
--- a/src/content/deployment/flavors.md
+++ b/src/content/deployment/flavors.md
@@ -8,20 +8,25 @@ description: >
 
 ## What are flavors
 
-Have you ever wondered how to set up different environments in your Flutter app?
-Flavors (known as _build configurations_ in iOS and macOS), allow you (the developer) to
+Have you ever wondered how to set up different environments
+in your Flutter app? Flavors (known as _build configurations_
+in iOS and macOS), allow you (the developer) to
 create separate environments for your app using the same code base.
-For example, you might have one flavor for your full-fledged production app,
-another as a limited "free" app, another for testing experimental features, and so on.
+For example, you might have one flavor for your
+full-fledged production app,
+another as a limited "free" app,
+another for testing experimental features, and so on.
 
 Say you want to make both free and paid versions of your Flutter app.
 You can use flavors to set up both app versions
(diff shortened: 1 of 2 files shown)


src/content/deployment/flavors.md
```diff
@@ -114,6 +116,34 @@ build configurations already created.
 </li>
 </ol>
 
+### Manually update podfile
```
jmagman, line 119: Casing nit: Podfile
```suggestion
### Manually update Podfile
```

src/content/deployment/flavors.md
```diff
@@ -114,6 +116,34 @@ build configurations already created.
...
+
+After running `flutter clean`, `flutter pub`, and `pod install`,
+you receive a build error indicating that it is unable to load
+the contents of a file list,
+you will need to manually update your podfile.
```
jmagman, line 124:
```suggestion
you will need to manually update your Podfile.
```

src/content/deployment/flavors.md
```diff
@@ -114,6 +116,34 @@ build configurations already created.
...
+  'Release-free' => :release,
+  'Debug-prod' => :debug,
+  'Profile-prod' => :release,
+  'Release-prod' => :release,
+}
+```
+
+2. Run `flutter clean` and `flutter pub get`.
+
+3. Run `pod install`.
```
jmagman, line 143: Flutter developers should never need to run `pod install` (except for add-to-app).  If we have instructions to explicitly run it anywhere else in the docs, we should remove it.
You can delete steps 2 and 3 and go right to "Build the app."

jmagman (changes requested):
https://github.com/flutter/website/issues/10713 is reporting an issue in adding a flavor (prod) that isn't described in these docs.  These instructions are only for adding the "free" flavor, they don't say anything about prod.  

And we already have the correct instructions for updating the Podfile here:
https://docs.flutter.dev/deployment/flavors#plugin-configurations

jmagman:
Reading it again, the fix for https://github.com/flutter/website/issues/10713 would be to just add the instructions for updating the xcconfig, as they say in their workaround section:
```diff
- #include? "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug.xcconfig"
+ #include? "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug-free.xcconfig"
```
Example from our integration docs:
https://github.com/flutter/flutter/blob/32da263b1d5590a98e03d0ea1374f85e0c18de3f/dev/integration_tests/flavors/ios/Flutter/Debug.xcconfig#L3

qwadrox:
> Reading it again, the fix for #10713 would be to just add the instructions for updating the xcconfig, as they say in their workaround section:
> 
> ```diff
> - #include? "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug.xcconfig"
> + #include? "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug-free.xcconfig"
> ```
> 
> Example from our integration docs: https://github.com/flutter/flutter/blob/32da263b1d5590a98e03d0ea1374f85e0c18de3f/dev/integration_tests/flavors/ios/Flutter/Debug.xcconfig#L3

What is the purpose of xcconfig files?
If I have 3 scheme like prod, uat, dev, should I duplicate the 'Debug' => :debug block in podfile 3 times? like  'Debug-prod' => :debug, ... etc. 
First I thought I have to have 3 line per scheme as well in the xcconfig files: #include? like 
```
"Pods/Target Support Files/Pods-Runner/Pods-Runner.debug-prod.xcconfig"
"Pods/Target Support Files/Pods-Runner/Pods-Runner.debug-dev.xcconfig"
"Pods/Target Support Files/Pods-Runner/Pods-Runner.debug-uat.xcconfig"
```
But it seems one is enough

jmagman:
> What is the purpose of xcconfig files?

It's an Xcode concept to add build configurations to your project:
https://developer.apple.com/documentation/xcode/adding-a-build-configuration-file-to-your-project

You likely only needed one because they are all identical.

qwadrox:
> > What is the purpose of xcconfig files?
> 
> It's an Xcode concept to add build configurations to your project:
> https://developer.apple.com/documentation/xcode/adding-a-build-configuration-file-to-your-project
> 
> You likely only needed one because they are all identical.

I understand now. Thanks. 
The error occurred because I renamed both schemes. I renamed the original Runner scheme to 'prod', which caused the imported Pod generated config to be incorrectly referenced in Debug.xconfig.

The original documentation is correct; it only advises creating a new scheme, leaving the Runner scheme unchanged. This maintains the correct reference paths in Debug.xconfig.

It would be only beneficial to add a cautionary note to the docs: 

"If you rename the original Runner scheme, ensure you update Debug.xcconfig and Release.xcconfig in the ios/Flutter directory to reference the newly generated Pods/Target Support Files/Pods-Runner xcconfig file."

atsansone:
@sfshaza2 : I have a larger refactor of this whole page in flight. I'll integrate your changes to it and then we can close this PR. Does that work for you?

sfshaza2:
@atsansone, I'd have preferred if you had talked about this with me ahead of time.

PR #10717, opened 2024-06-06
Adjustments after default renderer text merged too early

parlough:
Follow up to https://github.com/flutter/website/pull/10714#issuecomment-2153102352


sfshaza2 (approved):
okaaaay. lgtm

PR #10737, opened 2024-06-11
Fix getOS() for all operating systems

PondWader:
In the current website getOS() returns "macos" if the user agent is on macOS or iOS and otherwise windows, since -1 is truthy, `if (userAgent.indexOf('Win'))` is always true and so "windows" is always returned. This PR fixes that behaviour as well as fixing an android check that would only return linux if the device is android, rather than the opposite.

## Presubmit checklist

### DIFF
--- a/src/content/assets/js/main.js
+++ b/src/content/assets/js/main.js
@@ -52,13 +52,13 @@ function getOS() {
     return 'macos';
   }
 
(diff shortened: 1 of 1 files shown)


atsansone (approved):

parlough (approved):
Nice job catching this, and thanks for the fix!!

PR #10738, opened 2024-06-11
Update dart-devtools-survey-metadata.json for Q2-2024 DevTools Survey

kenzieschmoll:
@jayoung-lee lmk what you think about the end date, and if this is the final survey link.


jayoung-lee (approved):
LGTM

PR #10763, opened 2024-06-19
Prevent text wrap on cookie notice button

parlough:
Also expand saving of click from 30 to 90 days.

Before:

Ok, got it button but text wrapped


sfshaza2 (approved):
lgtm!

PR #10764, opened 2024-06-19
Add scroll-margin-top to headers

parlough:
This allows headers to be visible in the viewport when navigating to them, rather than below the navbar.


sfshaza2 (approved):
lgtm

Issue #10363, opened 2024-04-04
Determine pattern of tooling go links and create important ones


parlough:
If you have destinations you currently _(or would like to)_ link to from tooling, please comment here with the destination/topic/reason or similar. Thanks!

DanTup:
These are the URLs I found in the Dart-Code currently (some are Dart, but since I had them I thought I'd post them for completeness):

- https://dart.dev/get-dart
- https://dart.dev/tools/dart-devtools
- https://dart.dev/tools/pub/glossary#system-cache
- https://docs.flutter.dev/f/flutter-survey-metadata.json
- https://docs.flutter.dev/get-started/install/linux#update-your-path
- https://docs.flutter.dev/get-started/install/macos#update-your-path
- https://docs.flutter.dev/get-started/install/windows#update-your-path
- https://flutter.dev/docs/development/tools/web-renderers
- https://flutter.dev/setup/
- https://api.flutter.dev/flutter/${sample.libraryName}/${sample.className}-${suffix}.html${fragment}

The last one is used to power the "Open interactive sample" CodeLens feature. We get the library/classname from the analyzers Outline data (and use `class`, `mixin` or `extension-type` as a suffix based on the element kind, and build a matching fragment like `"widgets.RestorationMixin.1"` to scroll down to the embedded sample.

parlough:
In the various linked PRs, I've added tooling `/to/` links for links already used in the tool and framework, and with https://github.com/flutter/flutter/commit/b6cd31e6b5a61125ef0e78d866af673f404a84ea, https://github.com/flutter/flutter/blob/main/docs/contributing/use-reliable-links.md now has guidelines on adding and using them.

PR #10765, opened 2024-06-19
Adding tooling links using new `/to/` format

parlough:
The links added here will be used from various locations that need consistently functional links that are sometimes manually copy and pasted. The current set includes links the `flutter` tool, DevTools, the VS Code plugin(s), the Flutter framework, as well as some surrounding documentation.

It's particularly important to have these links for areas relying on fragments, as those redirects can't happen on the server side, and developers may continue to run SDKs or tools with old links long after their destination needs to change.

After this PR, `/go/<design-doc>` should only be used for design docs going forward. Only one item long, top-level redirects such as `docs.flutter.dev/<top-level>` should also be avoided from now on.

Once this lands, I will update the tools where necessary and land documentation in the `flutter/flutter` repository that ensures this format of link is considered.

Contributes to https://github.com/flutter/website/issues/10363


sfshaza2 (approved):
lgtm

PR #10782, opened 2024-06-21
Delete obsolete survey metadata file

kenzieschmoll:
This survey metadata file is actually tracked at https://github.com/flutter/uxr/blob/master/surveys/devtools-survey-metadata.json, so we do not need this one anymore.


parlough:
@kenzieschmoll Are all relevant versions of DevTools updated to use the new destination, as well as other metnions? Should we add a redirect?

kenzieschmoll:
Good catch. Added a redirect.

Looks like there was some prior work here I had forgotten about: https://github.com/flutter/website/pull/9935.

PR #10791, opened 2024-06-24
Updated Gradle info

### LINKED ISSUE #9234: Add documentation for common changes to generated gradle files
### What information is missing?

Example: One of the [most liked flutter-related questions on stack overflow](https://stackoverflow.com/questions/52060516/flutter-how-to-change-android-minsdkversion-in-flutter-project) asks how to change the minSdkVersion in a Flutter app for android. Note that while the top answer here is correct, it doesn't cover how plugin authors should change the minSdkVersion (to be fair the question doesn't ask that, but the post will be the top result on google for a plugin author who searches about changing the minSdkVersion in their plugin). Also, some of the other (highly liked) answers have different suggestions that aren't ideal.

Another one of the most liked: [changing your application id](https://stackoverflow.com/questions/51534616/how-to-change-package-name-in-flutter). This even has a TODO: in the [template for the module level build.gradle](https://github.com/flutter/flutter/blob/b0f90db6e1e66441039da8ebb474a137d4ab0fbe/packages/flutter_tools/templates/app_shared/android-java.tmpl/app/build.gradle.tmpl#L36C58-L36C120).

### How would you like us to fix this problem?

We should add a page covering these common cases when a developer will need to interact with their gradle files, clarifying the steps they need to take and providing examples (maybe also covering changes to the android manifest in the same page).

atsansone:
Fixes #9234

### DIFF
--- a/src/content/deployment/android.md
+++ b/src/content/deployment/android.md
@@ -78,14 +78,14 @@ dependencies {
 
 To find out the latest version, visit [Google Maven][].
 
-2. Set the light theme in `<my-app>/android/app/src/main/res/values/styles.xml`:
+1. Set the light theme in `<my-app>/android/app/src/main/res/values/styles.xml`:
 
 ```diff
 -<style name="NormalTheme" parent="@android:style/Theme.Light.NoTitleBar">
 +<style name="NormalTheme" parent="Theme.MaterialComponents.Light.NoActionBar">
 ```
 
-3. Set the dark theme in `<my-app>/android/app/src/main/res/values-night/styles.xml`
+1. Set the dark theme in `<my-app>/android/app/src/main/res/values-night/styles.xml`
 
 ```diff
 -<style name="NormalTheme" parent="@android:style/Theme.Black.NoTitleBar">
@@ -150,7 +150,7 @@ If not, create one using one of the following methods:
      notation for the names. For example, on Mac/Linux
      use `Program\ Files`, and on Windows use
      `"Program Files"`.
-   
+
    * The `-storetype JKS` tag is only required for Java 9
      or newer. As of the Java 9 release,
      the keystore type defaults to PKS12.
@@ -184,9 +184,12 @@ don't check it into public source control.
 When building your app in release mode, configure gradle to use your upload key.
 To configure gradle, edit the `<project>/android/app/build.gradle` file.
 
-1. Add the keystore information from your properties file before the `android` block:
+1. Define and load the keystore properties file before the `android`
+   property block.
 
-   ```diff
+1. Set the `keystoreProperties` object to load the `key.properties` file.
+
+   ```diff title="[project]/android/app/build.gradle"
    +   def keystoreProperties = new Properties()
    +   def keystorePropertiesFile = rootProject.file('key.properties')
    +   if (keystorePropertiesFile.exists()) {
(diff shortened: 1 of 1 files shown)


domesticmouse:
Can you please fix the conflicts?

src/content/deployment/android.md
```diff
@@ -92,8 +92,7 @@ To find out the latest version, visit [Google Maven][].
 +<style name="NormalTheme" parent="Theme.MaterialComponents.DayNight.NoActionBar">
 ```
 
-<a id="signing-the-app"></a>
-## Sign the app
+## Sign the app {:#signing-the-app}
```
parlough, line 95: Please keep the separate anchor here so they both continue to work. There's old links to both headers from outside the site.

atsansone: In the site checks out as OK.

parlough: Yes in the site is fine, but these extra anchors are here for outside the side, such as from old framework and tool versions, other sites, etc.

src/content/deployment/android.md
```diff
@@ -296,66 +302,81 @@ check out the official [Android documentation][multidex-docs].
...
+| `application`        | Edit the `android:label` in the [`application`][applicationtag] tag to reflect the final name of the app. |
+| `uses-permission`    | Add the `android.permission.INTERNET` [permission][permissiontag] value to the `android:name` attribute if your app needs Internet access. The standard template doesn't include this tag but allows Internet access during development to enable communication between Flutter tools and a running app. |
+
+{:.table .table-striped}
+
+## Review or change the Gradle build configuration {:#review-the-gradle-build-configuration}
+
+To verify the Android build configuration,
+review the `android` block in the default
+[Gradle build file][gradlebuild].
```
reidbaker, line 331: There can be multiple build.gradle files and it is hard to know which one is default. 
This code is normally in the android/app/build.gradle file

src/content/deployment/android.md
```diff
@@ -296,66 +302,81 @@ check out the official [Android documentation][multidex-docs].
...
 
-`applicationId`
-: Specify the final, unique [application ID][].
-  
-`minSdk`
-: Specify the [minimum API level][] on which you designed the app to run.
-  Defaults to `flutter.minSdkVersion`.
+| Property             | Purpose                                                                                                                                                                                                            | Default Value               |
+|----------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------|
+| `compileSdk`         | The Android API level with which your app should be compiled.                                                                                                                                                      | `flutter.compileSdkVersion` |
```
reidbaker, line 368:
```suggestion
| `compileSdk`         | The Android API level with which your app should be compiled.  This should normally be the highest version available.         
```

src/content/deployment/android.md
```diff
@@ -296,66 +302,81 @@ check out the official [Android documentation][multidex-docs].
...
+| `defaultConfig`      |  |  |
+| `.applicationId`     | The final, unique [application ID][] that identifies your app.                                                                                                                                                     |                             |
+| `.minSdk`            | The [minimum Android API level][] for which you designed the app to run.                                                                                                                                           | `flutter.minSdkVersion`     |
+| `.targetSdk`         | The target Android API level for which you designed the app to run.                                                                                                                                                | `flutter.targetSdkVersion`  |
```
reidbaker, line 372: https://stackoverflow.com/questions/26694108/what-is-the-difference-between-compilesdkversion-and-targetsdkversion
This stackoverflow question does a good job explaining the difference

reidbaker (approved):

Issue #9424, opened 2023-09-19
Explain how to create macOS PlatformViews


cbracken:
Once we've completed support for PlatformViews on macOS we'll need a page similar to https://docs.flutter.dev/platform-integration/ios/platform-views covering the creation of platform views on macOS.

For the most part, the framework code should be near-identical other than substituting `AppKitView` where `UiKitView` is used for iOS and ensuring that any platform-specific checks reference [Platform.macOS](https://api.flutter.dev/flutter/package-platform_platform/Platform/macOS-constant.html) instead of [Platform.iOS](https://api.flutter.dev/flutter/package-platform_platform/Platform/iOS-constant.html).

On the Swift side, there are some small differences in where and how to register platformViews, and the fact that the macOS view being created is an `NSView` rather than a `UIKitView`.

For reference, I have a macOS PlatformView example that wraps an Apple Maps MKMapView here:
https://github.com/cbracken/platformview_test.

Part of: https://github.com/flutter/flutter/issues/41722

cbracken:
/cc @domesticmouse 

Code-wise, I think everything we need for creating/registering macOS PlatformViews, both on the framework and engine side, is in place. There are still a bunch of bugfixes and some implementation work to do for touch/gesture/a11y support, but API-wise, I think we're probably in good enough shape to have a go at the docs.

Is this something you'd prefer I take a look at, or something you'd like to do?

atsansone:
@cbracken : I'd be up for doing this. I'll send you an invite for next week, if that's not too late?

domesticmouse:
@atsansone I'm happy to help with this =)

domesticmouse:
Silly question, what is the state of this issue?

domesticmouse:
Unassigning this as we are not currently promoting this capability. Re-assign when this changes.

sfshaza2:
Just to be clear, the new page exists here: https://docs.flutter.dev/platform-integration/macos/platform-views

PR #10796, opened 2024-06-25
[macOS] Add platform view documentation

cbracken:
This adds documentation on authoring and using platform views for Flutter apps targeting macOS.

The initial version of this documentation is Swift-only, but a follow-up will add Objective-C example code.

Fixes: #9424

## Presubmit checklist

- [X] This PR is marked as draft with an explanation if not meant to land until a future stable release.

### DIFF
--- a/examples/platform_integration/platform_views/lib/native_view_example_4.dart
+++ b/examples/platform_integration/platform_views/lib/native_view_example_4.dart
@@ -1,0 +1,53 @@
+// ignore_for_file: unused_local_variable
+
+// #docregion import
+import 'package:flutter/foundation.dart';
+import 'package:flutter/services.dart';
+// #enddocregion import
+import 'package:flutter/widgets.dart';
+
+class MacOSCompositionWidget extends StatelessWidget {
+  const MacOSCompositionWidget({super.key});
+
+  @override
+  // #docregion macos-composition
+  Widget build(BuildContext context) {
+    // This is used in the platform side to register the view.
+    const String viewType = '<platform-view-type>';
+    // Pass parameters to the platform side.
+    final Map<String, dynamic> creationParams = <String, dynamic>{};
+
+    return AppKitView(
+      viewType: viewType,
+      layoutDirection: TextDirection.ltr,
+      creationParams: creationParams,
+      creationParamsCodec: const StandardMessageCodec(),
+    );
+  }
+  // #enddocregion macos-composition
+}
+
+class TogetherWidget extends StatelessWidget {
+  const TogetherWidget({super.key});
+
+  @override
+  // #docregion together-widget
+  Widget build(BuildContext context) {
+    // This is used in the platform side to register the view.
+    const String viewType = '<platform-view-type>';
+    // Pass parameters to the platform side.
+    final Map<String, dynamic> creationParams = <String, dynamic>{};
+
+    switch (defaultTargetPlatform) {
+      case TargetPlatform.android:
+      // return widget on Android.
+      case TargetPlatform.iOS:
+      // return widget on iOS.
+      case TargetPlatform.macOS:
+      // return widget on macOS.
+      default:
+        throw UnsupportedError('Unsupported platform view');
+    }
+  }
+  // #enddocregion together-widget
+}
--- a/src/content/platform-integration/macos/platform-views.md
+++ b/src/content/platform-integration/macos/platform-views.md
@@ -194,7 +194,7 @@ When implementing the `build()` method in Dart,
 you can use [`defaultTargetPlatform`][]
 to detect the platform, and decide which widget to use:
 
-<?code-excerpt "lib/native_view_example_3.dart (together-widget)"?>
+<?code-excerpt "lib/native_view_example_4.dart (together-widget)"?>
 ```dart
 Widget build(BuildContext context) {
   // This is used in the platform side to register the view.
--- a/src/content/platform-integration/macos/platform-views.md
+++ b/src/content/platform-integration/macos/platform-views.md
@@ -1,0 +1,240 @@
+---
+title: Hosting native macOS views in your Flutter app with Platform Views
+short-title: macOS platform-views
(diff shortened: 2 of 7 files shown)


examples/platform_integration/platform_views/lib/native_view_example_4.dart
```diff
@@ -0,0 +1,53 @@
+// ignore_for_file: unused_local_variable
+
+// #docregion import
+import 'package:flutter/foundation.dart';
```
domesticmouse, line 4: Is this part of a project? We use flutter projects to confirm the code is up to date as flutter changes.

atsansone:
Does this fix https://github.com/flutter/website/issues/9424?

examples/platform_integration/platform_views/lib/native_view_example_4.dart
cbracken: I couldn't find any related projects for this, but can confirm the code passes analysis and compiles, similar to the iOS code in `native_view_example_3.dart`. I couldn't find any project on GitHub that included this code though.

cbracken:
@atsansone yes, this will fix that bug; added to the description -- thanks!

stuartmorgan:
> The initial version of this documentation is Swift-only, but a follow-up will add Objective-C example code.

Technically we've never supported Obj-C as a language for macOS. In practice it works because of interop of course, but we haven't traditionally done documentation or examples in anything but Swift.

src/content/platform-integration/macos/platform-views.md
```diff
@@ -0,0 +1,243 @@
...
+
+Implement the factory and the platform view. The `FLNativeViewFactory` creates
+the platform view, and the platform view provides a reference to the `NSView`.
+For example, `FLNativeView.swift`:
+
+```swift
+import Cocoa
+import FlutterMacOS
+
+class FLNativeViewFactory: NSObject, FlutterPlatformViewFactory {
```
stuartmorgan, line 92: Prefixing is an Obj-C-ism; I would remove the `FL`s from these classes.

src/content/platform-integration/macos/platform-views.md
```diff
@@ -0,0 +1,243 @@
...
+        super.init()
+        // macOS views can be created here
+        createNativeView(view: _view)
+    }
+
+    func view() -> NSView {
+        return _view
+    }
+
+    func createNativeView(view _view: NSView){
```
stuartmorgan, line 134: It would be good to run all this code through the Swift autoformatter so it has standard spacing, consistent indenting with [our other example code](https://docs.flutter.dev/platform-integration/platform-channels#step-6-add-a-macos-platform-specific-implementation), etc.

src/content/platform-integration/macos/platform-views.md
cbracken: Done.

src/content/platform-integration/macos/platform-views.md
cbracken: Done.

cbracken:
> Technically we've never supported Obj-C as a language for macOS.

Thanks. Was thinking it might be handy just in case people had some existing Obj-C NSViews they'd like to wrap but you're entirely right: we don't even have an Obj-C template for macOS, and people in that boat can just expose their own Swift-friendly header to the plugin. Not only that, but publishing Obj-C source implies that we're willing to commit to supporting that use-case and we aren't in a position to do so.

Removed the tabs and sending this out as Swift-only.

cbracken:
Alright; this is ready for a proper review.

examples/platform_integration/platform_views/lib/native_view_example_4.dart
domesticmouse: We should wrap it in a project here. @khanhnwin can give progress on how to do this in a way that makes sure this code will stay correct going forward.

sfshaza2 (approved):
lgtm. Thanks so much for this, @cbracken!

jmagman (approved):

parlough:
Does this need any version note or similar clarification?

examples/platform_integration/platform_views/lib/native_view_example_4.dart
cbracken: @khanhnwin currently the Dart code is checked into the examples directory similar to the iOS/Android versions, but the Swift code is not (also similar to the iOS/Android versions). Where's the best place to land that code?

@domesticmouse do you think we should hold off landing this until that happens, or follow up with separate PRs that lands macOS, iOS, and Android versions of the native code as well?

Agreed we should do this though -- writing the Swift code for this involved basically recreating the exact project you're asking for, just to be sure the code is actually correct.

sfshaza2:
@cbracken, are these docs valid for the current stable release? If not, I can add a note at the top stating that fact.

sfshaza2:
@loic-sharma, do you know if this info is valid as of now, or if it's valid after the next release? Before I land it, I want to be clear. If it's for the next release, I can still land it, but will first add a note.

stuartmorgan:
@sfshaza2 In general we don't want to tell people to use macOS platform views until after the next stable release (due to critical missing functionality like gesture support).

src/content/platform-integration/macos/platform-views.md
```diff
@@ -0,0 +1,236 @@
...
+app.
+If you'd like to embed native Android views in your Flutter app,
+see [Hosting native Android views][].
+If you'd like to embed native iOS views in your Flutter app,
+see [Hosting native iOS views][].
+:::
+
+[Hosting native Android views]: /platform-integration/android/platform-views
+[Hosting native iOS views]: /platform-integration/ios/platform-views
+
```
sfshaza2, line 26:
```suggestion

:::note Version note
Platform view support on macOS isn't fully functional as of the current release.
For example, gesture support isn't yet available on macOS.
Stay tuned for a future stable release.
:::
```

sfshaza2 (approved):
I've added a version note. Thanks for that info, @stuartmorgan!

PR #10833, opened 2024-07-02
Fix iOS platform views AppDelegate code excerpt

Urielbp:

The current example does not work, as the `registrar` variable is confused with the `registrar` method causing a `'NSInternalInconsistencyException', reason:
'Duplicate plugin key: <...>'`  exception.

The `register(
    _ factory: any FlutterPlatformViewFactory,
    withId factoryId: String)` method needs to be called with the created `weak var registrar`, not with a new `registrar` method.

I also propose a change to the variable name, not to be confused with the `registrar` method.

_Issues fixed by this PR (if any):_ None

_PRs or commits this PR depends on (if any):_ None

## Presubmit checklist


parlough:
Thanks for the PR and explanation!

@jmagman or @hellohuanlin would you be able to review this?

src/content/platform-integration/ios/platform-views.md
```diff
@@ -174,10 +174,10 @@ import UIKit
     ) -> Bool {
         GeneratedPluginRegistrant.register(with: self)
 
-        weak var registrar = self.registrar(forPlugin: "plugin-name")
+        weak var pluginRegistrar = self.registrar(forPlugin: "plugin-name")
 
-        let factory = FLNativeViewFactory(messenger: registrar!.messenger())
-        self.registrar(forPlugin: "<plugin-name>")!.register(
+        let factory = FLNativeViewFactory(messenger: pluginRegistrar!.messenger())
+        pluginRegistrar?.register(
```
hellohuanlin, line 180: nit: can be `guard let pluginRegisterar = self.registrar(forPlugin: "plugin-name") else { return false }`

hellohuanlin (approved):

sfshaza2 (approved):
@Urielbp, can you make the change that @hellohuanlin suggests? I can then land this.

thx!

src/content/platform-integration/ios/platform-views.md
Urielbp: Nice idea ;)
I didn't suggest this change before because I thought the `weak var` was needed to avoid some sort of retain cycle. In this example it's not possible to happen so I'll change to a `guard let`.

Done.

Urielbp:
@hellohuanlin @sfshaza2 can you re-review the PR? I made the suggested changes and I'm waiting for the pipeline to be started. Thanks.

hellohuanlin (approved):

parlough (approved):
Thanks so much for the PR and applying the suggestions @Urielbp! I'll land this once the staging build passes :)

Issue #1897, opened 2018-12-05
Clean up `_data/catalog/widgets.json` and `index.json`


chalin:
For example, the `image` field should be a URL to an image, not an image element. In cases where we'd like to use an embedded SVG, that should be assigned to field other than `image` (maybe `svg` :).

Also, we should consider at least partly generating the data used in this list.

chalin:
@Hixie - any comments about this?

micpap25:
Hi @chalin ,
I can definitely work on this; is this still something that should be done?

PR #10837, opened 2024-07-03
Clean up and simplify Material widget catalog setup

parlough:
- Remove the specially added `pagecontent` fields, which also wasn't working on the Material 2 page, and inline similar content directly into the relevant Markdown pages.
- Use shorter description for Material 3 on the index page, as it was causing other cards to be too large, and the transition info isn't super relevant to those navigating the index.
- Update "Text Input" subcategory of Material 3 to "Text input" for consistency with other headers on the site

@sfshaza2 There are copy changes to the content on both the `material.md` and `material2.md` pages. Please take a look. Thanks!

Contributes to https://github.com/flutter/website/issues/1897


domesticmouse (approved):

src/content/ui/widgets/material.md
```diff
@@ -1,8 +1,36 @@
...
   A catalog of Flutter's widgets implementing Material 3 design guidelines.
 ---
 
+Flutter provides a variety of visual, behavioral, and motion-rich widgets
+that implement the [Material 3][] design specification.
+Material 3 is the default design language of Flutter,
+enabling you to design and build beautiful, usable apps
+that can adapt to any platform.
+
+:::secondary
```
sfshaza2, line 14: Doesn't secondary require an alternate title? It used to.

sfshaza2 (approved):
lgtm, assuming that the secondary note doesn't need a title.

src/content/ui/widgets/material.md
parlough: It's optional :) - In this case I didn't want to add a title to reduce how much attention the aside attracted.

PR #10842, opened 2024-07-04
Clean up and simplify widget catalog source files

parlough:
- Switches the files from JSON to YAML for an easier editing and reviewing experience
- Uses consistence capitalization across category and subcategory titles
- Don't include raw HTML in the file for images
- Don't require an image source to be specified if using the placeholder image
- Add some alt text, with opportunity for more specific text in the future

Fixes https://github.com/flutter/website/issues/1897


domesticmouse (approved):

sfshaza2 (approved):
lgtm

PR #10847, opened 2024-07-06
Fix last updated date on production deploy

parlough:
Cloud Build seems to only do a shallow clone, so "unshallowing" the checkout is required to get last modified information.


domesticmouse (approved):

sfshaza2 (approved):
lgtm

PR #10853, opened 2024-07-07
Improved card responsiveness

parlough:
- Use a responsive grid system, so card decks don't need to manually specify row breaks, margins, etc
- Prevent cards being too narrow/squished when browser is narrow but side menu is still present
- Stops relying on some bootstrap styles for cards

Contributes to https://github.com/flutter/website/issues/8030


sfshaza2 (approved):
Wow! lgtm

PR #10900, opened 2024-07-15
Fixing title and superscripts

sfshaza2:
As it says

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "markdown-it-attrs": "^4.1.6",
     "markdown-it-container": "^4.0.0",
     "markdown-it-deflist": "^3.0.0",
+    "markdown-it-footnote": "^4.0.0",
     "sass": "^1.77.6",
     "shiki": "^1.10.1"
   }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -51,6 +51,9 @@ importers:
       markdown-it-deflist:
         specifier: ^3.0.0
         version: 3.0.0
+      markdown-it-footnote:
+        specifier: ^4.0.0
+        version: 4.0.0
       sass:
         specifier: ^1.77.6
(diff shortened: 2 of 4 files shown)


atsansone (changes requested):
This may not have been an error with the Markdown other than there was no space between the footnote and the link next to it. Consider fixing that instead of a full rewrite with HTML. We're trying to remove inline HTML as we find it as it isn't proper markdown.

See [Markdown Guide](https://www.markdownguide.org/extended-syntax/#footnotes) and [GitHub Markdown Guide](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#footnotes).

sfshaza2:
@atsansone, please remove the "request changes" tag. We've discussed this.

domesticmouse (approved):

atsansone:
> @atsansone, please remove the "request changes" tag. We've discussed this.

Yes, but this should be checked to see why the normal footnotes don't work. I'd prefer not to push this through when it may have been another issue, like spacing.

@parlough : This appears to be an issue with how 11ty handles footnotes. Its preprocessor goes Markdown -> Liquid -> HTML. This can be changed or the [`markdown-it-footnote` npm package](https://www.npmjs.com/package/markdown-it-footnote) can be installed.

sfshaza2:
@parlough, I defer to however you want to fix this. But the page is pretty messed up currently.

sfshaza2:
I created issue https://github.com/flutter/website/issues/10919

parlough (approved):
Thanks for the fix and for waiting @sfshaza2!

I updated the PR to enable the Markdown renderer to support the footnote syntax.

atsansone (approved):
Perfect! Thanks @parlough

PR #10941, opened 2024-07-24
Update CHANGELOG name

itsjustkevin:
This change removes references to "Hotfixes to the stable channel"

### Context

Hotfixes-to-the-stable-channel previously lived in the Flutter Wiki.  The decision was made to migrate the wiki to docs which opened an opportunity to align with Dart on how we post our changelog. 

https://github.com/flutter/flutter/pull/151931 was merged to align the processes by moving Hotfixes-to-the-stable-channel to the root of Flutter and renaming it CHANGELOG.md to align with Dart.

Some missed links outside of flutter/flutter were called out in https://github.com/flutter/flutter/issues/152162.

### DIFF
--- a/src/content/release/release-notes/index.md
+++ b/src/content/release/release-notes/index.md
@@ -8,8 +8,8 @@ This page links to announcements and release notes for
 releases to the stable channel.
 
 :::note
-For information about bug-fix releases, check out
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
lgtm

PR #10994, opened 2024-08-04
Cleanup extra breaking change page anchors from 11ty migration

parlough:
These extra anchors were supposed to be missing the dashes.


sfshaza2 (approved):
lgtm.  thx!

PR #10995, opened 2024-08-04
Fix IDs on widget catalog pages

parlough:
Since these are header elements, we should make sure their IDs can be properly used as a URI fragment by using the same slugify function as elsewhere on the site.


sfshaza2:
OOOhhhh. 18 missing anchors!

sfshaza2 (approved):
lgtm, once the missing anchors are fixed...

domesticmouse:
Looks like Sluggify is breaking the anchor targets. Do not want.

PR #11003, opened 2024-08-06
3.24 Release Notes

itsjustkevin:
Updates the 3.24 release notes.


domesticmouse (approved):

itsjustkevin:
@parlough in the future, how should I handle these links if I already know they are going to fail?

parlough:
> @parlough in the future, how should I handle these links if I already know they are going to fail?

No need to worry about them, as we'll make sure the breaking change header/anchor is landed before or at the same time as this PR. If this ever needs to land before that though, feel free to just drop the link to the breaking change entry temporarily and we can add it back.

PR #11012, opened 2024-08-06
Replace "Uri.parse" with "Uri.https"

tatsuyafujisaki:
This pull request replaces
```dart
Uri.parse('https://jsonplaceholder.typicode.com/<path>');
```
with
```dart
Uri.https('jsonplaceholder.typicode.com', '<path>')
```

as suggested below.
> To create a URI with https scheme, use [Uri.https](https://api.flutter.dev/flutter/dart-core/Uri/Uri.https.html) or [Uri.http](https://api.flutter.dev/flutter/dart-core/Uri/Uri.http.html):

https://api.dart.dev/stable/dart-core/Uri-class.html

I don't replace `Uri.parse('https://...')` of other URLs to avoid affecting a lot fo existing code.

## Presubmit checklist

### DIFF
--- a/examples/cookbook/networking/authenticated_requests/lib/main.dart
+++ b/examples/cookbook/networking/authenticated_requests/lib/main.dart
@@ -7,7 +7,7 @@ import 'package:http/http.dart' as http;
 Future<Album> fetchAlbum() async {
   // #docregion get
   final response = await http.get(
-    Uri.parse('https://jsonplaceholder.typicode.com/albums/1'),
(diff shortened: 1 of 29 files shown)


domesticmouse (changes requested):
I can't see the text in the linked page that states to use `Uri.https` instead of `Uri.parse`. The `parse` method is more general than the `https` method, but that's the only difference I can see. If you already have the url partially parsed, the `https` method makes sense, but it obfuscates the URL, making it harder to read.

tatsuyafujisaki:
I understand. Thank you for your explanation!

Issue #11011, opened 2024-08-06
First time build appears to hang for a while


RedBrogdon:
### Describe the problem

I just did a fresh checkout of the repo, and the first time I built the site, I got this output:

```shell
Building package executable...
Built flutter_site:flutter_site.
[11ty] Benchmark   3392ms  13%   512× (Configuration) "generate_toc" Liquid Filter
[11ty] Copied 966 Wrote 594 files in 26.70 seconds (44.9ms each, v3.0.0-beta.1)
[11ty] Watching…
[11ty] Server at http://localhost:4000/
```

Between the second line ("Built flutter_site...") and the third, there was about half a minute of hang time with no output, which caused me to wonder if I had a problem.

### Expected fix

This is a DevExp nit for the infra, but if it's possible to have some kind of progress indicator during that time, it would improve the experience of building the site for new contributors.

### On which platform(s) did you experience this issue?

macOS

RedBrogdon:
@parlough, let me know what you think. If this would be non-trivial to address, feel free to close.

PR #11030, opened 2024-08-08
Add some indicator that site is building

parlough:
This isn't a perfect solution as it doesn't provide progress updates, but it will at least indicate to the command runner that the command is working and might take a while.

If I have time in the future, I can revisit this to be more robust.

Also adds a verbose flag to the `serve` command to get 11ty debug output.

Resolves https://github.com/flutter/website/issues/11011


domesticmouse (approved):

RedBrogdon (approved):
👍

PR #11058, opened 2024-08-18
Switch special tabs from SwiftUI page to inline button

parlough:
Also update the referenced gists by forking them and using the new URLs.


sfshaza2 (approved):
lgtm!

PR #11063, opened 2024-08-19
rephrase "instances of any screen size"; other fixes

yjbanov:

"instances of any screen size" doesn't quite make sense. What's an "instance", and how can it have a "screen size"? For the purposes of add-to-app a _Flutter view_ is embedded into the app, and it indeed [has size](https://api.flutter.dev/flutter/dart-ui/FlutterView/physicalSize.html).

I also digested what "hybrid navigation stack" and "partial-screen views" mean. They are not self-explanatory.


sfshaza2 (approved):
lgtm

PR #11090, opened 2024-08-28
Fix build failure and update other deps

parlough:
This PR works around the error caused by https://github.com/11ty/eleventy/issues/3425 by pinning the `liquidjs` dependency to an earlier version.


sfshaza2 (approved):
lgtm

PR #11093, opened 2024-08-30
Migration guide for updated Material 3 `Slider`

TahaTesser:
Related to https://github.com/flutter/flutter/pull/152237

## Presubmit checklist


src/content/release/breaking-changes/introduce-new-slider-shapes.md
```diff
@@ -0,0 +1,102 @@
...
+shape includes an updated height, a gap between the active and inactive tracks,
+and stop indicators to show the end value of the inactive track.
+Pressing the thumb adjusts its width and the track adjusts its shape. The
+new value indicator shape is a rounded rectangle. New color mappings are also
+introduced for some of the `Slider` shapes.
+
+## Migration guide
+
+You automatically get the new `Slider` shapes and color mappings.
+To use the previous `Slider` shapes and color mappings, set the following
```
Piinks, line 25: Is going through the theme the only option? Can folks configure directly on `Slider`?

TahaTesser: Theme only. There are too many slider components parameters so looks like they weren't passed to Slider directly.

Piinks: Hmm, working with some customer that are affected by this, some of them don;t like having to insert a theme to maintain backwards compatibility.
How many properties would be exposed on Slider in order to achieve backwards compatibility?

TahaTesser: Pretty all of them, otherwise they would be inconsistent. 

If necessary, what about if introduce single customization parameter in the Slider widget that take all of the Slider components and customizes the Slider. Something like TextStyle, AnimationStyle, we could've SliderStyle?

Piinks: That is also a really good idea!

Piinks: If you feel like that is better than a bunch of new constructors, let's give it a try and see how it looks. I trust your assessment of the two options. :)

TahaTesser:
This is updated for new flag.

sfshaza2 (approved):
lgtm.

PR #11116, opened 2024-09-09
Remove/disable broken scorecard workflow

parlough:
The Flutter EngProd team is not currently using or maintaining the setup for this, causing the workflow to fail and the site build to appear as failing.

Mirrors the change in https://github.com/flutter/samples/pull/2430


sfshaza2 (approved):
Thanks, @parlough!!

Xanhum-wymra5-buxjer (commented):
L

PR #11118, opened 2024-09-10
Fix footnotes on architectural overview page

parlough:
- Use the recent footnote support we added as a Markdown plugin
- Remove custom styling on the page
- Add a scroll margin so the footnote reference is in view once navigated back to
- Some other related cleanup on the page


sfshaza2 (approved):
lgtm

Issue #10525, opened 2024-05-07
Migrate Jekyll `include` to liquidjs `render` where possible


parlough:
The `include` syntax we're using is supported as a fallback for migration, and [`render`](https://liquidjs.com/tags/render.html) is the appropriate replacement tag. Its included content doesn't inherit its parents scope, reducing the chance of mistakes.

parlough:
This has been completed over a variety of commits. Some following clean up will be completed in https://github.com/flutter/website/issues/12405.

PR #11121, opened 2024-09-10
Fix headers and toc duplication in widget catalog

parlough:
Headers weren't being modified by our 11ty/markdown to add anchors, so this PR switches them to use Markdown and enable the automatic processing.

The pages also had duplicate TOCS, one manual and another automatically generated. This PR removes the manual one to standardize with the rest of the site.

Also slightly improves the rendering on the catalog pages with only one or two entries in a subcategory.

Contributes to https://github.com/flutter/website/issues/10525 and makes future site redesign changes easier


sfshaza2 (approved):
lgtm

PR #11125, opened 2024-09-10
Update Dart platforms diagram

parlough:


kevmoo:
Couldn't we just reuse the SVG image? It'll scale fine, right?

parlough:
This image is used in a few spots outside the website that expect the added Dart logo and a PNG.

For now I'll update this for the legacy references and see if I can update those external usages to use the svg hosted on dart.dev.

sfshaza2 (approved):
lgtm

PR #11135, opened 2024-09-13
Update `firebase.json` with `/to/spm` for Swift Package Manager page

khanhnwin:
We're releasing a short video that promotes the new Swift Package Manager docs. We'd like a shorter URL to send folks to the docs.


khanhnwin:
cc @loic-sharma

firebase.json
```diff
@@ -84,6 +84,7 @@
       { "source": "/sdk-archive", "destination": "/release/archive", "type": 301 },
       { "source": "/setup-:os*", "destination": "/get-started/install/:os*", "type": 301 },
       { "source": "/setup", "destination": "/get-started/install", "type": 301 },
+      { "source": "/spm", "destination": "/packages-and-plugins/swift-package-manager", "type": 301 },
```
parlough, line 87: What will this be used for? I'd prefer to use a `/to/spm` [tooling redirect](https://github.com/flutter/flutter/blob/master/docs/contributing/use-reliable-links.md) if possible.

khanhnwin: It'll be used in a YouTube Short video. Good to use whichever format you prefer!

parlough (approved):
Thanks @khanhnwin!

I've updated the redirect to use `/to/spm`, so after this lands `https://flutter.dev/to/spm` should work :)

khanhnwin:
Thank you @parlough!

PR #11137, opened 2024-09-13
Remove use of third-party dependency for cookie banner

parlough:
Switch implementation of banner to use `localStorage` instead of cookies (through `js-cookie`) to:

- Reduce the amount of dependencies we have
- Reduce page loading time
- Make upcoming site design changes a bit easier


sfshaza2 (approved):
lgtm

PR #11185, opened 2024-09-26
Replace C cast in example with more correct use of standard macro

robert-ancell:
The existing example was bad practice.


robert-ancell:
Discovered when checking if the same issues in https://github.com/flutter/flutter/pull/155735 were in the example (they aren't).

parlough (approved):
Looks good to me from a website perspective, thanks!

sfshaza2 (approved):
lgtm

PR #11191, opened 2024-09-26
Automatically run `pnpm install` in dash_site tool

parlough:
After the initial setup/install, this will run fast enough to not impede the serving process, while improving the contribution experience.


sfshaza2 (approved):
lgtm. Thx!!!

Issue #11178, opened 2024-09-25
Flavors documentation omits important section for `appFlavor` when building from XCode


navaronbracke:
### Page URL

https://docs.flutter.dev/deployment/flavors

### Describe the problem

The flavors documentation does not include a step to set the `FLUTTER_APP_FLAVOR` Dart define for the `appFlavor` constant. This is important to include when building from XCode, as otherwise the `appFlavor` is null.

### Expected fix

When setting up Schemes in XCode, we should also add the following to the documentation:
1) Navigate to the `Build Settings` of the target(s)
2)  If `DART_DEFINES` does not exist yet
2a) Click the `+` button in the top left
2b) Click `Add User-Defined Setting`
2c) For the name enter `DART_DEFINES`
2d) For the value take the base64 encoded version of `FLUTTER_APP_FLAVOR=<value>`, where `<value>` is the name of the flavor that is expected to be the `appFlavor` in Dart
3) If `DART_DEFINES` does exist, the values are delimited by a comma (this should be verified, I am not sure) But it is important to encode the _entire string_ as base64
4) If `DART_DEFINES` is already defined in `Generated.xcconfig`, delete the `Generated.xcconfig` and enter the Dart defines in XCode, before trying again.

### Additional context

I confirmed that this is not a tool bug, as the Flutter tool already picks up existing `DART_DEFINES` from the iOS/MacOS Scheme when present.

![Screenshot 2024-09-25 at 14 30 16](https://github.com/user-attachments/assets/bf5174fc-45a9-42fa-8bd5-71cdac276d45)

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

sfshaza2:
Excellent bug report. Thx!

jmagman:
This means the `DART_DEFINES` would have to be kept in sync with the Xcode configurations and schemes, which already encodes which flavor is being used. And it also conflicts with the existing only-half-implemented `FLAVORS` environment variable.
https://github.com/flutter/flutter/blob/9441f9d48fce1d0b425628731dd6ecab8c8b0826/packages/flutter_tools/bin/xcode_backend.dart#L489
https://github.com/flutter/flutter/blob/9441f9d48fce1d0b425628731dd6ecab8c8b0826/packages/flutter_tools/lib/src/build_system/targets/ios.dart#L554

[...]

PR #11192, opened 2024-09-26
Add setting DART_DEFINES to flavors page

sfshaza2:
Fixes https://github.com/flutter/website/issues/11178

Please review, @loic-sharma and/or @jmagman 

FYI @sethladd 

**Staged:** https://flutter-docs-prod--pr11192-flavors-qqqcgtun.web.app/deployment/flavors

### DIFF
--- a/src/content/deployment/flavors.md
+++ b/src/content/deployment/flavors.md
@@ -77,7 +77,7 @@ If `DART_DEFINES` doesn't exist yet:
 You must encode the entire string as `base64`.
 :::
 
-If DART_DEFINES exists, separate the values with a comma.
+If `DART_DEFINES` exists, separate the values with a comma.
 
 If `DART_DEFINES` is already defined in `Generated.xcconfig`,
 delete that file and enter the `DART_DEFINES` setting in Xcode
--- a/src/content/deployment/flavors.md
+++ b/src/content/deployment/flavors.md
@@ -58,6 +58,34 @@ Open your project in Xcode.
 </li>
 <li>
 
+Navigate to the **Build Settings** of the target(s)
+and follow these instructions:
+
+</li>
+
+If `DART_DEFINES` doesn't exist yet:
+
+* Click the **+** button in the top left.
+* Select **Add User-Defined**.
+* Enter `DART_DEFINES` as the name.
+* Enter the **base64 encoded version**
+  of `FLUTTER_APP_FLAVOR=<value>` as the value,
+  where `value` is the name of the flavor
+  that is expected to be the `appFlavor` in Dart.
+
+:::note Important
+You must encode the entire string as `base64`.
+:::
+
+If DART_DEFINES exists, separate the values with a comma.
+
+If `DART_DEFINES` is already defined in `Generated.xcconfig`,
+delete that file and enter the `DART_DEFINES` setting in Xcode
+before trying again.
+
+</li>
+<li>
+
 Select **Product** > **Scheme** > **New Scheme** from the menu
 to add a new `Scheme`.
 
--- a/src/content/deployment/flavors.md
+++ b/src/content/deployment/flavors.md
@@ -8,11 +8,15 @@ description: >
 
 ## What are flavors
 
-Have you ever wondered how to set up different environments in your Flutter app?
-Flavors (known as _build configurations_ in iOS and macOS), allow you (the developer) to
-create separate environments for your app using the same code base.
-For example, you might have one flavor for your full-fledged production app,
-another as a limited "free" app, another for testing experimental features, and so on.
+Have you ever wondered how to set up different environments
+in your Flutter app?
+Flavors (known as _build configurations_ in iOS and macOS),
+allow you to create separate environments for your app
+using the same code base.
+For example, you might have one flavor for your
+full-fledged production app,
+another as a limited "free" app,
+another for testing experimental features, and so on.
 
 Say you want to make both free and paid versions of your Flutter app.
 You can use flavors to set up both app versions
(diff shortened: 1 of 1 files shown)


src/content/deployment/flavors.md
```diff
@@ -52,32 +58,63 @@ Open your project in Xcode.
...
+and follow these instructions:
+
+</li>
+
+If `DART_DEFINES` doesn't exist yet:
+
+* Click the **+** button in the top left.
+* Select **Add User-Defined**.
+* Enter `DART_DEFINES` as the name.
+* Enter the **base64 encoded version**
```
loic-sharma, line 71: Wikipedia seems to always capitalize "Base64", should we as well?

See: https://en.wikipedia.org/wiki/Base64

```suggestion
* Enter the **Base64 encoded version**
```

src/content/deployment/flavors.md
```diff
@@ -52,32 +58,63 @@ Open your project in Xcode.
...
+  of `FLUTTER_APP_FLAVOR=<value>` as the value,
+  where `value` is the name of the flavor
+  that is expected to be the `appFlavor` in Dart.
+
+:::note Important
+You must encode the entire string as `base64`.
+:::
+
+If DART_DEFINES exists, separate the values with a comma.
```
loic-sharma, line 80:
```suggestion
If `DART_DEFINES` exists, separate the values with a comma.
```

src/content/deployment/flavors.md
```diff
@@ -52,32 +58,63 @@ Open your project in Xcode.
...
+* Click the **+** button in the top left.
+* Select **Add User-Defined**.
+* Enter `DART_DEFINES` as the name.
+* Enter the **base64 encoded version**
+  of `FLUTTER_APP_FLAVOR=<value>` as the value,
+  where `value` is the name of the flavor
+  that is expected to be the `appFlavor` in Dart.
+
+:::note Important
+You must encode the entire string as `base64`.
```
loic-sharma, line 77: If I want my app flavor to be `foo`, should I set `DART_DEFINES` to `FLUTTER_APP_FLAVOR=Zm9v` (only `foo` is encoded) or to `RkxVVFRFUl9BUFBfRkxBVk9SPWZvbw==` (all of `FLUTTER_APP_FLAVOR=foo` is encoded)?

I believe it's the latter, perhaps an example would help clarify?

```suggestion
You must encode the entire string as `base64`.

For example, for flavor `foo` set `DART_DEFINES` to `RkxVVFRFUl9BUFBfRkxBVk9SPWZvbw==` (base64 encoding of `FLUTTER_APP_FLAVOR=foo`).
```

src/content/deployment/flavors.md
loic-sharma, line 80: Noob question: are you supposed to put the commas before doing the base64 encoding? For example, if I want the Dart defines `HELLO=world` and `FOO=bar`, would my `DART_DEFINES` be `SEVMTE89d29ybGQsRk9PPWJhcg==` (base64 encoding of `HELLO=world,FOO=bar`)?

I think an example would help clarify this case too

src/content/deployment/flavors.md
```diff
@@ -52,32 +58,63 @@ Open your project in Xcode.
...
+
+If `DART_DEFINES` doesn't exist yet:
+
+* Click the **+** button in the top left.
+* Select **Add User-Defined**.
+* Enter `DART_DEFINES` as the name.
+* Enter the **base64 encoded version**
+  of `FLUTTER_APP_FLAVOR=<value>` as the value,
+  where `value` is the name of the flavor
+  that is expected to be the `appFlavor` in Dart.
```
loic-sharma, line 74: It looks like this is the first time we reference `appFlavor` in this doc, consider making this a link too:

```suggestion
  that is expected to be the [`appFlavor`][] in Dart.
```

src/content/deployment/flavors.md
sfshaza2: I don't know! That hadn't occurred to me. Who would know in SWE land?

sfshaza2: @jmagman, do you know?

sfshaza2:
I've added a link to a DartPad where you can encode the text. Can't land this until we know the status of comma delimited strings, encoding-wise.

jmagman (changes requested):
1. I filed https://github.com/flutter/flutter/issues/155951 since this solution seems like a bulky workaround to a feature hole in the Flutter tool (I didn't even know `FLUTTER_APP_FLAVOR` existed--we don't explicitly set it in our iOS flavors samples), and it would be very easy to get out-of-sync with the Xcode configurations.  


2. If we do use this as a workaround, it should use the "free" flavor example on the rest of this page.  These steps need to happen after the flavor is already set up in Xcode, not at the beginning.  https://flutter-docs-prod--pr11192-flavors-qqqcgtun.web.app/deployment/flavors#using-flavors-in-ios-and-macos near the `Product Bundle Identifier` instructions would be better.

3. Over-riding the DART_DEFINES being passed in from the tool in the Xcode configuration is also going to get users into a wonky state.  How are users supposed to know what else the tool is passing into DART_DEFINES? Do they need to know to update it in Xcode if it changes?

> If DART_DEFINES is already defined in Generated.xcconfig, delete that file and enter the DART_DEFINES setting in Xcode before trying again.

I'm not sure what bug is happening in this step in the reported issue https://github.com/flutter/website/issues/11178 but it's concerning, I've never seen this advice before.

sfshaza2:
OK, given @jmagman's comments and the bug she's filed against the s/w, I'm closing this PR, as it can create new problems for a dev. Thanks, @navaronbracke, for surfacing a bug that we didn't know about!

PR #11209, opened 2024-09-30
Use our standard footnote support for 'Inside Flutter' doc

parlough:


sfshaza2 (approved):
Thanks for this cleanup! lgtm

PR #11211, opened 2024-09-30
Updating iOS 18 supported feature table

sfshaza2:
As it says and updated with input from @sethladd 

cc @jmagman


jmagman (approved):
Much better, thank you!

jmagman:
@sfshaza2 I also think the "Flutter on iOS" in the navigation should be like "Flutter on latest iOS" and be moved below (or well-below) "Add iOS as a build target".  The latter describes how to actually deploy to Flutter on iOS, which is more likely what people are looking for.

Screenshot 2024-09-30 at 7 45 45 PM

PR #11221, opened 2024-10-03
Update url-strategies.md

chrismercredi:
You can't add flutter_web_plugins directly using pub add. Its only included as part of the project if required by other dependencies like go_router.

Added additional information on how to add flutter_web_plugins to project to use usePathUrlStrategy() function.  

You can't add flutter_web_plugins directly using pub add or other methods. 

## Presubmit checklist

### DIFF
--- a/src/content/ui/navigation/url-strategies.md
+++ b/src/content/ui/navigation/url-strategies.md
@@ -17,8 +17,15 @@ For example, `flutterexample.dev/#/path/to/screen`.
 ## Configuring the URL strategy
 
 To configure Flutter to use the path instead, use the
-[usePathUrlStrategy][] function provided by the [flutter_web_plugins][] library
-in the SDK:
+[usePathUrlStrategy][] function provided by the [flutter_web_plugins][] library,
+which is part of the Flutter SDK.
+
+You cannot directly add `flutter_web_plugins` via `pub add`,
+as it is only available through dependencies like `go_router`. 
+When you add the `go_router` package, it automatically includes
+`flutter_web_plugins`. 
+
+Here's an example of how to use it:
 
 ```dart
 import 'package:flutter_web_plugins/url_strategy.dart';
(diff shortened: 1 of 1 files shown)


parlough (commented):
Thanks for the PR! It definitely makes sense to clarify that it doesn't need to be added from pub.

However, I believe relying on it as a transitive dependency isn't the only option. You can also specify it as a [sdk dependency](https://dart.dev/tools/pub/dependencies#sdk), similar to `flutter_test`.

```
dependencies:
  flutter:
    sdk: flutter
  flutter_web_plugins:
    sdk: flutter
```

What do you think about suggesting this instead? Since the `depend_on_referenced_packages` lint would report that it's not depended on if it it's retrieved through a transitive dependency.

Let me know what you think. Thanks again! :D

chrismercredi:
> Thanks for the PR! It definitely makes sense to clarify that it doesn't need to be added from pub.
> 
> However, I believe relying on it as a transitive dependency isn't the only option. You can also specify it as a [sdk dependency](https://dart.dev/tools/pub/dependencies#sdk), similar to `flutter_test`.
> 
> ```
> dependencies:
>   flutter:
>     sdk: flutter
>   flutter_web_plugins:
>     sdk: flutter
> ```
> 
> What do you think about suggesting this instead? Since the `depend_on_referenced_packages` lint would report that it's not depended on if it it's retrieved through a transitive dependency.
> 
> Let me know what you think. Thanks again! :D

Thanks for the feedback, I've updated it based on your suggestions :)

src/content/ui/navigation/url-strategies.md
```diff
@@ -17,8 +17,21 @@ For example, `flutterexample.dev/#/path/to/screen`.
 ## Configuring the URL strategy
 
 To configure Flutter to use the path instead, use the
-[usePathUrlStrategy][] function provided by the [flutter_web_plugins][] library
-in the SDK:
+[usePathUrlStrategy][] function provided by the [flutter_web_plugins][] library,
+which is part of the Flutter SDK.
+
+You cannot directly add `flutter_web_plugins` using `pub add`.
```
parlough, line 23:
```suggestion
You can't directly add `flutter_web_plugins` using `pub add`.
```

parlough (approved):
Thanks so much for those adjustments! Clarifying this will definitely be super helpful :)

Issue #11250, opened 2024-10-09
404 on Android Migration docs


feinstein:
### Page URL

https://flutter.dev/to/to/java-gradle-incompatibility

### Describe the problem

I got a compilation error on my project, and the flutter tool have me this output:

<img width="843" alt="image" src="https://github.com/user-attachments/assets/b3bc34a0-15d6-43e8-9d39-ef9d6c890d35">

But when I click the link, I get a 404

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

sfshaza2:
Ah! You, sir, have found an error in the tooling. The actual link has only one "to":

https://flutter.dev/to/java-gradle-incompatibility

Please feel free to file a bug against the tooling. Meanwhile, I was add a redirect on our site.

thx!

PR #11257, opened 2024-10-09
Working around a broken redirect in the tooling

sfshaza2:
Fixes https://github.com/flutter/website/issues/11250

I didn't stage this locally, so double check when staged.

cc @parlough


parlough (approved):
Good idea to add this workaround!

PR #11264, opened 2024-10-12
Remove codelabs submodule and usages

parlough:
The usages are very minimal and the website repo is much smaller and easier to manage without the huge codelabs repo.

Also fixes some copy formatting and punctuation issues near the updated code snippets.


sfshaza2 (approved):
Wow! lgtm

PR #11265, opened 2024-10-12
Remove unused example code and update dependabot config

parlough:
Removes the unused samples in `examples/basics` and `examples/platform_integration/ffi`. Then updates the dependabot config to remove them.

Also updates the dependabot config to rely on the simpler `directories` key, so we don't need to configurate each directory separately.


sfshaza2 (approved):
lgtm

PR #11278, opened 2024-10-16
Follow up clean up to URL strategy config doc

parlough:
Fixes one issue I didn't notice (code block being `dart` instead of `yaml`), and some other minor enhancements, primarily highlighting.


sfshaza2 (approved):
lgtm

PR #11282, opened 2024-10-16
Turn flutter_site tool into part of root workspace

parlough:
This will allow us to share one package resolution for all tools on the site.

In Dart 3.6, we'll do the same for the examples directory.


parlough:
Ah nevermind, I forgot we use the `flutter` tool for pub get. Rather than add an exception, I'll just wait until 3.6 for this change.

PR #11283, opened 2024-10-16
Update lints for example snippets

parlough:
Removes some deprecated lints, enables a few new ones that are popular, and fixes the resulting diagnostics :)


sfshaza2 (approved):
lgtm!

PR #11293, opened 2024-10-20
Fix a few typos across the site

parlough:


sfshaza2 (approved):
Honestly, given that we don't use AI for spelling, this isn't bad!

lgtm

Issue #11341, opened 2024-10-30
assetBase is incorrectly documented


dowski:
### Page URL

https://docs.flutter.dev/platform-integration/web/initialization/

### Page source

https://github.com/flutter/website/tree/main/src/content/platform-integration/web/initialization.md

### Describe the problem

The documentation for "The `_flutter.loader.load()` API" says that the `assetBase` value is part of the `config` object. That seems incorrect. `assetBase` needs to be passed to `engineInitializer.initializeEngine`.

### Expected fix

The `assetBase` field description should be removed from the `config` object description table. Instead it should be mentioned in the **Initializing the Flutter engine** section below.

### I would like to fix this problem.

- [X] I will try and fix this problem on docs.flutter.dev.

dowski:
Welp, for anyone ever looking at this, see https://github.com/flutter/website/pull/11342#issuecomment-2452777564.

The docs for `assetBase` aren't really wrong, but if you use `onEntryPointLoaded` in your call to `_flutter.loader.load()`, passing `assetBase` in the `config` won't have any effect.

dowski:
I'm going to close this and open a new more focused doc bug.

Issue #11320, opened 2024-10-24
Update "Flutter web app initialization" page


johnpryan:
- [ ] Tweak outline / headers to be more clear:
	- Bootstrapping
	- Customize initialization
	- Write a custom bootstrap script
	- Customize the Flutter Loader script
	- remove serviceWorkerSettings
	- Example: Customizing Flutter configuration based on URL query parameters
	- The onEntrypointLoaded callback
	- "Example: Display a progress indicator"
- [ ] Remove "Upgrade an older project" section
- [ ] Remove initialization-legacy

PR #11342, opened 2024-10-30
Fix incorrect assetBase documentation

dowski:
_Description of what this PR is changing or adding, and why:_ Fixing a documentation bug.

_Issues fixed by this PR (if any):_ #11341 

_PRs or commits this PR depends on (if any):_ N/A

## Presubmit checklist

- [ ] This PR is marked as draft with an explanation if not meant to land until a future stable release.

### DIFF
--- a/src/content/platform-integration/web/initialization.md
+++ b/src/content/platform-integration/web/initialization.md
@@ -116,6 +116,7 @@ arguments to customize initialization behavior:
 
 | Name                    | Description                                                                                                                   | JS&nbsp;type |
 |-------------------------|-------------------------------------------------------------------------------------------------------------------------------|--------------|
+|`assetBase`              | The base URL of the `assets` directory of the app. Add this when Flutter loads from a different domain or subdirectory than the actual web app. You might need this when you embed Flutter web into another app, or when you deploy its assets to a CDN. |`string`|
 | `config`                | The Flutter configuration of your app.                                                                                        | `Object`     |
 | `onEntrypointLoaded`    | The function called when the engine is ready to be initialized. Receives an `engineInitializer` object as its only parameter. | `Function`   |
 | `serviceWorkerSettings` | The configuration for the `flutter_service_worker.js` loader. (If not set, the service worker isn't used.)                    | `Object`     |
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
lgtm

dowski:
So actually, I just double-checked my code and the doc change I made is not right. `assetBase` is actually an argument to engine intializer `initializeEngine`, not `_flutter.loader.load()`.

dowski:
The `assetBase` arg to `initializeEngine` is still documented, although the whole page that contains those docs say they are for legacy versions of Flutter.

https://docs.flutter.dev/platform-integration/web/initialization-legacy#initializing-the-engine

dowski:
Okay, I tweaked this PR to make it correct. It does feel like some restructuring of the docs is in order (`assetBase` is nicely documented on the "legacy" initialization page). Maybe that's part of the goal of #11320?

dowski:
*Sigh*

I think this is doc change is still wrong, after reading more of the code.

As long as you **don't supply a custom `onEntryPointLoaded` callback**, you *can* supply `assetBase` in the `config`. It will get passed to `initializeEngine` [here](https://github.com/flutter/engine/blob/05cb5d7f7939da58ef9897997004f181f2699075/lib/web_ui/flutter_js/src/entrypoint_loader.js#L63-L65).

If you **do supply a custom `onEntryPointLoaded` callback**, you must manually pass it the `config` that was passed to `_flutter.loader.load()` (or instead of passing it there? -- I'm not clear on that).

----

Maybe the real fix here is to document that the `initializeEngine` call should receive the same config that `_flutter.loader.load()` receives? What do you all think?

dowski:
I'm going to close this and come back w/ a fresh PR focused on what I discovered above.

johnpryan:
Thanks for sharing your findings @dowski! I haven't validated this, but am I understanding correctly that when you configure`assetBase` in`flutter.loader.load()`, you need to also provide it to the `onEntrypointLoaded` callback? cc: @ditman @eyebrowsoffire

dowski:
Hey @johnpryan - you're welcome.

It's more like if you pass an `onEntryPointLoaded` callback, `assetBase` gets ignored unless you manually pass it to the `initializer. initializeEngine` call in the callback.

I wound up filing a new focused bug that includes sample code:

https://github.com/flutter/flutter/issues/158041

I didn't file it here in the `website` repo as I'm not sure it's strictly a documentation issue. The API feels plain hard to use.

PR #11353, opened 2024-11-01
Update the "setState called during build" in common-errors.md

554richard:
I ran into the error "setState called during build" while writing an app which uses the pluto_grid package. I had the need to validate a cell in the onChanged callback of the grid, but got the error when I tried to do a showDialog to alert the user to an error. 
To solve it I went to the common-errors page in the flutter docs, but found a) the code describing the problem was a snippet with no obvious functionality and b) the solution code is for a different situation, not a direct solution to either my problem or the one in the given snippet.

I can see that there are different solutions in different scenarios (Navigator, FutureBuilder, addPostFrameCallback) but I feel the addPostFrameCallback is the most intuitive and direct, particularly in the showDialog scenario.

The example code I give is a bit arbitrary and useless, but I can see that including code with the pluto_grid package would be inappropriate, and I wasn't sure what else I could use for an illustration.

This is my first attempt at contributing to an Open Source project of any sort. Please let me know if I've messed up., so I can improve at the procedure.

### DIFF
--- a/src/content/testing/common-errors.md
+++ b/src/content/testing/common-errors.md
@@ -443,46 +443,27 @@ attempting to trigger a `Dialog` from within the
 immediately show information to the user,
 but `setState` should never be called from a `build` method.
 
-The following code illustrates a common culprit of this error:
+The following snippet seems to be a common culprit of this error:
 
 <?code-excerpt "lib/set_state_build.dart (problem)"?>
 ```dart
-import 'package:flutter/material.dart';
-
-void main() => runApp(const ShowDialogExampleApp());
-
-class ShowDialogExampleApp extends StatelessWidget {
-  const ShowDialogExampleApp({super.key});
-
-  @override
-  Widget build(BuildContext context) {
-    return const MaterialApp(
-      home: DialogExample(),
-    );
-  }
-}
-
-class DialogExample extends StatelessWidget {
-  const DialogExample({super.key});
-
-  @override
-  Widget build(BuildContext context) {
-    // Don't do this:    
-    showDialog(
+Widget build(BuildContext context) {
+  // Don't do this.
+  showDialog(
       context: context,
-      builder: (BuildContext context) {
+      builder: (context) {
         return const AlertDialog(
           title: Text('Alert Dialog'),
         );
-      },
-    );
-    return Scaffold(
-      appBar: AppBar(title: const Text('This does not work')),
-      body: const Center(
-        child: Text('Does not Work'),
-      ),
-    );
-  }
+      });
+
+  return const Center(
+    child: Column(
+      children: <Widget>[
+        Text('Show Material Dialog'),
+      ],
+    ),
+  );
 }
 ```
 
@@ -496,7 +477,8 @@ framework for every frame, for example, during an animation.
 
 One way to avoid this error is instruct flutter to finish rendering the page
 before calling `showDialog()`. This can be done by using
-addPostFrameCallback() method. The following code illustrates this on the broken example just given:
+addPostFrameCallback() method. 
+The following code illustrates this on the broken example just given:
 
 <?code-excerpt "lib/set_state_build.dart (solution)"?>
 ```dart
--- a/src/content/testing/common-errors.md
+++ b/src/content/testing/common-errors.md
@@ -443,27 +443,48 @@ attempting to trigger a `Dialog` from within the
 immediately show information to the user,
 but `setState` should never be called from a `build` method.
 
-The following snippet seems to be a common culprit of this error:
+The following code illustrates a common culprit of this error:
 
 <?code-excerpt "lib/set_state_build.dart (problem)"?>
 ```dart
-Widget build(BuildContext context) {
-  // Don't do this.
-  showDialog(
+import 'package:flutter/material.dart';
+
+void main() => runApp(const ShowDialogExampleApp());
+
+class ShowDialogExampleApp extends StatelessWidget {
+  const ShowDialogExampleApp({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return MaterialApp(
+      theme: ThemeData(
+          colorSchemeSeed: const Color(0xff6750a4), useMaterial3: true),
+      home: const DialogExample(),
+    );
+  }
+}
+
+class DialogExample extends StatelessWidget {
+  const DialogExample({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    // Don't do this:    
+    showDialog(
       context: context,
-      builder: (context) {
+      builder: (BuildContext context) {
         return const AlertDialog(
           title: Text('Alert Dialog'),
         );
-      });
-
-  return const Center(
-    child: Column(
-      children: <Widget>[
-        Text('Show Material Dialog'),
-      ],
-    ),
-  );
+      },
+    );
+    return Scaffold(
+      appBar: AppBar(title: const Text('This does not work')),
+      body: const Center(
+        child: Text('Does not Work'),
+      ),
+    );
+  }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


src/content/testing/common-errors.md
```diff
@@ -443,27 +443,48 @@ attempting to trigger a `Dialog` from within the
 immediately show information to the user,
 but `setState` should never be called from a `build` method.
 
-The following snippet seems to be a common culprit of this error:
+The following code illustrates a common culprit of this error:
 
 <?code-excerpt "lib/set_state_build.dart (problem)"?>
 ```dart
```
nate-thegrate, line 449: If all the code required for running an app is being added here, maybe it could be structured as an interactive DartPad sample! I'm not sure exactly how it works but I did see another page that was formatted this way:

https://github.com/flutter/website/blob/93fe19aca35566afda7bb12b78ff952d401586b8/src/content/cookbook/design/tabs.md?plain=1#L96-L136

src/content/testing/common-errors.md
```diff
@@ -475,43 +496,49 @@ framework for every frame, for example, during an animation.
...
-One way to avoid this error is to use the `Navigator` API
-to trigger the dialog as a route. In the following example,
-there are two pages. The second page has a
-dialog to be displayed upon entry.
-When the user requests the second page by
-clicking a button on the first page,
-the `Navigator` pushes two routes–one
-for the second page and another for the dialog.
+One way to avoid this error is instruct flutter to finish rendering the page
+before implementing the showDialog. This can be done by using
```
nate-thegrate, line 500:
```suggestion
before calling `showDialog()`. This can be done by using
```

"implementing" = creating the API
"calling" = invoking the function

src/content/testing/common-errors.md
```diff
@@ -475,43 +496,49 @@ framework for every frame, for example, during an animation.
...
+addPostFrameCallback() method. The following code illustrates this on the broken example just given:
```
nate-thegrate, line 501: Generally we want files to have lines of <= 80 characters. ([more info here](https://github.com/dart-lang/site-shared/blob/main/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks))

src/content/testing/common-errors.md
```diff
@@ -475,43 +496,49 @@ framework for every frame, for example, during an animation.
...
-            // Immediately show a dialog upon loading the second screen.
-            Navigator.push(
-              context,
-              PageRouteBuilder(
-                barrierDismissible: true,
-                opaque: false,
-                pageBuilder: (_, anim1, anim2) => const MyDialog(),
-              ),
-            );
-          },
```
nate-thegrate, line 513: Here's a possible solution that doesn't involve stateful widgets (using the Navigator's `context` for the dialog, to ensure it isn't disposed of by the end of the frame):

```dart
onPressed: () {
  final NavigatorState navigator = Navigator.of(context);
  navigator.pushNamed('/second');

  WidgetsBinding.instance.addPostFrameCallback((_) {
    showDialog(
      context: navigator.context,
      builder: (context) => const MyDialog(),
    );
  });
},
```

nate-thegrate (commented):
Thanks for the contribution!

I don't normally review flutter/website PRs, but I did want to share some thoughts regarding the Dart code here :)

src/content/testing/common-errors.md
554richard: That's a great idea, but I don't know how I'd be able to test the functionality before submitting it for review. I can get common-errors.md onto my PC, but how would I serve it to myself for testing, especially with the an interactive DartPad in place?

src/content/testing/common-errors.md
554richard: I split line 501 into two in my fork of the code, but I'm not sure if this change makes its way automatically into the PR, or whether I need to do something to update it.

src/content/testing/common-errors.md
554richard: The difficulty I'm having with implementing this is that my example code doesn't have an onPressed callback anywhere in it, and if I create a button in the Scaffold and put the showDialog in its onPressed:,, the "setState error during build" error disappears without the need of a addPostFrameCallback.

I'm afraid that the broken code I give as an example is crazy - nobody in their right mind would write it. The actual code that caused me the problem was with pluto_grid, and I don't know how to create a sensible example of the error without it - and I can see that pluto_grid should not be included in a page like this. I guess the solution snippet in the current version is a good solution to a problem which was also too specific - which is why I didn't find it helpful.

I suppose I've got into trouble because I'm trying to post a solution to a specific problem in a "common errors" page whose purpose has to be much more general. 

If I can't find a sensible example of broken code - which uses a callback like onPressed or onChanged and causes this particular error without using pluto_grid, I don't think I can make a sensible contribution. What do you think?

sfshaza2:
@nate-thegrate, THANK YOU for jumping in here!!! I've seen the discussion and the updates. I've just kicked off the CI testing to see if it's all happy.

sfshaza2:
Hi, @554richard! According to the CI results, you need to refresh the code excerpts for the site. If you click **Details** on the "test" line, it tells you how to do this. Just run the script locally and commit the changes. Also, it's not 100% clear to me if you and @nate-thegrate have a solution that you are both happy with.

554richard:
I'm afraid my naivety in this business is a problem. I see the test failure and found the line to I need to run and then recommit, but I made all the original changes by forking, editing and PR within the GitHub website, and I don't know how or where to run the script and recommit.

Also after discussion I am not happy with the solution I have presented. I feel that my pull request has weaknesses which make it maybe no better than the original document, and I don't know how to correct them (see my last comment above). I am perfectly happy to continue but I need expert advice.  Should I maybe drop this request and file some kind of request/bug report?

src/content/testing/common-errors.md
nate-thegrate: It looks like this PR is using the [`554richard:patch-1`](https://github.com/554richard/website/tree/patch-1) branch. If you push changes to that branch, they'll show up in this pull request!

src/content/testing/common-errors.md
nate-thegrate: Thanks for sharing your thoughts.

I think you made a good point about the post-frame callback not being necessary here 🙂
I edited the comment above accordingly, thanks!

nate-thegrate:
```
Error: Some code excerpts needed to be updated!
  Run `./dash_site refresh-excerpts` to update.
```

My guess would be that you'd need to get [your branch](https://github.com/554richard/website/tree/patch-1) downloaded to a local machine in order to run that script, but I'm not really familiar with it either 😅

Overall, I think this is a nice contribution—at the very least, it'd be nice to update this file so it uses the `showDialog()` function rather than a page route builder. And in the case of showing a popup on launch, I think it does make sense to do it via post-frame callback, but inside `initState` rather than `build`.

Hopefully we can get this figured out!

554richard:
I think the pop-up has to come from an onChanged or onPressed callback to make any sense. This could set a flag to tell the initState to execute the showDialog. I think this may be a working solution even without the the Postframe callback. I can experiment with that in my pluto_grid code. The problem I have is how to make a "before" example which triggers the "setState called during build" without using pluto_grid, but with an onPressed callback or the like. I guess we could leave the "before" code snippet which is on the page as it is - somebody else can improve it if they get any bright ideas. 

As for the test failure, I'm figuring out how to get my branch onto my local machine - I failed with VSCode, I'm trying now with GitHub Desktop. I guess I'll learn a bit more about how GitHub works ;-)

sfshaza2:
I've just kicked off the build again and have caught up on all the comments. @554richard, would you still like to follow through on this?

goderbauer:
(triage): @554richard do you still have plans to come back to this PR?

sfshaza2:
Thanks for your PR, @554richard, but I'm closing.

PR #11397, opened 2024-11-14
Add caching and service worker docs to the web FAQ

johnpryan:
This adds some information about web caching and service workers to the web FAQ.

cc/ @ditman @kevmoo


src/content/platform-integration/web/faq.md
```diff
@@ -120,18 +120,109 @@ See [Preparing a web app for release][].
 
 Not currently.
 
+### Why doesn't my app update immediately after it's deployed?
+
+You may need to configure the Cache-Control header returned by your web server.
```
sfshaza2, line 125:
```suggestion
You might need to configure the `Cache-Control` header returned by your web server. 
```

"may" is used primarily for legal docs.

src/content/platform-integration/web/faq.md
```diff
@@ -120,18 +120,109 @@ See [Preparing a web app for release][].
 
 Not currently.
 
+### Why doesn't my app update immediately after it's deployed?
+
+You may need to configure the Cache-Control header returned by your web server. 
+For example, if this header is set to 3600, then the browser
+and CDN will cache the asset for 1 hour, and your users may see an out-of-date
```
sfshaza2, line 127:
```suggestion
and CDN will cache the asset for 1 hour, and your users might see an out-of-date
```

src/content/platform-integration/web/faq.md
```diff
@@ -120,18 +120,109 @@ See [Preparing a web app for release][].
...
+version of your app up to 1 hour after you deploy a new version. For
+more information about caching on the web,
+see [this web.dev article][http-cache] on HTTP caching.
```
sfshaza2, line 130:
```suggestion
check out [Prevent unnecessary network requests with the HTTP Cache][http-cache].
```

Nice use of the URL shortcut. :)

src/content/platform-integration/web/faq.md
```diff
@@ -120,18 +120,109 @@ See [Preparing a web app for release][].
...
+
+It is a good idea to be aware of this behavior to avoid breaking users.
```
sfshaza2, line 132:
```suggestion
It is a good idea to be aware of this behavior to avoid an undesirable user experience.
```

?? They aren't actually breaking users.  :)

src/content/platform-integration/web/faq.md
```diff
@@ -120,18 +120,109 @@ See [Preparing a web app for release][].
...
+After you deploy your app, users may see a
```
sfshaza2, line 133:
```suggestion
After you deploy your app, users might use a 
```

src/content/platform-integration/web/faq.md
```diff
@@ -120,18 +120,109 @@ See [Preparing a web app for release][].
...
+version of your app up to 1 hour after you deploy a new version. For
+more information about caching on the web,
+see [this web.dev article][http-cache] on HTTP caching.
+
+It is a good idea to be aware of this behavior to avoid breaking users.
+After you deploy your app, users may see a 
+cached version of your app (cached by the browser or CDN)
+for the duration defined by your cache headers.
+This can lead to users using a version of your app that
+are incompatible with changes that have been deployed to backend services.
```
sfshaza2, line 137:
```suggestion
is incompatible with changes that have been deployed to backend services.
```

src/content/platform-integration/web/faq.md
```diff
@@ -120,18 +120,109 @@ See [Preparing a web app for release][].
...
+
+### How do I bust the web cache after each deployment?
```
sfshaza2, line 139:
```suggestion
### How do I "bust the web cache" after a deployment and force an app download?
```

OK, I was going to reword "bust the web cache" but I googled and it's definitely a thing. But I'm adding a tad of an explanation to the anchor title.

src/content/platform-integration/web/faq.md
```diff
@@ -120,18 +120,109 @@ See [Preparing a web app for release][].
...
+If you wish to defeat these cache headers after each deployment, a common
+technique is to append a build ID of some sort to the links of your static
+resources, or update the filenames themselves.
+(`logo.png` would become `logo.v123.png`.)
```
sfshaza2, line 143:
```suggestion
For example, `logo.png` might become `logo.v123.png`.
```

src/content/platform-integration/web/faq.md
```diff
@@ -120,18 +120,109 @@ See [Preparing a web app for release][].
...
+<script src="flutter_bootstrap.v123.js" async></script>
+```
+
+Flutter does not currently support appending build IDs to resources
+automatically.
+
+### How do I configure my cache headers?
+
+If you are using Firebase Hosting,
+The shared cache (CDN) is invalidated when you deploy a new version of your
```
sfshaza2, line 159:
```suggestion
the shared cache (CDN) is invalidated when you deploy a new version of your
```

src/content/platform-integration/web/faq.md
```diff
@@ -120,18 +120,109 @@ See [Preparing a web app for release][].
...
+app. But you may choose to configure your cache headers as follows,
```
sfshaza2, line 160:
```suggestion
app. But you might choose to configure your cache headers as follows,
```

src/content/platform-integration/web/faq.md
```diff
@@ -120,18 +120,109 @@ See [Preparing a web app for release][].
...
+flutter build web --pwa-strategy=none
+```
+
+If you would like to continue to use a service worker, you can
+[build your own][using-service-workers] or try third-party tools
+such as [Workbox][workbox].
+
+If your service worker is not refreshing,
+configure your CDN and browser cache by setting
+Cache-Control header to a small value such as 0 or 60 seconds.
```
sfshaza2, line 209:
```suggestion
the `Cache-Control` header to a small value such as 0 or 60 seconds.
```

sfshaza2 (approved):
A few little things. I'll go ahead and LGTM, but please fix.  :)

src/content/platform-integration/web/faq.md
johnpryan: I think I'm going to change this to "clear the web cache" and remove the quotes.

ditman:
This is already published!

https://docs.flutter.dev/platform-integration/web/faq#why-doesnt-my-app-update-immediately-after-its-deployed

Thanks @johnpryan!

slavap:
@johnpryan @ditman 
Unfortunately this section https://docs.flutter.dev/platform-integration/web/faq#how-do-i-clear-the-web-cache-after-a-deployment-and-force-an-app-download
does not provide a working solution and only adds **confusion**.
There are three ways to solve this:
1. Properly configure cache headers on hosting - it is not so obvious and needs multiple examples.
2. Use "subfolder version" approach - it is simple and works without any headers https://github.com/flutter/flutter/issues/127459#issuecomment-2378870433
3. Post-processing script for adding ?v=nnn for everything - it is kind of proper implementation of the mentioned section, i.e.
```html
<script src="flutter_bootstrap.js?v=123" async></script>
```
is **naïve**, and has to be like bash script from https://github.com/flutter/flutter/issues/127459#issue-1723105251

johnpryan:
> Unfortunately this section does not provide a working solution and only adds confusion.
Do you think we should just remove it? This section is only necessary if you need to clear the web cache immediately after deploying a new version, which is probably not a requirement for most apps. If you have a link we can point to for better guidance I'd be happy to replace it.

> Properly configure cache headers on hosting - it is not so obvious and needs multiple examples.
What would those additional examples look like? We are providing a Firebase example in the FAQ now.

> Use "subfolder version" approach
 This is an interesting approach - have you had any issues with it? Are you using a routing package like go_router?
 
 > Post-processing script for adding ?v=nnn for everything - it is kind of proper implementation of the mentioned section, i.e. is naïve, and has to be like bash script from https://github.com/flutter/flutter/issues/127459#issue-1723105251
I think it's a commonly used technique, but you're right it's not as simple as we describe in the docs. I'm happy to provide a link to a blog post or article that explains this technique.

slavap:
@johnpryan 
>> Do you think we should just remove it?

Of course not! But this section is very important and must provide a complete solution for Flutter Web's caching problems.
Just search how **many** issues are opened about caching, deferred loading, pubspec version, web worker, ...
Developers are just lost and need a clear detailed explanation.

>> We are providing a Firebase example in the FAQ now.

Firebase is fine, but it has to be more "generic" - i.e. which headers for which files, and then how to implement this **strategy** on different hosting platforms (e.g. Firebase, AWS, Azure, self-hosted Apache, Tomcat, ...). Perhaps the community could help here.

>> This is an interesting approach - have you had any issues with it? Are you using a routing package like go_router?

For me, it works great. But I'm using auto_route and not bothering about # in urls.
The only reported problem was about using it with "path URL strategy", but even for that case there is a workaround https://github.com/flutter/flutter/issues/127459#issuecomment-2394842274
So I believe this solution could be added to the docs, considering it does not require any cache headers config on hosting.

>> not as simple as we describe in the docs. I'm happy to provide a link to a blog post or article that explains this technique.

you may directly add this link to the docs https://github.com/flutter/flutter/issues/127459#issuecomment-2378870433
OR just copy bash script from there to the docs.
The problem is this script is patching compiled js results, which is **scary** for people, and also new versions of Flutter may require adjustments of this script. Ideally, it would be real nice if "flutter build web" just had compile option for adding this "?v=nnn" in ALL proper places during the build.

ditman:
> @ditman not sure I’ve got your question. index.html is placed exactly the same as usual Flutter Web deployment, only the rest of files go to subfolder on the same level as index.html

Ah, so you're updating your index.html to inject assets from `/versionA` to `/versionB` when you deploy? Understood! (I thought you were moving the whole `build/web` under a versioned folder)

slavap:
@ditman The main point is that there is no need to do "heavy" editing of index.html, it is usually just enough to change "base href" as:
```html
<head>
  <base href="web1234/">
```

philitell:
Please support documenting the "subfolder"-approach. This a simple workaround for a major problem that has been around way too long. 
You just need to change base href e.g. by using the "--base-href" argument for flutter build and move the index.html file to the parent folder.

1. Create a build
```
flutter build web --base-href "/my_project_1_1_2/"
```
2. Copy content of build/web to a folder on your server (folder name contains the version number in one or other way)  and move the index.html file to the parent folder. So the final file/folder-structure on your server looks like:
```
my_project/
   index.html
   my_project_1_1_2/
```

For Deployment of e.g. version 1.1.3 you deploy your files to subfolder "my_project_1_1_3" and update index.html in the parent directory.

philitell:
@zs-dima What do you consider a "good" workaround for this issue?

PR #11408, opened 2024-11-19
Remove unused alt config from widget catalog yaml file

parlough:
An appropriate `alt` value is specified in `catalog-page.md`. The important part of each card is the description which a separate alt text would mostly duplicate.


sfshaza2 (approved):
lgtm

PR #11424, opened 2024-11-21
Fix Web support page highlighting in left nav

wrunk:
The page "Web support for Flutter" wasn't showing an active highlight in the left nav (see screenshots). Looks like there was an extraneous trailing slash.

VERY minor change

Web support for Flutter: No active highlight:
Screenshot 2024-11-21 at 10 08 16

Other pages have:
Screenshot 2024-11-21 at 10 09 00

### DIFF
--- a/src/_data/sidenav.yml
+++ b/src/_data/sidenav.yml
@@ -513,6 +513,7 @@
       children:
         - title: Web support in Flutter
           permalink: /platform-integration/web
+          match-page-url-exactly: true
         - title: Add web as build target
           permalink: /platform-integration/web/install-web
         - title: Build a web app
--- a/src/_data/sidenav.yml
+++ b/src/_data/sidenav.yml
@@ -512,7 +512,7 @@
       permalink: /platform-integration/web
       children:
         - title: Web support in Flutter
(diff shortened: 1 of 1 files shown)


wrunk:
Apologies, my fix was insufficient. For a third level nav item/nav link to get tagged/classed as active, it needs to have 3 path elements (currently only has 2: /platform-integration/web). Options:
- Move the current platform-integration/web/index.md to something else like platform-integration/web/support.md then add in a TOC-like index.md
- Change the permalink for the platform-integration/web drop down from /platform-integration/web -> /platform-integration/web/support but this would likely require a redirect for platform-integration/web 
- Something else?

If you let me know your preference I'm happy to update.

sfshaza2:
@parlough, would you mind owning this one? It looks like some files weren't placed correctly and it's affected the side nav.

parlough (approved):
Thanks for noticing and fixing this!

Due to the current implementation of our sidenav, when an entry's link matches its parent, it needs the `match-page-url-exactly: true` option set as well.

I've gone ahead and pushed that addition. Thanks again :D

PR #11427, opened 2024-11-22
Updated Doc: changes Center to Container

daveragos:
it was wrongly written as a Center widget when in the code its Container


sfshaza2 (approved):
lgtm

PR #11430, opened 2024-11-22
Fix typo in concepts.md: 'principals' to 'principles'

itsJimothy:
Change 'principals' to the contextually correct 'principles' on the App Architecture / Architecture Concepts page.


parlough (approved):
Nice catch. Thanks for fixing this!

PR #11445, opened 2024-11-27
Update required Node version and npm deps

parlough:
After this PR lands, you might need to update your version of Node to the latest LTS or newer.

Please let me know if you have any other issues running the site after this lands. Thanks!


sfshaza2 (approved):
lgtm. And thanks for the heads up about Node!

PR #11453, opened 2024-12-02
Remove the 'Photo filter carousel' recipe

antfitch:

Remove a recipe that references non-Flutter code.

Fixes https://github.com/flutter/website/issues/8204

n/a

## Presubmit checklist


sfshaza2 (commented):
This PR looks great except for a couple things:

- We don't like broken links. So, after this PR lands, trying to access the page gives you a 404, as you can see in the staged version:  https://flutter-docs-prod--pr11453-fw-8204-remove-recipe-ygynfeuc.web.app/cookbook/effects/photo-filter-carousel
To address this, add a redirect to the firebase.json file at the top of the repo. You can redirect to the cookbook index.
- You shouldn't need to delete this from the index file. It's my understanding that this file is autogenerated. Right, @parlough? Or is that no longer the case?

Thanks, @antfitch!

parlough:
> You shouldn't need to delete this from the index file. It's my understanding that this file is autogenerated. Right, @parlough? Or is that no longer the case?

This index file is still manually built, I believe maybe because the automatic order based on name isn't exactly what we wanted.

PR #11455, opened 2024-12-03
Hide banner for now

parlough:
The included content happened a while ago. We can reenable the banner when needed.


sfshaza2 (approved):
sg. I'm still used to the bug where having no banner caused issues. lgtm

PR #11459, opened 2024-12-03
Don't require match-url-exactly in sidenav for active page highlighting

parlough:
No longer require `match-url-exactly: true` to be specified if an entry has the same permalink as the collapsible section it is a child of.


sfshaza2 (approved):
Thanks for fixing this! lgtm

PR #11460, opened 2024-12-03
Clean up duplicated tab panes from test-drive page

parlough:
Also clean up some formatting and copy from the includes.


sfshaza2 (approved):
lgtm

PR #11468, opened 2024-12-05
Design pattern index page

### LINKED ISSUE #11374: ☂ Land updated app architecture guidance
We're in the process of creating new code examples and docs covering application architecture (e.g. state management, networking, etc.). This issue represents the work of getting those docs reviewed and landed:

* Land the initial, high-level documentation for content. 
  * #11300
* Update navigation to include a link to the above documentation. 
  * #11300
* Land Compass app tutorial. 
  * #11414 
* Land design recipes.
* Update design recipe index page with a card for each recipe. The card should include 1-2 sentences about how the design pattern relates to the bigger picture of the architecture docs.
* Move code snippets for case-study from flutter/samples to website.
  * #11441 
* Add survey to architecture pages
  * #11452  

Design recipes:
* Optimistic state - #11394 
* Persistent storage part 1: Key-value store - #11410
* Persistent storage part 2: SQL - #11410
* Offline First support: #11425
* Command pattern: #11444
* Result pattern: #11444

ericwindmill:

This PR adds an 'index' page for the Design Pattern cookbook recipes under App Architecture in the side nav.
* Adds the page
* Adds a `design-pattern.yml` file with 
* Adds an "Expansion Panel List" component (html + css)

**This is still very much a work in progress.** Before merging:
- [x] Land #11444 + add them to the data file, and create icons for their rows
- [x] Update the `design-pattern.yml` with real data
- [x] Create the remaining needing icons (right now all four list items use the same icon)
- [x] Replace 'LOREM IPSUM, BROTHER' 😅 

* This will resolve #11374


ericwindmill:
@parlough I've made this PR before it's complete because I'm hoping you can review the SCSS updates I made. Also, I'm generally curious how you feel about adding a component like this. I think something like this will be useful for FWE.

src/_data/cookbook/design_patterns.yml
```diff
@@ -0,0 +1,74 @@
...
+  path: /cookbook/architecture/key-value-data
+  img: /assets/images/docs/app-architecture/design-patterns/kv-store-icon.svg
+  body: |
+    Most Flutter applications, no matter how small or big they are, 
+    require storing data on the user’s device at some point, 
+    such as API keys, user preferences or data that should be available offline.
+    
+    In this recipe, you will learn how to integrate persistent storage for key-value
+    data in a Flutter application by implementing a switchable light-mode and dark-mode theme 
+    in an application and storing the users preference.
```
ericwindmill, line 16: In some SSGs, there are ways to extract the data of a page during the build process and use that data in templates. I assume its possible in 11ty, but I need to learn about how 11ty works before I can look into that. WDYT?

src/content/assets/images/docs/app-architecture/design-patterns/sql-icon.svg
parlough, line 1: Seems like there are some artifacts to the left of the purple stack.

src/_sass/components/_expansion-list.scss
```diff
@@ -0,0 +1,143 @@
...
+      margin-bottom: 3rem;
+    }
+  }
+
+  :last-child {
+    .expansion-panel-body {
+      border-bottom: none;
+     }
+  }
+}
```
parlough, line 144:
```suggestion
}

```

diagrams/app-architecture/architecture-docs.fig
ericwindmill, line 1: Updated to contain the icon SVG files.

src/_data/cookbook/design_patterns.yml
parlough: Can you explain a bit more what data you'd want to access and where you want to use it? Definitely sounds possible :)

ericwindmill: Rather than having to write a yml file with the data, you could access the frontmatter and content of any given page during the build process by its path. So in the the `design-patterns.md` file you could write something like:

```
assign pages = pageNodes.where(path.contains('/cookbook/architecture')
for page in pages:
  var intro = page.content.takeChars(500)
  render expansionPanel(title: paage.frontmatter.title, body: intro, etc...)
```

This assumption that its _possible_ is based on my experience (long ago) with Gatsby. Gatsby is, for better or worse, extremely feature rich and not light-weight at all, so as I type this I'm realizing that this may be an unrealistic expectation any given SSG.

In Gatsby you can configure multiple build steps, so the first step can take all the data from your markdown files and put it into an object that's available globally in your templates. Given that the data from these yml files are available in the templates, I'm wondering if its possible to access the md content as well if we configure 11ty in some way.

ericwindmill: I've watched some 11ty videos this morning, and I think this is possible using the tags/collections feature. I don't want to implement it in this PR, but that feature might allow us to clean up the website code and content in the future. I'm still learning about 11ty and learning the implementation details of our website, so take this all with several grains of salt.

src/_data/cookbook/design_patterns.yml
```diff
@@ -0,0 +1,120 @@
+- title: "Persistent storage architecture: Key-value data"
+  tags:
+    - data
+    - shared-preferences
+    - dark mode
```
ericwindmill, line 5: `tags` has meaning in 11ty, and shouldn't be used in this way.

src/_data/cookbook/design_patterns.yml
ericwindmill: Nevermind, its really straight-forward, I implemented it in this PR.

sfshaza2 (approved):
If @parlough is happy, then I'm happy. And it seems like he approves (based on the thumbs up and heart emojis). lgtm, but I will let you land, when ready.

parlough (approved):
Thanks @ericwindmill! This is super cool. I pushed a few minor css changes, particularly to improve the focus outline. Most changes have a comment with more info.

I do wonder if these should just be in a `/app-architecture/design-patterns/` directory though and not called recipes?

ericwindmill:
> I do wonder if these should just be in a `/app-architecture/design-patterns/` directory though and not called recipes?

I agree, and I moved them.

PR #11487, opened 2024-12-12
Typo fix in restore-state-ios.md

elitree:
Small typo fix.


parlough (approved):
Thanks for noticing and fixing this!

PR #11494, opened 2024-12-13
Update DevTools release notes for cherry pick release 2.40.2

kenzieschmoll:
Some notes in the 2.41.0 release needed to be moved to 2.40.2 and some could be deleted since they fixed bugs that we did not actually release since the cherry pick release landed.


sfshaza2 (approved):
lgtm

PR #11500, opened 2024-12-16
Increase required version of node and pnpm

parlough:
If you installed your version of `node` before November, you might need to update to the latest LTS or newer after this lands.

So the site now requires:

- Flutter 3.27 (Dart 3.6) or later
- Node 22.11 or later
- pnpm 9.12.3 or later

The dependency upgrades in this PR should also resolve https://github.com/flutter/website/security/dependabot/109


sfshaza2 (approved):
lgtm

PR #11507, opened 2024-12-17
Adding updated info about Impeller and a link to the "can I use" page

sfshaza2:
@chinmaygarde, can you take a look? I've added the "Can I use Impeller" page. The website impeller page needs further work, but this is a start...


mariamhas (commented):
Overall this looks good to me.

For android, I think we should keep the previous text about fallback support.

"On devices that don't support Vulkan, Impeller will fallback to the the legacy OpenGL renderer. No action on your part is necessary for this fallback behavior. Consider trying Impeller on Android before it becomes the default on stable, you can explicitly opt into it."

@zanderso should be added as a reviewer

jonahwilliams (approved):
LGTM

I think the change seems good. I agree with @mariamhas that we should keep the line about the fallback, just to answer the questions folks may have about why their particular device isn't using Impeller

sfshaza2:
Thanks so much, @mariamhas and @jonahwilliams! Would you like to weigh in, @zanderso?

zanderso (approved):

parlough (approved):

sfshaza2:
I spoke offline to @mariamhas and she said that @jonahwilliams's approval is enough to land. So, I'm landing.  :)

PR #11520, opened 2024-12-19
Update shader compilation jank guide

### LINKED ISSUE #6415: Explain benefit of shader compilation warmup on 'Reduce shader compilation jank on mobile' page
### Page URL

https://flutter.dev/docs/perf/rendering/shader/
https://docs.flutter.dev/perf/shader

### Page source

https://github.com/flutter/website/tree/main/src/docs/perf/rendering/shader.md

### Describe the problem

We likely need to update our guidance for when to use / what benefits users will get from shader compilation warmup. Among other things, shader warmup done on higher end phones may not be applicable to lower end phones. We'd expect this to be more of an issue for Android devices (in contrast to iOS ones).

### Expected fix

_No response_

### Additional context

cc @zanderso

### LINKED ISSUE #6504: Update `flutter drive` to `flutter test` on 'Reduce shader compilation jank on mobile' page
# Page Url
https://docs.flutter.dev/perf/rendering/shader/

# Page Source
https://github.com/flutter/website/tree/main/src/perf/rendering/shader.md

## Describe the problem

The current docs (https://docs.flutter.dev/perf/rendering/shader) has a method to do warmup using integration tests, but this uses the older `flutter driver` way of running integration tests instead of the newer and now standard `flutter test` way.

## Expected Fix

Migrate these integration tests to use new `flutter test`

antfitch:
The shader compilation jank guide was out of date and needed a few updates:

* Impeller is now the default graphics renderer for mobile devices. Updated wording to shift the focus around how we talk about Impeller and Skia.
* Removed integration test steps.
* Added a list of important considerations to capture potential issues that might come up when using cache-sksl.
* Completed a light refactor (guide's a bit old and needed a fresh up).

---

Fixes #6504
Closes #6415


sfshaza2 (approved):
lgtm
Thanks, @antfitch!

src/content/perf/shader.md
```diff
@@ -69,29 +40,39 @@ The following screenshot shows an example timeline tracing.
...
+
+If you're not using Flutter's Impeller graphics renderer and you're creating an
+iOS app, you can minimize shader compilation jank by bundling precompiled
+shaders with your iOS app. To learn more, see the iOS app development
+documentation.
 
-On iOS, "first run" means that the user might see
-jank when an animation first occurs every time
-the user opens the app from scratch.
+If you're not using Flutter's Impeller graphics renderer and you're creating an
```
jonahwilliams, line 50: We don't actually recommend/support using the cached sksl on android - we found that the shaders generated weren't compatible across different devices so its not safe to do. I don't know where we documented this

jonahwilliams: I would vote for removing all documentation on SkSL warmup, as we plan to remove all of the tooling for it in the next quarter or so.

src/content/perf/shader.md
```diff
@@ -69,29 +40,39 @@ The following screenshot shows an example timeline tracing.
...
+Android app.
+
+Important considerations:
+
+* On Android, precompiled shaders are device or GPU-specific.
+* The precompiled shaders for Android might only work with a subset of devices.
+* Precompiled shaders could make jank worse or produce errors on some
+  types of Android devices while making jank better on others.
+* With `cache-sksl`, you can precompile shaders for one specific class of
+  Android device at a time.
```
parlough, line 68: If keeping this section, update "Android device" to "Android-powered device". Not super important, but the style guide does call it out: https://developers.google.com/style/word-list#android-powered

src/content/perf/shader.md
```diff
@@ -6,61 +6,32 @@ description: What is shader jank and how to minimize it.
...
-which is the default renderer on iOS.
-You can preview Impeller on Android by passing
-`--enable-impeller` to `flutter run.
+This document describes what shader compilation jank is and how to minimize it.
+
+Shader compilation jank is the annoying stutter you sometimes see in
+animations, especially the first time you run an app. This happens because the
+app has to compile shaders while it's running.
+
+If you are using [Impeller][], Flutter's default mobile renderer, you shouldn't
```
parlough, line 15:
```suggestion
If your app is using [Impeller][], Flutter's default mobile renderer, you shouldn't
```

parlough (approved):
Nice cleanup for some long-standing issues, thanks!!

Looks good to me, but do consider making more adjustments in relation to Jonah's comments. Feel free to do it in a follow-up PR if you prefer to land this first :)

---

Another thing to note, but doesn't need changes for this PR to land, do try to use semantic line breaks for text in Markdown.

It helps make change review and history tracking easier and cleaner. I find it also helps me notice and form the logic of sentences as I write them. Eventually it becomes natural.

https://sembr.org/ has some guidelines to loosely follow and https://rhodesmill.org/brandon/2012/one-sentence-per-line/ has some discussion and history about them too :D

src/content/perf/shader.md
antfitch: I was wondering about this. I'll remove the documentation since it appears to be obsolete.

jonahwilliams (approved):
LGTM

sfshaza2 (approved):
It looks to me like you covered everything including a redirect. lgtm

src/content/perf/shader.md
sfshaza2: Even betterer:
If your app uses [Impeller][], Flutter's default mobile renderer,
you shouldn't

PR #11523, opened 2024-12-19
Reference `EnableImpeller` not `DisableImpeller`.

matanlurey:
Closes https://github.com/flutter/flutter/issues/160595.

The latter never existed.

Side-note: We should never have _negative_ flags (they are confusing to parse).


jonahwilliams (approved):
LGTM

sfshaza2 (approved):
lgtm

PR #11526, opened 2024-12-20
Update set-up-app-links.md (add missing line)

Reza-Babakhani:
`<meta-data android:name="flutter_deeplinking_enabled" android:value="true" />`

This one line of code took me 3 days.
Without this, deep linking doesn't work.

### DIFF
--- a/src/content/cookbook/navigation/set-up-app-links.md
+++ b/src/content/cookbook/navigation/set-up-app-links.md
@@ -88,6 +88,7 @@ It provides a simple API to handle complex routing scenarios.
     Replace `example.com` with your own web domain.
 
     ```xml
+    <meta-data android:name="flutter_deeplinking_enabled" android:value="true" />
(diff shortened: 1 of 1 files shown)


parlough:
Thanks for the PR to fix this and sorry about the confusion you met. I'll defer to others to review:

@chunhtai or @hannah-hyj would you mind taking a look at this PR? Thank you!

hannah-hyj:
@Reza-Babakhani  Sorry for the confusion you encountered and thank you for filing this.

Starting from Flutter version 3.27, the deep link flag is set to `true` by default, so this line is no longer needed.
However, for Flutter versions earlier than 3.27, this line is still necessary.

I think it makes sense to add back this line, but include it as a note:
"If you are using a Flutter version earlier than 3.27, add the following metadata tag:
<meta-data android:name="flutter_deeplinking_enabled" android:value="true" />"

PR #11588, opened 2025-01-13
Adding Amanda to the "blessed" codeowners.

sfshaza2:
It is time.


parlough (approved):
I also had Amanda added to https://github.com/orgs/flutter/teams/website-reviewers, so this PR isn't fully necessary, but helps as it will result in her getting pinged if no specific code owner category was added.

PR #11620, opened 2025-01-24
[SwiftPM] Update instructions to increase OS version

loic-sharma:

Tell users to build their apps using the Flutter tool after they update their app's iOS/macOS minimum deployments.

This is currently necessary due to this bug: https://github.com/flutter/flutter/issues/162196

## Presubmit checklist

### DIFF
--- a/src/content/packages-and-plugins/swift-package-manager/for-app-developers.md
+++ b/src/content/packages-and-plugins/swift-package-manager/for-app-developers.md
@@ -151,8 +151,24 @@ the project, you might get an error like this:
 Target Integrity (Xcode): The package product 'plugin_name_ios' requires minimum platform version 14.0 for the iOS platform, but this target supports 12.0
 ```
 
-To use the plugin, increase the **Minimum Deployments** of your app's target.
+To use the plugin:
 
-{% render docs/captioned-image.liquid,
-image:"development/packages-and-plugins/swift-package-manager/minimum-deployments.png",
-caption:"The target's **Minimum Deployments** setting" %}
+1. Increase the **Minimum Deployments** of your app's target.
+
+   {% render docs/captioned-image.liquid,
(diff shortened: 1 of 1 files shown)


vashworth (approved):
LGTM

src/content/packages-and-plugins/swift-package-manager/for-app-developers.md
```diff
@@ -151,8 +151,27 @@ the project, you might get an error like this:
...
+
+   {% render docs/captioned-image.liquid,
+   image:"development/packages-and-plugins/swift-package-manager/minimum-deployments.png",
+   caption:"The target's **Minimum Deployments** setting" %}
+
+1. If you updated your iOS app's **Minimum Deployments**,
+   regenerate the iOS project's configuration files:
+
+   ```sh
+   flutter build ios --config-only
```
loic-sharma, line 169: FYI this can be simplified to just `flutter build ios`, but that does more stuffs and takes longer to run. Let me know if I should favor simplicity here!

sfshaza2 (approved):
lgtm

sfshaza2:
You know, I don't know why this didn't occur to me before, but why not just use "enable/disable" rather than "turn on/turn off"? It struck me as I reviewed the page when it says in the warning at the top "Swift Package Manager support is off by default." That strikes me as a sloppy sentence. It's rather idiomatic and I wondered off how? Off where? I would prefer "disabled by default." I mean, "off" is a fairly loaded word. Another way to fix it would be "turned off by default". But "disabled/enabled" is pithier.  ;)

loic-sharma:
The original guide used enable/disable, but I received feedback that turn on/off was preferable language. I personally don't have a strong preference either ways, I'm happy to go with whichever folks prefer 😄

PR #11629, opened 2025-01-28
Add documentation for the new Flutter inspector

elliette:
Updates the copy and images of the Flutter DevTools inspector documentation for the new inspector. 

Moves the legacy inspector documentation to a new page, `legacy-inspector`.


src/content/tools/devtools/legacy-inspector.md
```diff
@@ -0,0 +1,573 @@
+---
+title: Use the legacy Flutter inspector
+description: Learn how to use the legacy Flutter inspector to explore a Flutter app's widget tree.
+---
+
+<?code-excerpt path-base="visual_debugging/"?>
+
+:::note
```
sfshaza2, line 8: I'm wondering if this little note, at the top, is enough warning. Let's say that someone googles something and lands _inside_ this page. They wouldn't even see that note at the top. @parlough, is there a watermark or something we can do to make this more obvious? (I'd love a subtle-but-legible watermark.)

Issue #3632, opened 2019-06-21
Add dark mode to the official documentation


shunjid:
It's a **feature request**. I don't know anyone felt the need of it or, not but I think it would be nicer if @flutter team adds a 'Dark Mode' in the official documentation.

Thank you.
😊

![flutter dev_docs(iPad Pro)](https://user-images.githubusercontent.com/29749035/59942463-884dc180-9481-11e9-993a-8a1abdf71bbd.png)

legalcodes:
+1 I'd support this!

shunjid:
Super **excited**. Can't wait anymore :heart_eyes:

tillchen:
Any updates :D?

shunjid:
Hi @sfshaza2 
Any updates over this?

sfshaza2:
I'm afraid not, @shunjid. It is just not a priority at the moment. We have no support for website infrastructure right now.

kissu:
I guess that dark reader is still the best answer regarding that issue: https://darkreader.org/

toto6038:
It's 2023. I think this issue deserves priority.

sfshaza2:
Well, do have more infrastructure support right now, though I still suspect this will have to wait until after Flutter Forward and probably also after Google I/O.

parlough:
We should likely finish https://github.com/flutter/website/issues/8030 first as well, since the migration has to happen eventually anyway and will make the implementation of dark mode easier.

sfshaza2:
sg!

Ali-Oliaee:
> It's 2023. I think this issue deserves priority.

well, we're going to 2025 =]

apravint:
i see dark mode now in the website.. :)

PR #11639, opened 2025-01-29
More accessible alert/aside style

parlough:
Only colors the title and a left border instead of the background and the text. This is more likely to be visually accessible and is the style many doc websites are adapting.

Also removes dependence on bootstrap-provided styles. Contributes to https://github.com/flutter/website/issues/8030 and https://github.com/flutter/website/issues/3632


sfshaza2 (approved):
Exciting! lgtm

PR #11660, opened 2025-02-01
Ensure body is scrollable if sidenav is dismissed

parlough:
Fixes the body not being scrollable if you have a narrow window, open the menu, then keep the menu open and make your window wide (causing it to close).


sfshaza2 (approved):
lgtm

PR #11675, opened 2025-02-05
[Draft] UI thread going away on iOS and Android

sfshaza2:
@jonahwilliams, can you advise? Is the UI thread only going away on mobile? But not desktop, for example? These two pages (with code) need updating and I'm not quite sure how far to go.

### DIFF
--- a/src/content/perf/ui-performance.md
+++ b/src/content/perf/ui-performance.md
@@ -38,6 +38,12 @@ steps to take, and tools that can help.
 
 ## Diagnosing performance problems
 
+:::note                              
+Starting in 3.29, Flutter on Android and iOS execute Dart code
+on the application’s main thread,
+and there is no longer a separate UI thread. 
+:::
+
 To diagnose an app with performance problems, you'll enable
 the performance overlay to look at the UI and raster threads.
 Before you begin, make sure that you're running in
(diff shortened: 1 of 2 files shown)


jonahwilliams:
For now, only Android and iOS has this behavior - desktop does not. We plan to change desktop/embedder in the future, but no set date there. I don't think we should necessarily pre-announce anything yet

src/content/perf/ui-performance.md
```diff
@@ -38,6 +38,12 @@ steps to take, and tools that can help.
 
 ## Diagnosing performance problems
 
+:::note                              
+Starting in 3.29, Flutter on Android and iOS execute Dart code
```
jonahwilliams, line 42: We probably _don't_ need a note here, because even though the threads will be changed, the tooling will still communicate that these work is happening on the "UI thread", the devtools traces will be the same, et cetera. Its just that this thread happens to be the same as the platform main thread.

src/content/platform-integration/platform-channels.md
```diff
@@ -8,14 +8,17 @@ description: Learn how to write custom platform-specific code in your app.
...
-see [using packages][].
+through existing packages. For more information,
+visit [using packages][].
 
 [using packages]: /packages-and-plugins/using-packages
 
 :::note
-The information in this page is valid for most platforms,
-but platform-specific code for the web generally uses
+Starting in 3.29, Flutter on Android and iOS execute Dart code
```
jonahwilliams, line 17: Likewise, for this section - changing the threads does not change how platform channels behaves. It will enable ffi interop in the future, but that is not done yet.

sfshaza2:
Thanks, @jonahwilliams! Closing this PR.

PR #11686, opened 2025-02-06
Make sure install help button styles take priority

parlough:
This slightly improves the styling of "Install help" buttons on pages like https://docs.flutter.dev/get-started/install/windows/mobile, placing them more inline with the titles.

We should likely consider replacing these, but that can be revisited with later install work.


sfshaza2 (approved):
lgtm

PR #11690, opened 2025-02-07
Update "Obfuscate your app" topic to be more platform agnostic.

antfitch:

This topic only had Android code snippets. To make it easier for devs creating apps for other platforms, I reformatted the commands to be platform agnostic.

Notes:
- We had a few limitations outside of the Limitations section, so moved them there.
- Noticed we had a warning inside of the Limitations section, so updated the heading to include warnings.
- Updated some of the variable names in the commands and created a bulleted list below each command with the variable names and what they need to be replaced with.
- 'Read an obfuscated name' contents looked like they needed a revision, so took care of that. Essentially it looks like that section is about generating an obfuscation map that contains keys and values. It's still not quite right, but I hope it's closer to where things are at now. I'm also not sure if we still need the `--extra-gen-snapshot-options=` before the `--save-obfuscation-map=` command.

## Presubmit checklist

### DIFF
--- a/src/content/deployment/obfuscate.md
+++ b/src/content/deployment/obfuscate.md
@@ -14,23 +14,32 @@ compiled Dart code, replacing each symbol with
 another symbol, making it difficult for an attacker
 to reverse engineer your proprietary app.
 
+[Code obfuscation]: https://en.wikipedia.org/wiki/Obfuscation_(software)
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
lgtm

PR #11705, opened 2025-02-11
Allow any attributes from markdown-attrs

parlough:
Such as `width` when including an image, such as:

```
![Rename a Flutter flavor](/assets/images/docs/flavors/flavors-ios-app-names.png){:width="50%"}
```


sfshaza2 (approved):
lgtm

PR #11725, opened 2025-02-16
Fix count variable assignment bug in code example

andreynazarov3333:
There’s a bug in the increment() method.
### Issue in increment()
`var count = this.count;`
	•	Here, count (a local variable) gets a copy of this.count.
	•	This means modifying count++ only affects the local variable, not this.count.

### DIFF
--- a/src/content/get-started/fundamentals/state-management.md
+++ b/src/content/get-started/fundamentals/state-management.md
@@ -474,13 +474,14 @@ class CounterViewModel extends ChangeNotifier {
   }
 
   Future<void> increment() async {
-    var count = this.count;
-    if (count == null) {
+    final currentCount = count;
+    if (currentCount == null) {
       throw('Not initialized');
     }
     try {
-      await model.updateCountOnServer(count + 1);
-      count++;
+      final incrementedCount = currentCount + 1;
+      await model.updateCountOnServer(incrementedCount);
+      count = incrementedCount;
     } catch(e) {
       errorMessage = 'Count not update count';
     }


src/content/get-started/fundamentals/state-management.md
```diff
@@ -474,13 +474,12 @@ class CounterViewModel extends ChangeNotifier {
   }
 
   Future<void> increment() async {
-    var count = this.count;
```
parlough, line 477: Creating a new local variable here allows for it to be promoted to a non-nullable type, avoiding the non-null assertion operators.

What do you think about keeping the local variable, just with a different name for clarity, and instead fixing the assignment?

Something like this:

```dart
final currentCount = count;
if (currentCount == null) {
  throw 'Not initialized';
}

try {
  final incrementedCount = currentCount + 1;
  await model.updateCountOnServer(incrementedCount);
  count = incrementedCount;
} catch (e) {
  errorMessage = 'Count not update count';
}

notifyListeners();
```

parlough (commented):
Thanks for noticing this and proposing a fix!

I'm a bit hesitant to have non-null assertions in example snippets though as it can be a risky practice. What do you think about adding one or two new variables to make this more clear and avoiding the potential mistake:

src/content/get-started/fundamentals/state-management.md
andreynazarov3333: Oh, I just noticed that since we've already checked that count is not null at the beginning of the method, we don't need to use the force unwrap operator ( !) later in the method. 
But I also agree that maybe more verbose version is easier to understand for beginners so I replaced my initial commit with yours proposal.

parlough (approved):
Thanks for making those adjustments and again for fixing this in the first place! Looks good to me :)

Issue #11672, opened 2025-02-04
Auto-detect the user's host machine to prevent extra clicks to install the SDK


johnpryan:
Right now, when a user installs Flutter, they need to click through multiple steps before they can begin reading instructions:
1. "Choose your platform to get started" (https://docs.flutter.dev/get-started/install)
2. "Choose your first type of app" (https://docs.flutter.dev/get-started/install/macos)
3. "Start building Flutter iOS Apps on macOS" (https://docs.flutter.dev/get-started/install/macos/mobile-ios)

Instead, we could reorganize the "Set up Flutter" page to immediately show content, but select the correct "tab" for the user's host machine, detected by the browser using the `user-agent` header. We could also store this value in localStorage so that whenever that developer returns to docs.flutter.dev, they get their previously selected host device (if they are reading the docs on a machine that isn't their host device, such as a mobile phone)

This would also let us reorganize the TOC to be something like this:

* Set up Flutter
  * Windows
  * macOS
  * Linux
  * ChromeOS
* Run Flutter on a target device
  * iOS
  * Android
  * web
  * macOS (Desktop)
  * Windows (Desktop)
  * Linux (Desktop)

In fact, we could even consider dropping the subsections from the TOC, since those will be accessible via the tabs.

Here's an example in the GCP docs where they select a tab based on the browser's user-agent: https://cloud.google.com/sdk/docs/install

sfshaza2:
cc @parlough @antfitch

antfitch:
I think splitting the current flow up to look like this is a fine idea.

PR #11732, opened 2025-02-19
Show set up entries in sidenav when child page is open

parlough:
An improvement for the flow of someone navigating between the set up pages in the sidenav. Before this PR, they didn't show up at all in the sidenav, but now they do if any of the set up pages are open.

Contributes to https://github.com/flutter/website/issues/11672 as an improvement while I work on a longer term, automated solution.


johnpryan (approved):

MaryaBelanger (approved):
Works as described, and a good addition too!

PR #11745, opened 2025-02-26
Revert "Add iOS 18.4 beta warnings"

loic-sharma:
Reverts flutter/website#11740


jmagman (changes requested):
I think we should keep the warning since it's impacting anyone on iOS 18.4 beta 1, which is the latest

PR #11752, opened 2025-02-27
Fix broken link for iOS 18.4 warning

loic-sharma:


parlough (approved):
Thanks!

Unfortunately GitHub is excluded from the linkchecker, so this wasn't caught.

Issue #11735, opened 2025-02-22
Red on black is unreadable


admarwick:
### Page URL

https://docs.flutter.dev/release/breaking-changes/android-java-gradle-migration-guide/

### Page source

https://github.com/flutter/website/tree/main/src/content/release/breaking-changes/android-java-gradle-migration-guide.md

### Describe the problem

Screenshots were made with a red-on-black color scheme. This is very difficult to read, to put it mildly.

### Expected fix

Make new screen shots with a black-on-white color scheme for maximum readability, and not just for this page!

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

PR #11769, opened 2025-03-04
Replacing screenshot of error with text

sfshaza2:
Not only is the red-on-black error hard to read, but it's bad for SEO.

Fixes https://github.com/flutter/website/issues/11735


parlough (approved):
👍

Issue #11789, opened 2025-03-07
Remove "Test drive" page


parlough:
It's a remnant of the old get started experience.

PR #11777, opened 2025-03-05
Create a reference for creating a new, empty app

parlough:
Resolves https://github.com/flutter/website/issues/11332

PR #11790, opened 2025-03-07
Remove 'Test drive' page and references to it

parlough:
Now that we have the [Create a new Flutter app reference](https://docs.flutter.dev/reference/create-new-app) landed in https://github.com/flutter/website/pull/11777, we can retire the "Test drive" page that was part of a previous get started flow.

Resolves https://github.com/flutter/website/issues/11789


sfshaza2 (approved):
lgtm!

PR #11793, opened 2025-03-08
Update os-settings.md for windows set mirror env

CaiJingLong:
On Windows devices, the current document requires manual modification of the current directory. 
After modification, it can achieve the same effect as using pwd on Linux to read the current working directory.


sfshaza2 (commented):
@jwill, can you weigh in on this?

jwill:
Looks fairly reasonable, don't have a windows machine to 100% verify but LGTM.

Issue #10535, opened 2024-05-08
Fix list structures


atsansone:
From https://buganizer.corp.google.com/issues/337264374

### Observed Results

Navigating through the left menu specifically through the lists inside drop downs, I noticed how they are verbalized by SR.

* [Screencast VoiceOver](http://screencast/cast/NjMxNjcyODEzNTk3NDkxMnxkZDQ2MzU0Yy1iYw)
* [Screencast ChromeVox](http://screencast/cast/NTMwNDE3MTU4NTYwMTUzNnw5MWMzNmJhYS02OA)

### Expected Results

Ensure that the lists from drop downs should have a correct structure in this way they will be announced better in screen readers.

`<ul>` and `<ol>` must only directly contain `<li>`, `<script>` or `<template>` elements

### User Impact

Screen reader and Cognitive users are affected because they do not have clearly how many items are inside every list.

### Steps to Reproduce

1. Enabled VoiceOver using <kbd>Command</kbd> + <kbd>F5</kbd> also ChromeVox using <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>Z</kbd>
1. Open the following pages:
   * https://docs.flutter.dev/
   * https://docs.flutter.dev/get-started/install
1. Continue the linear navigation using the <kbd>Tab</kbd> key and reach the left menu.
1. Expanded the drop downs and listen to how the displayed list are verbalized by SR.

### GAR Assessment

[Flutter Docs](https://a11ytracker.corp.google.com/products/c2fdcf18-53d2-475d-b8b0-e5dedf8cb900?e=ProductLabels)

### GAR Details

[1.16. (c) Without Visible Screen (With Screen Reader)](go/GAR-web#run-through-screen-reader)

### Environment Details

Machine: MAC/ChromeBook
Browser: Chrome Version 120.0.6099.272 (Official Build) (64-bit)
Screen reader: VoiceOver / ChromeVox.
Bug filling time: 15 mins.

PR #11802, opened 2025-03-11
Only use `<li>` as children of `<ul>` elements in sidebar

parlough:
Uses a hidden `<li>` for divider elements only used for visual distinction. Matches what is done at the top level of the sidebar as well.

Fixes https://github.com/flutter/website/issues/10535


sfshaza2 (approved):
lgtm

PR #11813, opened 2025-03-13
Remove unnecessary configuration of tabindex

parlough:
This was called out by the lighthouse accessibility report.


sfshaza2 (approved):
lgtm

Issue #11828, opened 2024-03-01
Document that older Android versions have unmitigatable security risks on our supported platforms page


Hixie:
Subcomponent of https://github.com/flutter/flutter/issues/63559.

We should update our public documentation to mention that versions of Android older than Android 9 (aka Pie, SDK level 28) leave applications vulnerable to the StrandHogg attack, and then link to <https://developer.android.com/privacy-and-security/risks/strandhogg>.

antfitch:
Transferred this to the website.

ash2moon:
@antfitch I added this PR https://github.com/flutter/website/pull/11829 to add a new page for vulnerabilities.

PR #11829, opened 2025-03-18
Add index of platform vulnerabilities that Flutter developers should be aware of

ash2moon:
Adds documentation for an Android specific vulnerability ([StrandHogg](https://developer.android.com/privacy-and-security/risks/strandhogg) in a new index of platform-specific vulnerabilities.

Resolves https://github.com/flutter/website/issues/11828

## Presubmit checklist

- [ ] This PR is marked as draft with an explanation if not meant to land until a future stable release.

### DIFF
--- a/src/_data/sidenav.yml
+++ b/src/_data/sidenav.yml
@@ -860,6 +860,8 @@
     - title: Reference
       permalink: /reference
       children:
+        - title: Platform Vulnerabilities
+          permalink: /reference/vulnerabilities
         - title: Who is Dash?
           permalink: /dash
         - title: Widget index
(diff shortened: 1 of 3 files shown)


jmagman:
My gut here is that we shouldn't be directly documenting platform issues that aren't Flutter specific, but instead link to external resources where it makes sense. Our own version will be very incomplete and likely out-of-date.
I think targeted warnings in the tool are more effective (https://github.com/flutter/flutter/issues/144484) where we can directly make an actionable suggestion (raise your min SDK).

I'd love to hear @sfshaza2 opinion though.

(and sorry @ash2moon I know this one looked like an easy one to pick off, my bad)

PR #11837, opened 2025-03-20
Add property to feedback widget to capture page location.

antfitch:

We want developers to be able to rate our pages. We can do this now, but we're not capturing the page location that they want to rate. Added an additional parameter to our event tag to capture this.

@RedBrogdon are you able to take a quick look?

## Presubmit checklist


antfitch:
Never mind. I don't think we need this field.

antfitch:
Actually, I think we might need this. Can someone review?

Issue #11862, opened 2025-04-03
Widget titles in the catalog are clipped


Piinks:
### Page URL

https://docs.flutter.dev/ui/widgets/cupertino

### Describe the problem

![Image](https://github.com/user-attachments/assets/f1c38e05-96d5-464a-a485-c50e68a4ca9e)

Widget names get cut off in the tiles of the widget catalog.

### Expected fix

A new layout, or line breaks would be preferred, or maybe something else? Right now it does not look great for folks browsing the catalog.

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

sfshaza2:
FYI, @parlough

PR #11866, opened 2025-04-03
Prevent long class names in widget catalog from being clipped

parlough:
Does this by inserting word break opportunities before each capital letter.

Resolves https://github.com/flutter/website/issues/11862


sfshaza2 (approved):
Nice solution! lgtm

PR #11871, opened 2025-04-04
Adding campaign redirect

RedBrogdon:

Adds a single redirect link.

## Presubmit checklist

(none of them applied 😄 )


firebase.json
```diff
@@ -249,6 +249,8 @@
       { "source": "/go/add-missing-features-to-selection-controls", "destination": "https://docs.google.com/document/d/1QizXwBiO-2REIcEovl5pK06BaLPOWYmNwOE5jactJZA/edit?usp=sharing&resourcekey=0-1pb9mJiAq29Gesmt25GAug", "type": 301 },
       { "source": "/go/add-to-app-samples", "destination": "https://docs.google.com/document/d/1M_GsAL8C1NxbsmzI35w2Th1wvX1KMoPHY5yeeNbzjY0", "type": 301 },
       { "source": "/go/allow-datatable-row-height-to-expand", "destination": "https://docs.google.com/document/d/1qr1IgaU2z9eaM6zCa4u-Q8kJD5Zq7IcdxVyrgR7CwZc/edit?usp=sharing&resourcekey=0-1bNp0ocF4AwGBj6NvRfrEA", "type": 301 },
+      { "source": "/go/agentive-apps-campaign", "destination": "https://flutter.dev/events/building-agentic-apps", "type": 301 },
```
khanhnwin, line 252: should this be `/go/agentive-apps-campaign` --> `/go/agentic-apps-campaign`?

khanhnwin (approved):

parlough (approved):

firebase.json
domesticmouse, line 252: This should be duplicated for `/go/agentic-apps-campaign`

domesticmouse (changes requested):

firebase.json
RedBrogdon: In hindsight, it would have been better to use "agentic," but this one was sent to the printer four weeks ago. On the bright side, it's only being used for a QR code destination, so most people will never even see the /go version on the URL as text.

I wasn't planning on having a shortlink we pass around as text, but we could also make one. Maybe /go/baa?

domesticmouse: Or `/go/bwai` or similar. It is part of the Build with AI push, yeah?

PR #11878, opened 2025-04-07
Add alternate redirects for agentic campaign links

parlough:
An alternative to https://github.com/flutter/website/pull/11871

Adds a short link at flutter.dev/to/baa for users to type when QR code isn't option.

Closes https://github.com/flutter/website/pull/11871


sfshaza2 (approved):
lgtm

Issue #11883, opened 2025-04-08
Validate all images for usefulness and dark mode compatibility


parlough:
We should go through most images on the site and determine if they are:

1. Used
2. Necessary
3. Accessible in light mode
4. Accessible in dark mode or have an alternate for dark mode

If they aren't used or necessary, they (and their usages) should be removed and adjusted as necessary.

If they aren't accessible in dark mode, either the image should be updated to be accessible in both light and dark mode or an alternate image for dark mode should be added.

<details>
<summary><code>src/content/assets/images</code></summary>

- [x] `flutter-chevron-bg.svg`
- [x] `flutter-logo-sharing.png`

<details>
<summary><code>src/content/assets/images/404</code></summary>

- [x] `dash_404.png`

</details>

<details>
<summary><code>src/content/assets/images/branding</code></summary>

<details>
<summary><code>src/content/assets/images/branding/dart</code></summary>

- [x] `64.png`
- [x] `logo.svg`

</details>

<details>
<summary><code>src/content/assets/images/branding/flutter</code></summary>

<details>
<summary><code>src/content/assets/images/branding/flutter/icon</code></summary>

- [x] `1080.png`
- [x] `64.png`
- [x] `mono.svg`

</details>

<details>
<summary><code>src/content/assets/images/branding/flutter/logo</code></summary>

- [x] `1080.svg`
- [x] `default.svg`
- [x] `flutter-lockup-small.png`
- [x] `flutter-lockup.png`
- [x] `flutter-logomark-1080px.png`
- [x] `flutter-logomark-320px.png`
- [x] `flutter-mono-81x100.png`
- [x] `square.svg`

</details>

<details>
<summary><code>src/content/assets/images/branding/flutter/logo+text</code></summary>

<details>
<summary><code>src/content/assets/images/branding/flutter/logo+text/horizontal</code></summary>

- [x] `default.png`
- [x] `default.svg`
- [x] `white.png`
- [x] `white.svg`

</details>

<details>
<summary><code>src/content/assets/images/branding/flutter/logo+text/vertical</code></summary>

- [x] `default.png`
- [x] `default.svg`
- [x] `white.png`
- [x] `white.svg`

</details>

</details>

</details>

</details>

<details>
<summary><code>src/content/assets/images/dash</code></summary>

- [x] `BigDashAndLittleDash.png`
- [x] `DartHummingbird.jpg`
- [x] `Dash.png`
- [x] `DashClawMachine.png`
- [x] `Dashatars.png`
[...]

PR #11886, opened 2025-04-09
Remove outdated trailing comma mention

parlough:
This section is no longer relevant as the formatter automatically adds or removes trailing commas as necessary in Dart 3.7 and later.

The images also include inaccessible text, so just remove it and point to the kept up to date dart.dev docs.

Contributes to https://github.com/flutter/website/issues/11883


sfshaza2 (approved):
lgtm

PR #11889, opened 2025-04-09
Clean up more unused images

parlough:
Also replaces an image that was mostly about the text with a code block.

Contributes to https://github.com/flutter/website/issues/11883


sfshaza2 (approved):
Yowza! lgtm

PR #11894, opened 2025-04-10
Fix markdown usage in aside titles

parlough:
Previously custom aside titles were used as raw text, but a few usages expected inline Markdown syntax to work. That's a fair assumption and sometimes useful, so this PR updates the aside rendering to use markdown-it to render the title to HTML.

**Before:**

A screenshot of an aside title including unrendered markdown

**After:**

A screenshot of an aside title being properly rendered to HTML


sfshaza2 (approved):
lgtm

PR #11895, opened 2025-04-10
Update IDX mentions, links, and images to Firebase Studio

parlough:


sfshaza2 (approved):
Thanks, @parlough! lgtm

Issue #11784, opened 2025-03-06
I can't tell when a tab ends


antfitch:
### Page URL

n/a

### Describe the problem

The new tabs are great, but I can't tell when I've reached the end of the tab. In the following example, the top  of this content is part of a tab, but the bottom of this content is not. How do we make it clear what isn't inside a if tab?

<img width="1360" alt="Image" src="https://github.com/user-attachments/assets/2bc1ecb7-d4d4-465b-b7f9-3ba4b2f801f7" />

Note: The following text is not part of the tab, but it looks like it is:

```
That's it! When you run the app, you should see Hello World.
...
````

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

sfshaza2:
Good point! Maybe insert a horizontal line at the end?

sfshaza2:
All I can say is, it must be because of this stoopid wrist brace. ;)

PR #11897, opened 2025-04-11
Add an optional wrapper to tabs

parlough:
Adds an optional argument to the `tabs` shortcode that when set to `true`, wraps tabs in a border and slight background color. This allows us to opt specific tabs into being more visually isolated from the surrounding content.

To better facilitate this, slightly adjusts the inset (code block) background color to be more distinct from the background of the tab.

Resolves https://github.com/flutter/website/issues/11784

---

Example from [/ui/layout](https://flutter-docs-prod--pr11897-feat-tab-wrapper-aofkivvh.web.app/ui/layout#common-layout-widgets):

Light mode version of a wrapped tab


sfshaza2 (approved):
lgtm

PR #11898, opened 2025-04-11
Reduce node dependencies for quicker site setup

parlough:
Simplifies the node dependency setup of the site to simplify contribution setup and reduce command completion time.

- Updates 11ty to 3.0.1 which updates various dependencies
- Removes `firebase-tools` from the `package.json` and uses a globally installed version if/when needed
- Switches the HTML minifier to a lighter weight, up-to-date alternative


sfshaza2 (approved):
Wow, this is great! lgtm

PR #11904, opened 2025-04-15
Keep quotes when minifying HTML attributes

parlough:
https://github.com/flutter/website/commit/99d9296b22b7e3703e917fdb86fe4f982b823960 switched to a different HTML minifier which stripped quotes from HTML attributes when technically browsers could parse it, but generally shouldn't be dropped.


sfshaza2 (approved):
lgtm

ericwindmill (approved):

sfshaza2:
cc @bkonyi

PR #11906, opened 2025-04-15
Add animations codelab to codelabs.yml

johnpryan:


sfshaza2 (commented):
I'm confused. When I look at the staged version of this PR... on the "Tutorials and samples" page, I can filter by "animations" and there it is!! But when I filter by "How to" (and only "How to"), it's NOT there. (I scrolled through several times to make sure I didn't miss it.)

johnpryan:
Interesting, I don't see _any_ of our codelabs showing up when I filter for "how-to", but I think they probably should. @ericwindmill WDYT?

ericwindmill:
@sfshaza2 @johnpryan -- I think you're right. In fact, I'm not sure that the "type" of 'how to' makes sense. I'll make a follow up PR to make the terminology clearer.

ericwindmill (approved):

PR #11914, opened 2025-04-17
Optimize remaining PNGs

parlough:
The primary pages benefiting here are the newer architecture pages as they have some recently added large images.


jmagman:
Is the right way to optimize these images documented somewhere? I remember trying to find docs but failing:
https://github.com/flutter/website/pull/8265#discussion_r1107480892

sfshaza2 (approved):
yowza. lgtm

PR #11916, opened 2025-04-17
Add `colorist` codelab

domesticmouse:

## Presubmit checklist

### DIFF
--- a/src/content/codelabs/index.md
+++ b/src/content/codelabs/index.md
@@ -156,6 +156,10 @@ Learn how to use Flutter with other technologies.
 
 ### Flutter and Firebase
 
+* [Build a Gemini powered Flutter app][]—**NEW**<br>
+  In this codelab, you’ll learn how to integrate Gemini API into your Flutter
+  app using Vertex AI in Firebase.
+
 * [Add a user authentication flow to a Flutter app using FirebaseUI][]<br>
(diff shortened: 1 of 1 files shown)


src/content/codelabs/index.md
```diff
@@ -156,6 +156,10 @@ Learn how to use Flutter with other technologies.
 
 ### Flutter and Firebase
 
+* [Build a Gemini powered Flutter app][]—**NEW**<br>
+  In this codelab, you’ll learn how to integrate Gemini API into your Flutter
```
RedBrogdon, line 160: Maybe "the Gemini model" or just "Gemini?"

sfshaza2 (approved):
lgtm. However, I defer to @RedBrogdon on any special Gemini language.

src/content/codelabs/index.md
domesticmouse: I'm going with "integrate Gemini" because in my mind there is no such thing as "the Gemini model". Currently we have Gemini as a branch for a family of models, but I'd argue that even `gemini-2.0-turbo` is a series of models.

RedBrogdon: Ah, you're right. "A" Gemini model would be more accurate than "the." What you've got in the latest commit is better than both of those, though. Nice work.

Issue #11177, opened 2024-09-25
Powershell command is not working properly


tamilcselvan:
### Page URL

https://docs.flutter.dev/get-started/uninstall/

### Page source

https://github.com/flutter/website/tree/main/src/content/get-started/uninstall/index.md

### Describe the problem

In this section https://docs.flutter.dev/get-started/uninstall#remove-flutter-configuration-files, the below command is not working fine in the powershell

```powershell
Remove-Item -Recurse -Force -Path $env:APPDATA'.flutter-devtools
```

### Expected fix

When I tried this, it is working fine.

Remove-Item -Recurse -Force -Path "$env:APPDATA\flutter-devtools"

### Additional context

$PSVersionTable

Name                           Value
----                           -----
PSVersion                      7.4.5
PSEdition                      Core
GitCommitId                    7.4.5
OS                             Microsoft Windows 10.0.22631
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

sfshaza2:
@parlough, the source file uses some rather confusing JS. Can you take a look? I'd be down with just simplifying the source. I defer to you.

PR #11927, opened 2025-04-22
Fix PS uninstall instructions

parlough:
Mostly fixes improperly escaped backslashes.

Fixes https://github.com/flutter/website/issues/11177


sfshaza2 (approved):
lgtm

Issue #11923, opened 2025-04-22
First example in "Introduction to Widgets" is difficult to see


aaronshenhao:
![Image](https://github.com/user-attachments/assets/7da59b02-cf3c-4c3b-aba3-ad23e475257e)

The issue occurs in both Firefox and Microsoft Edge. The text is imperceptible against the black background. Link: https://docs.flutter.dev/ui

PR #11948, opened 2025-04-28
Update first widgets example to work with both themes

parlough:
Sets the text style to blue so the text is visible in both the light and dark mode versions of DartPad.

Screenshot of the more visible text

Fixes https://github.com/flutter/website/issues/11923


sfshaza2 (approved):
lgtm!

Issue #10913, opened 2024-07-16
Provide more guidance on website best practices


johnpryan:
We should have a docs page that explains everything a maintainer / contributor needs to know to be a productive contributor to the website. For example:

- [Using includes properly](https://github.com/flutter/website/pull/10911#discussion_r1680082201)
- Embedding images (when do I use markdown syntax vs something else? and how do I decide?)
- Other idiosyncratic Jekyll / 11ty things we are doing that aren't obvious to a new contributor

speedywarp:
is the issue resolved? if not i would like to work on this

johnpryan:
Hi @speedywarp, thanks for your interest, this probably needs to be written by someone that has knowledge about our website infrastructure and patterns that are in use. Is that something you can help with?

PR #11983, opened 2025-05-07
Block out docs contribution docs

parlough:
These won't show up in the sidenav, sitemap, or search, so we should feel free to land these docs in a work-in-progress state. I'll keep adding to as I have time, but others should feel free to as well.

Contributes to https://github.com/flutter/website/issues/10913


sfshaza2 (approved):
interesting. I must have missed discussion about this... :)

PR #11989, opened 2025-05-08
Shorten samples & tutorials sidenav name

parlough:


sfshaza2 (approved):
That's not shorter! Just a different order. lgtm

PR #12006, opened 2025-05-12
Adjust theme toggle icon

parlough:
Switch to an icon that's more obviously related to light and dark mode. This new icon is also used by a few other Google sites, like the Gemini web app.

Screenshot of the new theme select toggle


sfshaza2 (approved):
Nice! lgtm

Issue #12047, opened 2025-05-20
Sidebar scrolling is broken on mobile


loic-sharma:
### Describe the problem

1. Navigate to a sub-page on an iOS device using the Chrome or Safari browser. For example: https://docs.flutter.dev/get-started/install/windows

2. Open the sidebar
3. Try to scroll using a swipe gesture

Result:

https://github.com/user-attachments/assets/9f7190a8-20a0-4718-82fc-d0f1657d668c

### Expected fix

The sidebar should scroll

### Additional context

This reproduces on an iOS device using either Safari or Chrome browser

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

PR #12048, opened 2025-05-20
Fix unscrollable sidenav due to height calculation

parlough:
Fixes https://github.com/flutter/website/issues/12047


sfshaza2 (approved):
lgtm

PR #12051, opened 2025-05-20
Web hot reload also works on stable

sfshaza2:
As it says


sfshaza2:
Oops, @kevmoo and I duplicated efforts. Closing this one.

PR #12054, opened 2025-05-21
Update and include changes to `generate-i10n` in `3.32.0`.

matanlurey:
This launched in 3.32.0.

Also clear up some confusion in https://github.com/flutter/flutter/issues/169209#issuecomment-2898790845.


parlough (approved):
Ah I missed this one, thanks for updating this!

PR #12078, opened 2025-05-27
Clean up liquid config and variable naming

parlough:
- Use consistent naming scheme for Liquid variables, filters, etc.
- Make widget filter filter more generalized
- Remove some unused logic and variables


sfshaza2 (approved):
lgtm

PR #12086, opened 2025-05-30
Remove separate wide layout

parlough:
Rather than use a separate layout for the wide class needed for the learning index, specify it as an extra class on the `body` element. This reduces maintenance of the separate layout and prevents it from otherwise getting out of date.


sfshaza2 (approved):
lgtm

PR #12110, opened 2025-06-09
Fixed swift code

gaaclarke:

Updated the swift example code to match the compiled swift example: https://github.com/flutter/flutter/blob/1b9502b007d01e1cc79dd86fc3300061522803a5/examples/platform_channel_swift/ios/Runner/AppDelegate.swift#L33

There were syntactic errors in the swift code which isn't checked like the dart code.

## Presubmit checklist

### DIFF
--- a/src/content/release/breaking-changes/uiscenedelegate.md
+++ b/src/content/release/breaking-changes/uiscenedelegate.md
@@ -161,7 +161,7 @@ Apps that rely on Storyboards (and XIBs) to create platform channels in
   func register(with registry: any FlutterPluginRegistry) {
     let registrar = registry.registrar(forPlugin: "battery")
     let batteryChannel = FlutterMethodChannel(name: "samples.flutter.dev/battery",
-                                              binaryMessenger: registrar.messenger)
+                                              binaryMessenger: registrar!.messenger())
     batteryChannel.setMethodCallHandler({
       [weak self] (call: FlutterMethodCall, result: FlutterResult) -> Void in
       // This method is invoked on the UI thread.


src/content/release/breaking-changes/uiscenedelegate.md
```diff
@@ -161,7 +161,7 @@ Apps that rely on Storyboards (and XIBs) to create platform channels in
   func register(with registry: any FlutterPluginRegistry) {
     let registrar = registry.registrar(forPlugin: "battery")
     let batteryChannel = FlutterMethodChannel(name: "samples.flutter.dev/battery",
-                                              binaryMessenger: registrar.messenger)
+                                              binaryMessenger: registrar!.messenger())
```
stuartmorgan-g, line 164: There is still a problem here that there is no reason that engine API clients should assume that an API that returns a nullable type should be force unwrapped. We are demonstrating and encouraging an anti-pattern.

gaaclarke: Let's have that conversation outside of this PR.  This is to bring the migration guide in alignment with the example code at `main`.

stuartmorgan-g: > Let's have that conversation outside of this PR.

I disagree strongly. The number of people who will see this page is orders of magnitude more than the number of people who will go look at an example in the flutter/flutter tree. The point at which we are going out of our way to make sure that as many people as possible see the code is an important checkpoint to make sure we aren't showing people code that we do not actually want people to write.

I do not want us writing high-profile documentation that encourages people to blindly ignore our own API's contracts.

sfshaza2 (approved):
lgtm, once you and @stuartmorgan-g reach agreement

src/content/release/breaking-changes/uiscenedelegate.md
stuartmorgan-g: I will defer this review to @vashworth to decide if this is really the pattern the iOS team wants to recommend to developers.

vashworth: I think we should definitely discuss this more, but let's at least get the documentation to a non-broken state. 

Perhaps, we could make this more defensive and swift-y
```
func register(with registry: any FlutterPluginRegistry) {
        let registrar = registry.registrar(forPlugin: "battery")
        guard let registrar = registry.registrar(forPlugin: "battery") else {
            print(
                "Error: Registrar for 'battery' plugin is nil. Cannot set up battery channel."
            )
            return  // Exit the function if registrar is nil
        }
        let batteryChannel = FlutterMethodChannel(
            name: "samples.flutter.dev/battery",
            binaryMessenger: registrar.messenger()
        )
    }
```

gaaclarke: > Perhaps, we could make this more defensive and swift-y

I think the force unwrapping serves as a better example for the code and is the proper behavior in most cases.  It will ultimately be up to the developer to decide if they want that or not and we cannot give a "one size fits all" solution.

The reason I'm recommending the force unwrap is because of the barrier between the host and the Flutter app it's easy to keep executing in a non-obvious broken state which can lead to data loss or corruption.

Right now nil is never returned from `registrarForPlugin`, but imagine we started returning nil if there was a collision with the name of an existing plugin.  The fix would be something like changing `registrarForPlugin:"foo"` to `registrarForPlugin:"flutter.dev/foo"`.  That is a fix that is in the purview of the developer and giving them the most forthwith information that there is a problem is the most helpful.  In the `guard` situation they'd have to derive that problem from secondary effects.

johnmccutchan: > Right now nil is never returned from registrarForPlugin

Littering code with error handling paths that _will never run_ is bad and only adds confusion for the developer. 

Does the example `guard let` code lead to a working app? What happens when batteryChannel isn't initialized?

I'm asking because Aaron's code was very clear "foo!.bar()" indicates to me that foo isn't ever expected to be nil in practice. Where as the new code makes me think it does happen in practice and that I should worry about at this code site and other code sites.

What should my app do in the case that the battery plugin isn't available? 

Why would that ever happen if I added a dependency on the batter plugin?

Aaron's code is much clearer and I strongly suspect most apps won't function correctly if necessary plugins aren't available.

jtmcdole: I'm OK with `!()`. If folks want to get defensive, they can use the guard statement. ultimately we want the example code to compile.

gaaclarke: Okay, marking this resolved.  I think at the crux of this comment is design feedback and there is a follow up meeting for that.

PR #12158, opened 2025-06-27
Fix PR template

sfshaza2:
Specifically, asking folks that if they can't (or really don't want to) sign the CLA, they should create an issue instead of a PR.


sfshaza2:
@parlough @antfitch can one of you review?

parlough:
Good idea, thanks!

PR #12185, opened 2025-07-07
Fix broke link for dynamic_theme in quickstarts_flutter.yml

liuchong:
Fix broke link for dynamic_theme in quickstarts_flutter.yml, should be typo.


parlough (approved):
Thanks for catching and fixing these! Looks good to me :)

PR #12261, opened 2025-08-02
Remove codelab and cookbook index in favor of learning resource index

parlough:


antfitch:
@parlough do we have a project for this? If not, can you add one for this?

parlough:
> @parlough do we have a project for this? If not, can you add one for this?

It's follow-up cleanup from a variety of Eric and other's projects, so I'm not sure if there's a specific project. I went ahead and created https://github.com/flutter/website/issues/12367 as an epic, consolidated issue for these pieces and added it to the Flutter docs 2025 project. I'll add some more details and links to other portions to that issue soon.

parlough:
@sfshaza2 Do you have any concerns with retiring these pages in favor of the combined [learning resource index](https://docs.flutter.dev/reference/learning-resources)? If so, let me know how we can improve the index to get to a point where you'd be okay with this change. Thanks :D

sfshaza2 (approved):
I dived in to the learning resources page. I don't like it as well as the pages we're replacing, but I _do_ like that it will be easier to maintain. Having said that, I'm a bit surprised that using Flame is considered a good tutorial for beginners...

parlough:
Thanks for taking a look @sfshaza2!

>  I don't like it as well as the pages we're replacing, but I do like that it will be easier to maintain.

That's definitely fair. They were a bit more focused. After the Jaspr migration, I'll explore further improvements to this page, but if you have any specific issues with or ideas for the index, do let me know.

> Having said that, I'm a bit surprised that using Flame is considered a good tutorial for beginners...

I think that specific codelab is more about showing the possibilities, as it directly gives most of the code. We can definitely remove that tag though if you prefer!

PR #12263, opened 2025-08-04
docs: add `intelligence` to apple-frameworks.md

wrbl606:
Enrich Apple Frameworks list with `intelligence` as a bridge for `AppIntents` framework.


parlough:
Thanks for the PR and adding this! This generally looks good to me, but I'd like to confirm with some others first.

@jmagman @vashworth Do either of you have any concerns about adding this to the [framework -> package table](https://flutter-docs-prod--pr12263-patch-1-t1uvf92d.web.app/platform-integration/ios/apple-frameworks#flutter-plugins-and-apple-frameworks) we have in the docs?

wrbl606:
I don't want to rush things but it has been a while since the last message here. Is there something I can do to push this towards production?

sfshaza2:
I pinged some folks offline...

PR #12265, opened 2025-08-04
Update key-value-data.md

tomasbaran:
Remove unnecessary try & catch blocks since the repo classes always return the `Result` type (and never leak an `Exception`).


ericwindmill:
I'm still considering the change, but the runnable example code at https://github.com/flutter/website/tree/main/examples/app-architecture/todo_data_service needs to be updated to match the suggested changes in order to pass CI.

PR #12308, opened 2025-08-14
Update app-extensions.md

fcontreras:
**Description:**

In some environments the value of the Minimum Deployments is not set for the Runner Target, and the ShareExtension one takes the latest one it can, so they not match causing the Extension not to be shown on the Photos Share sheet.

So I added an extra step for people following the tutorial to check that small detail that can cause frustration.

**Issues solved**: None
**Commits dependency**: None


fcontreras:
CLA Signed

sfshaza2 (commented):
This change looks great. But I will check the staged version when the build has completed.

PR #12336, opened 2025-08-20
Add a Rules section to our new Create with AI topic

antfitch:

Adds a Rules section and an Overview section.

## Presubmit checklist

  of 80 characters or fewer.

### DIFF
--- a/src/content/ai/create-with-ai.md
+++ b/src/content/ai/create-with-ai.md
@@ -9,6 +9,21 @@ description: >
 This guide covers how you can leverage AI tools to build AI-powered features for
 your Flutter apps and streamline your Flutter and Dart development.
 
+## Overview
+
+AI can be used for building AI-powered apps with Flutter and
+for accelerating your development workflow. You can
+integrate AI-powered features like
+natural language understanding and content generation
+directly into your Flutter app using powerful SDKs,
+like the Firebase SDK for Generative AI. You can also use
+AI tools, such as Gemini Code Assist and Gemini CLI, to help
+with code generation and scaffolding. These tools are
+powered by the Dart and Flutter MCP Server, which provides
+AI with a rich context about your codebase. Additionally,
+rules files help fine-tune the AI's behavior and enforce
+project-specific best practices.
+
 ## Build AI-powered experiences with Flutter
 
 Using AI in your Flutter app unlocks new user experiences that allow your app
@@ -24,7 +39,9 @@ resources:
 * [Flutter AI Toolkit][] - A sample app with pre-built widgets to help you build
   AI-powered features in Flutter
 
----
+[Firebase AI Logic]: {{site.firebase}}/docs/ai-logic
+[firebase-ai-logic-docs]: {{site.firebase}}/docs/ai-logic/get-started
+[Flutter AI Toolkit]: {{site.url}}/ai-toolkit
 
 ## AI development tools
 
@@ -34,9 +51,9 @@ you write code faster, understand complex concepts, and reduce boilerplate.
 
 ### Gemini Code Assist
 
-[Gemini Code Assist][] is an AI-powered collaborator available in Visual Studio Code
-and JetBrains IDEs (including Android Studio). It has a deep understanding of
-your project's codebase and can help you with:
+[Gemini Code Assist][] is an AI-powered collaborator available in
+Visual Studio Code and JetBrains IDEs (including Android Studio). It has a deep
+understanding of your project's codebase and can help you with:
 
 * **Code completion and generation**: It suggests and generates entire blocks of
   code based on the context of what you're writing.
@@ -46,6 +63,7 @@ your project's codebase and can help you with:
   Code Assist to explain it and suggest a fix, and
   [Dart and Flutter MCP Server][dart-mcp-flutter-docs]
 
+[Gemini Code Assist]: https://codeassist.google/
 
 ### Gemini CLI
 
@@ -60,15 +78,16 @@ environment. You can use it to:
 To get started, visit the [Gemini CLI][] website, or try this
 [Gemini CLI codelab][].
 
----
+[Gemini CLI]: {{site.github}}/google-gemini/gemini-cli
+[Gemini CLI codelab]: https://codelabs.developers.google.com/gemini-cli-hands-on
 
 ## Dart and Flutter MCP Server
 
 To provide assistance during Flutter development, AI tools such as
 Gemini Code Assist and Gemini CLI need to communicate with Dart
 and Flutter's developer tools. The Dart and
-Flutter MCP (model context protocol) Server facilitates this
-communication. The model context protocol is a specification that
+Flutter MCP Server facilitates this communication.
+MCP (model context protocol) is a specification that
 outlines how development tools can share the context of a user's code
 with an AI model, which allows the AI to better understand and interact
 with the code.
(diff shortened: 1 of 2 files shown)


src/content/ai/create-with-ai.md
```diff
@@ -78,18 +97,86 @@ and fix errors, hot reload, get the selected widget, and more. This bridges
...
+    *   If desired, copy the contents of `airules.md` to an
+        LLM like [Gemini][] and provide a prompt to reformat
+        the file for another editor.
+
+        Example prompt:
+
+        ```text
+        Convert the following airules.md file into a
+        guidelines.md file for JetBrains IDE. Make sure to
+        use the styles required for a guidelines.md file:
```
twerske: Could we run this prompt ourselves then link to a downloadable version for each of the environments instead of telling developers how to DIY?

Especially because this is making everyone create a Firebase Studio project to the take the rules and convert them.

src/content/ai/create-with-ai.md
```diff
@@ -78,18 +97,86 @@ and fix errors, hot reload, get the selected widget, and more. This bridges
...
-[Gemini CLI codelab]: https://codelabs.developers.google.com/gemini-cli-hands-on
-[Firebase AI Logic]: {{site.firebase}}/docs/ai-logic
-[firebase-ai-logic-docs]: {{site.firebase}}/docs/ai-logic/get-started
-[How to build agentic apps with Flutter and Firebase AI Logic]: {{site.yt.watch}}/watch?v=xo271p-Fl_4
-[Flutter AI Toolkit]: {{site.url}}/ai-toolkit
 [dart-mcp-dart-docs]: {{site.dart-site}}/tools/mcp-server
 [dart-mcp-github]: {{site.github}}/dart-lang/ai/tree/main/pkgs/dart_mcp_server
 [dart-mcp-flutter-docs]: #dart-and-flutter-mcp-server
+
+## Rules
```
twerske: I'd love to ground this in Rules for Flutter (and Dart?)

src/content/ai/create-with-ai.md
```diff
@@ -78,18 +97,86 @@ and fix errors, hot reload, get the selected widget, and more. This bridges
...
+| Windsurf | `guidelines.md` | [Configure guidelines.md][] |
+
+[Configure airules.md]: https://firebase.google.com/docs/studio/set-up-gemini#custom-instructions
+[Configure .github/copilot-instructions.md]: https://code.visualstudio.com/docs/copilot/copilot-customization#_custom-instructions
+[Configure cursorrules.md]: https://docs.cursor.com/en/context/rules
+[Configure guidelines.md]: https://www.jetbrains.com/help/junie/customize-guidelines.html
+[Configure .instructions.md]: https://code.visualstudio.com/docs/copilot/copilot-customization#_custom-instructions
+[Configure guidelines.md]: https://docs.windsurf.com/windsurf/cascade/memories#rules
+
+### Convert rules between editors
```
twerske: I don't think we need to provide instructions on converting between editors. This is a detail that feels better fit for IDE and Rules project docs than Flutter's.

twerske:
One other q, the double nested header of "Create with AI" is a little funky here, should it be Introduction or something?

[image](https://github.com/user-attachments/assets/ce761cfc-314b-4f73-9434-c5c5edd263df)

src/content/ai/create-with-ai.md
antfitch: I want to avoid creating a downloadable version for each product for a few reasons:

(a) The versions are likely to get outdated quickly, which will require maintenance.

(b) What matters to each product in their rules file might also change over the coming months and years. I want to avoid checking in with various products to make sure we're using their latest standards.

(b) New tools are coming out rapidly and the list we have may grow or change every few months. Each addition would require a new rules file and more maintenance.

Essentially, while I love how the Angular team set up a similar section with the requested style, I want to avoid what the Angular team is doing because of the long-term maintenance cost. So leaving this as is for now, but I get where you're coming from.

What we could do is this: I could take the current Firebase Studio rules file and reduce it (it's very long). I could add it to these steps and then have people use whatever LLM they want to convert the rules file to any format that works for them.

src/content/ai/create-with-ai.md
antfitch: This is something I definitely want to keep, but I'll make a change so that people don't have to go to Firebase Studio to grab some rules to work with.

antfitch:
> One other q, the double nested header of "Create with AI" is a little funky here, should it be Introduction or something?
> 
> [image](https://github.com/user-attachments/assets/ce761cfc-314b-4f73-9434-c5c5edd263df)

I agree. It does look odd. I don't want to call this an Intro because it's essentially a guide to using AI. I'll look into changing the category title or finding some other title for this topic.

src/content/ai/create-with-ai.md
antfitch: Sounds great. Updating!

src/content/ai/create-with-ai.md
```diff
@@ -78,18 +97,81 @@ and fix errors, hot reload, get the selected widget, and more. This bridges
...
+[Configure guidelines.md]: https://www.jetbrains.com/help/junie/customize-guidelines.html
+[Configure .instructions.md]: https://code.visualstudio.com/docs/copilot/copilot-customization#_custom-instructions
+[Configure guidelines.md]: https://docs.windsurf.com/windsurf/cascade/memories#rules
+
+### Create rules for your editor
+
+If you want to create Flutter and Dart rules for an
+AI-driven environment, follow these steps:
+
+1.  Download the attached file,
```
antfitch: @parlough, this step requires the ability to download text files. It looks like we might need to update `eleventy.config.ts`. Can you check on a fix for this? We'll need to get this fix in and then re-look at this PR.

antfitch:
@sfshaza2, @twerske, and @parlough I've made some updates. Please look closely. Here's a quick overview:

1. Updated category name from "Create with AI" to "AI solutions". This preserves the topic name "Create with AI".
2. I've used "Rules for Flutter and Dart" in various places, including a section heading.
3. I've removed "Convert rules between editors" and replaced this with "Create rules for your editor". This is a new space for most people, so I want to help them understand quickly how they can take our template and convert it so that it is useful for various environments.
4. We don't have the ability to download files from our site (txt, md) that I can see, so @parlough, we'll need to get that figure out with our system.
5. @twerske we need someone on the development side of things to look very closely at that TXT file I've added. It's essentially the Firebase Studio rules file for Flutter with Firebase Studio bits pulled out.

And that's the summary!

src/content/ai/create-with-ai.md
twerske: I like your proposal and I see you moved forward with that approach (ie one simplified version and instructions to put it into your AI tool of choosing) consider this resolved

twerske:
@redbrogdon plans to eval to support item #5 on the list.

Additionally, what are thoughts about putting the download link higher as well of like "Here are the Flutter AI Rules" so it's not only under the subheader section

antfitch:
> Additionally, what are thoughts about putting the download link higher as well of like "Here are the Flutter AI Rules" so it's not only under the subheader section

I like this idea. I've done something a little edgy. Maybe it works, maybe it doesn't. Take a close look.

src/content/ai/create-with-ai.md
antfitch: Thank you @parlough for making this update!

src/content/ai/ai-rules.md
```diff
@@ -0,0 +1,81 @@
...
+corresponding rule file names:
+
+| Environment | Rules File | Installation Instructions |
+| :--- | :--- | :--- |
+| Copilot powered IDEs | `copilot-instructions.md` | [Configure .github/copilot-instructions.md][] |
+| Cursor | `cursor.md` | [Configure cursorrules.md][] |
+| Firebase Studio | `airules.md` | [Configure airules.md][] |
+| JetBrains IDEs | `guidelines.md` | [Configure guidelines.md][] |
+| VS Code | `.instructions.md` | [Configure .instructions.md][] |
+| Windsurf | `guidelines.md` | [Configure guidelines.md][] |
```
johnpryan, line 40: What about Gemini CLI? I think you are supposed to put those rules in a GEMINI.md file anywhere in your project, according to [this codelab](https://codelabs.developers.google.com/gemini-cli-hands-on#9).

antfitch: Good catch. I've added this to the list.

johnpryan (approved):

parlough:
Woo! Thanks everyone for the hard work on this. Exciting!

PR #12345, opened 2025-08-22
Rename 'toc' to 'showToc' in frontmatter

parlough:
This will help enable the site infrastructure migration, which reserves the `toc` variable for other purposes.


sfshaza2 (approved):
lgtm

PR #12350, opened 2025-08-22
[Android] Add breaking changes page on default ABI filters

camsim99:
_Description of what this PR is changing or adding, and why:_
Adds a breaking change for https://github.com/flutter/flutter/pull/168293 which landed in 3.35. TLDR: To set custom Android ABI filters, you must clear the defaults that Flutter sets or use the Flutter-provided flag for doing so.

_Issues fixed by this PR (if any):_
Fixes https://github.com/flutter/flutter/issues/174004

_PRs or commits this PR depends on (if any):_
N/A

## Presubmit checklist

- [ ] If you are unwilling, or unable, to sign the CLA, even for a _tiny_, one-word PR, please file an issue instead of a PR.
- [ ] If this PR is not meant to land until a future stable release, mark it as draft with an explanation.
- [ ] This PR follows the [Google Developer Documentation Style Guidelines](https://developers.google.com/style)—for example, it doesn't use _i.e._ or _e.g._, and it avoids _I_ and _we_ (first-person pronouns).
- [ ] This PR uses [semantic line breaks](https://github.com/dart-lang/site-shared/blob/main/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks)
  of 80 characters or fewer.

PR #12379, opened 2025-09-02
Fix "Flutter now sets default `abiFilters` in Android builds" link

camsim99:
Fixes link added in https://github.com/flutter/website/pull/12350.


parlough (approved):
Thanks for catching and fixing this!

Issue #12413, opened 2025-09-08
Add link to GenUI SDK for Flutter


antfitch:
Under AI Solutions on the Flutter docs site, add a link to https://github.com/flutter/genui.

PR #12415, opened 2025-09-08
Adding a link to the experimental GUI SDK

sfshaza2:
Fixes https://github.com/flutter/website/issues/12413

cc: @sethladd @twerske @antfitch

### DIFF
--- a/src/_data/sidenav.yml
+++ b/src/_data/sidenav.yml
@@ -229,7 +229,6 @@
           permalink: /ai-toolkit/custom-llm-providers
         - title: Chat client sample
           permalink: /ai-toolkit/chat-client-sample
-    - divider
     - title: Dart & Flutter MCP server
       permalink: https://dart.dev/tools/mcp-server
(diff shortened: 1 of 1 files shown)


src/_data/sidenav.yml
```diff
@@ -232,6 +232,9 @@
     - divider
     - title: Dart & Flutter MCP server
       permalink: https://dart.dev/tools/mcp-server
+    - divider
```
antfitch:
```suggestion
```

src/_data/sidenav.yml
```diff
@@ -232,6 +232,9 @@
     - divider
```
antfitch:
```suggestion
```

src/_data/sidenav.yml
parlough: I think the two dividers in a row looks a bit awkward. Can we remove this second one and keep the external links grouped together?

```suggestion
```

Issue #12382, opened 2025-09-03
The iOS platform channel code sample is outdated


navaronbracke:
### Page URL

https://docs.flutter.dev/platform-integration/platform-channels/

### Page source

https://github.com/flutter/website/tree/main/src/content/platform-integration/platform-channels.md

### Describe the problem

The code sample for setting up an iOS MethodChannel for an application is slightly outdated.

Most notably it still refers to `@UIApplicationMain` instead of `@main`.

### Expected fix

The code sample is updated, so that it no longer uses the deprecated annotation.
Preferably it should match what is defined in 

https://github.com/flutter/flutter/blob/master/packages/flutter_tools/templates/app/ios.tmpl/Runner/AppDelegate.swift

modulo the changes for the MethodChannel itself.

### I would like to fix this problem.

- [x] I will try and fix this problem on docs.flutter.dev.

parlough:
I believe fixed by https://github.com/flutter/website/pull/12383. Thanks @navaronbracke!

PR #12416, opened 2025-09-08
Fixed a Swift code snippet

sfshaza2:
Fixes https://github.com/flutter/website/issues/12382


parlough:
I believe the issue at hand was already fixed by https://github.com/flutter/website/issues/12382.

The code removed here is necessary for the steps. I think the mention in the issue was saying the surrounding setup should be updated to match the template, not the body.

PR #12429, opened 2025-09-15
Fix definition of Flutter SDK's versioning system

sfshaza2:
This replaces https://github.com/flutter/website/pull/12108


parlough (approved):
I pushed a commit removing a duplicate `/flutter` in the GitHub link, but otherwise looks good to me!

PR #12456, opened 2025-09-22
Add back website-reviewers team as fallback codeowner

parlough:
This will allow members of `@flutter/website-reviewers` to still approve and land PRs if necessary while still only assigning @sfshaza2 as the initial reviewer since she is in the group as well.

\cc @antfitch


sfshaza2 (approved):
lgtm

Issue #12472, opened 2025-09-26
Add-to-app metrics n the docs are old and potentially inaccurate.


asanchezmobiledev-cyber:
### Page URL

https://docs.flutter.dev/add-to-app/performance/

### Page source

https://github.com/flutter/website/tree/main/src/content/add-to-app/performance.md

### Describe the problem

The metrics for add-to-app increased memory load are from 2021 using Flutter 1.X. Can we please have the latest metrics posted?

### Expected fix

Metrics from 2025 are added using latest version of Flutter.

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

sfshaza2:
Spoke offline to SWEs and was asked to remove the memory and latency section.

PR #12492, opened 2025-10-01
Removing an out of date section

sfshaza2:
Fixes https://github.com/flutter/website/issues/12472


taaha161:
Was this section removed because we can expect Flutter no longer has a latency cost when starting a FlutterEngine?

PR #12493, opened 2025-10-01
Adding survey to banner

sfshaza2:
Once it goes live...


parlough (approved):
Thanks for adding this back. Approving so you can land when it's published.

If you'd like, I think could even remove the 3.35 mention now to keep it more focused.

sfshaza2:
Removing the last release would be fine, too. But, for now, I want to land this :)

Issue #12498, opened 2025-10-02
Fix standard auto-formatting issues


guidezpl:
### Describe the problem

I wanted to contribute but in editing a few files, I realized they were hand formatted.

Examples

-  HTML <img width="872" height="172" alt="Image" src="https://github.com/user-attachments/assets/23d31a18-612b-44b5-9dcd-95e9ab68970c" />
- YML <img width="561" height="47" alt="Image" src="https://github.com/user-attachments/assets/13f92408-4375-4491-9cfb-eb319920f5e3" />
- <img width="614" height="37" alt="Image" src="https://github.com/user-attachments/assets/f8e4050b-d541-4c63-84b2-f564071988d6" />

### Expected fix

Fix standard auto-formatted issues (no trailing spaces, ensure ending newlines, HTML indented properly)

### On which platform(s) did you experience this issue?

macOS

### Additional context

VS Code out of the box formatters

parlough:
Yeah we eventually need to add some formatting tools and checks, particularly for HTML and Markdown, but I don't want to require contributors to have Node or another tool.

For now, if you're updating a file, feel free to keep the auto formatting changes in your PR if you're changing those files and it's easier to keep them.

PR #12501, opened 2025-10-02
Format all files

guidezpl:
This is what formatting all files and re-organizing imports does to all files, with the following additional changes

- revert `firebase.json`
- revert edits where imports would change `docregion`

_Issues fixed by this PR (if any):_ #12498 

## Presubmit checklist

  of 80 characters or fewer.


guidezpl:
Seeking feedback from reviewers. Initial thoughts:

- is it possible to add a test to check this for each file type?
- which file types don't need to be formatted
  - .svg
  - ...?

guidezpl:
@sfshaza2 @parlough any thoughts on what should be/should not be formatted?

parlough:
Sorry about the delay on this @guidezpl! We want to hold off on this until after https://github.com/flutter/website/pull/12548 lands.

> Is it possible to add a test to check this for each file type?

It would be if we had a way to automate formatting from Dart, as we don't want to require contributors to have Node installed. 

> any thoughts on what should be/should not be formatted?

I think formatting SCSS, Markdown, yaml, and what remains of HTML files after the migration makes sense :)

guidezpl:
sure, pls land that awesome PR, and I'll circle back afterwards!

PR #12548, opened 2025-10-14
Migrate the site infrastructure to Jaspr and Jaspr Content

parlough:
**Staged:** https://flutter-docs-prod--pr12548-feat-jaspr-migration-kz7c0jfj.web.app/

Migrates the site from being built with JS/TS, Node, and 11ty to being built with purely Dart, [Jaspr](https://jaspr.site/), and [Jaspr Content](https://docs.jaspr.site/content).

---

As part of this work, the `dash_site` bash script has been removed from the root of the repository.

After the removal, you can access the site tooling, including building and staging with `dart run dash_site <command>` from anywhere within the repository workspace. If you still want a shorter command, you can add an alias for `dart run dash_site` in your shell, such as `dst`.

---

**Staged:** https://flutter-docs-prod--pr12548-feat-jaspr-migration-kz7c0jfj.web.app/

Resolves https://github.com/flutter/website/issues/12405
Fixes https://github.com/flutter/website/issues/12590

PR #12568, opened 2025-10-20
feat: migrate archive table to jaspr

schultek:
This migrates:

- SDK archive tables in /install/archive
- Latest release download buttons in /install/manual

Part of #12548

### DIFF
--- a/site/lib/src/components/client/archive_table.dart
+++ b/site/lib/src/components/client/archive_table.dart
@@ -2,48 +2,10 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'dart:convert';
-
-import 'package:http/http.dart' as http;
 import 'package:jaspr/jaspr.dart';
 import 'package:meta/meta.dart';
-import 'package:universal_web/js_interop.dart';
 
-class Release {
-  Release({
-    required this.url,
-    required this.version,
-    required this.channel,
-    required this.dartSdkVersion,
-    required this.architecture,
-    required this.hash,
-    required this.releaseDate,
-  });
-
-  factory Release.fromJson(Map<String, Object?> json, String baseUrl) {
-    return Release(
-      url: '$baseUrl/${json['archive'] as String}',
-      version: json['version'] as String,
-      channel: json['channel'] as String,
-      dartSdkVersion:
-          (json['dart_sdk_version'] as String?)?.split(' ')[0] ?? '-',
-      architecture: (json['dart_sdk_arch'] as String?) ?? 'x64',
-      hash: (json['hash'] as String).substring(0, 7),
-      releaseDate: Date(Date.parse(json['release_date'] as String)),
-    );
-  }
-
-  final String url;
-  final String version;
-  final String channel;
-  final String dartSdkVersion;
-  final String architecture;
-  final String hash;
-  final Date releaseDate;
-}
-
-const _baseReleasesUrl =
-    'https://storage.googleapis.com/flutter_infra_release/releases/';
+import '../../models/flutter_release_model.dart';
 
 @client
 class ArchiveTable extends StatefulComponent {
@@ -54,26 +16,6 @@ class ArchiveTable extends StatefulComponent {
 
   @override
   State<ArchiveTable> createState() => _ArchiveTableState();
-
-  static final Map<String, Future<List<Release>>> _flutterReleasesFutures = {};
-
-  /// Fetches Flutter release JSON for the given OS and caches the result.
-  static Future<List<Release>> fetchFlutterReleases(String os) {
-    return _flutterReleasesFutures[os] ??= http
-        .get(Uri.parse('${_baseReleasesUrl}releases_$os.json'))
-        .then((response) {
-          if (response.statusCode == 200) {
-            final releases = jsonDecode(response.body) as Map<String, Object?>;
-            final baseUrl = releases['base_url'] as String;
-            final releasesList = releases['releases'] as List<Object?>;
-            return releasesList.map((release) {
-              return Release.fromJson(release as Map<String, Object?>, baseUrl);
-            }).toList();
-          } else {
-            throw Exception('Failed to load Flutter releases');
-          }
-        });
-  }
 }
 
 class _ArchiveTableState extends State<ArchiveTable> {
@@ -82,7 +24,7 @@ class _ArchiveTableState extends State<ArchiveTable> {
 
   bool isLoading = true;
   String? error;
-  List<Release> releases = [];
+  List<FlutterRelease> releases = [];
 
   @override
   void initState() {
@@ -96,7 +38,7 @@ class _ArchiveTableState extends State<ArchiveTable> {
   @awaitNotRequired
   Future<void> loadReleases() async {
     try {
-      final releasesData = await ArchiveTable.fetchFlutterReleases(
+      final releasesData = await FlutterRelease.fetchFlutterReleases(
         component.os,
       );
       final filteredReleases = releasesData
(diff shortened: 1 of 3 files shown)


site/lib/src/components/client/archive_table.dart
```diff
@@ -0,0 +1,229 @@
...
+  @override
+  State<ArchiveTable> createState() => _ArchiveTableState();
+
+  static final Map<String, Future<List<Release>>> _flutterReleasesFutures = {};
+
+  /// Fetches Flutter release JSON for the given OS and caches the result.
+  static Future<List<Release>> fetchFlutterReleases(String os) {
+    if (!_flutterReleasesFutures.containsKey(os)) {
+      final url = Uri.parse(
+        'https://storage.googleapis.com/flutter_infra_release/releases/releases_$os.json',
```
parlough: The base of this URI is used in the `buildProvenanceLink` below as well. Consider pulling out the shared portion into a top-level const.

site/lib/src/util.dart
```diff
@@ -12,6 +12,8 @@ const productionBuild = bool.fromEnvironment('PRODUCTION');
 /// Path to the `/src` directory where site content is located.
 final siteSrcDirectoryPath = path.join('..', 'src');
 
+const globalSdkChannel = 'stable';
```
parlough: Can you add a doc comment to this indicating how its used.

src/content/install/manual.md
```diff
@@ -156,7 +156,7 @@ then extract the SDK to where you want it stored.
     Download the following installation bundle to get the
     latest stable release of the Flutter SDK.
 
-    [(loading...)](#){:.download-latest-link-windows .filled-button}
+    <DownloadLatestButton os="windows" />
```
parlough, line 159: I love this as a custom Jaspr Content component, much cleaner!

site/lib/src/components/client/download_latest_button.dart
```diff
@@ -0,0 +1,82 @@
...
+
+  @override
+  Component build(BuildContext context) {
+    if (!isLoading && error == null && latestRelease == null) {
+      return const Component.empty();
+    }
+    return a(
+      href: latestRelease?.url ?? '#',
+      classes: 'filled-button',
+      events: {'click': (_) {}},
```
parlough: If this empty event handler is needed, can you add an inline comment explaining why?

site/lib/src/components/client/archive_table.dart
```diff
@@ -0,0 +1,229 @@
...
+    return a(
+      href:
+          '$baseUrl$channel/$os/flutter_${os}_${release.version}-$channel.$extension.intoto.jsonl',
+      target: Target.blank,
+      [text('Attestation bundle')],
+    );
+  }
+}
+
+/// The JavaScript Date object.
```
parlough: Good idea to use this rather than `package:intl` for this focused use case. Would it make sense to do the same for the dart.dev archive as well?

site/lib/src/components/client/archive_table.dart
```diff
@@ -0,0 +1,229 @@
...
+
+    if (os == 'windows' && dateValue < windowsCutoff) {
+      // Provenance not available before 4/3/2023 for Windows
+      return span([text('-')]);
+    } else if (dateValue < otherOsCutoff) {
+      // Provenance not available before 12/15/2022 for macOS and Linux
+      return span([text('-')]);
+    }
+
+    final extension = os == 'linux' ? 'tar.xz' : 'zip';
```
parlough: Even though `extension` isn't reserved in this context, consider renaming to `archiveExtension` to avoid any confusion.

site/lib/src/components/client/archive_table.dart
```diff
@@ -0,0 +1,229 @@
...
+  final String channel;
+
+  @override
+  State<ArchiveTable> createState() => _ArchiveTableState();
+
+  static final Map<String, Future<List<Release>>> _flutterReleasesFutures = {};
+
+  /// Fetches Flutter release JSON for the given OS and caches the result.
+  static Future<List<Release>> fetchFlutterReleases(String os) {
+    if (!_flutterReleasesFutures.containsKey(os)) {
```
parlough: Since the value isn't nullable, could we do something like the following to avoid the extra lookup and not-null assertion.

```dart
return _flutterReleasesFutures[os] ??= http.get(url).then((response) {
  // ...
});
```

site/lib/src/components/client/archive_table.dart
```diff
@@ -0,0 +1,229 @@
...
+    required this.url,
+    required this.version,
+    required this.channel,
+    required this.dartSdkVersion,
+    required this.architecture,
+    required this.hash,
+    required this.releaseDate,
+  });
+
+  factory Release.fromJson(Map<String, dynamic> json, String baseUrl) {
```
parlough:
```suggestion
  factory Release.fromJson(Map<String, Object?> json, String baseUrl) {
```

site/lib/main.dart
```diff
@@ -127,6 +129,22 @@ List<CustomComponent> get _embeddableComponents => [
...
+      final channel = attributes['channel'] as String;
+      return ArchiveTable(os: os.toLowerCase(), channel: channel.toLowerCase());
+    },
+  ),
+  CustomComponent(
+    pattern: RegExp('DownloadLatestButton', caseSensitive: false),
+    builder: (_, attributes, _) {
+      final os = attributes['os'] as String;
+      final arch = attributes['arch'];
+      return DownloadLatestButton(os: os.toLowerCase(), arch: arch?.toLowerCase());
```
parlough: Can you run `dart format` over the site directory before merging this PR in as well.

site/lib/src/util.dart
schultek: I don't actually know why this would ever be changed, I just made it a variable since its also specified in site.yaml. Do you know why it exists there?

site/lib/src/components/client/download_latest_button.dart
schultek: Oops, leftover from testing

site/lib/src/components/client/archive_table.dart
schultek: I will take a look

site/lib/src/util.dart
parlough: I think we used to use the variable for staged versions of the site documenting other channels, but we don't do that anymore.

I opened https://github.com/flutter/website/pull/12575 to remove the site config value. Feel free to inline this variable as well.

site/lib/main.dart
schultek: Done

site/lib/src/components/client/archive_table.dart
schultek: Done

site/lib/src/components/client/archive_table.dart
schultek: Done

site/lib/src/components/client/archive_table.dart
schultek: Done

site/lib/src/components/client/archive_table.dart
schultek: Done

site/lib/src/components/client/archive_table.dart
schultek: Added as a enhancement in the project

site/lib/src/util.dart
schultek: Done

site/lib/src/components/client/archive_table.dart
```diff
@@ -0,0 +1,223 @@
...
+// found in the LICENSE file.
+
+import 'dart:convert';
+
+import 'package:http/http.dart' as http;
+import 'package:jaspr/jaspr.dart';
+import 'package:meta/meta.dart';
+import 'package:universal_web/js_interop.dart';
+
+class Release {
```
parlough: Consider moving this to `/site/lib/src/models/release_model.dart` since its shared between both components and renaming the class something like `FlutterRelease` to more clearly signify what it represents.

parlough (approved):
One more suggestion to consider, but still looks good to me. Thanks for making those adjustments!

Feel free to merge into the migration branch when you're ready.

PR #12576, opened 2025-10-20
Fix typo in proxy config docs

nshahan:
The section of the example configuration with the replacement of `/data/` to `/report/` points to port `:3000`.

cc @kevmoo


kevmoo:
Thanks @nshahan !

PR #12578, opened 2025-10-21
Migrate client interactivity of learning resource index

schultek:
Migrates the filtering of learning resources (at /reference/learning-resources) to Jaspr.

Part of #12548 

## Presubmit checklist

- [ ] If you are unwilling, or unable, to sign the CLA, even for a _tiny_, one-word PR, please file an issue instead of a PR.
- [ ] If this PR is not meant to land until a future stable release, mark it as draft with an explanation.
- [ ] This PR follows the [Google Developer Documentation Style Guidelines](https://developers.google.com/style)—for example, it doesn't use _i.e._ or _e.g._, and it avoids _I_ and _we_ (first-person pronouns).
- [ ] This PR uses [semantic line breaks](https://github.com/dart-lang/site-shared/blob/main/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks)
  of 80 characters or fewer.


site/web/assets/js/learning-resources-index.js
```diff
@@ -1,344 +0,0 @@
...
-    // The items inside the value array correspond to the 'type' property that
-    //      can exist in the metadata (on items in the files in src/_data/learning-resources-index)
-    resourceTypeMapping: {
-        'tutorial': ['codelab', 'tutorial'],
-        'sample code': ['quickstart', 'demo', 'sample', 'sample code'],
-        'workshop': ['workshop', 'video'],
-        'recipe': ['recipe', 'how to', 'cookbook']
-    },
-    resourceTagMapping: {
-        'ai': ['ai', 'gemini', 'llm'],
```
schultek, line 17: @parlough I noticed these 'alternative tags' are not part of the new resource data at `lib/src/data/learning_resources.dart`. Is that ok or even intentional?

The effect is that for some specific search queries or filters the results are slightly different. (E.g. try selecting the "Cupertino" or "Design" tags)

site/lib/src/components/client/learning_resource_filters_sidebar.dart
```diff
@@ -0,0 +1,221 @@
...
+      final element =
+          web.document.getElementById('filter-${tag.id}') as HTMLInputElement?;
+      element?.checked = false;
+    }
+    for (final type in LearningResourceType.values) {
+      final element =
+          web.document.getElementById('filter-${type.id}') as HTMLInputElement?;
+      element?.checked = false;
+    }
+  }
```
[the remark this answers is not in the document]
schultek: `checked` is not a Jaspr api atm.

site/web/assets/js/learning-resources-index.js
parlough: When I migrated the data from YAML to Dart I didn't realize this system existed.

I think it's beneficial to keep this though. Perhaps we can have two enums, one for the source tags, then another for the filter categories that are determined from the source tags?

Would you mind adding and hooking something like that up? Then after this lands I can go back and add the original tags back.

schultek: How did you create the dart file from the original yml files for all resources?

schultek: I want to bring back the yaml files, I think that's easiest and also cleaner I think to have the data in yaml and not dart files.

parlough: > How did you create the dart file from the original yml files for all resources?

I used an LLM to generate a bespoke migration script but I didn't keep it around.

> I want to bring back the yaml files, I think that's easiest and also cleaner I think to have the data in yaml and not dart files.

That works for me @schultek! Feel free to copy them from the `main` channel or whatever works best for you

schultek: Done, I think its cleaner now. I could also reuse the LearningResource model across server and client.

parlough (approved):
Love it!

Thanks for considering the ergonomics of the data and noticing the regressed tag functionality.

PR #12587, opened 2025-10-22
feat: migrate remaining logic from main.js

schultek:
Part of #12548

Migrate the remaining logic from main.js to Dart.


site/lib/src/client/global_scripts.dart
```diff
@@ -235,40 +245,6 @@ void _activateTabWithSaveId(web.HTMLElement tabWrapper, String saveId) {
   }
 }
 
-enum _ClientOperatingSystem {
```
parlough, line 238: Thanks for consolidating this logic!

PR #12588, opened 2025-10-22
feat: organize components

schultek:
Part of #12548

Reorganizes the Jaspr components to the following:

- Common components (Button, Card, etc) under `components/common`
  - @client components of these under `components/common/client`
-  Dartpad related under `components/dartpad`
- Layout components (Header, Footer, SideNav, etc.) under `components/layout`
- Page specific components (ArchiveTable, etc.) under `components/pages`

Also cleans up the initialization of custom components for `jaspr_content`, by moving any attribute parsing logic to the respective component.

### DIFF
--- a/site/lib/src/components/common/youtube_embed.dart
+++ b/site/lib/src/components/common/youtube_embed.dart
@@ -44,10 +44,25 @@ class YoutubeEmbed with CustomComponentBase {
     // we use https://github.com/justinribeiro/lite-youtube which
     // lazily loads the video, significantly reduces page load times,
     // and enables configurability through element attributes.
-    return raw('''
-<lite-youtube videoid="$videoId" videotitle="$videoTitle" videoStartAt="$startTime" ${playlistId != null ? 'playlistid="$playlistId"' : ''}>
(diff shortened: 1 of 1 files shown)


site/lib/src/components/common/youtube_embed.dart
```diff
@@ -0,0 +1,53 @@
...
+      startTime = 0;
+      videoId = rawVideoId;
+    }
+
+    // Instead of directly including a YouTube embed iframe,
+    // we use https://github.com/justinribeiro/lite-youtube which
+    // lazily loads the video, significantly reduces page load times,
+    // and enables configurability through element attributes.
+    return raw('''
+<lite-youtube videoid="$videoId" videotitle="$videoTitle" videoStartAt="$startTime" ${playlistId != null ? 'playlistid="$playlistId"' : ''}>
```
parlough: Not necessary for this PR, but could this actually use `Component.element` so handling the attributes is cleaner than all this string interpolation? Or would that not work for a custom element like this?

schultek: On the client idk, but since this is pre-rendered probably yet, let me try

schultek: Works.

PR #12600, opened 2025-10-28
feat: add a tooltip for each api docs link

schultek:
This adds a dynamic tooltip to each link pointing to the flutter api docs. The tooltip shows a short preview of the docs on hover (or focus, or tap on mobile). 

Bildschirmfoto 2025-10-28 um 12 14 22

### DIFF
--- a/site/lib/src/components/common/client/api_link_tooltip.dart
+++ b/site/lib/src/components/common/client/api_link_tooltip.dart
@@ -1,0 +1,173 @@
+// Copyright 2025 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:http/http.dart' as http;
+import 'package:jaspr/jaspr.dart';
+import 'package:meta/meta.dart';
+import 'package:universal_web/js_interop.dart';
+import 'package:universal_web/web.dart' as web;
+
+import '../tooltip.dart';
+
+@client
+class ApiLinkTooltip extends StatefulComponent {
+  const ApiLinkTooltip({required this.url, required this.text, super.key});
+
+  final String url;
+  final String text;
+
+  @override
+  State<ApiLinkTooltip> createState() => _ApiLinkTooltipState();
+}
+
+class _ApiLinkTooltipState extends State<ApiLinkTooltip> {
+  Component? tooltipContent;
+
+  @override
+  void initState() {
+    super.initState();
+
+    if (kIsWeb) {
+      setupTooltip();
+    }
+  }
+
+  @awaitNotRequired
+  Future<void> setupTooltip() async {
+    final (extractedHeader, extractedDescription) = await scrapeApiDocs(
+      component.url,
+    );
+
+    if (extractedHeader == null && extractedDescription == null) {
+      return;
+    }
+
+    if (!mounted) return;
+    setState(() {
+      tooltipContent = fragment([
+        if (extractedHeader != null)
+          span(classes: 'tooltip-header', [raw(extractedHeader)]),
+        if (extractedDescription != null)
+          span(classes: 'tooltip-content', [raw(extractedDescription)]),
+      ]);
+    });
+  }
+
+  @override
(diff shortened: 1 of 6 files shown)


parlough:
Exciting! I haven't reviewed the code yet, but is there some common popover/tooltip component we can extract out that this, glossary terms, and whatever else can share?

schultek:
Yes, but I would do it when we actually have a second use-case for this. But I tried to keep this and the glossary cards in site-www already pretty similar.

site/lib/src/components/common/client/api_link_tooltip.dart
```diff
@@ -0,0 +1,256 @@
...
+
+@awaitNotRequired
+Future<(String?, String?)> scrapeApiDocs(String url) async {
+  try {
+    final response = await http.get(Uri.parse(url));
+    var content = response.body;
+
+    final startIndex = content.indexOf(RegExp('<div\\s+id="$contentId"'));
+    if (startIndex == -1) {
+      print(
```
parlough, line 81: When it fails to fetch the API docs can it handle it be handled more gracefully? Like remove the underline and hover state.

schultek: Done

parlough:
@schultek When you get a chance can you rebase this on `main`?

schultek:
@parlough Done

parlough:
Overall looks great to me, but won't we need to update the CORS configuration on api.flutter.dev for this to work? How did you test this?

schultek:
> Overall looks great to me, but won't we need to update the CORS configuration on api.flutter.dev for this to work? How did you test this?

Oh yes good that you caught this, I fully forget to raise this. 
Because locally I use a cors extension in chrome that mitigates this for testing.

site/lib/src/components/common/client/api_link_tooltip.dart
```diff
@@ -0,0 +1,252 @@
...
+  }
+}
+
+const contentId = 'dartdoc-main-content';
+// This seems to be a good limit to avoid overly small or large tooltips.
+const maxDescriptionLength = 400;
+const minTrailingParagraphLength = 20;
+
+@awaitNotRequired
+Future<(String?, String?)> scrapeApiDocs(String url) async {
```
ditman, line 74: Why scrape the docs? We also control the generation of the docs, could this snippet be added to the API doc generation? This sounds like useful functionality for the whole ecosystem, not just api.fluter.dev

parlough: > We also control the generation of the docs, could this snippet be added to the API doc generation?

Good point! I suppose it could or at least more easily extracted from a general Markdown version of each page.

My hopeful thinking was to extract the API information directly with https://github.com/dart-lang/labs/tree/main/pkgs/jot or something else more we build. 

For now this scraping is a great way to quickly show off what this could look like though :D

schultek: I fully agree that this can and should be improved in the future. I was mainly going for a quick win here to show the concept.

This will anyways have to change when the api docs are being integrated in the docs (which as I understood is the mid-term goal).

site/lib/src/extensions/registry.dart
```diff
@@ -20,4 +21,5 @@ const List<PageExtension> allNodeProcessingExtensions = [
   TableWrapperExtension(),
   CodeBlockProcessor(),
   GlossaryLinkProcessor(),
+  ApiLinkProcessor(),
```
kevmoo, line 24: How does this affect load times?
Is it lazy?

schultek: Yes its lazy and only happens after load so it doesn't block anything.

kevmoo: It's like you know how to write software or something. Cool! 😜

parlough:
I'm going to close this PR to keep the queue up-to-date, but let's keep this branch open and revisit it again soon. I'm still super excited about this, no matter how we retrieve the information :D

PR #12613, opened 2025-10-30
Added impellerc example for auditing fragment shaders

gaaclarke:

Adds low level example of how to audit the output of impellerc.

https://github.com/flutter/flutter/issues/171471

## Presubmit checklist

  of 80 characters or fewer.


gaaclarke:
@chinmaygarde does this satisfy what you were thinking for https://github.com/flutter/flutter/issues/171471?  I tried to use `--reflection-json` but it doesn't work for `--runtime-stage-metal`, it doesn't print out the uniforms.  We should probably fix that but this is the correct instructions for `stable`.

sfshaza2 (approved):
This lgtm, so I will approve, but don't land until @chinmaygarde has had a chance to review.

chinmaygarde (approved):
This lgtm. I suppose we could also make impellerc do the same thing instead of depending on flatc. But that can be an additional improvement for later. This is a good start.

Issue #12660, opened 2025-11-07
Opening the CHANGELOG link inside release notes leads to the 404 page


iakdis:
### Page URL

https://docs.flutter.dev/release/release-notes/release-notes-3.35.0

### Page source

https://github.com/flutter/website/blob/main/src/content/release/release-notes/release-notes-3.35.0.md

### Describe the problem

When opening the **CHANGELOG** link of any version inside the release notes (_3.35.0 release notes & change log_), it leads to the 404 page. This applies to all versions of the release notes. For the sake of describing this issue, the examples are based on the latest version 3.35.0.

Location to reproduce the issue: https://docs.flutter.dev/release/release-notes/release-notes-3.35.0 (see screenshot)

<img width="1790" height="814" alt="Image" src="https://github.com/user-attachments/assets/3fa638e6-78fb-449b-be47-4f43d46e8e41" />

Destination location of the **CHANGELOG** link: https://docs.flutter.dev/release/release-notes/{{site.repo.flutter}}/blob/main/CHANGELOG.md (see screenshot)

<img width="1792" height="823" alt="Image" src="https://github.com/user-attachments/assets/7f06623f-f3ab-49cc-a741-f435666b7768" />

### Expected fix

The **CHANGELOG** link should open the correct GitHub URL of the CHANGELOG.md file (https://github.com/flutter/flutter/blob/main/CHANGELOG.md).

### Additional context

The problem seems to occur due to the Templating system not working for the snippet `{{site.repo.flutter}}`. What differentiates the release notes pages from others is the option `skipTemplateRendering` in the front matter, which is set to `true`. While testing, removing the option from the front matter makes the link open as expected. 

### I would like to fix this problem.

- [x] I will try and fix this problem on docs.flutter.dev.

PR #12661, opened 2025-11-07
Remove `skipTemplateRendering` from release notes front matter

iakdis:
_Description of what this PR is changing or adding, and why:_ The problem that is described in the linked issue seems to occur due to the Templating system not working for the snippet `{{site.repo.flutter}}`. What differentiates the release notes pages from others is the option `skipTemplateRendering` in the front matter, which is set to `true`. While testing, removing the option from the front matter makes the link open as expected.

_Issues fixed by this PR (if any):_ https://github.com/flutter/website/issues/12660

## Presubmit checklist

  of 80 characters or fewer.

### DIFF
--- a/src/content/release/release-notes/release-notes-3.0.0.md
+++ b/src/content/release/release-notes/release-notes-3.0.0.md
@@ -9,7 +9,7 @@ This page has release notes for 3.0.0.
 For information about subsequent bug-fix releases,
 see our [CHANGELOG][].
 
-[CHANGELOG]: {{site.repo.flutter}}/blob/main/CHANGELOG.md
+[CHANGELOG]: https://github.com/flutter/flutter/blob/main/CHANGELOG.md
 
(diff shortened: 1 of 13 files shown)


parlough (commented):
Thanks for noticing this and opening a PR to fix the issue.

Unfortunately, there's a performance issue with the Liquid templating package that we use that results in these pages taking too long to render with it. Since we haven't got around to fixing it yet, we'll need to keep skipping this for now.

Instead of doing this, could you instead replace the usage of the Liquid variables on these pages with the underlying values? I think it's fine if these usages aren't configurable.

iakdis:
I see, thank you for the information @parlough! I have reverted the changes and added a new commit which replaces the Liquid variables used on these pages according to the variable defined in `site.yml`.

parlough (approved):
Thanks again for noticing this and fixing this! I really appreciate :D

PR #12671, opened 2025-11-11
Updating the banner

sfshaza2:
Promoting the live stream

### DIFF
--- a/src/data/banner.yml
+++ b/src/data/banner.yml
@@ -1,5 +1,5 @@
 text: >-
-  Don't miss the Flutter 3.38, Dart 3.10, and
+  Don't miss the live Flutter and Dart
   live Q & A happening Wednesday, Nov 12 at 11am PST.
 link:
   text: "Be there!"


src/data/banner.yml
```diff
@@ -1,7 +1,7 @@
 text: >-
-  Build high quality, feature-rich apps with
-  the new Flutter Extension for Gemini CLI.
+  Don't miss the Flutter 3.38, Dart 3.10, and
```
parlough: Should this say release somewhere or are these meant to describe the live Q&A?

src/data/banner.yml
```diff
@@ -1,7 +1,7 @@
...
+  live Q & A happening Wednesday, 11/12 at 11am PT.
```
parlough:
```suggestion
  live Q & A happening Wednesday, Nov 12 at 11am PST.
```

src/data/banner.yml
sfshaza2: I'd rather leave the word "release" out of it.

parlough: That's fine, but the sentence reads a bit awkward to me right now.

> Don't miss the Flutter 3.38, Dart 3.10, and live Q & A ...

So are "Flutter 3.38", "Dart 3.10", and "live" all separately describing the Q & A?

johnpryan: If we want to leave "release" out, maybe we should move the "the":

> Don't miss Flutter 3.38, Dart 3.10, and the live Q & A happening Wednesday, Nov 12 at 11am PST

craiglabenz:
Should we revert this now, @sfshaza2 and @parlough? The banner is still up, but event is no longer _today_.

PR #12676, opened 2025-11-12
Switch to a more flexible banner format

parlough:
Switch from defining the banner in `src/data/banner.yml` to the `bannerHtml` field in the `src/data/site.yml` banner alongside `showBanner`.

I'm not sure this is the solution I want to use long term, but it works for now.


sfshaza2 (approved):
lgtm. THANKS!

PR #12680, opened 2025-11-13
Add folding support to code blocks

schultek:
This adds support for defining **folding ranges** in any code block.

Folding ranges can be defined by specifying `foldable` on the code block and using `[*` and `*]` lines in the code, and will be transformed to collapsible sections with a toggle marker and indicator line. Folding ranges can be nested. By default folding ranges are open, but can be set to initially closed using `[* -`.

When using line numbers, the folding range markers and lines will be shifted to the right of the line numbers.

- Code block with line numbers and collapsed folding ranges: 
Bildschirmfoto 2025-11-13 um 13 13 29
- Code block with line numbers and expanded folding ranges: 
Bildschirmfoto 2025-11-13 um 13 13 40
- Code block without line numbers and partly collapsed folding ranges: 
Bildschirmfoto 2025-11-13 um 13 14 02

### DIFF
--- a/src/content/get-started/fundamentals/widgets.md
+++ b/src/content/get-started/fundamentals/widgets.md
@@ -38,11 +38,9 @@ can receive context from the parent.
 This structure carries all the way up to the root
 widget, as this trivial example shows:
 
-```dart foldable showLineNumbers
-[* -
-import 'package:flutter/material.dart'; 
+```dart
+import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
-*]
 
 void main() => runApp(const MyApp());
 
@@ -51,20 +49,17 @@ class MyApp extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    [* -
     return MaterialApp( // Root widget
       home: Scaffold(
         appBar: AppBar(
           title: const Text('My Home Page'),
         ),
         body: Center(
-          [*
           child: Builder(
             builder: (context) {
               return Column(
                 children: [
                   const Text('Hello, World!'),
-                  [* -
                   const SizedBox(height: 20),
                   ElevatedButton(
                     onPressed: () {
@@ -72,16 +67,13 @@ class MyApp extends StatelessWidget {
                     },
                     child: const Text('A button'),
                   ),
-                  *]
                 ],
               );
             },
           ),
-          *]
         ),
       ),
     );
-    *]
   }
 }
 ```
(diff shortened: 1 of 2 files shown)


src/content/get-started/fundamentals/widgets.md
schultek, line 1: This is only for demoing this feature and should be reverted before merge.

site/lib/_sass/components/_code.scss
```diff
@@ -142,13 +142,110 @@ pre {
...
+
+      summary {
+        margin: 0;
+        font-weight: initial;
+        user-select: initial;
+        list-style: none;
+        position: relative;
+
+        // Caret at the beginning of the folding range. 
+        span.material-symbols {
```
parlough, line 193: Can you add `user-select: none` to this icon as I kept auto-selecting it when opening.

site/lib/src/components/common/wrapped_code_block.dart
```diff
@@ -43,14 +46,67 @@ final class WrappedCodeBlock extends StatelessComponent {
...
+        ];
+
+        children[start - 1] = details(
+          open: open,
+          styles: Styles(raw: {'--level': '$level'}),
+          [
+            summary(
+              classes: 'fold-summary',
+              [
+                const MaterialIcon('keyboard_arrow_up'),
```
parlough: Seems this should be switched to `keyboard_arrow_down` or the rotation in CSS swapped.

site/lib/src/components/common/wrapped_code_block.dart
```diff
@@ -43,14 +46,67 @@ final class WrappedCodeBlock extends StatelessComponent {
...
+    if (foldingRanges.isNotEmpty) {
+      for (final (:start, :end, :level, :open) in foldingRanges) {
+        final foldingSummary = children.remove(start - 1);
+        final foldedChildren = [
+          for (var i = start + 1; i <= end; i += 1) ?children.remove(i - 1),
+        ];
+
+        children[start - 1] = details(
+          open: open,
+          styles: Styles(raw: {'--level': '$level'}),
```
parlough, line 88: Clever way of going about this!

parlough (commented):
Overall looks great to me, thanks!

Also gives me some ideas of how to best extract the code block logic.

site/lib/_sass/components/_code.scss
schultek: done

site/lib/src/components/common/wrapped_code_block.dart
schultek: I changed it to right for closed and down for open, to match the behavior of vscode.

PR #12687, opened 2025-11-14
Update site banner to link to umbrella blog post for Flutter.

antfitch:


parlough:
I think Shams wants to add some other text and links, but let's land this and then she can update it again.

Issue #12628, opened 2025-11-03
Update links on FWE index page


parlough:
### Page URL

https://docs.flutter.dev/tutorial

### Page source

https://github.com/flutter/website/blob/main/src/content/tutorial/index.md

### Describe the problem

The links on the index page aren't functional. I guess due to the directory structure changing at some point?

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

parlough:
As a side note, I believe the link checker didn't catch these because these pages are _(purposefully)_ not linked to in the sitemap for now. I'll investigate how we can still get the checking for hidden pages once I have a chance :)

ericwindmill:
> As a side note, I believe the link checker didn't catch these because these pages are _(purposefully)_ not linked to in the sitemap for now. I'll investigate how we can still get the checking for hidden pages once I have a chance :)

This is strange because I've used these links many times in the past. Could it be your hunch + a difference with the new Jaspr infra? 

Either way, I'll fix the links when we're ready to land FWE, because theres a good chance we'll change the URLs to something like /learn/lesson-title or whatever

parlough:
Closing this as the links work now after https://github.com/flutter/website/commit/e528aea11e2d0158430181e6082e9bb46ca163e0, thanks Eric!

Once we release the docs we can ensure we're happy with their published paths/links.

PR #12707, opened 2025-11-17
Fix links on FWE index page

ericwindmill:

Since the migration to Jaspr, the links on the (unlisted) FWE index page  no longer work. I believe this is a problem with permalinks, but this PR _only updates the link URLS_.  

We'll need to update the links anyway when FWE lands, so this is not creating more work.

This is a **temporary** fix for #12628

## Presubmit checklist

  of 80 characters or fewer.


parlough (approved):
Yeah, we don't currently support for the `permalink` frontmatter option.

Thanks for updating these.

PR #12711, opened 2025-11-18
Simplify design systems section on widget catalog index page

parlough:
This loop previously looped through items to render them but manually checked their values anyway. Instead of relying on harder to understand Liquid templating logic, just inline the intended cards.

**Staged:** https://flutter-docs-prod--pr12711-misc-widget-catalog-index-s-dzp1gfca.web.app/ui/widgets#design-systems


sfshaza2 (approved):
lgtm

PR #12715, opened 2025-11-18
Add optional page navigation to top toc dropdown (for FWE)

schultek:
This extends the top floating table-of-contents dropdown by an optional page navigation list, for use inside FWE tutorials.

Bildschirmfoto 2025-11-18 um 14 04 38

---

Differences to the current toc:

- Instead of "On this page" shows the current page title
- When open, shows a list of pages (including chapters).
  - The current page's toc is nested under the entry of the current page (design & behavior stays the same) 

To test this, I also created a new `TutorialLayout` that hides the sidebar and forces the floating toc bar on all screen sizes. This should be improved in the future.

### DIFF
--- a/site/lib/_sass/components/_pagenav.scss
+++ b/site/lib/_sass/components/_pagenav.scss
@@ -24,6 +24,10 @@
       color: var(--site-base-fgColor-alt);
       font-size: 20px;
     }
+
+    >.material-symbols:first-child {
+      margin-right: 0.25rem;
+    }
   }
 
   .toc-breadcrumb {
@@ -31,16 +35,25 @@
     white-space: nowrap;
     overflow: hidden;
 
-    &:first-child .material-symbols {
-      margin-right: 0.25rem;
+    &.toc-hide-medium {
+      @media (width < 576px) {
+        display: none;
+      }
     }
 
-    span:last-child {
+    &.toc-hide-small {
+      @media (width < 420px) {
+        display: none;
+      }
+    }
+
+    span:not(.material-symbols) {
       overflow: hidden;
       text-overflow: ellipsis;
     }
 
     .page-number {
+      flex-shrink: 0;
       height: 1.3rem;
       width: 1.3rem;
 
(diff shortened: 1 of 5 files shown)


parlough:
Thanks for working on this @schultek :D - The direction looks great.

Let me know when it's ready for review.

schultek:
@parlough It's ready now

parlough:
I really like it. I haven't looked at the code yet, but a few styling issues I noticed:

1.  Hovering top-level entries in the dropdown underlines the text, including the number in the circle. Perhaps use the same hover/active styles of TOC dropdown portion?

    Screenshot of hovering an entry in the dropdown.

2.  It's nice the chapter circle is visible in the TOC box, but it's not centered. Can we vertically center it? If needed, we can make this version of the box taller.

    Screenshot of the chapter circle not being centered.

3.  The background of the uncompleted chapter circles is not visible in dark mode.

    Screenshot of the uncompleted chapter circle in dark mode being mostly not visible.

schultek:
1. and 3. is fixed.

2. I agree that it looks slightly off, but it is actually centered. I think its looks like this because the "4" is visually unbalanced, it looks perfect with an "8".

parlough:
Thanks for those fixes! Looks good to me @schultek :D 

Now that the `tutorial.yml` setup you made has landed, can you update this so the actual pages show this setup and use that data? Thank you!

parlough:
@schultek Mostly looks amazing to me! Thank you. I pushed a few small style changes and hid the other breadcrumbs on the page.

One remaining issue is on narrower layouts, the navigation gets quite cramped:

Screenshot of cramped page navigation on narrow screen layout.

On narrower tutorial layouts, could we perhaps drop the other portions and just show the current lesson/page title? Or do you have any other ideas?

schultek:
I'm now hiding breadcrumbs on smaller screens:

https://github.com/user-attachments/assets/51935bbe-edbb-47c2-bd98-7d9db0bcd696

PR #12726, opened 2025-11-19
Refactor to more idiomatic Jaspr code

schultek:
This refactors some parts of the code to have more idiomatic Jaspr components and reduce logic in `global_scripts.dart`.

- Implement a new `ComponentRef` system that in some capacity works similar to server components (more limited and less elegant, but functionally working). This is to unblock some improvements (some of which are included in this PR) to the interactive components of the website, as we can already start architecting towards server components and write less manual dom logic.
- Change the `Dropdown` component to accept direct `toggle` and `content` components, which simplifies it's implementation. Migrate existing uses of `Dropdown`.
- Migrate the top TOC (renamed to `PageNav`) to use the `Dropdown` component, as well as the new `ComponentRef`s. Split into separate "site-subheader" and "pagenav" elements. Preparations for #12715 
- Add a new `Tooltip` component and migrate the glossary tooltips to it using the new `ComponentRef`s. Can be reused for #12600

### DIFF
--- a/site/lib/src/style_hash.dart
+++ b/site/lib/src/style_hash.dart
@@ -2,4 +2,4 @@
 // dart format off
 
 /// The generated hash of the `main.css` file.
-const generatedStylesHash = 'Nx/ufc4onkpD';
+const generatedStylesHash = 'VAQsUT7crZAw';
--- a/src/content/ai/genui/get-started.md
+++ b/src/content/ai/genui/get-started.md
@@ -178,18 +178,13 @@ The main components in this package include:
 
 Follow these instructions:
 
- 1. Add the following to your `pubspec.yaml` file:
+ 1. Set up dependencies:
+    Use `dart pub add` to add `genui`, `genui_a2ui`, and `a2a` as
+    dependencies in your `pubspec.yaml` file.
 
-    ```yml
-    dependencies:
-      flutter:
-        sdk: flutter
-      genui: ^latest_version      # Replace with the actual latest version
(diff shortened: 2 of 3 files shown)


site/lib/src/components/common/dropdown.dart
```diff
@@ -41,93 +44,45 @@ final class _DropdownState extends State<Dropdown> {
...
+            final keydownEvent = e as web.KeyboardEvent;
+            if (_expanded && keydownEvent.key == 'Escape') {
+              toggle(to: false);
+            }
+          },
+          'focusout': (e) {
+            final relatedTarget =
+                (e as web.FocusEvent).relatedTarget as web.HTMLElement?;
+            if (relatedTarget != null &&
+                relatedTarget.closest('#${component.id}') == null) {
```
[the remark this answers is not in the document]
schultek: done

site/lib/src/components/layout/client/pagenav.dart
```diff
@@ -0,0 +1,89 @@
...
+      });
+    }
+  }
+
+  @override
+  Component build(BuildContext context) {
+    return Dropdown(
+      key: dropdownKey,
+      id: 'pagenav',
+      toggle: button(
```
parlough, line 60: This button used to be across the entire bar, making the whole thing clickable, but now just the text is clickable.

Can you update the styles so the button spans the whole bar again? Other than that, looks good to me. Thank you!

schultek: Done

PR #12733, opened 2025-11-20
Add DownloadableSnippet component

schultek:
Adds a component for downloading a code snippet as a file. This is displayed as a collapsed code block with additional action buttons in the title.

Provides three options:
- Download directly using browser apis
- Copy code to clipboard
- Expand to show full code

Bildschirmfoto 2025-11-21 um 15 24 22

---

I also updated the `CopyButton` to extract its content from the actual code element instead of taking it as an parameter, to avoid duplicating the data in the output html.

### DIFF
--- a/site/lib/_sass/components/_button.scss
+++ b/site/lib/_sass/components/_button.scss
@@ -94,3 +94,22 @@ button {
     }
   }
 }
+
+.segmented-button {
+  display: inline-flex;
+
+  a,
+  button {
+    &:not(:last-child) {
+      border-top-right-radius: 0;
+      border-bottom-right-radius: 0;
+
+      border-right: 1px solid var(--site-outline-variant);
+    }
+
+    &:not(:first-child) {
+      border-top-left-radius: 0;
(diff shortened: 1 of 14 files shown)


site/lib/src/components/tutorial/client/file_download_button.dart
```diff
@@ -0,0 +1,140 @@
...
+      anchor.remove();
+    }
+  }
+
+  void copyToClipboard() {
+    web.window.navigator.clipboard.writeText(component.text);
+  }
+
+  void openInNewTab() {
+    web.window.open(objectUrl, '_blank');
```
schultek: It seems like on Chrome when opening an Object URL in a new tab like this, it blocks the "Save as" option (both through using Cmd+S and the context menu), so this is actually not that useful. Wdyt?

schultek:
@parlough This is now updated to be a collapsed code block instead of just a button.

parlough (approved):
This is perfect and works well across the browsers I tested. Thanks for making those adjustments @schultek!

PR #12750, opened 2025-11-25
Add redirect from `/download` to `/install`

ericwindmill:
This PR adds a redirect from docs.flutter.dev/download -> /install

This is precautionary, as we're adding a /download redirect from the marketing site to docs.flutter.dev/install, which is being advertised in a FWE video, and folks might accidentally try the url on the docs site.


parlough (approved):
Makes sense. Thanks for thinking about this!

PR #12752, opened 2025-11-26
Fix style of leading console token in code blocks

schultek:
Makes console tokens (`$`) in code blocks a different color and prevents text selection.


schultek:
@parlough The nicer way would of course be to do this in the `opal` tokenizer as a separate `console` language syntax, but I opted for the quick win. We can improve this in the future.

PR #12754, opened 2025-11-26
add page navigation and prefetching to the tutorial

schultek:
Adds a new `tutorial.yml` file that contains the full unit structure of the tutorial (and in the future maybe additional details). It is used to automatically create:

- The tutorial outline on the tutorial index page.
- Prev and Next page buttons for each unit.
- A prefetch link for the next unit.

### DIFF
--- a/src/data/tutorial.yml
+++ b/src/data/tutorial.yml
@@ -28,7 +28,7 @@ units:
         url: /tutorial/state/listenable-builder/
   - title: Flutter UI 102
     chapters:
-      - title: Set up your project
+      - title: Advanced UI features
         url: /tutorial/ui-102/intro/
       - title: "`LayoutBuilder` and adaptive layouts"
         url: /tutorial/ui-102/adaptive-layout/
--- a/site/lib/src/components/common/prev_next.dart
+++ b/site/lib/src/components/common/prev_next.dart
@@ -4,6 +4,7 @@
 
 import 'package:jaspr/jaspr.dart';
 
+import '../../markdown/markdown_parser.dart';
 import 'material_icon.dart';
 
 /// Previous and next page buttons to display at the end of a page
@@ -49,7 +50,9 @@ class _PrevNextCard extends StatelessComponent {
           attributes: {'aria-label': ariaLabel},
           [text(subtitle)],
         ),
-        span(classes: 'prev-next-title', [text(page.title)]),
+        span(classes: 'prev-next-title', [
+          DashMarkdown(inline: true, content: page.title),
+        ]),
       ]),
       if (!isPrevious) const MaterialIcon('chevron_right'),
     ]);


src/data/tutorial.yml
```diff
@@ -0,0 +1,38 @@
+title: Learn Flutter
+chapters:
+  - title: Introdution to Flutter UI
+    units:
```
parlough: Let's flip these two. Generally units contain chapters rather than the other way around.

src/data/tutorial.yml
parlough, line 1: Now that we have this file to set the order, can you remove the numbers from the paths?

site/lib/src/extensions/tutorial_prefetch_processor.dart
```diff
@@ -0,0 +1,95 @@
...
+      },
+    );
+
+    if (nextUnit == null) {
+      return nodes;
+    }
+
+    return [
+      ComponentNode(
+        Document.head(
```
parlough, line 75: I haven't had a chance to use this functionality a lot, but this ends up being super nice.

site/lib/src/extensions/tutorial_prefetch_processor.dart
```diff
@@ -0,0 +1,95 @@
...
+    };
+
+    var chapterIndex = -1;
+    var unitIndex = -1;
+
+    final normalizedPageUrl = page.url.endsWith('/')
+        ? page.url
+        : '${page.url}/';
+
+    outer:
```
parlough: This and the following logic with nesting ends up a bit hard to follow. Since we only need the order of the smaller portions _(the chapters, but in this code they're called units)_, could you instead create an ordered list of just the chapters first:

```dart
final allChapters = [
  for (final unit in tutorial.units) 
    ...unit.chapters,
];
```

Then the following logic doesn't need to deal with the nesting.

site/lib/src/components/tutorial/tutorial_outline.dart
```diff
@@ -0,0 +1,50 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:jaspr/jaspr.dart';
+import 'package:jaspr_content/jaspr_content.dart';
+
+import '../../markdown/markdown_parser.dart';
+import '../../models/tutorial_model.dart';
+
+class TutorialOutline extends CustomComponentBase {
```
parlough, line 11: Nice idea tackling this all together! Ended up great.

src/data/tutorial.yml
```diff
@@ -0,0 +1,38 @@
...
+        url: /tutorial/state/set-up-project/
+      - title: Make Http Requests
+        url: /tutorial/state/http-requests/
+      - title: Use `ChangeNotifier` to update app state
+        url: /tutorial/state/change-notifier/
+      - title: Use `ListenableBuilder` to update app UI
+        url: /tutorial/state/listenable-builder/
+  - title: Flutter UI 102
+    chapters:
+      - title: Set up your project
```
parlough:
```suggestion
      - title: Advanced UI features
```

src/data/tutorial.yml
```diff
@@ -0,0 +1,38 @@
...
+        url: /tutorial/ui/stateful-widget/
+      - title: Add implicit animations
+        url: /tutorial/ui/implicit-animations/
+  - title: State in Flutter apps
+    chapters:
+      - title: Set up a new project
+        url: /tutorial/state/set-up-project/
+      - title: Make Http Requests
+        url: /tutorial/state/http-requests/
+      - title: Use `ChangeNotifier` to update app state
```
parlough, line 25: It seems the previous and next page buttons aren't set up to format inline markdown yet. Can you add that support?

Issue #12739, opened 2025-11-24
Migrate widget catalog pages and index away from Liquid logic


parlough:
- [x] https://github.com/flutter/website/blob/main/src/content/ui/widgets/index.md
  - Migrated by https://github.com/flutter/website/pull/12755
- [ ] https://github.com/flutter/website/blob/main/src/content/reference/widgets.md
- [ ] Does it make sense to generate the empty Markdown files in https://github.com/flutter/website/tree/main/src/content/ui/widgets from the data?

PR #12755, opened 2025-11-26
Add `WidgetCatalogCategories` component

schultek:
Contributes to #12739


parlough:
To fix the failing test, I pushed a change that uses the specified IDs instead of the lowercase names. After that, looks good to me. Thanks :D

Issue #12746, opened 2025-11-25
Hot reload on web no longer needs experiment


sigurdm:
### Page URL

https://docs.flutter.dev/platform-integration/web/faq#does-hot-reload-work-with-a-web-app

### Page source

https://github.com/flutter/website/blob/b9c666a87ed87e35da453a31e889766f35d33388/src/content/platform-integration/web/faq.md?plain=1#L43

### Describe the problem

According to https://docs.flutter.dev/platform-integration/web/building#hot-reload-web hot reload on web is enabled by default. But the faq note seems to have not been updated accordingly.

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

Issue #12644, opened 2025-11-05
Add glossary entries for "Hot reload" and "Hot restart"


parlough:
Add a glossary entry for "Hot reload" and "Hot restart" in the context of Flutter to https://github.com/flutter/website/blob/main/src/data/glossary.yml.

PR #12757, opened 2025-11-26
Add glossary entries for hot reload and hot restart

sfshaza2:
Also updates the web faq regarding hot reload.

Fixes https://github.com/flutter/website/issues/12644
Fixes https://github.com/flutter/website/issues/12746

### DIFF
--- a/src/content/tools/hot-reload.md
+++ b/src/content/tools/hot-reload.md
@@ -75,9 +75,11 @@ Your app continues to execute from where it was prior
 to run the hot reload command.
 The code updates and execution continues.
 
+<a id="hot-restart" aria-hidden="true"></a>
+
 :::secondary
 **What is the difference between hot reload, hot restart,
-and full restart?** {:#hot-restart}
+and full restart?**
 
 * **Hot reload** loads code changes into the VM or the browser,
   and re-builds the widget tree, preserving the app state;
--- a/src/data/glossary.yml
+++ b/src/data/glossary.yml
@@ -5,6 +5,7 @@
     a running application in the Dart VM and see the changes immediately
     while maintaining application state.
   long_description: |-
+    This feature is also called "stateful hot reload".
     After the Dart runtime updates classes with the new versions
     of fields and functions, the Flutter framework automatically
     rebuilds the widget tree, allowing you to quickly view the effects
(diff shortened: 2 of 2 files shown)


ericwindmill (approved):
LGTM, but a link is broken

src/content/tools/hot-reload.md
```diff
@@ -77,7 +77,7 @@ The code updates and execution continues.
 
 :::secondary
 **What is the difference between hot reload, hot restart,
-and full restart?**
+and full restart?** {:#hot-restart}
```
parlough: I'm not sure this will work for adding a fragment to text. Consider just adding a fragment before the note:


```suggestion
<a id="hot-restart" aria-hidden="true"></a>

:::secondary
**What is the difference between hot reload, hot restart,
and full restart?**
```

src/data/glossary.yml
```diff
@@ -1,3 +1,49 @@
+- term: "Hot reload"
+  short_description: |-
+    Also called "stateful hot reload",
+    is a Flutter feature that allows you to inject updated code into
```
parlough: I think you could maybe remove the "also called ..." portion of the short description and instead mention that in the beginning of the long description.

src/data/glossary.yml
```diff
@@ -1,3 +1,49 @@
...
+    - "fast development"
+    - "tooling"
+
+- term: "Hot restart"
+  short_description: |-
+    Similar to hot reload, but when you need to re-run `main` or
+    `initState`, use hot restart, which does not maintain app state.
+  long_description: |-
+    Hot restart is still faster than a full restart, which also
+    recompiles the Java / Kotlin / Objective-C / Swift / JavaScript code.
```
parlough: So we don't have to keep this list up to date, maybe consider just saying "native code" or "platform code".

If needed, it could provide a singular example, something like:

"recompiles the native, platform code (such as Swift)".

src/data/glossary.yml
```diff
@@ -1,3 +1,49 @@
...
+  long_description: |-
+    After the Dart runtime updates classes with the new versions
+    of fields and functions, the Flutter framework automatically
+    rebuilds the widget tree, allowing you to quickly view the effects
+    of your changes. Hot reload greatly increases the speed of development.
+
+    Hot reload works on mobile, web, and desktop apps that are
+    running in debug mode and is fully supported in VS Code,
+    Android Studio, and IntelliJ IDEA. It does not re-run `main` or
+    `initState`; for that, use hot restart.
```
parlough:
```suggestion
    `initState`; for that, use [hot restart][].
    
    [hot restart]: /resources/glossary#hot-restart
```

PR #12811, opened 2025-12-11
Move fwe content into new learn/tutorial directory

parlough:
Also flatten the structure of the directory, relying on the structuring in `tutorial.yml`. This results in simpler paths and easier changes of order or grouping in the future.


sfshaza2 (approved):
Should we have a redirect for this?  lgtm

parlough:
> Should we have a redirect for this? lgtm

Doesn't hurt! Added :)

PR #12821, opened 2025-12-14
Add initial preview and review cards to FWE

parlough:
@ericwindmill These are rather rough, so please do either modify these now or as a follow-up to match your vision for each lesson.


ericwindmill:
+1 to Shams comment... We should be intentional about the icons. If there aren't suitable icons then we can maybe have someone create them.

tbh these look pretty good to me. I'd prefer to just land them and edit them holistically when I do a final pass

PR #12845, opened 2025-12-19
[hooks] Document `flutter create --template=package_ffi`

dcharkes:
The Flutter docs were still recommending `plugin_ffi` for interop instead of `package_ffi`.

This leads to LLMs generating Plugin FFI code instead of build hooks. @antfitch believes making our documentation up to date should lead to LLMs generating the right thing.

The situation is not as clear-cut as simply deleting the old docs, as there are still valid use cases for using FFI plugins such as accessing the Flutter plugin API. However, it should be clear that the new recommended way for bundling dynamic libraries or accessing system libraries via build hooks and `@Native() external` functions.

cc @stuartmorgan-g How do you feel about marking the old FFI plugins as deprecated in the docs in accordance with https://github.com/flutter/flutter/issues/131209? Flutter Plugins (non-FFI) are not deprecated. It's the use cases using `dart:ffi` and FFIgen that should use the new path mostly.

FYI @jwill You might also be looking into docs.

cc @mariamhas This is intended to address the EAP feedback.

### DIFF
--- a/src/content/platform-integration/bind-native-code.md
+++ b/src/content/platform-integration/bind-native-code.md
@@ -8,8 +8,8 @@ for [_foreign function interface_][FFI]. Other terms for similar functionality
 include _native interface_ and _language bindings._
 
 Since Flutter 3.38, the recommended way to bind to native code is to use the
-`flutter create --template=package_ffi` command. This template uses [build
-hooks](https://dart.dev/tools/hooks) to configure the native build in a
+`flutter create --template=package_ffi` command. This template uses
+[build hooks][] to configure the native build in a
 `build.dart` script, and no longer requires OS-specific build files. This
 approach works for both Flutter and Dart standalone projects.
 
@@ -20,6 +20,7 @@ documentation:
 - [iOS](/platform-integration/ios/c-interop)
 - [macOS](/platform-integration/macos/c-interop)
 
+[build hooks]: https://dart.dev/tools/hooks
 [dart:ffi]: {{site.dart.api}}/dart-ffi/dart-ffi-library.html
 [FFI]: https://en.wikipedia.org/wiki/Foreign_function_interface
 
@@ -57,8 +58,7 @@ file. The function is marked to be exported so that it can be called from Dart.
 ## The build hook
 
 The native code is compiled and bundled with your app automatically. This is
-done by the `hook/build.dart` script, which is a [build
-hook](https://dart.dev/tools/hooks).
+done by the `hook/build.dart` script, which is a [build hook][build hooks].
 
 This means you no longer need to write OS-specific build files (like
 `CMakeLists.txt` for Linux/Windows, `.podspec` for iOS/macOS, or `build.gradle`


src/content/packages-and-plugins/developing-packages.md
```diff
@@ -710,7 +718,21 @@ check out [Flutter in plugin tests][].
...
+
+Since Flutter 3.38, the recommended way for using FFI is to use the
+`flutter create --template=package_ffi` command, which uses
+[build hooks](/platform-integration/c-interop).
+
+The legacy FFI plugins (`flutter create --template=plugin_ffi`) described here
+are still useful for some situations:
+1. to access the Flutter Plugin API,
+2. if you need to configure a Google Play services runtime on Android, or
+3. if you need to use static linking on iOS or macOS.
```
sfshaza2:
```suggestion
- To access the Flutter Plugin API.
- If you need to configure a Google Play services runtime on Android.
- If you need to use static linking on iOS or macOS.
```

The original list is a UK English style and we use American style. Also, we use unordered lists unless a specific ordering is required.

sfshaza2 (approved):
Some nits about a list. Otherwise lgtm

stuartmorgan-g:
> The situation is not as clear-cut as simply deleting the old docs, as there are still valid use cases for using FFI plugins such as accessing the Flutter plugin API.

That's not an FFI plugin though, that's an actual plugin that happens to use FFI as a communication mechanism. It shouldn't require any special template that I'm aware of.

> cc @stuartmorgan-g How do you feel about marking the old FFI plugins as deprecated in the docs in accordance with [flutter/flutter#131209](https://github.com/flutter/flutter/issues/131209)?

Yes, we should definitely update the docs as part of deprecating the template.

(To clarify, by "as part of" I just mean as part of the overall process; I don't see any reason we can't switch docs to recommending hooks immediately, even though template deprecation will take a while to reach `stable`.)

src/data/sidenav.yml
antfitch, line 1: I would remove the older c-interop topics from the TOC completely. Keep them & have links to them in the new c-interop guide, but let's not draw attention to them in the left navigation pane.

dcharkes:
> > The situation is not as clear-cut as simply deleting the old docs, as there are still valid use cases for using FFI plugins such as accessing the Flutter plugin API.
> 
> That's not an FFI plugin though, that's an actual plugin that happens to use FFI as a communication mechanism. It shouldn't require any special template that I'm aware of.

Right, we should point to the normal plugin documentation then.

However, if users end up adding C/C++ code to their plugins to use via `dart:ffi`, the documentation in the old pages on how to not get your code tree-shaken by the native compiler. Let's discuss in the next sync.

dcharkes:
@antfitch The preview website doesn't want to load https://flutter-docs-prod--pr12845-build-hooks-docs-o35es4kr.web.app/platform-integration/c-interop ? It forwards to https://flutter-docs-prod--pr12845-build-hooks-docs-o35es4kr.web.app/platform-integration/ .

PR #12852, opened 2025-12-19
Add definition for Embedder

### LINKED ISSUE #12645: Add glossary entry for "Embedder"
Add a glossary entry for "Embedder" in the context of Flutter to https://github.com/flutter/website/blob/main/src/data/glossary.yml.

sfshaza2:
Fixes #12645. Adds the glossary entry for 'Embedder' with relevant links.

### DIFF
--- a/src/data/glossary.yml
+++ b/src/data/glossary.yml
@@ -1,3 +1,27 @@
+- term: "Embedder"
+  short_description: |-
+    The platform-specific component that coordinates with the operating system
+    and manages the Flutter engine.
+  long_description: |-
+    The _embedder_ provides the entrypoint for a Flutter app.
+    It coordinates with the underlying operating system to access
+    services like rendering surfaces, accessibility, and input.
+    It also manages the message event loop.
+
+    The embedder is written in a language appropriate for the platform,
+    such as Java/Kotlin for Android, Swift/Objective-C for iOS/macOS,
+    and C++ for Windows and Linux.
+  related_links:
+    - text: "Architectural overview: The Embedder"
+      link: "/resources/architectural-overview#platform-embedding"
+      type: "doc"
+    - text: "Flutter on embedded devices"
+      link: "/embedded"
+      type: "doc"
+  labels:
+    - "architecture"
+    - "engine"
+
 - term: "Engine"
   short_description: |-
     The portable runtime for Flutter apps.


src/data/glossary.yml
```diff
@@ -1,3 +1,27 @@
+- term: "Embedder"
+  short_description: |-
+    The platform-specific component that coordinates with the operating system
+    and manages the Flutter engine.
+  long_description: |-
+    The _embedder_ provides the entrypoint for a Flutter app.
```
loic-sharma, line 6: Strictly speaking, this is the runner app, not the embedder. I'd consider removing this sentence.

---

The runner app provides the native entry point for your Flutter project. It is responsible for launching Flutter.

Your Flutter project has a runner app for each platform, contained in the `android/`, `ios/`, `linux/`, `macos`, `web`, or `windows` directories.

The runner app's code is created when you run the `flutter create` command to create your Flutter project. However, you are free to modify the runner app as much as needed.

The runner app uses APIs exposed by the platform's embedder to launch the Flutter engine.

---

What do you think of something like this?

```yaml
- term: "Embedder"
  short_description: |-
    The platform-specific component to support Flutter on a native platform.
  long_description: |-
    Each native platform supported by Flutter has an _embedder_ for
    platform-specific logic.
    The embedder is the bridge that coordinates with the underlying
    operating system.
    It implements services such as input, accessibility, message event loops,
    and more.
    The embedder also launches and manages the Flutter engine.

    Each embedder is written in the platform's native language:
    Java and Kotlin for Android, Swift and Objective-C for iOS and macOS,
    and C++ for Windows and Linux.

    Each embedder also expose plugin APIs.
    These APIs let plugin packages add additional platform-specific
    functionality to the app.

    The embedder is launched and managed by the runner app.
```

src/data/glossary.yml
```diff
@@ -1,3 +1,27 @@
...
+    It coordinates with the underlying operating system to access
+    services like rendering surfaces, accessibility, and input.
+    It also manages the message event loop.
+
+    The embedder is written in a language appropriate for the platform,
+    such as Java/Kotlin for Android, Swift/Objective-C for iOS/macOS,
```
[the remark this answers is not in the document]
loic-sharma: This is factually correct, but I'd consider sweeping that under the rug.

The point the reader should takeaway here is that the Android embedder is the chunk of code that ports Flutter to Android, and is written in the language that's native for Android.

The Android embedder does use C++, but that's an implementation detail mainly to interoperate with the C++ Flutter engine. (Same story with iOS/macOS embedders using Objective-C++)

src/data/glossary.yml
sfshaza2: Beautiful!

sfshaza2:
Feedback incorporated in this PR https://github.com/flutter/website/pull/12896

PR #12862, opened 2025-12-24
Fix(docs): Use 'prop drilling' terminology and add glossary entry (fixes #12828)

### LINKED ISSUE #12828: Is there a incorrect terminology in the section "Using widget constructors"
### Page URL

https://docs.flutter.dev/get-started/fundamentals/state-management

### Page source

https://github.com/flutter/website/blob/main/src/content/get-started/fundamentals/state-management.md

### Describe the problem

In the overview, right before the "Using widget constructors" section, you mention that using widget constructors is sometimes called "prop drilling" in other frameworks. However, at the end of the "Using widget constructors" section, you mention dependency injection. Is there a misuse of the term in the doc?

Please find the attached image. Shouldn’t it be referred to as “prop drilling” in this section instead of “dependency injection”?

Image #

### Expected fix

_No response_

### Additional context

_No response_

sfshaza2:
Resolves #12828 by updating 'state-management.md' to explicitly use the term 'prop drilling' for the constructor-based data passing pattern, replacing the confusing 'dependency injection' reference in that specific context. Also adds 'Prop drilling' to the glossary.

### DIFF
--- a/src/content/get-started/fundamentals/state-management.md
+++ b/src/content/get-started/fundamentals/state-management.md
@@ -157,9 +157,9 @@ Column(
 
 Passing the shared data for your app through widget constructors
 makes it clear to anyone reading the code that there are shared dependencies.
-However, creating a chain of widgets that only pass data down to their children
-is often referred to as _prop drilling_.
-
+Creating a chain of widgets that only pass data down to their children
+is often referred to as [_prop drilling_](/resources/glossary#prop-drilling).
+ 
 ### Using InheritedWidget
 
 Manually passing data down the widget tree can be verbose
--- a/src/data/glossary.yml
+++ b/src/data/glossary.yml
@@ -149,11 +149,14 @@
 
 - term: "Prop drilling"
   short_description: |-
-    The process of passing data through multiple layers of widgets via constructors.
+    The process of passing data through multiple layers of widgets
+    through constructor parameters.
   long_description: |-
-    The process of passing data through multiple layers of widgets via constructors,
-    usually to reach a deeper descendant. This pattern can become verbose, which is
-    why other state management solutions (like `InheritedWidget` or `Provider`) are often used.
+    The process of passing data through multiple layers of widgets
+    through constructor pareameterss, usually to reach a deeper descendant.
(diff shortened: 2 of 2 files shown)


src/content/get-started/fundamentals/state-management.md
```diff
@@ -157,8 +157,8 @@ Column(
 
 Passing the shared data for your app through widget constructors
 makes it clear to anyone reading the code that there are shared dependencies.
-This is a common design pattern called _dependency injection_
-and many frameworks take advantage of it or provide tools to make it easier.
+However, creating a chain of widgets that only pass data down to their children
```
parlough: What is "However" referring to here? Could it just be dropped?

src/content/get-started/fundamentals/state-management.md
```diff
@@ -157,8 +157,8 @@ Column(
...
+is often referred to as _prop drilling_.
```
parlough: Consider cross linking this to the glossary entry you added: `/resources/glossary#prop-drilling`

src/data/glossary.yml
```diff
@@ -128,6 +128,22 @@
     - "performance"
     - "smooth rendering"
 
+- term: "Prop drilling"
+  short_description: |-
+    The process of passing data through multiple layers of widgets via constructors.
```
parlough: https://developers.google.com/style/word-list#via suggests not using "via".

Could "through constructor parameters" work instead?

src/data/glossary.yml
```diff
@@ -128,6 +128,22 @@
...
+  long_description: |-
+    The process of passing data through multiple layers of widgets via constructors,
```
parlough: Here as well.

parlough (commented):
Generally looks good to me, just two small questions/concerns:

src/data/glossary.yml
sfshaza2: I do know this, but was being lazy.

parlough (approved):
Thanks for making those updates. Looks great to me :)

PR #12903, opened 2026-01-08
Update banner to call out AI best practices docs

parlough:
@sfshaza2 What do you think about a banner like this to call out the new docs section? Also happy to reword it if you have any suggestions :D


sfshaza2 (approved):
lgtm!

PR #12909, opened 2026-01-09
Docs: Add Adaptive definition to glossary

### LINKED ISSUE #12646: Add glossary entry for "Adaptive"
Add a glossary entry for "Adaptive" in the context of Flutter to https://github.com/flutter/website/blob/main/src/data/glossary.yml.

sfshaza2:
Fixes #12646. Adds a glossary entry for 'Adaptive' design, distinguishing it from 'Responsive' design.

### DIFF
--- a/src/data/glossary.yml
+++ b/src/data/glossary.yml
@@ -7,8 +7,8 @@
     as opposed to responsive design which is about fitting the UI _into_
     the space.
     An adaptive app selects the appropriate layout
-    (such as bottom nav vs side panel)
-    and input devices (for example, mouse vs touch)
+    (such as having a bottom nav instead of a side panel)
+    and input devices (for example, mouse versus touch)
     to feel natural on the current device.
   related_links:
     - text: "Adaptive vs responsive design"


sfshaza2:
@reidbaker, can you double check?

src/data/glossary.yml
```diff
@@ -1,3 +1,26 @@
...
+  short_description: |-
+    A design concept where the UI adapts to be usable in the available space,
+    often changing layout or input methods based on device capabilities.
+  long_description: |-
+    Adaptive design is about the UI being _usable_ in the space,
+    as opposed to responsive design which is about fitting the UI _into_
+    the space.
+    An adaptive app selects the appropriate layout
+    (such as bottom nav vs side panel)
+    and input devices (for example, mouse vs touch)
```
parlough: Perhaps something like this instead of "vs":

```suggestion
    (such as having a bottom nav instead of a side panel)
    and input devices (for example, mouse versus touch)
```

Issue #12537, opened 2025-10-11
[Android build file location error]


321paranoiawhy:
### Page URL

https://docs.flutter.dev/deployment/android#build-an-apk

### Describe the problem

In older flutter version, the android build apk file location is right. But in newer flutter version (I do not know above which flutter version), the apk file location should be changed from:

```bash
[project]/build/app/outputs/apk/release/app-armeabi-v7a-release.apk
[project]/build/app/outputs/apk/release/app-arm64-v8a-release.apk
[project]/build/app/outputs/apk/release/app-x86_64-release.apk
```

to:

```bash
[project]/build/app/outputs/flutter-apk/app-armeabi-v7a-release.apk
[project]/build/app/outputs/flutter-apk/app-arm64-v8a-release.apk
[project]/build/app/outputs/flutter-apk/app-x86_64-release.apk
```

that is remove `/release/` and change `/apk/` to `/flutter-apk/`.

the `AAB` format is not influenced.

and shoule include `[project]/build/app/outputs/flutter-apk/app-release.apk` when not split per ABI.

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

piedcipher:
<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/ff00e648-d133-4258-8d52-d56567ab3953" />

CLI prints the one with **flutter-apk/** but generated APKs are present in both directories.

321paranoiawhy:
> <img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/ff00e648-d133-4258-8d52-d56567ab3953" />
> 
> CLI prints the one with **flutter-apk/** but generated APKs are present in both directories.

So may be described these behavior in the doc somewhere else？

PR #12930, opened 2026-01-13
Update android apk build output paths

sfshaza2:
Fixes https://github.com/flutter/website/issues/12537


sfshaza2:
@camsim99, can you review?

parlough (approved):
Matches what I see when testing. Thanks for updating these!

camsim99 (approved):
Also matches what I see when testing. Thanks!

PR #12948, opened 2026-01-16
Add back other install pages to sidenav

parlough:
@ericwindmill Do you have concerns with adding these back? It's a bit awkward if they aren't in the sidenav and also hard to find them.

Another option would be for this existing section to be "Set up Flutter" and have a separate "Manage installation" section, perhaps at the bottom of the sidenav.


ericwindmill (approved):
I think we should find ways to simplify the side nav in general, but ultimately my opinion on the matter is loosely held and I defer to you on all choices.

PR #12961, opened 2026-01-20
Fix typo in flavors-ios.md

cdeil:
Second attempt following up #12946 . What happened there is that I didn't notice that Gemini reversed the typo fix in the second commit.


sfshaza2 (approved):
lgtm

PR #12985, opened 2026-01-24
Fixed small typo "therefor"

adriancuadrado:


adriancuadrado:
Oh, I didn't know "[therefor](https://www.wordreference.com/definition/therefor)" is an actual word. My bad 😅.

PR #12999, opened 2026-01-27
Add "Cupertino" and "Material" glossary entries

sfshaza2:
Adding glossary entries for Material and Cupertino.

Staged: https://flutter-docs-prod--pr12999-glossary-iak66wqc.web.app/resources/glossary

### DIFF
--- a/site/lib/src/client/global_scripts.dart
+++ b/site/lib/src/client/global_scripts.dart
@@ -390,6 +390,9 @@ void _setUpSteppers() {
 
   for (var i = 0; i < steppers.length; i++) {
     final stepper = steppers.item(i) as web.HTMLElement;
+
+    final collapsible = !stepper.classList.contains('non-collapsible');
+
     final children = stepper.childNodes;
     final steps = [
       for (var j = 0; j < children.length; j++)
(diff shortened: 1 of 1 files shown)


src/data/glossary.yml
```diff
@@ -21,6 +21,28 @@
     - "ui"
     - "design"
     
+- term: "Cupertino"
+  short_description: |-
+    The design language used by Apple on iOS.
```
khanhnwin: This could be misleading since Apple technically doesn't call their design language "Cupertino"?

khanhnwin: I would maybe replace it with "Flutter's implemention of the iOS design language"

src/data/glossary.yml
```diff
@@ -21,6 +21,28 @@
...
+  long_description: |-
+    Flutter's `cupertino` library implements the current iOS design language,
```
khanhnwin: "current" could be misleading since folks might think Liquid Glass?

sfshaza2:
Good feedback, @khanhnwin! I thought you could give me a blessed review, but I see not. @parlough, can you help?

src/data/glossary.yml
```diff
@@ -21,6 +21,28 @@
...
     - "design"
     
+- term: "Cupertino"
+  short_description: |-
+    Flutter's implementation of the iOS design language.
+  long_description: |-
+    Flutter's `cupertino` library implements the iOS design language,
+    comprising a set of widgets that implement Apple's Human Interface Guidelines.
+
+    The `cupertino` library, originally part of the main Flutter repo,
```
parlough, line 31: Good idea to mention this here :)

PR #13001, opened 2026-01-28
docs Tile constructor params wrong

tmpreston:
Constructor doesn't match earlier changes at this point in the docs.


parlough (approved):
Looks good to me. Thanks for noticing and fixing these!

PR #13029, opened 2026-02-04
fix non-collapsible steppers

schultek:
Fixes the issue where the steps at https://docs.flutter.dev/learn/pathway are collapsed after re-loading.


sfshaza2 (approved):
lgtm, @schultek. Though maybe @parlough should review.  :)

PR #13033, opened 2026-02-04
Add 'Primary constructor' to glossary

sfshaza2:
Adds a definition for 'Primary constructor' to the glossary, focusing on its usage in Dart extension types.


src/data/glossary.yml
kallentu, line 1: @eernstg to review as well.

src/data/glossary.yml
```diff
@@ -291,6 +291,26 @@
...
     - "ui"
 
+- term: "Primary constructor"
+  short_description: |-
+    A constructor declared as part of the type declaration.
+  long_description: |-
+    A _primary constructor_ is a concise way to declare a constructor
+    directly in the type declaration.
+    
+    Currently, this feature is only available for [extension types][].
```
kallentu: This is inaccurate -- primary constructors can be used for classes, extension types, and enums as well.

src/data/glossary.yml
eernstg: Thanks for the heads up!

I'll assume that this glossary entry should describe the primary constructors feature which is about to be released&mdash;it doesn't seem particularly useful to have one description of the concept for some weeks, and then updating it substantially to cover the much more general notion of primary constructors that we're about to release.

src/data/glossary.yml
```diff
@@ -291,6 +291,26 @@
     - "design"
     - "ui"
 
+- term: "Primary constructor"
+  short_description: |-
+    A constructor declared as part of the type declaration.
```
eernstg: The phrase 'type declaration' is rather vague. It's often used to designate a type annotation (as in "the variable has a type declaration" when talking about declarations like `String s = '';`), as well as the meaning which is intended here. It would be nice if we could use a phrase which is more precisely denoting a class declaration, a mixin class declaration, an enum declaration, or an extension type declaration.

Next, a class declaration starts at a keyword (often `class`, but we can also have `final class`, `sealed class`, and the like), and it ends at the final `}` of the class body (or, with the upcoming enhancements, a semicolon which is the entire class body on its own). This implies that every member declaration is also part of "the type declaration". The new thing is that the primary constructor is part of the _header_ of the class declaration, it's not in the body.

Similarly for mixin class declarations etc.

Perhaps this would work:
```suggestion
    A constructor declared in the header of a class or a similar type declaration.
```

src/data/glossary.yml
```diff
@@ -291,6 +291,26 @@
...
+  long_description: |-
+    A _primary constructor_ is a concise way to declare a constructor
+    directly in the type declaration.
```
eernstg, line 300:
```suggestion
    directly in the header of a class, mixin class, enum, or extension type declaration.
```

src/data/glossary.yml
```diff
@@ -291,6 +291,26 @@
...
+    
+    Currently, this feature is only available for [extension types][].
+    It allows you to define the underlying representation object
+    and bind it to a field in one step.
```
eernstg: I think it's really great that this glossary entry is so short. It should allow developers to get to know about the concept of a primary constructor. There would be other pages about all the details.

So how do we communicate that the primary constructors of class, mixin class, and enum declarations have a large number of properties in common, and the primary constructors of extension types are different in several ways?

Here is one attempt at saying very little, but still enough to give the reader a relevant hint:
```suggestion
    A primary constructor in a class, mixin class, or enum declaration is a concise
    way to specify a constructor and a number of instance variables in a single
    constructor declaration in the header.
    A primary constructor in an extension type declaration allows you to define the 
    underlying representation variable and its type.
```

We can't even really say "in a single declaration" when it comes to extension types, because they don't support declaring the representation variable using a regular instance variable declaration in the first place. There is simply no other way to declare the representation variable (name and type) of an extension type than having a primary constructor.

src/data/glossary.yml
```diff
@@ -291,6 +291,26 @@
...
+
+    [extension types]: https://dart.dev/language/extension-types
+  related_links:
+    - text: "Extension types constructors"
+      link: "https://dart.dev/language/extension-types#constructors"
+      type: "external"
```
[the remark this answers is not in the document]
eernstg: 'Dart 3.4' is not accurate, the new, generalized primary constructors feature hasn't yet been released, but it's being implemented and will definitely be released.

sfshaza2:
I've made updates to the PR but am closing it for now until the feature is released.

PR #13035, opened 2026-02-05
Remove Rosetta installation step from iOS setup

sfshaza2:
Rosetta is no longer required for Flutter tooling on Apple Silicon Macs. This change removes the installation step from the iOS setup guide.


sfshaza2:
I feel like this change is premature. Closing.

Issue #13084, opened 2026-02-17
Learn section not linked on mobile


schultek:
### Describe the problem

When on the main docs page on mobile, the learn section is not linked in the sidebar. So there is currently no way to get to the learn section on mobile.

<img width="456" height="310" alt="Image" src="https://github.com/user-attachments/assets/09d050b1-a40b-409a-bd8a-f8b5e8ae4306" />

When on the learn section, one can navigate back. The same nav options should be shown on the regular docs.

<img width="459" height="270" alt="Image" src="https://github.com/user-attachments/assets/7ee612e4-090a-4c90-a341-fa5b539017da" />

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

PR #13085, opened 2026-02-17
Include site sections on mobile sidenav

parlough:
Also renames "User Guides" to just "Guides" as we prefer "Developers" over "Users" and the title case is inconsistent with what we use elsewhere.

Fixes https://github.com/flutter/website/issues/13084


sfshaza2:
Yeah, I was tempted to change "User Guides" to "User guides", but this is better.

PR #4458, opened 2020-08-07
New breaking change: Network Policy on iOS and Android

mehmetf:

PR #13133, opened 2026-03-03
Remove documentation about metadata

mboetger:
It turns out the metadata tag for network policy in the Android embedder is not connected to the engine.  

Addresses: https://github.com/flutter/flutter/issues/183153

## Presubmit checklist

- [X] If you are unwilling, or unable, to sign the CLA, even for a _tiny_, one-word PR, please file an issue instead of a PR.
- [X] If this PR is not meant to land until a future stable release, mark it as draft with an explanation.
- [X] This PR follows the [Google Developer Documentation Style Guidelines](https://developers.google.com/style)—for example, it doesn't use _i.e._ or _e.g._, and it avoids _I_ and _we_ (first-person pronouns).
- [X] This PR uses [semantic line breaks](https://github.com/dart-lang/site-shared/blob/main/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks)
  of 80 characters or fewer.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -33,7 +33,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@89a39a4e59826350b863aa6b6252a07ad50cf83e
+      uses: github/codeql-action/init@c793b717bc78562f491db7b0e93a3a178b099162
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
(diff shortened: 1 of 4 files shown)


mboetger:
@Sfshaza 

This [documentation](https://docs.flutter.dev/release/breaking-changes/network-policy-ios-android) was [added](https://github.com/flutter/website/pull/4458) to support the breaking changes in:

https://github.com/flutter/engine/pull/20218
https://github.com/flutter/engine/pull/20492
https://github.com/flutter/engine/pull/20733
Since then, all of the functionality has been removed.

"--disallow-insecure-connections" is used in [neither Android nor iOS](https://github.com/search?q=repo%3Aflutter%2Fflutter%20disallow-insecure-connections&type=code).
"--domain-network-policy" is used by the Android embedder, but not by the engine. So it does nothing.

This breaking change documentation only re-iterates information about the Android, iOS networking policies and therefore only affects Android, iOS sockets - not Dart/Flutter sockets. This is stated on the page, but the page also makes it seem like there's a way around it.

I have attempted to clear this up and highlight this.  But I can't help asking: Should we just remove the documentation?

Issue #13146, opened 2026-03-05
Suggest using package:os_detect


jonasfj:
### Page URL

https://docs.flutter.dev/platform-integration/web/faq

### Page source

https://github.com/flutter/website/blob/main/src/content/platform-integration/web/faq.md

### Describe the problem

FAQ: Does Platform.is work on the web?

Not currently.

### Expected fix

Suggest people use:

https://pub.dev/packages/os_detect

Or maybe package:platform when https://github.com/dart-lang/core/pull/936
has landed.

Both are maintained by the Dart team, and are better options than `Platform.is...` from `dart:io`.

CC: @lrhn

As far as I recall, os_detect should allow tree shaking. But again that's not really something the language promises.

### I would like to fix this problem.

- [ ] I will try and fix this problem on docs.flutter.dev.

PR #13160, opened 2026-03-09
Dusting and cleaning of the web pages

sfshaza2:
Fixes https://github.com/flutter/website/issues/13146

Also, fixes formatting and updates guidance around static websites.

### DIFF
--- a/.agents/skills/stage-site/SKILL.md
+++ b/.agents/skills/stage-site/SKILL.md
@@ -1,0 +1,54 @@
+---
+name: stage-site
+description: >-
+  Stages the Flutter site locally. This includes checking for broken links,
+  verifying excerpts, updating formatting, and serving the dev site.
+---
+
+# Stage the Flutter site
+
+Before committing changes or reviewing a PR locally,
+it is important to stage the site and ensure everything works correctly. 
+Follow these steps to stage the site:
+
+## 1. Check for broken links
+
+Run the following command to check for any unlinked or broken
+Markdown link references in the site output:
+
+```bash
+dart run dash_site check-link-references
+```
+
+Alternatively, to verify all links between pages on the site work:
+
+```bash
+dart run dash_site check-links
+```
+
+If any broken links are found, try to patch them or alert the user.
+
+## 2. Sync code excerpts
+
+Ensure that any changed code excerpts are properly run and synced to the Markdown files:
+
+```bash
+dart run dash_site refresh-excerpts
+```
+
+## 3. Update formatting
+
+Update the formatting of the site examples and tools to ensure
+they follow the official Dart format:
+
+```bash
+dart run dash_site format-dart
+```
+
+## 4. Stage the site locally
+
+Finally, serve a local development server of the site so the user can preview the changes:
+
+```bash
+dart run dash_site serve
+```
--- a/.agents/skills/validate-pr/SKILL.md
+++ b/.agents/skills/validate-pr/SKILL.md
@@ -1,0 +1,49 @@
+---
+name: validate-pr
+description: >-
+  Stages the Flutter site locally. This includes checking for broken links,
+  verifying excerpts, updating formatting, and serving the dev site.
+---
+
+# Stage the Flutter site
+
+Before committing changes or reviewing a PR locally,
+it is important to stage the site and ensure everything works correctly. 
+Follow these steps to stage the site:
+
+## 1. Sync code excerpts
+
+Ensure that any changed code excerpts are properly run and synced to the
+Markdown files:
+
+```bash
+dart run dash_site refresh-excerpts
+```
+
+## 2. Update formatting
+
+Update the formatting of the site examples and tools to ensure
+they follow the official Dart format:
+
+```bash
+dart run dash_site format-dart
+```
+
+## 3. Check for broken links
+
+Run the following command to check for any unlinked or broken
+Markdown link references in the site output:
+
+```bash
+dart run dash_site check-link-references
+```
+
+If any broken links are found, try to patch them or alert the user.
+
+## 4. Stage the site locally
+
+Finally, serve a local development server of the site so the user can preview the changes:
+
+```bash
+dart run dash_site serve
+```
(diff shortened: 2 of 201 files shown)


src/content/platform-integration/web/faq.md
```diff
@@ -25,33 +26,38 @@ see [Web support for Flutter][].
...
-search-engine optimized HTML.
+In general, Flutter is geared towards dynamic application experiences.
+Flutter's web support is no exception.
+Flutter web prioritizes performance, fidelity, and consistency.
+This means application output doesn't align with what search
+engines need to properly index.
 
-That said, as mentioned in the [roadmap][], the Flutter team plans to
-investigate search engine indexability of Flutter web.
+However, a recently released Dart package, [Jaspr][] _does_
```
abdallahshaban557: not super recently release at this point - and also, we should add the proper link - and mention it is a community package.

src/content/platform-integration/web/faq.md
```diff
@@ -25,33 +26,38 @@ see [Web support for Flutter][].
...
+support static websites.
+In fact, both the [Dart][] and [Flutter][] documentation
```
abdallahshaban557: links - and also, not just documentation - but also the marketing sites!

src/content/platform-integration/web/faq.md
```diff
@@ -25,33 +26,38 @@ see [Web support for Flutter][].
...
+websites were recently migrated to using the Jaspr package.
+
+To summarize, for web content that is static or document-like,
+we recommend _either_ using:
+
+1. [Jaspr][], if you want to use Dart but want a more traditional
+   DOM-based website.
+1. HTML—in this case, consider separating your primary application
```
abdallahshaban557, line 46: this HTML thing i don't understand

src/content/platform-integration/web/faq.md
```diff
@@ -106,28 +113,31 @@ See [Preparing a web app for release][].
 
 ### Does `Platform.is` work on the web?
 
-Not currently.
+Not currently. Instead,
+consider using the [`os_detect`][] package.
```
abdallahshaban557: I think support exists for kIsWeb - if this is for platform detection would should perhaps update this question here.

src/content/platform-integration/web/index.md
```diff
@@ -61,25 +65,36 @@ rather than the app-centric services that a UI framework like Flutter
 can deliver. However, you _can_ use Flutter to embed interactive
 experiences into these websites.
 
+However, if you want to use Dart and wish to implement a traditional
+DOM-based website, a recently released Dart package,
+[Jaspr][], supports static websites; in fact,
```
abdallahshaban557, line 70: links in this paragraph are missing

src/content/platform-integration/web/faq.md
sfshaza2: No, the links are there. They are below the paragraph. That's primarily how we do links on the site.

src/content/platform-integration/web/faq.md
sfshaza2: Interesting, @jonasfj asked for this fix. I'll see what he says when we reviews.

src/content/platform-integration/web/index.md
sfshaza2: We put the links _below_ so as not to mess up the text.

src/content/platform-integration/web/faq.md
jonasfj: `kIsWeb` looks fine to use.

But this question is written because historically lots of code has been using [`Platform.isAndroid`](https://api.dart.dev/dart-io/Platform/isAndroid.html) and [`Platform.isIOS`](https://api.dart.dev/dart-io/Platform/isIOS.html), while these are decidedly not deprecated (one can dream I suppose).

Using [`Platform.isAndroid`](https://api.dart.dev/dart-io/Platform/isAndroid.html) will **not** work when targeting web with.

It's sadly technically allowed to `import 'dart:io'` when compiling for web. This is and should be strongly discouraged. And if you import `'dart:io'` calling any method from it will throw an `UnsupportedError`. It is possible to work around this with [`IOOverrides`](https://api.dart.dev/dart-io/IOOverrides-class.html). This is not encouraged.

On pub.dev, we will score packages as **not supporting web** if they import `dart:io` (they may import it with conditional imports).

----------------

If you're coding a flutter app using `kIsWeb`, is fine. Maybe, even desirable, as it's right there.

If you are coding a package for use by others, you should probably prefer `os_detect` (unless you have a flutter dependency, then it probably doesn't matter).

I think it's important to mention `os_detect`, because:
 * Packages that don't have a flutter dependency should not add one to detect platform, instead they should use `os_detect`.
 * Packages with a flutter dependency may prefer `os_detect`, and it may be an easier migration path for packages that has been using [`Platform.isAndroid`](https://api.dart.dev/dart-io/Platform/isAndroid.html) and friends.


I suspect the FAQ is here because `dart:io` can be imported when targeting web, it just becomes a runtime error to call `Platform.isAndroid` and friends. This is a bit of trap.

---------------

Sorry, for all the rambling. This is complex to put it mildly -- generally, I think we should tell users to avoid `Platform.isXYZ`, and instead prefer `package:os_detect` (or `package:platform` when the next PR lands).

abdallahshaban557: Got it - appreciate the explanation here, @jonasfj! Since this is in the main FAQ for the web part of our docs, I think we should mention both then. kIsWeb for an app, and `os_detect` for a package. Does you agree?

src/content/platform-integration/web/faq.md
```diff
@@ -93,41 +102,57 @@ on [dart.dev]({{site.dart-site}}).
...
+although such support isn't built in.
 
 ### How do I deploy a web app?
 
 See [Preparing a web app for release][].
 
 ### Does `Platform.is` work on the web?
 
-Not currently.
+No. While you can technically import `dart:io` when compiling for the web,
```
kevmoo, line 118: @srujzs @biggs0125 – we throw for `dart:io` with wasm, right?

srujzs: No, it's the same as with dart2js and ddc where you can import it, but any usage throws.

sfshaza2: So, I'm interpreting the suggested text as correct. Yes, @kevmoo?

kevmoo: Yep! just checking

PR #13166, opened 2026-03-11
Adding a staging skill

sfshaza2:
As it says

### DIFF
--- a/.agents/skills/validate-pr/SKILL.md
+++ b/.agents/skills/validate-pr/SKILL.md
@@ -1,5 +1,5 @@
 ---
-name: stage-site
+name: validate-pr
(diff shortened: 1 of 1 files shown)


.agents/skills/validate-pr/SKILL.md
parlough, line 1: Maybe we should consider a different name and description for this skill? It slightly conflates the meaning of the word "staging": what the skill discusses and then its step 4, which is also staging the site. I wonder that could be a bit confusing.

Would this skill be better described as something like "Validate(or check?) your changes [to the site] [before committing]"?

conooi (approved):
love it! lgtm

PR #13199, opened 2026-03-20
Move section "update your app to include the article view" earlier

molulan:
Move the section up to just after creating the ArticleView widget. This way it's possible to follow the changes visually in the app, Which is more motivating when following along the tutorial.


sfshaza2 (approved):
lgtm. Thanks for this great idea!

PR #13203, opened 2026-03-21
Also fix versionName in example build.gradle.kts

solsticedhiver:
There is a call to .toInteger() on flutter.versionCode in the example of build.gradle.kts

This does not seem right because that change throws an error when trying to build the app with that change.

Note: There is no such a call to .toInteger() with the build.gradle.kts created with flutter for a brand new app

Remove a call to to toInteger() for flutter.versionCode because it throws an error when building and does not seem necessary (builds fine without)

See above

## Presubmit checklist

  of 80 characters or fewer.


solsticedhiver:
Oops I have signed the CLA after clicking the button. should be ok now

PR #13247, opened 2026-04-02
Add a missing import and a note about an edge-case crash

### LINKED ISSUE #13192: Minor problems with Contacts app (Flutter UI 102 tutorial)
### Page URL

https://docs.flutter.dev/learn/pathway/tutorial/navigation

### Page source

https://github.com/flutter/website/blob/main/src/content/learn/pathway/tutorial/navigation.md

### Describe the problem

When I completed the lesson, my code would not run because it was missing a couple of imports.  adaptive_layout.dart and contact_groups.dart were both missing importing  contacts.dart. Once these were added, it will compile and run. It's possible that I missed something in the lesson, but I was trying to be careful, so I think they got left out. This problem seems to first appear in section 16-5.

The second thing that I noticed is that the app will throw an exception under one circumstance. Start with the app sized for small screens and click on one of the groups. The app now switches to showing contacts in that group. Resize the app for a large screen (> 600 pixels) and then it the back < button.  An exception will be thrown (text is below). I'm not sure how to fix this as I'm a beginner with flutter.

══╡ EXCEPTION CAUGHT BY SCHEDULER LIBRARY ╞═════════════════════════════════════════════════════════
The following assertion was thrown during a scheduler callback:
There are multiple heroes that share the same tag within a subtree.
Within each subtree for which heroes are to be animated (i.e. a PageRoute subtree), each Hero must
have a unique non-null tag.
In this case, multiple heroes had the following tag: Default Hero tag for Cupertino navigation bars
with navigator NavigatorState#2efce(tickers: tracking 2 tickers)
├# Here is the subtree for one of the offending heroes: Hero

The app I'm planning to write will be using a similar structure to that of the tutorial so it would be nice if I understood how to code this without causing an exception. The exception is caught and doesn't crash the app, so should it be ignored? I'm not the expert on this one so I'll leave it up to the people that are.

### Expected fix

Add to adaptive_layout.dart and contact_groups.dart
import 'contacts.dart';

Thanks for putting these tutorials together. They are appreciated.

### Additional context

_No response_

sfshaza2:
Fixes https://github.com/flutter/website/issues/13192

The user has identified a missing import and a crash that can happen in a specific situation. Fixing this crash in a more sophisticated way involves state management so, in this case, we've added a note.

FYI: @lamek


src/content/learn/pathway/tutorial/navigation.md
```diff
@@ -68,6 +68,8 @@ Ensure that the `onListSelected` callback in
 `lib/screens/contact_groups.dart` is implemented as follows:
 
 ```dart title="lib/screens/contact_groups.dart"
+import 'package:rolodex/screens/contacts.dart';
```
johnpryan: I don't think this is needed? Something doesn't seem right here, the rolodex package on pub.dev is not what we want (it's old and out of date), and the Rolodex project we were working on is no longer maintained.

johnpryan (approved):

PR #13258, opened 2026-04-07
Add proofread_markdown skill and Dart wrapper script

antfitch:
`Proofread markdown` skill for the Flutter Docs repo. You can use this skill to update your markdown files to follow the Flutter doc teams formatting guidelines.

- Changes title case to sentence case in headings.
- Converts headings from passive case to active case (note: when working with how-to and tutorials, this is a good thing, but there are times when this should be avoided, such as when you're introducing concepts. Look carefully at the changes before accepting them.
- Ensures there is a blank line between heading and body text.
- Refactors steps for brevity.

This is an early version of this skill. As time goes on, we'll update it to be more complete, fix bugs, etc. Use this with your AI agents, but in a mode where you can approve the changes.


.agents/skills/proofread_markdown/SKILL.md
```diff
@@ -0,0 +1,90 @@
...
+## 1. Check line length
+
+Leadership does not want markdown files to have lines that exceed 80 characters.
+The reason is that it makes it easier to read for humans, and this is easier for
+agents, when a screenshot of a markdown file is needed.
+
+1. Run the `wrap_lines.dart` script on the file to automate line wrapping:
+
+   ```bash
+   dart /Users/afitch/flutter_web/website/.agents/skills/proofread_markdown/scripts/wrap_lines.dart <file_path>
```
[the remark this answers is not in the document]
antfitch: Should be fixed now.

.agents/skills/proofread_markdown/scripts/wrap_lines.dart
```diff
@@ -0,0 +1,96 @@
...
+      output.add('');
+      continue;
+    }
+
+    // Handle lists or quotes (matching python script's simple approach)
+    if (stripped.startsWith('-') ||
+        stripped.startsWith('*') ||
+        stripped.startsWith('1.') ||
+        stripped.startsWith('2.') ||
+        stripped.startsWith('3.')) {
```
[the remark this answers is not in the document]
antfitch: Should be fixed now.

.agents/skills/proofread_markdown/scripts/wrap_lines.dart
```diff
@@ -0,0 +1,96 @@
...
+      }
+    } else {
+      currentLine += (currentLine.isEmpty ? '' : ' ') + word;
+    }
+  }
+  if (currentLine.isNotEmpty) {
+    lines.add(currentLine.trim());
+  }
+  return lines;
+}
```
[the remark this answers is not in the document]
antfitch: should be fixed now.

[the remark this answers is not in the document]
antfitch: Should be fixed now.

.agents/skills/proofread_markdown/SKILL.md
```diff
@@ -0,0 +1,89 @@
...
+Change steps to be direct like this:
+
+```
+ ### Search for a Topic
+
+ 1.  Navigate to the **Search** tab (Magnifying Glass icon).
+ 2.  Select **View topics** from the dropdown (default).
+ 3.  Enter your query in the "Query" text box (e.g., "Flutter widgets").
+ 4.  Press **Enter**.
+ 5.  View results in the main content area.
```
sfshaza2: One thing that bugs me about the bots, is that they complain about this:

```
1. a
1. b
1. c
1. d
```

However, we PREFER this in a markdown file. Markdown numbers the steps and then if, later, you remove or add a line, it automatically fixes it.

.agents/skills/proofread_markdown/scripts/wrap_lines.dart
```diff
@@ -0,0 +1,96 @@
+import 'dart:io';
```
sfshaza2, line 1: OK, I've said this offline, but I really don't like this form of breaking lines. It's ugly, it's nasty to maintain. Semantic line breaks are much easier for a writer to process, read, update. Meaning, break on punctuation first, like commas, periods, semi-colons, colons. THEN if lines are still too long, break on natural pauses. Since that might be harder to teach a bot, I could live with breaking on 80 chars THEN.

It's MUCH easier to understand the diff of a file that uses semantic line breaks, rather than a straight 80 char line break. MUCH MUCH easier.

.agents/skills/proofread_markdown/SKILL.md
```diff
@@ -0,0 +1,89 @@
...
+# Proofread markdown files
+
+Hi there! You can use this skill to proofread markdown files. This particular
+project has documentation style rules that are slightly different than what you
+are used to, so please use the steps in this workflow to update any markdown
+files that have changed.
+
+## 1. Check line length
+
+Leadership does not want markdown files to have lines that exceed 80 characters.
```
sfshaza2: This is NOT about Leadership. Please change that. This is what GitHub wants and the convention that we've agreed to and adhered to.

.agents/skills/proofread_markdown/SKILL.md
```diff
@@ -0,0 +1,89 @@
...
+## 3. Use active voice in headings
+
+Let's use active voice in headings. So if you see a heading that looks like
+this:
+
+```
+## Project Configuration
+```
+
+Change it to active voice that looks like this:
```
sfshaza2: Also mention that we don't use gerunds in titles or headers. So: "Configure project" is cool but "Configuring project" is to be avoided.

.agents/skills/proofread_markdown/SKILL.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+## 5. Use less words in sections with steps
+
+If a section has steps in it, be brief. Keep it short, simple, and accurate.
+
+Change steps to be direct like this:
+
+```
+ ### Search for a Topic
```
sfshaza2: This should be:

```
### Search for a topic
```

.agents/skills/proofread_markdown/SKILL.md
```diff
@@ -0,0 +1,89 @@
...
+
+ 1.  Navigate to the **Search** tab (Magnifying Glass icon).
+ 2.  Select **View topics** from the dropdown (default).
+ 3.  Enter your query in the "Query" text box (e.g., "Flutter widgets").
+ 4.  Press **Enter**.
+ 5.  View results in the main content area.
+```
```
sfshaza2: I had feedback I don't see now. But a numbered list should use "1." for each line.

sfshaza2 (commented):
See comments.

antfitch:
Add the format that @parlough is using here to separate sections: https://github.com/flutter/website-cms/blob/main/.agents/skills/generate-showcase-from-video/SKILL.md

.agents/skills/proofread_markdown/scripts/wrap_lines.dart
parlough: Yes to this. Breaking at 80 characters without considering meaning or structure isn't really useful and I'm worried about maintaining this script as well.

I've found agents can write and update text to use semantic breaks quite well. You just need to tell them to and [reference a reference doc](https://agentskills.io/specification#file-references) that briefly explaina the format and provides some examples.

parlough:
> Add the format that @parlough is using here to separate sections: https://github.com/flutter/website-cms/blob/main/.agents/skills/generate-showcase-from-video/SKILL.md

Just clarifying: That new, fancy skill is by @abdallahshaban557 :)

.agents/skills/proofread_markdown/scripts/wrap_lines.dart
antfitch: A few thoughts I had around this today:

When I read a book (paperback or kindle), words generally break after a certain number of characters that don't fit on a line and this has never bothered me. So when I look at this discussion through that lens, breaking things semantically doesn't do much for me. Because I've been looking at markdown files for several years at work that simply follow the 80-character rule, I just don't have the issues of readability with non-semantic breaks.

Having an agent go line by line through a markdown file burns up a lot of tokens and is slower than a script that doesn't require tokens. On very large markdown files, a lot of tokens can be burned.

I've noticed from past recent experience that agents make mistakes when I ask them to manually go through a markdown file and fix line length. Especially in longer files, like us, they sometimes miss lines and I have to go back and double-check everything.

So, I feel strongly that we need to keep this, and especially for larger documents. However, I think we should have the ability for the author to choose.

antfitch: I feel very strongly about this. I just pulled up this answer from Antigravity (it has not seen our conversation)

"Can you assess how token expensive checking line lengths is? Let's assume you have a document with 500 lines."

Direct LLM Check: 8,000 to 20,000 tokens (Expensive and prone to counting errors).

Tool-Assisted Check: ~100 tokens (Extremely cheap and 100% accurate).
If you are planning to implement this as part of a skill, I strongly recommend writing a short Dart script or using a regex-based tool to do the line-length checking and reporting back to the agent!

antfitch: For this reason, I want to make sure we give contributors a choice. If they want to, they can choose one of three options and do some of the work by hand if they need to use semantic breaks.

.agents/skills/proofread_markdown/SKILL.md
antfitch: Deal! Actually I just grabbed @conooi's intro (love it!)

.agents/skills/proofread_markdown/SKILL.md
antfitch: Grabbed @conooi's version. I think you'll like!
Also, if a heading changes, I had the skill add the original link id after the heading to links don't break. `## Configure the project {: #project-configuration }`

.agents/skills/proofread_markdown/SKILL.md
antfitch: *facepalm* updated.

.agents/skills/proofread_markdown/SKILL.md
antfitch: Updated! Pulled in @conooi's changes.

.agents/skills/proofread_markdown/SKILL.md
antfitch: Captured!

.agents/skills/proofread_markdown/scripts/wrap_lines.dart
antfitch: BUT I will admit. When I chose the semantic line breaks option when presented by the agent, I love the results.

sfshaza2 (commented):
I can live with this! :D. lgtm

sfshaza2 (approved):
Oops! Forgot to click approve.

antfitch:
> I can live with this! :D. lgtm

YAY! I had forgotten for a moment that of course we could have a semantic line break script. Just took a few days. @conooi you could update the dart skill with this, too. Should save lots of tokens.

PR #13296, opened 2026-04-21
Prefetch and/or prerender pages when likely navigated to

parlough:
Implements a mechanism to specify pages to be prefetched and prerendered when likely navigated to, improving page loading speed and making website navigation feel more snappy and seamless.

- If a page has a `next` (page) specified, prerender it.
  - This for example, makes navigating to the next lesson in the tutorial feel immediate.
- If a page has `prev` (page) specified, prefetch it.
- If on the homepage, prefetch the pages linked to in the top nav.
- On any page, prefetch a page if hovering a link to it.
- On any page, prerender a page if beginning to click a link to it.

Note that the [Speculation Rules API](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API) this relies on mostly only works on Chromium-based browsers currently with WebKit having some initial support.

### DIFF
--- a/site/lib/src/extensions/registry.dart
+++ b/site/lib/src/extensions/registry.dart
@@ -10,7 +10,7 @@ import 'glossary_link_processor.dart';
 import 'header_extractor.dart';
 import 'header_processor.dart';
 import 'table_processor.dart';
-import 'tutorial_prefetch_processor.dart';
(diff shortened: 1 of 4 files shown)


sfshaza2 (approved):
lgtm

PR #13302, opened 2026-04-23
Update and simplify dependabot config

parlough:
Previously we didn't keep it up to date, so this PR uses more recent dependabot syntax to apply (and group) updates to entire directories.


sfshaza2 (approved):
lgtm

PR #13311, opened 2026-04-27
fix pagenav header for tutorial

### LINKED ISSUE #13188: TOC current header doesn't update when changing tutorial steps
For example, try to continue through the steps on a tutorial lesson such as https://docs.flutter.dev/learn/pathway/tutorial/advanced-ui

schultek:
Fixes #13188

Special handling for steppers that always shows the title of the currently open step, independent of scrolling - only when scrolling up the page title is shown instead.


parlough (approved):
I applied the Gemini review suggestions and some other clean up, but otherwise looks good to me. Thanks for tackling this @schultek!

PR #13322, opened 2026-04-28
Remove the -all flag from installation instructions. 

johnpryan:
This installs skills for all agents, not all the skills in the repository.

Instead, the `--skill '*'` flag installs all of the official skills, and the `--agent universal` flag puts them in `.agents/skills` folder.


lamek (approved):
LGTM

johnpryan:
Thank you for the quick review!

PR #13326, opened 2026-04-29
Updating the hcpp docs

### LINKED ISSUE #13323: Add experimental HCPP rendering mode to Android platform view docs
### What information needs to be added?

Description of mode is here: https://github.com/flutter/flutter/blob/master/docs/platforms/android/Android-Platform-Views.md
https://github.com/orgs/flutter/projects/201
Will be available in Flutter 3.44.

### Where should this new content appear?

https://docs.flutter.dev/platform-integration/android/platform-views

- [x] I will try and fix this problem on docs.flutter.dev.

sfshaza2:
Fixes https://github.com/flutter/website/issues/13323

### DIFF
--- a/src/content/platform-integration/android/platform-views.md
+++ b/src/content/platform-integration/android/platform-views.md
@@ -52,7 +52,7 @@ Flutter content is rendered directly into a Surface.
 * `+` good performance for Android Views
 * `+` best performance for Flutter rendering.
 * `+` all transformations work correctly.
-* `-` quick scrolling (e.g. a web view) will be janky
+* `-` quick scrolling (such as a web view) will be janky
 * `-` SurfaceViews are problematic in this mode and will be moved into a virtual
 display (breaking a11y)
 * `-` Text magnifier will break unless Flutter is rendered into a TextureView.
@@ -497,8 +497,7 @@ manually invalidate the view after they have been drawn to
 Manual view invalidation is done by calling `invalidate` on the View
 or one of its parent views.
 
-[`AndroidViewSurface`]:
-{{site.api}}/flutter/widgets/AndroidViewSurface-class.html
+[`AndroidViewSurface`]: {{site.api}}/flutter/widgets/AndroidViewSurface-class.html
 
 ### Issues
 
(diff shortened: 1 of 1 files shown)


src/content/platform-integration/android/platform-views.md
```diff
@@ -26,35 +26,98 @@ see [Hosting native macOS views][].
...
-* `-` quick scrolling (e.g. a web view) will be janky
-* `-` SurfaceViews are problematic in this mode and will be moved into a virtual display (breaking a11y)
+* `-` quick scrolling (such as a web view) will be janky
+* `-` SurfaceViews are problematic in this mode and will be moved into a virtual
+display (breaking a11y)
 * `-` Text magnifier will break unless Flutter is rendered into a TextureView.
 
-To create a platform view on Android,
-use the following steps:
+## [Hybrid Composition++ (HCPP)](#hcpp)
```
jmagman: Could you move this above `Texture Layer`?

src/content/platform-integration/android/platform-views.md
```diff
@@ -26,35 +26,98 @@ see [Hosting native macOS views][].
...
+
+:::note
+This feature is experimental and available starting from Flutter 3.44.
+:::
+
+HCPP is the latest hybrid composition strategy,
+designed to solve compositing performance and synchronization issues
+seen in the original Hybrid Composition mode.
```
jmagman, line 54: I think this is the first reference to `Hybrid Composition`, maybe link to the section describing it, which is below.

src/content/platform-integration/android/platform-views.md
```diff
@@ -418,15 +481,16 @@ android {
     }
 }
 ```
-### Surface Views
+### Surface views
 
-Handling SurfaceViews is problematic for Flutter and should be avoided when possible.
+Handling SurfaceViews is problematic for Flutter and should be avoided when
```
jmagman: This isn't true for HCPP, maybe we can either remove this section, or move it into each of the non-HCPP sections.

jmagman (commented):
@gmackall and @jwill  can you take a look at this?
Even when the HCPP details are correct, there may be other areas in the page that aren't relevant anymore with HCPP.

src/content/platform-integration/android/platform-views.md
sfshaza2: I've swapped the sections as requested and now Hybrid Composition comes first.

src/content/platform-integration/android/platform-views.md
sfshaza2: I've removed it.

gmackall (approved):

antfitch (approved):
Looks so good.

PR #13352, opened 2026-05-04
Adding a section on managing app lifecycle

### LINKED ISSUE #13078: How to manage lifecycle app-wide state?
### Page URL

https://docs.flutter.dev/app-architecture/guide

### Page source

https://github.com/flutter/website/blob/main/src/content/app-architecture/guide.md

### Describe the problem

I've read these docs many times, but I don't see anywhere that mentions lifecycle app-wide state. There are many cases where you want to use a state shared between multiple ViewModels, that you do not want to be storing in a database or leaving the application. You only want this shared state to be during the app lifecycle.

- **ViewModels**: Have a 1-1 with View so can't be shared
- **Services**: Only used for external data. `"They wrap API endpoints and expose asynchronous response objects, such as Future and Stream objects."`
- **Repositories**: Don't seem to hold data, just used to transform it from services. `"They're responsible for polling data from services, and transforming that raw data into domain models."`
- **Use-cases**: Seems like it could be here, but from reading the docs, it shouldn't hold state and should depend on repositories `"This method of adding use-cases is defined by the following rules: 1. Use-cases depend on repositories"`
 
Lifecycle app-wide state seems to be excluded from this MVVM approach.

### Expected fix

We have our own approach that we use, but I am sure the Flutter team who has worked on this framework with other companies can have a better answer.

If it's helpful, here's what we do: https://www.hungrimind.com/articles/the-definitive-guide-to-our-mvvm-architecture-in-flutter

sfshaza2:
Fixes https://github.com/flutter/website/issues/13078

### DIFF
--- a/src/content/app-architecture/guide.md
+++ b/src/content/app-architecture/guide.md
@@ -1,7 +1,7 @@
 ---
 title: Guide to app architecture
 shortTitle: Architecture guide
-description: >
+description: >-
   The recommended way to architect a Flutter app.
 prev:
     title: Common architecture concepts
@@ -17,8 +17,8 @@ making them easier to scale, test, and maintain.
 However, they're guidelines, not steadfast rules,
 and you should adapt them to your unique requirements.
 
-This section provides a high-level overview of how Flutter applications can be
-architected. It explains the layers of an application,
+This section provides a high-level overview of how Flutter applications
+can be architected. It explains the layers of an application,
 along with the classes that exist within each layer.
 The section after this provides concrete code samples and
 walks through a Flutter application that's implemented these recommendations.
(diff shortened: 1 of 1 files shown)


khanhnwin (approved):

conooi (approved):

src/content/app-architecture/guide.md
```diff
@@ -17,8 +17,8 @@ making them easier to scale, test, and maintain.
 However, they're guidelines, not steadfast rules,
 and you should adapt them to your unique requirements.
 
-This section provides a high-level overview of how Flutter applications can be
-architected. It explains the layers of an application,
+This section provides a high-level overview of how Flutter applications
+can be architected. It explains the layers of an application,
```
lamek, line 21: NIT: I would make this "how to architect Flutter applications". Even though it's our opinion and CAN do it other ways, I'd still make this more assertive.

lamek (approved):
LGTM - on suggestion

PR #13365, opened 2026-05-06
Ensure use of latest Jaspr and build deps

parlough:
Jaspr 0.23.1 includes some fixes when running an AOT installed version of `jaspr_cli`.


sfshaza2 (approved):
lgtm

PR #13377, opened 2026-05-09
Update breaking changes for upcoming 3.44 release

parlough:
Once I have a chance I'll open a follow-up PR with some clean up to each notice and guide.


sfshaza2 (approved):
Thank you, @parlough!

PR #13378, opened 2026-05-09
Fix a variety of spelling, grammar, and punctuation issues

parlough:


sfshaza2 (approved):
Thanks, @parlough! Gee, I hate to think how many of these are my fault.

PR #13396, opened 2026-05-18
Update banner for Flutter 3.44

johnpryan:


parlough (changes requested):
Thanks John!

Requesting changes so this doesn't land on accident.

sfshaza2 (approved):
lgtm, so land when ready!

parlough (approved):
Updated the text and added the blog link. Thanks!

PR #13408, opened 2026-05-19
Deduplicate install structions from archive page

parlough:
We don't want to duplicate installation instructions if not necessary. If a developer visiting the archive doesn't know how to install it, we can direct them to the dedicated page on [how to install Flutter manually](https://docs.flutter.dev/install/manual).


sfshaza2 (approved):
lgtm

Issue #13313, opened 2026-04-27
Remove outdated c-interop docs page


johnpryan:
### Page URL

https://docs.flutter.dev/platform-integration/ios/c-interop

### Page source

https://github.com/flutter/website/blob/main/src/content/platform-integration/ios/c-interop.md

### Describe the problem

This page is out of date, and should be removed before the upcoming Dart 3.12 / Flutter 3.44 release. They mention Cocapods, which we don't want to encourage now that Swift Package Manager is the default.

### Expected fix

No mentions of Cocoapods should be on the website

### I would like to fix this problem.

- [x] I will try and fix this problem on docs.flutter.dev.

vashworth:
@dcharkes Are there plans to remove the `plugin_ffi` template? Or do we plan to continue supporting it?

dcharkes:
> Are there plans to remove the `plugin_ffi` template? Or do we plan to continue supporting it?

We will remove it. We've already deprecated it and hid the documentation page from the menu that you've linked.

Edit: Maybe we should just go ahead and remove those old pages.

@johnpryan where did you find a link to this page?

johnpryan:
Updated the title to "Remove outdated c-interop docs page". I'm thinking we can remove this for the upcoming Dart 3.12 / Flutter 3.44 release?

dcharkes:
> Updated the title to "Remove outdated c-interop docs page". I'm thinking we can remove this for the upcoming Dart 3.12 / Flutter 3.44 release?

SGTM. FYI @stuartmorgan-g @goderbauer @conooi

PR #13469, opened 2026-06-05
Combine and simplify legacy ffi plugin pages

parlough:
Combine the 3 pages (iOS, macOS, Android) covering the legacy template and mechanisms into a single, cleaned up page so we have a smaller legacy surface to maintain while not confusing developers who already used the template or are still using the mechanisms used by it. Also completes some clean up, updates, and fixes to the consolidated content.

However, removes or updates the remaining links to the legacy pages to their more up-to-date counterparts.

Contributes to https://github.com/flutter/website/issues/13313


sfshaza2 (approved):
lgtm, thx!

PR #13483, opened 2026-06-09
Improve UX of sidenav

Anderson-Andre-P:
Improves the sidebar navigation UX with two visual changes:

  1. Increased indentation of nested nav items — the margin-left of collapsed sub-lists was too small, making the
  hierarchy hard to read. Increased from 0.675rem to 1.75rem for clearer visual nesting.
  2. Active page highlight — the previously active item had only a very subtle background (4% opacity black/white) that
  was hard to notice. The li containing the current page now receives a light blue background (color-mix with the site
  primary color at 12% opacity) and a 3px solid blue left border, making it immediately clear which page the user is
  on.

  Both changes use existing CSS variables (--site-primary-color) to respect light and dark mode automatically.
  
  ---
  Issues fixed by this PR (if any):

  None.

  ---
  PRs or commits this PR depends on (if any):

  None.


Anderson-Andre-P (commented):
Fixed. Moved background, border, and padding directly to the a tag, removing the :has() block. Added padding-inline-start: calc(0.4rem - 3px) to compensate for the border width and keep text aligned with inactive items.

sites/docs/lib/_sass/components/_sidebar.scss
```diff
@@ -157,6 +157,9 @@
 
           &:not(.collapsible) {
             color: var(--site-link-fgColor);
+            background-color: color-mix(in srgb, var(--site-primary-color) 12%, transparent);
+            border-inline-start: 3px solid var(--site-primary-color);
+            padding-inline-start: calc(0.4rem - 3px);
```
parlough: I think the new background color here is nice and likely sufficient. Could we try just that for now? The border ends up curving and looks a bit out of place.

```suggestion
```

sites/docs/lib/_sass/components/_sidebar.scss
```diff
@@ -171,7 +174,7 @@
     }
 
     ul {
-      margin-left: 0.675rem;
+      margin-inline-start: 1.75rem;
```
parlough: This ends up making double nested items have very little space and split too easily. I agree some extra indentation would be nice, but this ends up a bit too much. How about something like `.875rem` or `1rem`?

parlough (changes requested):
Thanks for thinking to improve this and sorry about the delay on the review!

I have two comments/suggestions. Let me know your thoughts :)

sites/docs/lib/_sass/components/_sidebar.scss
Anderson-Andre-P: I ran this project again and noticed that this border is indeed shifted slightly to the left compared to the other navigation items. The border that looks off is the one on the right side, correct? We can fix that later by increasing it from 3px to 4px, but for now we can remove it without any problems.

sites/docs/lib/_sass/components/_sidebar.scss
Anderson-Andre-P: That's fine, we can keep it at `1rem`. Maybe the fact that I use a very large monitor influenced me to make it much bigger than it actually needed to be.

Anderson-Andre-P (commented):
New submission for review.

sites/docs/lib/_sass/components/_sidebar.scss
```diff
@@ -157,6 +157,7 @@
 
           &:not(.collapsible) {
             color: var(--site-link-fgColor);
+            background-color: color-mix(in srgb, var(--site-primary-color) 12%, transparent);
```
parlough: Reducing this to 10% keeps us at a sufficient contrast ratio in light mode.

```suggestion
            background-color: color-mix(in srgb, var(--site-primary-color) 10%, transparent);
```

parlough (approved):
Thanks for those adjustments! Looks good to me.

PR #13489, opened 2026-06-10
Update and clarify command package suggestion

### LINKED ISSUE #12204: Confusing package recommendation: flutter_command in addition to package:provider
### Page URL

https://docs.flutter.dev/app-architecture/case-study/ui-layer/

### Page source

https://github.com/flutter/website/tree/main/src/content/app-architecture/case-study/ui-layer.md

### Describe the problem

In the Architecture Case Study you suggest package:provider - and then shortly after in the next section you recommend using flutter_command - is this not unnecessary if using provider ? If so, a little more clarity would be useful to the likely audience of these pages.

### Expected fix

Make it clear (if) flutter_command is not necessary in addition to using package:provider.

### Additional context

_No response_

parlough:
- Replaces `flutter_command` with its successor `command_it`.
- Clarify what the package provides.
- Use a softer title to present it as an option rather than requirement or even strong recommendation.

Fixes https://github.com/flutter/website/issues/12204

### DIFF
--- a/sites/docs/src/content/app-architecture/case-study/ui-layer.md
+++ b/sites/docs/src/content/app-architecture/case-study/ui-layer.md
@@ -530,12 +530,14 @@ the number of arguments that the underlying method expects.
 You can see examples of these implementation classes in
 the Compass app's [`utils` directory][].
 
-:::tip Package recommendation
-Instead of writing your own `Command` class,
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
lgtm

PR #13518, opened 2026-06-18
[docs] Consolidate to flex versions of Google Sans fonts

parlough:
Consolidate use of the original Google Sans fonts with the newer, publicly available ones that support variable axes. This increases consistency across the site and enables compliant use of the same fonts on forked sites.

Marketing site fonts were also updated to these in https://github.com/flutter/website/pull/13519.


sfshaza2 (approved):
lgtm. Thanks!

PR #13533, opened 2026-06-24
Reformat with latest formatter updates and fixes

parlough:
This fixes most cases of the `#enddocregion` comments after block bodies being unnecessarily moved during formatting.


sfshaza2 (approved):
lgtm

PR #13569, opened 2026-07-10
Secure GitHub Actions workflow

conooi:
This PR hardens the repository's GitHub Actions workflows against security risks flagged by the zizmor auditing tool. The default top-level permissions have been restricted to contents: read to enforce the principle of least privilege, and all repository checkout steps now set persist-credentials: false to prevent token leakage. A local audit verified that these updates resolve all findings with no outstanding issues.


parlough (approved):
Thanks for tackling this @conooi! Looks good to me.

PR #13579, opened 2026-07-15
Adding a page on dependency management.

sfshaza2:
As per @stuartmorgan-g.

New file staged here: https://flutter-docs-prod--docs-pr13579-versioning-ej79wt5n.web.app/packages-and-plugins/dependency-management


sites/docs/src/content/packages-and-plugins/dependency-management.md
```diff
@@ -0,0 +1,250 @@
...
+The `pubspec.lock` file contains the exact version of every direct and
+transitive dependency used in your project.
+This file ensures that every developer on your team, and your build servers,
+use the exact same package versions.
+Commit the `pubspec.lock` file to version control for application projects,
+but not for package projects.
+
+:::note Use version ranges, not exact versions
+Some developers specify exact versions in `pubspec.yaml`
+(for example, `foo: 1.0.0`) to guarantee that their builds are predictable.
```
stuartmorgan-g: I think we should avoid leading with describing the antipattern; if someone is skimming they might not realize this is us saying what *not* to do.

How about making the whole box just: "Since `pubspec.lock` always pins exact versions for predictable builds, you should use ranges in your `pubspec.yaml`. This allows the [....]"

sites/docs/src/content/packages-and-plugins/dependency-management.md
```diff
@@ -0,0 +1,250 @@
...
+However, `pubspec.lock` already pins exact versions for predictable builds.
+
+Specify version ranges (using caret syntax) in your `pubspec.yaml` file
```
stuartmorgan-g: I removed this parenthetical in my proposed text because the page hasn't introduced caret syntax yet.

sites/docs/src/content/packages-and-plugins/dependency-management.md
```diff
@@ -0,0 +1,250 @@
...
+### Traditional ranges
+
+You can define explicit version ranges using comparison operators:
+
+```yaml
+dependencies:
+  url_launcher: '>=5.4.0 <6.0.0'
+```
+
+### Any version
```
stuartmorgan-g, line 109: I'm torn on whether or not we should have this section. I don't really want to introduce the idea at all, but I guess people will get bad advice from other sources and this gives us a clear place to say why not to do it.

sites/docs/src/content/packages-and-plugins/dependency-management.md
```diff
@@ -0,0 +1,250 @@
...
+If you do not specify a version constraint or use `any`,
+the solver can choose any version:
+
+```yaml
+dependencies:
+  url_launcher: any
+```
+
+Avoid using `any` because a future package update might introduce breaking
+changes that break your build.
```
stuartmorgan-g: Maybe we make this a warning box?

sites/docs/src/content/packages-and-plugins/dependency-management.md
```diff
@@ -0,0 +1,250 @@
...
+A dependency conflict occurs when two packages in your dependency tree require
+incompatible versions of the same transitive dependency.
+
+Consider this example scenario:
+
+1. Your app depends on two direct dependencies: `package_a` and `package_b`.
+1. `package_a` depends on `foo: ^1.0.0`.
+1. `package_b` depends on `foo: ^2.0.0`.
+
+```mermaid
```
stuartmorgan-g: I assume this will turn into a diagram? It's just text on the staged version.

sites/docs/src/content/packages-and-plugins/dependency-management.md
```diff
@@ -0,0 +1,250 @@
...
+
+```console
+Because every version of package_a depends on foo ^1.0.0 and
+every version of package_b depends on foo ^2.0.0,
+package_a is incompatible with package_b.
+So, because my_app depends on both package_a ^1.0.0
+and package_b ^1.0.0, version solving failed.
+```
+
+To read this error message, trace the conflict from the bottom up:
```
stuartmorgan-g: Maybe reword this to make it clear that "bottom" here is referring to the dependency graph, not the console output? Because step 1 is about finding something that's at the start of the example output, not the end.

stuartmorgan-g (commented):
This is great!

sites/docs/src/content/packages-and-plugins/dependency-management.md
sfshaza2: Yeah. I know that we don't recommend it. I've strengthened the avoidance with a note.

sites/docs/src/content/packages-and-plugins/dependency-management.md
sfshaza2: D'oh! :D

stuartmorgan-g (approved):
LGTM!

conooi (approved):

abdallahshaban557:
This is an amazing addition to our docs! Thank you @stuartmorgan-g and @sfshaza2 !

PR #13589, opened 2026-07-17
Generate structured ld+json site data

parlough:
- Generates and attaches structured ld+json data to each page.
- Removes the original RDFa from the breadcrumbs since the ld+json data includes it already.
- Improves last updated dates for pages based on data, like the widget catalog and the glossary.


sfshaza2 (approved):
lgtm

PR #13591, opened 2026-07-18
Document multiple Flutter plugin instances

### LINKED ISSUE #150855: Document multiple engines and plugin instances in plugin development docs
We should explicitly document somewhere in the plugin-author-facing docs (probably [here](https://docs.flutter.dev/packages-and-plugins/developing-packages) or [here](https://docs.flutter.dev/platform-integration/platform-channels)) that plugins need to be prepared to handle the case where they are running in an application with multiple engine instances, which can mean that multiple instances of the plugin class are created, each with their own lifetimes.

See https://github.com/flutter/flutter/issues/150674 and https://github.com/flutter/flutter/issues/150674 for cases where this has caused confusion in the past.

awnoa:

Documents that an app can run multiple Flutter engines, that each engine creates its own plugin instance, and that those instances can have independent lifetimes. It also adds guidance for scoping engine-specific state and coordinating shared native resources.

Fixes flutter/flutter#150855.

None.

## Presubmit checklist

- [ ] If you are unwilling, or unable, to sign the CLA, even for a _tiny_, one-word PR, please file an issue instead of a PR.

### DIFF
--- a/sites/docs/src/content/packages-and-plugins/developing-packages.md
+++ b/sites/docs/src/content/packages-and-plugins/developing-packages.md
@@ -145,7 +145,8 @@ in global or static variables, and clean it up when the plugin
 detaches from its engine.
 
 If plugin instances share a native singleton or other global resource,
-coordinate access across the instances.
+coordinate access across the instances, for example by using
+reference counting.
 Release the resource only after every plugin instance
 has stopped using it.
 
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
lgtm. Thanks!

PR #13600, opened 2026-07-21
Fix unbalanced parens in SwiftUI devs Padding example

meliharik:
The `Padding` example in the "Flutter for SwiftUI developers" page closes with three parens where the snippet only opens one:

```diff
 Padding(                         // <-- This is a Widget
   padding: EdgeInsets.all(10.0), // <-- So is this
   child: Text("Hello, World!"),  // <-- This, too
-)));
+)
```

`EdgeInsets.all(...)` and `Text(...)` are each self-closing on their own line, so `Padding(` is the only outstanding paren. The block goes from 3 open / 5 close to 3 / 3.

The snippet has no `<?code-excerpt?>` region, so `dash_site analyze-dart` doesn't cover it — which is presumably how it survived.


sfshaza2 (approved):
You are exactly right about _how_ this survived! Our Swift code isn't analyzed. Thanks for catching this!

PR #13605, opened 2026-07-21
Fix and update a few links using template variables

parlough:


sfshaza2 (approved):
Thanks for cleaning this up! Soon the material and cupertino links will be to pub.dev/material_ui and pub.dev/cupertino_ui.

PR #13606, opened 2026-07-21
Update `SurfaceProducer` migration doc to use `onSurfaceCleanup`

### LINKED ISSUE #11574: SurfaceProducer migration doc should be updated to use new onSurfaceCleanup function
### Page URL

https://docs.flutter.dev/release/breaking-changes/android-surface-plugins 

### Page source

_No response_

### Describe the problem

When the `onSurfaceDestroyed` is deprecated in the future, and replaced by `onSurfaceCleanup`, the migration guide should be updated to reflect this.

There should also be a note that notifies readers that the old API was available in Flutter 3.27, but renamed in a to-be-named subsequent release.

There is prior art for a similar change in the document as of today:

```
In the upcoming stable release, 3.27, onSurfaceCreated is deprecated, and onSurfaceAvailable and handlesCropAndRotation are added.
```

so this change will closely follow that pattern.

### Expected fix

The migration guide's code sample should be updated to use `onSurfaceCleanup` instead, and a footnote should be added to clarify how the transition happened, akin to what we did for the change relating to `onSurfaceAvailable`.

### Additional context

The following issue talks about the aforementioned change: https://github.com/flutter/flutter/issues/160933
and the PR that adds the API change is https://github.com/flutter/flutter/pull/160937

parlough:
Resolves https://github.com/flutter/website/issues/11574


sfshaza2 (approved):
lgtm!

PR #13623, opened 2026-07-23
Enhance dependency upgrade instructions

mit-mit:
Updated instructions for upgrading dependencies in `pubspec.yaml` to use the `--major-versions` flag.


mit-mit:
This is consistent with what we have in https://docs.flutter.dev/install/upgrade#upgrade-packages

sites/docs/src/content/packages-and-plugins/dependency-management.md
```diff
@@ -179,9 +179,9 @@ constraints, run:
 flutter pub outdated
 ```
 
-If newer versions exist,
-update the version constraints in your `pubspec.yaml` to the newer versions and
-run `flutter pub get`.
+To upgrade to newer major versions, run `flutter pub upgrade --major-versions`.
+This command automatically upgrades the constraints in your `pubspec.yaml` and
+then runs `flutter pub get` to retrieve those upgraded dependencies.
```
parlough: Consider being a bit more specific here and turning the mention of `flutter pub get` into what is happening (retrieval).

```suggestion
This command updates the constraints in your `pubspec.yaml` file to
allow the latest resolvable versions, then retrieves those versions.
```

parlough (approved):
Thanks @mit-mit, suggesting this command makes a lot of sense.

One suggestion for you to consider, but looks good to me!

sites/docs/src/content/packages-and-plugins/dependency-management.md
mit-mit: I prefer using the word "upgrade" rather than "updates" to match the pub verb. But I can take the rest of the suggestion.

PR #13624, opened 2026-07-23
Remove iOS SWE position

parlough:
The linked careers page is no longer available.


jesskuras (approved):
Reviewed code and staged version, looks good!

PR #13636, opened 2026-07-28
Move docs firebase.json to sites/docs directory

parlough:
Moves the `firebase.json` file for the docs site from the root directory of the repository to the `sites/docs` directory to be consistent with the marketing site and avoid potential confusion.

For future purposes, use `docs.flutter.dev/to/site-redirects` for references to editing this file.


sites/docs/.firebaserc
```diff
@@ -0,0 +1,8 @@
+{
+  "projects": {
+    "default": "flutter-docs-prod",
+    "devtools-staging": "flutter-website-dt-staging"
+    "sz": "sz-flutter",
+    "sz2": "sz-flutter-2"
```
sfshaza2: I don't think we need these anymore...

abdallahshaban557 (approved):
LGTM

parlough:
@sfshaza2 Any concerns about moving the docs site's `firebase.json` file for consistency between the sites and a simpler root directory?

sfshaza2:
Well, just in general there are SO many changes to the repo that it's much hard to find things now. At least for me. But I don't have specific concerns about moving this file.

PR #13641, opened 2026-07-29
Fix trailing question mark added to URLs when updating query parameters

abdallahshaban557:
## Description

In Dart, calling `Uri.replace(queryParameters: {})` sets the query component to an empty string (``), which `Uri.toString()` serializes with a trailing question mark (e.g. `https://flutter.dev/showcase?`).

This PR updates `filters.dart` in `sites/www` and `global_scripts.dart` in `sites/docs` to construct a `Uri` without a query component when the updated query parameter map is empty, preventing unnecessary trailing question marks from being pushed to browser history.

## Verification
- Analyzed codebase with `dart analyze` (clean build, 0 issues).
- Tested query parameter stripping behavior in Dart.


parlough (approved):
Thanks @abdallahshaban557, looks good to me and makes sense.

Note I pushed a commit to ensure the correct `?` is stripped and moved it to a shared util extension.

abdallahshaban557:
Nice! will follow this pattern for things like this in the future @parlough!

PR #13648, opened 2026-07-30
Add note about deferred loading to flutter.dev/platform-integration/web/wasm

mkustermann:
We have recently added the `--enable-wasm-deferred-loading` flag and should mention it in the documentation to make users aware.

### DIFF
--- a/sites/docs/src/content/platform-integration/web/wasm.md
+++ b/sites/docs/src/content/platform-integration/web/wasm.md
@@ -80,7 +80,7 @@ $ flutter build web --wasm --enable-wasm-deferred-loading
 ```
 
 This is available on the flutter main branch and will become generally
-available later this year in the 3.50.0 flutter stable release.
+available later this year in the 3.50 flutter stable release.
 :::
 
 #### Wasm production debugging
--- a/sites/docs/src/content/platform-integration/web/wasm.md
+++ b/sites/docs/src/content/platform-integration/web/wasm.md
@@ -80,7 +80,7 @@ $ flutter build web --wasm --enable-wasm-deferred-loading
 ```
 
 This is available on the flutter main branch and will become generally
-available later this year in the 3.47.0 flutter stable release.
+available later this year in the 3.50.0 flutter stable release.
 :::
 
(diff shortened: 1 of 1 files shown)


biggs0125 (approved):

biggs0125 (commented):
I want to advertise this but should we either wait until this is on the stable branch or say something about this becoming available in Flutter 3.50? That way people don't try to do it today on the stable branch where it still doesn't work.

mkustermann:
> I want to advertise this but should we either wait until this is on the stable branch or say something about this becoming available in Flutter 3.50? That way people don't try to do it today on the stable branch where it still doesn't work.

If we mention the deferred loading work on the dart & flutter blog posts, then we're telling that this is available in preview now. If we tell that to users, I think we should have documentation for it, which this PR adds.

See also my response here: https://github.com/dart-lang/site-www/pull/7419#issuecomment-5135177976

(If we don't mention deferred loading in the dart & flutter blog posts - wdyt? - then there's no need to have it in the docs yet either)

biggs0125:
I don't mind mentioning it early here like you have but I do think we should mention a version number.  If people want to play around with it earlier on the main/dev branch they can do that based on the version number. Or they know to wait until that stable version. And going forward people will know how far back the support was there.

At some point in the far(-ish) future we can delete the version number for now I think it provides more clarity.

mkustermann:
> I don't mind mentioning it early here like you have but I do think we should mention a version number. If people want to play around with it earlier on the main/dev branch they can do that based on the version number. Or they know to wait until that stable version. And going forward people will know how far back the support was there.

Okay, so I've rephrased this now to

  _This is available on the flutter main branch and will become generally
  available later this year in the 3.50.0 flutter stable release._

So it should be clear to use it one should use main branch and it will tell which stable will ship it. Once the stable ships, we probably enable it by-default, making this note itself not necessary anymore and we can remove it.

mkustermann:
PTAL

sfshaza2 (approved):
This is a great addition, and is acceptable given the version mention. Thanks!!

PR #13658, opened 2026-08-03
Add "Why Flutter" whitepaper landing page to flutter.dev

vmkstudio:
Adds `/why-flutter`, a marketing page making the business case for adopting
Flutter, implemented from a design handoff.

## Approach

The goal was to reuse what `sites/www` already has rather than introduce a
parallel set of conventions:

| Concern | Reused |
| --- | --- |
| Header and footer | `DefaultLayout` — the page adds neither |
| Closing call to action | The shared `CTASection`, unchanged |
| Icons | The existing `Icon` component (Material Symbols Rounded) |
| Buttons | `.btn` from `components/_btn.scss` |
| Colour, radius, spacing | Tokens from `core/_vars.scss` |
| Breakpoints | `breakpoints.screen()` mixins, mobile-first |

Nine of the design's eleven type sizes map exactly onto the existing heading
scale. The section subhead moved 32px → 36px (`--font-size-heading-2`) and the
case quote landed on 36px exactly. Only the stat numerals (`3X`, `97%`) and the
fixed hero stage geometry are literals, as no token covers them.

## New files

- `content/why-us/index.md` — page content and frontmatter
- `lib/src/pages/why_us_page.dart` — composes the sections
- `lib/src/components/sections/why_us_*.dart` — six page sections
- `lib/src/utils/visibility_observer.dart` — scroll-into-view trigger and tween
- `lib/styles/pages/_why_us.scss` — page styles

## Note for reviewers

The four sections that hydrate on the client take their asset URLs as
constructor parameters instead of calling `context.asset()` themselves.
`utils/asset_utils.dart` imports `dart:io`, so calling it from a `@client`
component pulls the server-only SDK into the client bundle and
`build_web_compilers` silently skips compiling `main.client.dart.js` —
disabling interactivity across the whole site, not just this page. Worth
knowing about if other pages later add client-side sections.

## Verification

- `dart analyze` — clean
- `dart run dash_site --site=www build` — succeeds; `/why-us` generated
- Client bundle compiles (322,812 bytes vs 318,796 on `main`), no skipped
  entrypoints
- All 28 assets referenced by the page resolve in the build output

## Open questions

1. **The whitepaper PDF is 7.1 MB and committed to the repo.** Every other PDF
   the site links to is hosted externally. Happy to swap the link for a CDN URL
   if you'd prefer to host it there — probably the right call.
2. The "One Design System" card uses the `hub` symbol where the design had a
   bespoke SVG. Close, but not identical.
3. The page is not yet linked from the header nav; let me know where it belongs.


sfshaza2:
@vmkstudio, I didn't know that you did code, too! I'm telling the workflow to run so we can see the generated output.

craiglabenz:
Two issues I see with the PR as-is, then we should be good to go.

1. The download link for the PDF itself is currently dead. For me, clicking the button does nothing.

2. The "Design fidelity" section needs a little responsive UI love on mobile

Screenshot 2026-09-08 at 11 12 52 AM

**Update for posterity:** The responsive UI issue is probably only a P3 or P4 because horizontal scrolling works on a trackpad, but not via clicking and dragging. But, on a mobile workflow, few users will be clicking with a mouse; so the second item should be fine.

sfshaza2 (approved):
lgtm

PR #13664, opened 2026-08-03
Add redirect for android-local-area-permission

mboetger:
Add `/go/android-local-area-permission` design doc redirect.


parlough (approved):
Thanks!

PR #13665, opened 2026-08-03
[www] Surface past events on events page

### LINKED ISSUE #13499: Add a filter to the events page on flutter.dev to show all events including past ones
### Describe the problem

On fluter.dev/events, we currently remove events that have already passed. We need to add a filter to show even previous events.

### Expected fix

A new filter is available to show all past events or perhaps events from a specific time period.

### On which browser(s) did you experience this issue?

_No response_

### Additional context

_No response_

parlough:
Add a separate "Past events" section to the events page rather than an additional filter. In the future, we can consider if we want to change this to "Recent events" within a certain time span.

Resolves https://github.com/flutter/website/issues/13499

**Staged:** https://flutter-dev-230821--www-pr13665-feat-past-events-thr4ziro.web.app/events


abdallahshaban557 (approved):
LGTM

PR #13686, opened 2026-08-06
docs: recommend Homebrew for CocoaPods installation

### LINKED ISSUE #12446: iOS apps installation - cocoapods installation missing steps
### Page URL

### Page source

https://github.com/flutter/website/tree/main/src/content/get-started/install/macos/mobile-ios.md

### Describe the problem

Steps described in page to install `cocoapods` are incomplete. Also, the linked install guide is contradictory:
- [here](https://cocoapods.org/) it states:
   > We recommend you **use the default** ruby
   
   emphasis mine.
   Following this approach leads to `flutter doctor` complaining:
   > CocoaPods installed but not working
- while [here](https://guides.cocoapods.org/using/getting-started.html#getting-started) it states
   >we recommend **not using the system provided Ruby** and instead installing a newer Ruby [...] with `brew install ruby`

   emphasis mine.
   While this approach works, the page doesn't mention the fundamental `$PATH` update step.

### Expected fix

The page should enlist all the steps necessary to install `cocoapods` such as `flutter doctor` doesn't detect issues.
Steps I took to make it work:
- install newer ruby via Homebrew: `brew install ruby`
- update `$PATH`: `export PATH="/opt/homebrew/opt/ruby/bin:$PATH"`
- check that `$PATH` has been correctly updated:
   ```sh
   which gem
   # Should output /opt/homebrew/opt/ruby/bin/gem
   ```
- get `gem` installation directory: 
   ```sh
   gem environment home
   # Should output /opt/homebrew/lib/ruby/gems/3.4.0
   ```
- include it in `$PATH`: `export PATH=GEM_ENVIRONMENT_HOME_OUTPUT/bin:$PATH`
- install `cocoapods`: `gem install cocoapods`

### Additional context

Tested on MacBook Air M4

lamek:
Updates macOS and iOS setup guides to recommend installing CocoaPods via Homebrew to avoid Ruby permission errors. Applies style fixes for active voice and heading clarity.

Resolves https://github.com/flutter/website/issues/12446

### DIFF
--- a/sites/docs/src/content/platform-integration/ios/setup.md
+++ b/sites/docs/src/content/platform-integration/ios/setup.md
@@ -44,7 +44,7 @@ an iOS physical device or on the iOS Simulator.
     ```
 
     If you downloaded Xcode elsewhere or need to use a different version,
-    replace `/Applications/Xcode.app` with the path to there instead.
+    replace `/Applications/Xcode.app` with the path to your Xcode installation.
 
  1. <h3>Agree to the Xcode licenses</h3>
 
@@ -138,7 +138,7 @@ Set up each iOS device on which you want to test.
 
        ![Trust Mac](/assets/images/docs/setup/trust-computer.png)
 
- 1. <h3>Configure your physical iOS device</h3>
+ 1. <h3>Enable Developer Mode</h3>
 
     Apple requires enabling **[Developer Mode][]**
     on the device to protect against malicious software.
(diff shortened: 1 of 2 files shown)


parlough (commented):
I'm not sure we should suggest using Homebrew actually.

As the KPM docs [call out](https://kotlinlang.org/docs/multiplatform/multiplatform-cocoapods-overview.html#-5nhdn8_34:~:text=The%20CocoaPods%20installation%20with%20Homebrew%20might%20result%20in%20compatibility%20issues.) it can cause some issues.

Consider just relying on the guide or if you want, inline some instructions for popular methods to install Ruby, probably RVM, rbenv, and Homebrew.

sfshaza2:
@parlough has a good point. I look forward to removing Cocoapods entirely. :D

lamek:
Sounds good. I've reverted the Homebrew suggestion but kept the minor path changes.

sfshaza2 (approved):
lgtm

PR #13700, opened 2026-08-07
Remove poster image from We Spot Turtles showcase

abdallahshaban557:
Removes the `poster` image from the We Spot Turtles showcase frontmatter so the large logo image is no longer rendered at the top of the post.


lamek (approved):
LGTM

abdallahshaban557:
@lamek - I'll merge it after checking the preview!

PR #13713, opened 2026-08-12
Order arm64 builds first in SDK archive

parlough:
In the SDK archive, order arm64 builds before x64 builds within each release. This only affects macOS currently and helps promote arm64 builds over x64 builds on the road to their deprecation.

**Staged:** https://flutter-docs-prod--docs-pr13713-misc-order-arm64-first-3ovtrcs6.web.app/install/archive#stable-channel


sfshaza2 (approved):
lgtm

PR #13732, opened 2026-08-17
Support custom social blog images

parlough:
Allow specifying a custom social image for blog posts, as the animated images are often over the 5Mb limit of platforms like X. Also updates the blog import skill to account for this.


sfshaza2 (approved):
lgtm

PR #13735, opened 2026-08-17
[AI Docs] Integrate AI documentation into the default sidebar

jesskuras:
Integrate the standalone AI sidebar into the main default sidebar (`default.yml`)
under a dedicated `AI` top-level section header.

This migration:
- Adds a new `- header: AI` section in `default.yml` with top-level entries for
  Overview, Develop with AI, and Build AI-powered apps.
- Nests AI evaluations under Develop with AI.
- Updates all 25 AI content markdown files to use the default sidebar.
- Removes the standalone `src/data/sidenav/ai.yml`.


sfshaza2 (approved):
I know this is a draft, @jesskuras, but this was MUCH NEEDED. Thanks! I'm ok with adding this, sooner than later, and tweaking as things evolve.

lamek:
This looks great. I like keeping the top level "AI" tab.

I think we may need at least one redirect, for the default /ai path.

jesskuras:
> This looks great. I like keeping the top level "AI" tab.
> 
> I think we may need at least one redirect, for the default /ai path.

Thanks! I double-checked and since none of the URLs are changing (including the default /ai one) it looks like it still resolves fine.

PR #13736, opened 2026-08-17
Add docs on 'Create app flavors for Windows and Linux'

### LINKED ISSUE #13642: Add docs on "Create app flavors for Windows and Linux".
Flutter 3.47 will be released next month. This version adds the ability to setup app flavors on Windows and Linux. Please add relevant documentation.

PRs:
https://github.com/flutter/flutter/pull/187034
https://github.com/flutter/flutter/pull/187029

sfshaza2:
Closes #13642

Adds documentation for setting up app flavors on Windows and Linux desktop platforms.


sfshaza2:
@AngeloAvv, can you help with a review? The bot is suggesting a more detailed description and I could use some help.

AngeloAvv:
@sfshaza2 sure! I'll take a look in the upcoming days

sfshaza2:
This PR is holding up other PRs. @AngeloAvv, unless you can get to this soon-ish, I plan to land, because these changes are better than nothing. But in that event, I would still really appreciate if you could either file an issue or a PR with improvements!

sites/docs/src/content/deployment/flavors-windows-linux.md
```diff
@@ -0,0 +1,429 @@
...
+
+On Windows and Linux, Flutter uses [CMake][] to configure and build
+the native desktop runner.
+When you run `flutter run` or `flutter build` with the `--flavor` flag,
+Flutter provides the flavor name to CMake as the `FLUTTER_APP_FLAVOR`
+variable and isolates the build outputs in a flavor-specific directory.
+
+The following table illustrates the build directories that Flutter creates
+when a project defines two flavors (`staging`, `production`)
+and two build modes (`debug`, `release`):
```
parlough, line 34: This describes the following table as outlining `debug` mode as well, but it only includes entries for `release`.

sites/docs/src/content/deployment/flavors-windows-linux.md
```diff
@@ -0,0 +1,429 @@
...
+
+<table class="table table-striped">
```
parlough, line 36: Any reason to not use a Markdown table here?

It can be followed with `{:.table .table-striped}` to give it the desired styling.

sites/docs/src/content/deployment/flavors-windows-linux.md
```diff
@@ -0,0 +1,429 @@
...
+      }
+
+      runApp(const MyApp());
+    }
+    ```
+
+    :::note
+    The value of `appFlavor` matches the string passed to the `--flavor` flag.
+    If you run or build without specifying a flavor,
+    `appFlavor` returns `null`.
```
parlough, line 255: This should be updated to account for falling back to the default flavor if it's specified. Perhaps by cross-linking to the "Set a default flavor" section.

sites/docs/src/content/deployment/flavors-windows-linux.md
```diff
@@ -0,0 +1,429 @@
...
+To help distinguish between different flavors at runtime,
+you can customize the native window title for each flavor.
+
+<Tabs key="flavors-os" wrapped="true">
+<Tab name="Windows">
+
+1. Open `windows/runner/CMakeLists.txt` and pass `FLUTTER_APP_FLAVOR`
+   as a preprocessor definition:
+
+```cmake title="windows/runner/CMakeLists.txt"
```
parlough, line 274: This and the content of the other steps should be consistently indented with the rest of their list items/steps.

parlough (commented):
The fact that there's a set of tabs under most of the sections and a table to visual locations by platform, I'd highly recommend splitting this into two pages. It seems there's not much relating the platforms to each other in the case of flavors, so I think it's best for developers to tackle each separately anyway.

sites/docs/src/content/deployment/flavors-windows-linux.md
```diff
@@ -0,0 +1,429 @@
...
+  std::wstring title = L"flavors_example (" + std::wstring(L"" FLUTTER_APP_FLAVOR) + L")";
+  if (!window.Create(title.c_str(), origin, size)) {
+    return EXIT_FAILURE;
+  }
+  #else
+    if (!window.Create(L"flavors_example", origin, size)) {
+      return EXIT_FAILURE;
+    }
+  #endif
+```
```
AngeloAvv: Two issues here.

**1. `std::wstring(L"" FLUTTER_APP_FLAVOR)` relies on a deprecated concatenation rule.**

`FLUTTER_APP_FLAVOR` expands to a narrow literal (`"staging"`), so this is `L"" "staging"` — mixing a wide and a narrow literal. That is implementation-defined in C++11–C++20 and **ill-formed in C++23** ([P2201R1](https://wg21.link/p2201r1)). MSVC accepts it today, so it isn't broken *right now*, but it's a fragile pattern to enshrine in official docs — it breaks on a `/std:` bump.

Cleanest fix is to define a wide variant alongside the narrow one in `windows/runner/CMakeLists.txt`:

```cmake
target_compile_definitions(${BINARY_NAME} PRIVATE "FLUTTER_APP_FLAVOR_W=L\"${FLUTTER_APP_FLAVOR}\"")
```

then use `std::wstring(FLUTTER_APP_FLAVOR_W)` in `main.cpp`.

**2. Indentation is inconsistent** — `#else` / `#endif` and their bodies are indented as if nested inside the `#if` branch.

```suggestion
```cpp title="windows/runner/main.cpp"
#if defined(FLUTTER_APP_FLAVOR_W)
  std::wstring title = L"flavors_example (" + std::wstring(FLUTTER_APP_FLAVOR_W) + L")";
  if (!window.Create(title.c_str(), origin, size)) {
    return EXIT_FAILURE;
  }
#else
  if (!window.Create(L"flavors_example", origin, size)) {
    return EXIT_FAILURE;
  }
#endif
```
```

As an alternative worth considering: `flutter_flavorizr` sidesteps the preprocessor entirely on Windows by using `configure_file` on a `main.cpp.in` template. That one mechanism covers window title, icon, *and* PE metadata — see my comment on the icon section.

File: sites/docs/src/content/deployment/flavors-windows-linux.md [hunk 6]
```diff
@@ -0,0 +1,429 @@
...
+1. Prepare PNG icons for each flavor
+   (for example, `app_icon_staging.png` and `app_icon_production.png`).
+
+1. Create corresponding `.desktop` files for each flavor
+   (for example, `flavors_example_staging.desktop` and
+   `flavors_example_production.desktop`)
+   that reference the respective icon and executable binary name.
+
+1. Update `linux/CMakeLists.txt` to install the correct `.desktop` file
+   and icon based on `FLUTTER_APP_FLAVOR`.
```
AngeloAvv: This section is prose-only, and step 3 — *"Update `linux/CMakeLists.txt` to install the correct `.desktop` file and icon"* — is where all the actual work lives. A reader can't act on it.

It's genuinely harder than the Windows case: on Linux the icon isn't embedded in the binary, so it depends on `.desktop` entries, the `hicolor` icon theme layout, and the packaging format (snap/deb/flatpak) — which is why `flutter_flavorizr` doesn't implement Linux icons at all.

Two reasonable options:

**a)** Give the complete recipe, roughly:

```cmake title="linux/CMakeLists.txt"
if(DEFINED FLUTTER_APP_FLAVOR AND NOT FLUTTER_APP_FLAVOR STREQUAL "")
  set(DESKTOP_FILE "${CMAKE_CURRENT_SOURCE_DIR}/${BINARY_NAME}_${FLUTTER_APP_FLAVOR}.desktop")
  set(ICON_FILE "${CMAKE_CURRENT_SOURCE_DIR}/app_icon_${FLUTTER_APP_FLAVOR}.png")
else()
  set(DESKTOP_FILE "${CMAKE_CURRENT_SOURCE_DIR}/${BINARY_NAME}.desktop")
  set(ICON_FILE "${CMAKE_CURRENT_SOURCE_DIR}/app_icon.png")
endif()

install(FILES "${DESKTOP_FILE}"
  DESTINATION "${CMAKE_INSTALL_PREFIX}/share/applications" COMPONENT Runtime)
install(FILES "${ICON_FILE}"
  DESTINATION "${CMAKE_INSTALL_PREFIX}/share/icons/hicolor/256x256/apps"
  RENAME "${APPLICATION_ID}.png" COMPONENT Runtime)
```

(placed after `add_subdirectory(${FLUTTER_MANAGED_DIR})`, per the scope constraint)

**b)** Cut the Linux icon steps and link to [Build and release a Linux app](/deployment/linux), noting that icon handling is a packaging concern rather than a flavor concern.

Either is fine — but three non-actionable bullets set readers up to fail.

sites/docs/src/content/deployment/flavors-windows-linux.md
```diff
@@ -0,0 +1,429 @@
...
+
+```console title="console"
+$ flutter run -d linux --flavor production
+```
+</Tab>
+</Tabs>
+
+## Launch a flavor {: #launch-a-flavor }
+
+After you configure flavors for your Windows or Linux app,
```
AngeloAvv: Two additions worth making somewhere in this section:

**1. `--flavor` also works with `flutter test` and `flutter drive`** on desktop, not just `run`/`build`. The Linux PR touched `commands/test.dart` for exactly this, and both devicelab tasks (`flavors_test_linux.dart`, `flavors_test_windows.dart`) exercise `createFlavorsTest()` / `createIntegrationTestFlavorsTest()`. Readers wiring flavors into CI will look for this.

**2. Existing projects don't need to regenerate anything.** Neither desktop flavor PR modified the `windows.tmpl` / `linux.tmpl` templates, so `--flavor` works on any existing project as-is — no `flutter create .` re-run, no template migration. That's a genuinely reassuring detail and cheap to state.

sites/docs/src/content/deployment/flavors-windows-linux.md
```diff
@@ -0,0 +1,429 @@
...
+see [Set up Flutter flavors for iOS and macOS][].
+:::
+
+## Overview {: #overview }
+
+A Flutter flavor represents a collection of settings that define
+how a specific version of your app builds and runs.
+For example, a flavor can determine which window title, binary name,
+application ID, API endpoint, asset set, and logging configuration
+applies to a build.
```
AngeloAvv, line 24: "binary name" overpromises. As shipped, the flavor never affects the binary name — the generated templates were not modified by either desktop flavor PR (`grep FLUTTER_APP_FLAVOR packages/flutter_tools/templates/` returns nothing), and renaming it manually breaks executable path resolution (see my comment on the CMake section below).

Window title and application ID are achievable, but only as opt-in project-side customizations.

```suggestion
For example, a flavor can determine which window title,
application ID, API endpoint, asset set, and logging configuration
applies to a build.
```

sites/docs/src/content/deployment/flavors-windows-linux.md
```diff
@@ -0,0 +1,429 @@
...
+
+On Windows and Linux, Flutter uses [CMake][] to configure and build
+the native desktop runner.
+When you run `flutter run` or `flutter build` with the `--flavor` flag,
+Flutter provides the flavor name to CMake as the `FLUTTER_APP_FLAVOR`
+variable and isolates the build outputs in a flavor-specific directory.
```
AngeloAvv, line 30: "Flutter provides the flavor name to CMake as the `FLUTTER_APP_FLAVOR` variable" is accurate but omits the one detail that determines whether any of the examples in this page actually work: *where* in the CMake evaluation order the variable appears.

It is not a `-D` command-line argument. It is emitted into `flutter/ephemeral/generated_config.cmake` with `PARENT_SCOPE`, and that file is included from `<platform>/flutter/CMakeLists.txt`. Practically: `FLUTTER_APP_FLAVOR` is undefined until `add_subdirectory(${FLUTTER_MANAGED_DIR})` returns in the top-level `CMakeLists.txt`.

Stating this up front would prevent the most likely reader mistake — putting the `if(DEFINED FLUTTER_APP_FLAVOR)` block near the top of the file, where it silently does nothing.

```suggestion
On Windows and Linux, Flutter uses [CMake][] to configure and build
the native desktop runner.
When you run `flutter run` or `flutter build` with the `--flavor` flag,
Flutter writes the flavor name into `<platform>/flutter/ephemeral/generated_config.cmake`
as the `FLUTTER_APP_FLAVOR` variable, and isolates the build outputs
in a flavor-specific directory.

:::important
`generated_config.cmake` is included from `<platform>/flutter/CMakeLists.txt`,
so `FLUTTER_APP_FLAVOR` only becomes visible in your top-level
`CMakeLists.txt` **after** the `add_subdirectory(${FLUTTER_MANAGED_DIR})` line.
Any `if(DEFINED FLUTTER_APP_FLAVOR)` block placed before that line
is silently skipped.
:::
```

sites/docs/src/content/deployment/flavors-windows-linux.md
AngeloAvv, line 34: The intro promises two build modes but the table only lists `release` rows.

```suggestion
The following table illustrates the build directories that Flutter creates
when a project defines two flavors (`staging`, `production`)
and builds in release mode:
```

sites/docs/src/content/deployment/flavors-windows-linux.md
```diff
@@ -0,0 +1,429 @@
...
+</table>
+
+[CMake]: https://cmake.org/
+[Set up Flutter flavors for Android]: /deployment/flavors
+[Set up Flutter flavors for iOS and macOS]: /deployment/flavors-ios
+
+## Configure CMake for flavors {: #configure-cmake }
+
+Complete the following steps to configure two flavors called
+`staging` and `production` for a new or existing Flutter project.
```
AngeloAvv, line 78: Worth stating explicitly that this whole section is **optional**, because unlike Android and iOS, desktop flavors need no project configuration to be functional.

With zero CMake changes, `--flavor` already gives you:

- isolated build directories (`build/<platform>/<arch>/<flavor>/…`)
- a populated `appFlavor` constant at runtime
- flavor-filtered assets via the `flavors:` key in `pubspec.yaml`
- `default-flavor` support

CMake configuration is only needed for *native* differentiation — window title, app icon, application ID. Framing it as a required step 1–3 sequence makes desktop flavors look heavier than they are, and pushes readers into the failure modes flagged below.

```suggestion
Passing `--flavor` works on Windows and Linux without any project
configuration: Flutter isolates the build directory, populates
`appFlavor`, and filters flavor-specific assets automatically.

Complete the following optional steps only if you want to differentiate
the *native* runner per flavor — for example its window title,
application ID, or app icon.
This example configures two flavors called `staging` and `production`.
```

sites/docs/src/content/deployment/flavors-windows-linux.md
```diff
@@ -0,0 +1,429 @@
...
+1.  Create a new Flutter project called `flavors_example`:
+
+    ```console title="console"
+    $ flutter create flavors_example
+    $ cd flavors_example
+    ```
+
+1. Configure the CMake build files for your target platform:
+
+<Tabs key="flavors-os" wrapped="true">
```
AngeloAvv, line 89: Markdown list nesting looks off and will likely break the numbering in rendering.

Item 1 (line 80) uses `1.` + 4-space continuation indent, but items at lines 87 and 126 use `1. ` with no indented body, and the `<Tabs>` blocks between them start at column 0. That takes the following content out of the list context, so the ordered list restarts rather than continuing 1 → 2 → 3.

Same pattern in the "Create distinct window titles" section (lines 271/280 and 299/308).

Compare with `flavors-ios.md`, which keeps everything under a consistent 4-space continuation indent. Worth previewing the built page to confirm the step numbers render as 1/2/3.

sites/docs/src/content/deployment/flavors-windows-linux.md
```diff
@@ -0,0 +1,429 @@
...
+Update the configuration to append the flavor name to the binary name
+when a flavor is specified:
+
+```cmake title="windows/CMakeLists.txt"
+# The name of the executable created for the application.
+set(BINARY_NAME "flavors_example")
+if(DEFINED FLUTTER_APP_FLAVOR AND NOT FLUTTER_APP_FLAVOR STREQUAL "")
+  set(BINARY_NAME "${BINARY_NAME}_${FLUTTER_APP_FLAVOR}")
+endif()
+```
```
AngeloAvv, line 102: **Blocking — this snippet cannot work, for two independent reasons.**

**1. `FLUTTER_APP_FLAVOR` is not defined at this point in the file.**

It is not passed as `-D` on the CMake command line. `build_windows.dart` (`_runCmakeGeneration`) and `build_linux.dart` (`_runCmake`) pass no such flag. The variable comes from `cmake.dart`:

```dart
buffer.writeln('set(FLUTTER_APP_FLAVOR "${_escapeBackslashes(flavor)}" PARENT_SCOPE)');
```

which is written into `flutter/ephemeral/generated_config.cmake`. That file is `include()`d from `windows/flutter/CMakeLists.txt`, i.e. from inside `add_subdirectory(${FLUTTER_MANAGED_DIR})` — which sits around line 50 of `windows/CMakeLists.txt`. Because of `PARENT_SCOPE`, the variable only becomes visible in the top-level scope **after** `add_subdirectory` returns.

`set(BINARY_NAME ...)` is at line 7. At that point `DEFINED FLUTTER_APP_FLAVOR` is false, so the `if()` block never executes. It fails silently — the build succeeds and produces an unflavored binary, which is the worst possible failure mode for a docs example.

**2. Suffixing `BINARY_NAME` breaks `flutter run`.**

`getCmakeExecutableName()` in `packages/flutter_tools/lib/src/cmake.dart` parses the *first* literal `set(BINARY_NAME "...")` line with a regex and has no flavor awareness:

```dart
final nameSetPattern = RegExp(r'^\s*set\(BINARY_NAME\s*"(.*)"\s*\)\s*$');
```

Its unit test asserts exactly that (`build_windows_flavor_test.dart`: `expect(getCmakeExecutableName(project), 'myapp')`).

That value feeds `BuildableWindowsApp.executable()` → `WindowsDevice.executablePathForDevice()`. If CMake emits `flavors_example_staging.exe` but the tool looks for `flavors_example.exe`, `flutter run -d windows --flavor staging` fails to launch, and `flutter build` prints a `✓ Built …` line pointing at a non-existent path.

Worth noting: `flutter_flavorizr`, which implements desktop flavors in the wild, deliberately never touches `BINARY_NAME` on either platform, for this exact reason.

**Suggested replacement** — on Windows no top-level CMake change is needed at all:

````suggestion
On Windows, you don't need to modify `windows/CMakeLists.txt`.
Passing `--flavor` already isolates the build directory,
populates `appFlavor`, and filters flavor-specific assets
without any CMake changes.

To customize native behavior per flavor — such as the window title
or the app icon — edit `windows/runner/CMakeLists.txt`, which is
processed after `FLUTTER_APP_FLAVOR` becomes available.
See [Customize configurations](#customize-configurations).
````

File: sites/docs/src/content/deployment/flavors-windows-linux.md [hunk 13]
```diff
@@ -0,0 +1,429 @@
...
+# The name of the executable created for the application.
+set(BINARY_NAME "flavors_example")
+# The unique GTK application identifier for this application.
+set(APPLICATION_ID "com.example.flavors_example")
+
+if(DEFINED FLUTTER_APP_FLAVOR AND NOT FLUTTER_APP_FLAVOR STREQUAL "")
+  set(BINARY_NAME "${BINARY_NAME}_${FLUTTER_APP_FLAVOR}")
+  set(APPLICATION_ID "${APPLICATION_ID}.${FLUTTER_APP_FLAVOR}")
+endif()
+```
```
AngeloAvv, line 121: **Blocking — same two problems as the Windows tab.**

The `if(DEFINED FLUTTER_APP_FLAVOR ...)` block placed at the top of `linux/CMakeLists.txt` never executes: `generated_config.cmake` is included from `linux/flutter/CMakeLists.txt`, i.e. inside `add_subdirectory(${FLUTTER_MANAGED_DIR})` at line ~51, and `PARENT_SCOPE` only exposes the variable to the top-level scope after that call returns.

The `BINARY_NAME` suffix additionally breaks `BuildableLinuxApp.executable()` → `LinuxDevice.executablePathForDevice()`, which resolve the binary path through the flavor-unaware `getCmakeExecutableName()`.

`APPLICATION_ID` **is** safe to change — the tool never reads it — as long as it's set after `add_subdirectory(${FLUTTER_MANAGED_DIR})`. This placement is doubly required: `APPLICATION_ID` is consumed by `add_definitions(-DAPPLICATION_ID=...)` in `linux/runner/CMakeLists.txt`, pulled in by `add_subdirectory("runner")` further down, so the value must be final before that point.

The suggestion below is the shape `flutter_flavorizr` generates and that its example project builds and runs with today.

````suggestion
Open `linux/CMakeLists.txt` and locate the
`add_subdirectory(${FLUTTER_MANAGED_DIR})` line.
Add the flavor-aware `APPLICATION_ID` immediately after it —
`FLUTTER_APP_FLAVOR` only becomes visible at this point in the file,
and `APPLICATION_ID` must be final before `add_subdirectory("runner")` runs:

```cmake title="linux/CMakeLists.txt"
# Flutter library and tool build rules.
set(FLUTTER_MANAGED_DIR "${CMAKE_CURRENT_SOURCE_DIR}/flutter")
add_subdirectory(${FLUTTER_MANAGED_DIR})

if(DEFINED FLUTTER_APP_FLAVOR)
  if(FLUTTER_APP_FLAVOR STREQUAL "staging")
    set(APPLICATION_ID "com.example.staging")
  elseif(FLUTTER_APP_FLAVOR STREQUAL "production")
    set(APPLICATION_ID "com.example.production")
  endif()
endif()
```

Leave `set(BINARY_NAME ...)` unchanged. The Flutter tool resolves the
executable path by parsing that line literally, so renaming the binary
per flavor prevents `flutter run` from launching the app.
````

File: sites/docs/src/content/deployment/flavors-windows-linux.md (hunk 5 above)
Comment (AngeloAvv) [inline sites/docs/src/content/deployment/flavors-windows-linux.md:294]: **Blocking — the preprocessor is the wrong mechanism on Windows.**

Two concrete problems with the pair of snippets:

1. `target_compile_definitions(${BINARY_NAME} …)` requires the target to exist, so this must land **after** `add_executable(${BINARY_NAME} WIN32 …)`. As written, pasting it at the top of the file gives `Cannot specify compile definitions for target "…" which is not built by this project`.

2. `std::wstring(L"" FLUTTER_APP_FLAVOR)` concatenates a wide literal with a narrow one. That is implementation-defined in C++11–C++20 and ill-formed in C++23 ([P2201R1](https://wg21.link/p2201r1)). MSVC accepts it today, but it's fragile to enshrine in official docs.

More importantly, the define approach doesn't scale to the very next section: `Runner.rc` is a resource script and cannot read preprocessor defines set this way, so the icon section below has no way to reuse this mechanism — and ends up unimplementable (see my comment there).

`flutter_flavorizr` solves title, icon, and PE metadata with a single `configure_file` pass. This is what its example project builds and runs with. In `windows/runner/CMakeLists.txt`, **before** `add_executable`:

```cmake title="windows/runner/CMakeLists.txt"
if(DEFINED FLUTTER_APP_FLAVOR)
  if(FLUTTER_APP_FLAVOR STREQUAL "staging")
    set(RUNNER_APP_ICON "staging.ico")
    set(WINDOW_TITLE "Staging App")
  elseif(FLUTTER_APP_FLAVOR STREQUAL "production")
    set(RUNNER_APP_ICON "production.ico")
    set(WINDOW_TITLE "Production App")
  endif()
else()
  set(RUNNER_APP_ICON "app_icon.ico")
  set(WINDOW_TITLE "${BINARY_NAME}")
endif()

configure_file(
  "${CMAKE_CURRENT_SOURCE_DIR}/main.cpp.in"
  "${CMAKE_CURRENT_SOURCE_DIR}/main.cpp"
  @ONLY
)
```

Then rename `main.cpp` to `main.cpp.in` and parameterize the title:

```cpp title="windows/runner/main.cpp.in"
  if (!window.Create(L"@WINDOW_TITLE@", origin, size)) {
    return EXIT_FAILURE;
  }
```

Substitution happens at configure time, so there is no narrow/wide conversion and no preprocessor branch at all.

Note this generates `main.cpp` into the source tree, so `.gitignore` should cover it while the `.in` file is committed.

Since the Linux tab keeps using the preprocessor (correctly — see my comment there), it's worth saying explicitly in the page that the two platforms use different mechanisms, and why.

sites/docs/src/content/deployment/flavors-windows-linux.md
```diff
@@ -0,0 +1,429 @@
...
+if (use_header_bar) {
+  GtkHeaderBar* header_bar = GTK_HEADER_BAR(gtk_header_bar_new());
+  gtk_widget_show(GTK_WIDGET(header_bar));
+  gtk_header_bar_set_title(header_bar, title);
+  gtk_header_bar_set_show_close_button(header_bar, TRUE);
+  gtk_window_set_titlebar(window, GTK_WIDGET(header_bar));
+} else {
+  gtk_window_set_title(window, title);
+}
+```
```
AngeloAvv, line 327: The Linux mechanism here is **correct** — unlike the Windows tab. GTK takes narrow `const char*` strings, so the preprocessor define works directly with no conversion. `flutter_flavorizr` does exactly this, and its example project builds and runs with it.

Two adjustments to match what actually works:

**1. Placement.** `target_compile_definitions(${BINARY_NAME} …)` needs the target to exist, so it must come after `add_executable(${BINARY_NAME} …)`. `flutter_flavorizr` anchors on the existing `APPLICATION_ID` define, which is already correctly positioned:

```cmake title="linux/runner/CMakeLists.txt"
# Add preprocessor definitions for the application ID.
add_definitions(-DAPPLICATION_ID="${APPLICATION_ID}")

if(DEFINED FLUTTER_APP_FLAVOR AND NOT FLUTTER_APP_FLAVOR STREQUAL "")
  add_definitions(-DFLUTTER_APP_FLAVOR="${FLUTTER_APP_FLAVOR}")
endif()
```

Placing the new block right after the `APPLICATION_ID` line is an easy instruction to follow and is guaranteed to be in scope.

**2. `my_application.cc` snippet doesn't match the file it's editing.** The generated `my_application_activate` passes the title as a string literal in two places. The snippet introduces a `title` variable but shows the `if (use_header_bar)` block dedented to column 0, whereas in the real file it's indented two spaces inside the function — so this can't be pasted as-is.

The working shape, which keeps the diff to the existing code minimal:

```c title="linux/runner/my_application.cc"
  const gchar* window_title = "flavors_example";
#ifdef FLUTTER_APP_FLAVOR
  if (g_strcmp0(FLUTTER_APP_FLAVOR, "staging") == 0) {
    window_title = "Staging App";
  } else if (g_strcmp0(FLUTTER_APP_FLAVOR, "production") == 0) {
    window_title = "Production App";
  }
#endif
```

then replace the two hardcoded literals with `window_title`:

```c
    gtk_header_bar_set_title(header_bar, window_title);
    ...
    gtk_window_set_title(window, window_title);
```

Using `g_strcmp0` rather than string concatenation also lets each flavor have a fully custom title instead of `AppName (flavor)`, which is closer to what people actually want from flavors.

sites/docs/src/content/deployment/flavors-windows-linux.md
```diff
@@ -0,0 +1,429 @@
...
+  set(APP_ICON_NAME "app_icon_staging.ico")
+elseif(FLUTTER_APP_FLAVOR STREQUAL "production")
+  set(APP_ICON_NAME "app_icon_production.ico")
+else()
+  set(APP_ICON_NAME "app_icon.ico")
+endif()
+```
+
+1. Configure `windows/runner/Runner.rc` or your CMake target
+   to use `APP_ICON_NAME` for the application icon resource.
```
AngeloAvv, line 357: **Blocking — this section isn't implementable as written.**

Two problems:

1. The snippet targets `windows/CMakeLists.txt`, where `FLUTTER_APP_FLAVOR` is still undefined (same scope issue as the earlier section). `if(FLUTTER_APP_FLAVOR STREQUAL "staging")` on an undefined variable is always false, so `APP_ICON_NAME` silently falls through to `else()`.

2. Step 3 — *"Configure `windows/runner/Runner.rc` or your CMake target to use `APP_ICON_NAME`"* — is the entire difficulty, and it's left as an exercise. `Runner.rc` is a resource script; it cannot read CMake variables. There is no way to complete this step from what the page provides.

`flutter_flavorizr` handles this with the same `configure_file` pass proposed in my comment on the window-title section — one mechanism covering icon, title, and PE metadata. Extend that block in `windows/runner/CMakeLists.txt`, before `add_executable`:

```cmake title="windows/runner/CMakeLists.txt"
configure_file(
  "${CMAKE_CURRENT_SOURCE_DIR}/Runner.rc.in"
  "${CMAKE_CURRENT_SOURCE_DIR}/Runner.rc"
  @ONLY
)
```

Rename `Runner.rc` to `Runner.rc.in` and parameterize it — these are the exact substitutions its example project ships:

```rc title="windows/runner/Runner.rc.in"
IDI_APP_ICON            ICON                    "resources\\@RUNNER_APP_ICON@"

VALUE "FileDescription", "@WINDOW_TITLE@" "\0"
VALUE "ProductName",     "@WINDOW_TITLE@" "\0"
```

with the per-flavor `.ico` files sitting in `windows/runner/resources/` alongside the default `app_icon.ico`.

Same caveat as before: `Runner.rc` becomes a generated file, so it should be gitignored while `Runner.rc.in` is committed.

sites/docs/src/content/deployment/flavors-windows-linux.md
AngeloAvv, line 371: This section is prose-only, and step 3 — *"Update `linux/CMakeLists.txt` to install the correct `.desktop` file and icon"* — is where all the actual work lives. A reader can't act on it.

It's also genuinely harder than the Windows case: on Linux the icon isn't embedded in the binary, so it depends on `.desktop` entries, the icon theme layout, and the packaging format (snap/deb/flatpak).

I'd suggest cutting these three steps and linking to [Build and release a Linux app](/deployment/linux) instead, noting that per-flavor icons are a packaging concern rather than a flavor concern. Shipping three non-actionable bullets sets readers up to fail — and I'd rather not propose a concrete recipe here, since I don't have a verified working example to base one on.

sites/docs/src/content/deployment/flavors-windows-linux.md
AngeloAvv, line 161: Two additions worth making somewhere in this section:

**1. `--flavor` also works with `flutter test` and `flutter drive`** on desktop, not just `run`/`build`. The Linux PR touched `commands/test.dart` for exactly this, and both devicelab tasks (`flavors_test_linux.dart`, `flavors_test_windows.dart`) exercise `createFlavorsTest()` / `createIntegrationTestFlavorsTest()`. Readers wiring flavors into CI will look for this.

**2. Existing projects shouldn't need to regenerate anything.** Neither desktop flavor PR modified the `windows.tmpl` / `linux.tmpl` templates, so `--flavor` works on any existing project as-is — no `flutter create .` re-run, no template migration. That's a genuinely reassuring detail and cheap to state.

sites/docs/src/content/deployment/flavors-windows-linux.md
```diff
@@ -0,0 +1,429 @@
...
+```
+</Tab>
+<Tab name="Linux">
+
+```console title="console"
+$ flutter run -d linux --flavor <flavor_name>
+```
+
+Replace `<flavor_name>` with the name of your flavor
+(for example, `staging` or `production`).
```
AngeloAvv, line 184: This sentence sits inside the **Linux** tab, so readers on the Windows tab never see it — even though the Windows snippet above uses the same `<flavor_name>` placeholder.

Move it out of the `<Tabs>` block so it applies to both.

sites/docs/src/content/deployment/flavors-windows-linux.md
sfshaza2: Note:
#1 (the fact about --flavor working with `flutter test` and `flutter drive` is mentioned in both pages (flavors-linux and flavors-windows) under #test-a-flavor.
#2 (existing projects...) is mentioned in both pages under #configure-cmake.

sites/docs/src/content/deployment/flavors-windows-linux.md
sfshaza2: There are now two pages (no Tabs blocks) and each page now mentioned this under #run-a-flavor.

sites/docs/src/content/deployment/flavors-windows-linux.md
sfshaza2: I believe this is fixed.

sites/docs/src/content/deployment/flavors-windows-linux.md
sfshaza2: I _believe_ this is fixed.

sites/docs/src/content/deployment/flavors-windows-linux.md
sfshaza2: I _believe_ this is fixed.

sites/docs/src/content/deployment/flavors-windows-linux.md
sfshaza2: I _believe_ this is fixed.

sfshaza2:
@AngeloAVV, I want to thank you for your ENORMOUSLY detailed and helpful review!!! (In fact, if you could email me at shaza@google, I'd like to thank you tangibly....) I can tell that you are an educator!

I believe I have incorporated all of your feedback. Could you do a quick check and confirm?

Staged here:
- https://flutter-docs-prod--docs-pr13736-flavors-windows-linux-5airv7fi.web.app/deployment/flavors-linux
- https://flutter-docs-prod--docs-pr13736-flavors-windows-linux-5airv7fi.web.app/deployment/flavors-windows

sites/docs/src/content/deployment/flavors-windows.md
```diff
@@ -0,0 +1,336 @@
...
+:::
+
+## Overview
+
+A Flutter flavor represents a collection of settings that define
+how a specific version of your app builds and runs.
+For example, a flavor can determine the window title,
+application icon, API endpoint, asset set, and logging configuration
+for a build.
+
```
parlough, line 27: Consider adding a version note since this feature requires the latest stable release still.

```suggestion

:::version-note
Built-in support for flavors on Windows requires Flutter 3.47 or later.
:::

```

sites/docs/src/content/deployment/flavors-linux.md
```diff
@@ -0,0 +1,322 @@
...
+:::
+
+## Overview
+
+A Flutter flavor represents a collection of settings that define
+how a specific version of your app builds and runs.
+For example, a flavor can determine the window title,
+application ID, API endpoint, asset set, and logging configuration
+for a build.
+
```
parlough, line 27: Consider adding a version note since this feature requires the latest stable release still.

```suggestion

:::version-note
Built-in support for flavors on Linux requires Flutter 3.47 or later.
:::

```

parlough (approved):
I pushed some minor clean up and have two small suggestions to consider, but otherwise looks good to me. Thanks for splitting these up and addressing Angelo's (great) feedback!

PR #13741, opened 2026-08-18
Add mermaid support

ericwindmill:

Adds [Mermaid](https://mermaid.ai/live/edit#pako:eNpVjctugzAQRX_FmlUrkQjIi3hRqYE2m0jtIqtCFqMwYFSwkTFKU-Dfa4iitl756px7p4OzSgk4ZKW6nAVqw45RIpl9z3EodNGYCpsTm82e-j0ZVilJ157tHvaKNULVdSHzx5u_GyUWdodRI2ZEIT-HGwqn_puknkXxAWuj6tNfcryonr3Exbuw8_-J0GRbr3GGPMPZGTULUU8KOJDrIgVudEsOVKQrHCN0I03ACKooAW6_KWXYliaBRA62VqP8UKq6N7VqcwF2vmxsausUDUUF5hp_FZIp6VC10gBfTQvAO_gC7gXbeeAuXM_bbhdBsFn6Dlyts5h7Nmz8tev6_nK5Hhz4nm6682CzGn4A04l0EA) diagram support. "Yeehaw" for code and version control instead of saving diagrams as images.

Screenshot 2026-08-18 at 2 43 33 PM

## Presubmit checklist

  of 80 characters or fewer.


sites/docs/src/content/ai/evals.md
```diff
@@ -28,3 +28,20 @@ Evals measure both deterministic code correctness
...
+    subgraph two
+    b1-->b2
+    end
+    subgraph three
+    c1-->c2
+    end
+    one --> two
+    three --> two
+    two --> c2
+```
```
ericwindmill, line 69: I added this for demo purposes. It needs to be removed before this PR lands.

ericwindmill:
I'm pretty far out of my element here using  js interop and Jaspr, two things I'm only vaguely familiar with. I'd appreciate a thorough review @parlough !

packages/site_shared/lib/_sass/components/_mermaid.scss
```diff
@@ -0,0 +1,55 @@
...
+
+    // Sharper node outlines matching site borders
+    .node rect,
+    .node circle,
+    .node ellipse,
+    .node polygon,
+    .node path {
+      stroke-width: 1.5px;
+      fill:  var(--site-secondaryContainer-bgColor) !important;
+    }
```
ericwindmill, line 48: I'll update these base styles to match our themes once I know whether this entire approach is reasonable.

schultek:
As an alternative approach instead of using mermaid.js, we could use https://github.com/orestesgaolin/mermaid/tree/main/packages/mermaid_core to render the diagrams to SVGs at serve/build time.

This is the demo https://roszkowski.dev/mermaid/

Seems like this is not published yet, but if we were to use it maybe we can convince @orestesgaolin to publish it :)

ericwindmill:
> As an alternative approach instead of using mermaid.js, we could use https://github.com/orestesgaolin/mermaid/tree/main/packages/mermaid_core to render the diagrams to SVGs at serve/build time.
> 
> This is the demo https://roszkowski.dev/mermaid/
> 
> Seems like this is not published yet, but if we were to use it maybe we can convince @orestesgaolin to publish it :)

I would much rather use this, I'll try it out

orestesgaolin:
Would you like me to publish mermaid to pub.dev? For now I only published katex https://pub.dev/packages/katex

With mermaid I found that it sometimes does not well represent elk layouts and there are slight differences in rendering
screenshot_20260819170012@2x

ericwindmill:
> Would you like me to publish mermaid to pub.dev? For now I only published katex https://pub.dev/packages/katex
> 
> With mermaid I found that it sometimes does not well represent elk layouts and there are slight differences in rendering 

I'm not concerned with those small rendering differences. I have noticed one bug, the arrow label renders in the wrong spot when the flowchart is `LR` oriented. It works correctly when the orientation is TD

Screenshot 2026-08-19 at 8 13 44 AM

If you plan on maintaining this library, we'd likely want to use it over the JSInterop solution (and we can find a way to contribute and help out). But if you don't plan on maintaining, thats okay, no pressure.

packages/site_shared/pubspec.yaml
```diff
@@ -16,6 +16,11 @@ dependencies:
   jaspr_content: ^0.5.3
   markdown: ^7.3.1
   markdown_description_list: ^0.2.0
+  mermaid_core:
+    git:
+      url: https://github.com/orestesgaolin/mermaid.git
+      ref: round3-fixes-and-packaging
```
parlough, line 22: Any particular reason you're relying on this branch over the `main` branch? Seems like `main` contains additional updates.

If we're going to use it as a git dependency for now, pin it to a specific commit (like ddb951169592bf089518ba19dbf3bf7ffa8d3c35 from the current `main` branch).

sites/docs/src/content/contribute/docs/markdown.md
```diff
@@ -47,3 +47,62 @@ To learn more about customizing code blocks,
...
+
+```markdown
+```mermaid
+flowchart LR
+    A[Start] --> B(Process)
+    B --> C{Decision}
+    C -->|Yes| D[Done]
+    C -->|No| B
+```
+```
```
parlough, line 65:
`````suggestion
````markdown
```mermaid
flowchart LR
    A[Start] --> B(Process)
    B --> C{Decision}
    C -->|Yes| D[Done]
    C -->|No| B
```
````
`````

File: sites/www/lib/styles/styles.scss
Comment (parlough) [inline sites/www/lib/styles/styles.scss:1]: If you want support to be fully enabled on the marketing (www) site where these styles are added, the processor needs to be to be registered as an extension in `sites/www/lib/main.server.dart`.

File: sites/docs/src/content/contribute/docs/markdown.md [hunk 5]
```diff
@@ -47,3 +47,62 @@ To learn more about customizing code blocks,
...
+
+### Styling Mermaid diagrams
```
parlough, line 67:
```suggestion
### Style Mermaid diagrams
```

packages/site_shared/lib/src/extensions/mermaid_processor.dart
parlough, line 1: Can you add license headers to these Dart files.

packages/site_shared/lib/components/common/client/mermaid_diagram.dart
parlough, line 1: Rather than using this as a client dependency that listens for theme changes, can this instead be a server component that renders both versions up front and switches between them with css? That way we can drop the complicated theming logic, avoid shipping the renderer to the client, and improve page load time.

I imagine the component implementation could look something like this:

```dart
/// Renders a Mermaid diagram as server-generated SVG.
///
/// Both light and dark variants are included in the page
/// so that CSS can show the variant matching the site's current theme.
final class MermaidDiagram extends StatelessComponent {
  const MermaidDiagram({
    required this.diagram,
    super.key,
  });

  /// The Mermaid diagram definition to render.
  final String diagram;

  @override
  Component build(BuildContext context) {
    // Render both theme variants on the server.
    final lightSvg = _renderDiagram(theme: .defaultTheme);
    final darkSvg = _renderDiagram(theme: .darkTheme);

    return div(
      classes: 'mermaid-container',
      [
        if (lightSvg != null && darkSvg != null) ...[
          div(classes: 'mermaid-theme mermaid-theme-light', [
            RawText(lightSvg),
          ]),
          div(classes: 'mermaid-theme mermaid-theme-dark', [RawText(darkSvg)]),
        ] else
          // If rendering fails,
          // preserve the source in a fallback `<pre>` element.
          pre(
            classes: 'mermaid',
            attributes: {'data-source': diagram},
            [.text(diagram)],
          ),
      ],
    );
  }

  /// Renders [diagram] as SVG using the specified [theme].
  String? _renderDiagram({required MermaidTheme theme}) {
    try {
      // Server rendering doesn't have browser text metrics available, so use
      // the library's deterministic approximation when laying out labels.
      final mermaid = Mermaid(
        measurer: const ApproximateTextMeasurer(),
        theme: theme,
      );
      final scene = mermaid.render(diagram);
      return renderSceneToSvg(scene);
    } catch (error) {
      if (kDebugMode) {
        print('Failed to render Mermaid diagram: $error');
      }
      return null;
    }
  }
}
```


And the styles could then hide the other depending on the theme. Something like this:

```scss
.mermaid-container {
  // ...

  .mermaid-theme-dark {
    display: none;
  }

  @at-root body.dark-mode & {
    .mermaid-theme-light {
      display: none;
    }

    .mermaid-theme-dark {
      display: block;
    }
  }

  // ...
}
```

parlough (commented):
Thanks for exploring this @ericwindmill! It's exciting. Looks pretty good already, just some questions and comments, primarily about if we can move the rendering to the build?

packages/site_shared/lib/components/common/client/mermaid_diagram.dart
ericwindmill: Good call. I updated it and the experience is certainly better now

packages/site_shared/lib/_sass/components/_mermaid.scss
```diff
@@ -0,0 +1,71 @@
...
+
+    // Edge lines & arrows
+    .edgePath .path {
+      stroke-width: 1.5px;
```
parlough: These styles must have been specific to mermaid.js. These classes aren't applied here. Either adjust them to work or remove them.

packages/site_shared/pubspec.yaml
```diff
@@ -16,6 +16,11 @@ dependencies:
   jaspr_content: ^0.5.3
   markdown: ^7.3.1
   markdown_description_list: ^0.2.0
+  mermaid_core:
+    git:
+      url: https://github.com/orestesgaolin/mermaid.git
+      ref: 03759ce31539f87a1487e84d132d3ad1e8efa3cd
```
parlough: Consider trying to update to [90c39dd66b5e798c12dd77ec8b3fba4d583f0a05](https://github.com/orestesgaolin/mermaid/commit/90c39dd66b5e798c12dd77ec8b3fba4d583f0a05). It seems it might include a fix to an issue you mentioned.

sites/docs/src/content/contribute/docs/markdown.md
```diff
@@ -47,3 +47,64 @@ To learn more about customizing code blocks,
...
+  }
+
+  svg {
+    // Custom SVG element overrides (requires !important)
+    .highlightNode rect {
+      fill: var(--site-primary-color) !important;
+    }
+  }
+}
+```
```
parlough, line 110: Consider just dropping this snippet (but leaving the reference to `mermaid.scss`), since it seems the `.highlightNode rect {` example doesn't apply to the package we're using.

sites/docs/src/content/ai/evals.md
parlough, line 1: Remember to remove these tests before landing.

parlough (approved):
Thanks for making those updates @ericwindmill!

Seems we need some upstream fixes, but I'm fine with landing this and iterating upstream and here.

Seemingly upstream issues:

1. SVG part IDs for aren't unique or namespaced for each diagram, causing overlap.
2. Seems there's no consistent classes for configuring global styling.
3. In the light mode version, there's a white background generated that visually conflicts with the background of the container.

packages/site_shared/pubspec.yaml
ericwindmill: That commit doesn't seem to be ready, it has broken dependencies internally, I'm going to file an issue

orestesgaolin: it's now published on pub.dev with some extra fixes

ericwindmill: Awesome! thank you @orestesgaolin

parlough (approved):
Thanks for updating this to use the published package. Looks good to me!

Feel free to land when you're ready and we can iterate from there as we need to.

orestesgaolin:
> Thanks for updating this to use the published package. Looks good to me!

keep the issues coming, we'll make sure to bring full parity to mermaid.js

PR #13766, opened 2026-08-20
Add standardized SCSS breakpoints

ericwindmill:

Adds standardized scss breakpoints and mixins to use them.

### DIFF
--- a/packages/site_shared/lib/_sass/base/_breakpoints.scss
+++ b/packages/site_shared/lib/_sass/base/_breakpoints.scss
@@ -1,0 +1,31 @@
+@use 'sass:map';
+
+$breakpoints: (
+  'sm': 640px,
+  'md': 768px,
+  'lg': 1024px,
+  'xl': 1280px,
+  '2xl': 1536px,
+  'max': 1440px,
+  '3xl': 1920px,
+);
+
+@mixin screen($name) {
+  $value: map.get($breakpoints, $name);
+  @if $value == null {
+    @error 'Unknown breakpoint: #{$name}';
+  }
+  @media (min-width: $value) {
+    @content;
+  }
+}
+
+@mixin screen-below($name) {
+  $value: map.get($breakpoints, $name);
+  @if $value == null {
+    @error 'Unknown breakpoint: #{$name}';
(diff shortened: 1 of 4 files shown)


packages/site_shared/lib/_sass/base/_breakpoints.scss
```diff
@@ -0,0 +1,31 @@
...
+  'sm': 640px,
+  'md': 768px,
+  'lg': 1024px,
+  'xl': 1280px,
+  'max': 1440px,
+  '2xl': 1536px,
+  '3xl': 1920px,
+);
+
+@mixin screen($name) {
```
parlough: Consider adding a doc comment. Something like:

```suggestion
/// Emits a minimum-width media query for the specified named breakpoint.
///
/// @param {string} $name - A key in `$breakpoints`.
/// @content Styles to emit at or above the breakpoint width.
@mixin screen($name) {
```

packages/site_shared/lib/_sass/base/_breakpoints.scss
```diff
@@ -0,0 +1,31 @@
...
+  $value: map.get($breakpoints, $name);
+  @if $value == null {
+    @error 'Unknown breakpoint: #{$name}';
+  }
+  @media (min-width: $value) {
+    @content;
+  }
+}
+
+@mixin screen-below($name) {
```
parlough: Same here:


```suggestion
/// Emits a max-width media query that ends just below
/// the specified named breakpoint.
///
/// @param {string} $name - A key in `$breakpoints`.
/// @content Styles to emit below the breakpoint width.
@mixin screen-below($name) {
```

packages/site_shared/lib/_sass/base/_breakpoints.scss
parlough: Consider renaming the parameters in both functions to `$breakpoint-name` or similar to be clearer it's one of the predefined names.

parlough (approved):
Nice, thanks for extracting this out @ericwindmill!

Some small suggestions to consider, but otherwise looks good to me.

PR #13768, opened 2026-08-20
prefactor: generalize filterable index layout and components

ericwindmill:

Extract generic two-column filterable index layout styles and sidebar/search components from **learning resources index** into reusable filterable_index.dart and _filterable-index.scss. 

This change anticipates new content for FlutterBench, where I'm adding a CUJ index #13691


sites/docs/lib/src/components/pages/filterable_index.dart
```diff
@@ -0,0 +1,153 @@
...
+  @override
+  Component build(BuildContext context) {
+    return GlobalEventListener(
+      onClick: (event) {
+        final target = event.target as web.Element?;
+        if (target?.closest(_sidebarSelector) == null &&
+            target?.closest('.show-filters-button') == null) {
+          _toggle?.checked = false;
+        }
+      },
```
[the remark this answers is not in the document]
ericwindmill: Is this not exactly the same?

sites/docs/lib/src/components/pages/filterable_index.dart
```diff
@@ -0,0 +1,153 @@
...
+    ]);
+  }
+}
+
+/// The search field of a filterable index page,
+/// followed by the page-specific [children], such as a result count.
+///
+/// On narrow screens, the field also contains the button
+/// that opens the [FiltersSidebar] drawer.
+class FilterSearchGroup extends StatelessComponent {
```
parlough, line 84: Should the `_LearningResourceFiltersState` build use this instead of building its own `resource-search-group`?

sites/docs/lib/src/components/pages/filterable_index.dart
```diff
@@ -0,0 +1,153 @@
...
+      onClick: (event) {
+        final target = event.target as web.Element?;
+        if (target?.closest(_sidebarSelector) == null &&
+            target?.closest('.show-filters-button') == null) {
+          _toggle?.checked = false;
+        }
+      },
+      button(
+        classes: 'icon-button show-filters-button',
+        onClick: () {
```
parlough, line 154: Make sure to give this button an accessible label. Perhaps something like:


```suggestion
        classes: 'icon-button show-filters-button',
        attributes: const {'aria-label': 'Show filters'},
        onClick: () {
```

sites/docs/lib/src/components/pages/filterable_index.dart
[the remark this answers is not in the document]
parlough: The underlying of the comment is sort of right, but its solution isn't accurate since `is` also doesn't work for web types. Instead use the `isA` method. Something like this:

```dart
onClick: (event) {
  final target = event.target;
  if (target == null || !target.isA<web.Element>()) return;

  final element = target as web.Element;
  if (element.closest(_sidebarSelector) == null &&
      element.closest('.show-filters-button') == null) {
    _toggle?.checked = false;
  }
},
```

sites/docs/lib/src/components/pages/filterable_index.dart
```diff
@@ -0,0 +1,153 @@
...
+          if (_toggle case final toggle?) {
+            toggle.checked = !toggle.checked;
+          }
+        },
+        [
+          const MaterialIcon('filter_list'),
+        ],
```
parlough:
```suggestion
        const [MaterialIcon('filter_list')],
```

sites/docs/lib/src/components/pages/filterable_index.dart
```diff
@@ -0,0 +1,153 @@
...
+    super.key,
+  });
+
+  /// The id of the search field, so the page can label it.
+  final String searchId;
+
+  final String placeholder;
+  final String label;
+  final String value;
+  final void Function(String) onInput;
```
parlough, line 108: Can you add basic API doc comments to these members?

sites/docs/lib/_sass/components/_filterable-index.scss
```diff
@@ -0,0 +1,221 @@
+// The markup comes from the `FiltersSidebar` and `FilterSearchGroup`
+// components in `filterable_index.dart`, so everything here is keyed off
+// classes. Each page styles its own results list separately, keyed off the
+// id of that list.
+
+$mobile-breakpoint: 839px;
+$sidebar-width: 220px;
+
+.filterable-index {
```
parlough, line 6: Consider nesting all of the other top-level style entries in this file under this entry. That way unexpected conflicts don't show up in the future, for example for `.close-icon` which could definitely see use elsewhere.

parlough (commented):
Thanks for extracting this out @ericwindmill! Mostly looks good to me.

Should the learning resources index be updated to use the new `FilterSearchGroup`?

sites/docs/lib/_sass/components/_filterable-index.scss
```diff
@@ -0,0 +1,223 @@
...
+    }
+  }
+
+  // On narrow screens the sidebar becomes a drawer that slides in from the
+  // right. Opening and closing is driven entirely by the hidden
+  // `#open-filter-toggle` checkbox that precedes it.
+  @media (max-width: $mobile-breakpoint) {
+    .filterable-index {
+      flex-direction: column;
+    }
```
parlough: These styles are already nested under `.filterable-index` now.

```suggestion
    flex-direction: column;
```

sites/docs/lib/_sass/components/_filterable-index.scss
```diff
@@ -0,0 +1,223 @@
...
+      border: none;
+      border-radius: 0;
+    }
+
+    .filter-group .table-title {
+      background-color: var(--site-raised-bgColor-translucent);
+    }
+  }
+
+}
```
parlough: Add an extra new line at the end of the file:

```suggestion
}

```

sites/docs/lib/src/components/pages/filterable_index.dart
```diff
@@ -0,0 +1,163 @@
...
+
+  static web.HTMLInputElement? get _toggle =>
+      web.document.getElementById(_drawerToggleId) as web.HTMLInputElement?;
+
+  @override
+  Component build(BuildContext context) {
+    return GlobalEventListener(
+      onClick: (event) {
+        final target = event.target;
+        if (target == null || target.isA<web.Element>()) return;
```
parlough: I believe this intends to exit if the target isn't an `Element`:

```suggestion
        if (target == null || !target.isA<web.Element>()) return;
```

sites/docs/lib/src/components/pages/filterable_index.dart
```diff
@@ -0,0 +1,163 @@
...
+import 'package:universal_web/js_interop.dart';
+import 'package:universal_web/web.dart' as web;
+
+/// The id of the hidden checkbox that opens and closes the filter sidebar
+/// while it is a drawer on narrow screens.
+///
+/// The drawer is CSS-only: `_filterable-index.scss` keys off the checkbox's
+/// `:checked` state, and the close button within the sidebar is a `<label>`
+/// pointing back at it.
+const _drawerToggleId = 'open-filter-toggle';
```
parlough: So the component is properly reusable, perhaps this ID should be passed as a parameter to `FiltersSidebar` since each one would need a unique ID.

sites/docs/lib/src/components/pages/learning_resource_filters_sidebar.dart
parlough: I know this is pre-existing, but I noticed these aren't re-assigned. Consider making them `final`:


```suggestion
final Set<LearningResourceTag> selectedTags = {};
final Set<LearningResourceType> selectedTypes = {};
```

parlough (approved):
Thanks for those updates @ericwindmill! Looks good to me after a few final adjustments:

PR #13802, opened 2026-08-25
Clean up blog author images

parlough:
- Rename each blog author image to match their ID in the `authors.yaml` file.
- Standardize JPEG images to `.jpg` instead of a mix of `.jpg` and `.jpeg`.
- Convert the one-remaining PNG author image to WebP.
- Remove some unused and duplicate author images.
- Rename the image directory to `author_images` instead of `authors_images`.


sfshaza2 (approved):
Awesome!

PR #13824, opened 2026-08-27
Fix broken link in Flutter release blog

conooi:
Fix broken link in Flutter release blog


sfshaza2 (approved):
lgtm, but here's a question. Is the full URL required? Can a relative link work? @parlough

lamek (approved):
lgtm

parlough:
> lgtm, but here's a question. Is the full URL required? Can a relative link work? @parlough

@sfshaza2 Since they're on different sites, links to the docs from the blog need to be full links. I clarified this in the skill as well, so it should get it right as well.

PR #13845, opened 2026-09-02
Enable recent lints for site tooling and implementation

parlough:
- Enables some new and some pre-existing lints for site tooling and implementation.
- Fixes diagnostics reported after enabling those lints.
- Temporarily enabled some others to address certain cases of them.


sfshaza2 (approved):
lgtm!

PR #13846, opened 2026-09-02
Update deploy image to use Dart 3.13.3

parlough:
These versions include some dart2wasm compilation fixes, which we take advantage of on dart.dev currently.


abdallahshaban557 (approved):
LGTM

PR #13847, opened 2026-09-02
Misc example clean up and updates

parlough:
- Enable and fix some new lints.
- Tighten some dependency constraints.
- Use newer, preferred versions of some functions.
- Improve formatting and clarity of some inline and doc comments.


abdallahshaban557 (approved):
LGTM

PR #13852, opened 2026-09-02
Move away from build based style hash

parlough:
When building the full, static site, check the built `main.css` for its hash during the normal Jaspr site build rather than relying on a custom builder and a committed Dart file.


abdallahshaban557 (approved):
LGTM - THANK YOU!!!!

PR #13854, opened 2026-09-02
Consistently use jaspr_cli managed build options

parlough:
The `--no-managed-build-options` we were passing to the `jaspr build` command was meaning `--dart-define` wasn't respected so `productionBuild` was false for client-compiled code. We don't use the `build.yaml` files for much special configuration anyway, so we can instead use Jaspr's managed build options, passing what little we do need to configure. With that change, `productionBuild` is correctly set to `true` on release client build. This fixes code such as analytics sending that was incorrectly being stripped.

Supersedes and closes https://github.com/flutter/website/pull/13851


abdallahshaban557 (approved):
LGTM

jesskuras (approved):
Awesome, thanks for looking into this!!

PR #13870, opened 2026-09-09
Add table of contents and back-to-top button to blog posts

manuelzzz:
Adds a table of contents and a back-to-top button to long-form blog posts, to make navigating them easier.

- **Table of contents**: adds `BlogTableOfContents` (`packages/site_shared/lib/components/blog/blog_toc.dart`), which renders an automatically generated, collapsible "On this page" nav from the post's rendered headings. It's built on `jaspr_content`'s existing `TableOfContentsExtension` rather than a custom implementation, so it correctly builds anchor links relative to the current page path. Posts with fewer than 2 top-level headings don't show a TOC.
- **Back-to-top button**: adds `BackToTopButton` (`packages/site_shared/lib/components/common/client/back_to_top_button.dart`), a floating button that appears once the reader scrolls past the hero and smooth-scrolls back to the top on click. It reuses the site's existing `in-content` scroll-spy signal, so no new scroll-tracking JS was needed.
- **Fix**: heading anchors on `flutter.dev` (for example, TOC links, or `#some-heading` URLs) were landing behind the fixed header instead of scrolling clear of it. The existing `scroll-margin-top` rule in `_content.scss` referenced `--site-header-height`/`--site-subheader-height`, which are defined for `docs.flutter.dev` but were never defined for this site, silently making the rule a no-op. Defines them, aliased to the existing `--ui-header-height` variable.

Both features are implemented in the shared `site_shared` components (not `sites/www` or `sites/docs` directly) and only wired into the blog layout for now, so they're reusable for other long-form content later.

Closes #13737

PR #13881, opened 2026-09-11
Add auto-generated table of contents to blog posts

### LINKED ISSUE #13737: Adding TOC to blog posts
### Describe the problem

For long blog posts, like the release notes, we really need a toc of at least the h2 headers.

### Expected fix

Ideally, a toc can be added to the right side and will contain at least the h2 headers.

### On which platform(s) did you experience this issue?

_No response_

### Additional context

_No response_

manuelzzz:
Adds a table of contents to long-form blog posts, to make navigating them easier.

- **Table of contents**: adds `BlogTableOfContents` (`packages/site_shared/lib/components/blog/blog_toc.dart`), which renders an automatically generated, collapsible "On this page" nav from the post's rendered headings. It's built on `jaspr_content`'s existing `TableOfContentsExtension` rather than a custom implementation, so it correctly builds anchor links relative to the current page path. Posts with fewer than 2 top-level headings don't show a TOC.
- **Fix**: heading anchors on `flutter.dev` (for example, TOC links, or `#some-heading` URLs) were landing behind the fixed header instead of scrolling clear of it. The existing `scroll-margin-top` rule in `_content.scss` referenced `--site-header-height`/`--site-subheader-height`, which are defined for `docs.flutter.dev` but were never defined for this site, silently making the rule a no-op. Defines them, aliased to the existing `--ui-header-height` variable.

Fixes #13737

Requested to split from #13870

image


sites/www/lib/styles/pages/_blog_page.scss
```diff
@@ -33,6 +33,61 @@
...
+        }
+
+        .material-symbols {
+          transition: transform 0.15s var(--ui-anim-func);
+        }
+      }
+
+      details[open] summary .material-symbols {
+        transform: rotate(90deg);
+      }
```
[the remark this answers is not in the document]
manuelzzz: Thanks for the review! This selector is actually safe — `MaterialIcon` (`packages/site_shared/lib/components/common/material_icon.dart`) is the only icon component used here, and it unconditionally renders `class="material-symbols"`:

```dart
classes: [
  'material-symbols',
  if (filled) 'ms-filled',
  ...classes,
].toClasses,
```

There's no code path in this codebase that emits `material-symbols-outlined` or `material-icons` instead — those names only appear in the `--site-icon-fontFamily` CSS font-family fallback list, which is unrelated to the element's class attribute. Since this component can't produce a different class, broadening the selector would just be dead CSS guarding against something that can't happen. Keeping `.material-symbols` as-is here.

sites/www/lib/styles/pages/_blog_page.scss
```diff
@@ -33,6 +33,61 @@
...
+        cursor: pointer;
+        list-style: none;
+
+        &::-webkit-details-marker {
+          display: none;
+        }
+
+        .material-symbols {
+          transition: transform 0.15s var(--ui-anim-func);
+        }
```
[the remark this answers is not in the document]
manuelzzz: Thanks for the review! This selector is actually safe — `MaterialIcon` (`packages/site_shared/lib/components/common/material_icon.dart`) is the only icon component used here, and it unconditionally renders `class="material-symbols"`:

```dart
classes: [
  'material-symbols',
  if (filled) 'ms-filled',
  ...classes,
].toClasses,
```

There's no code path in this codebase that emits `material-symbols-outlined` or `material-icons` instead — those names only appear in the `--site-icon-fontFamily` CSS font-family fallback list, which is unrelated to the element's class attribute. Since this component can't produce a different class, broadening the selector would just be dead CSS guarding against something that can't happen. Keeping `.material-symbols` as-is here.

manuelzzz:
@parlough First things first, sorry for pinging you directly.

Following your recommendations from #13870, I split the PR into two: this one (#13870) and #13882.

The screenshot is available in the description, and the two PRs are independent.

packages/site_shared/lib/components/blog/blog_toc.dart
```diff
@@ -0,0 +1,56 @@
...
+/// Renders nothing if the page doesn't contain enough headings to warrant
+/// a table of contents.
+final class BlogTableOfContents extends StatelessComponent {
+  const BlogTableOfContents({super.key});
+
+  @override
+  Component build(BuildContext context) {
+    final toc = context.page.data['toc'];
+    if (toc is! TableOfContents || toc.entries.length < _minEntriesForToc) {
+      return const Component.empty();
```
parlough: You can use a dot shorthand here:

```suggestion
      return const .empty();
```

packages/site_shared/lib/components/blog/blog_toc.dart
```diff
@@ -0,0 +1,56 @@
...
+            const Component.element(
+              tag: 'summary',
+              children: [
+                MaterialIcon('chevron_right'),
+                Component.text('On this page'),
+              ],
+            ),
+            div(classes: 'toc-list', [toc.build()]),
+          ],
+        ),
```
parlough: I think you can use the built-in `details` and `summary` elements from `jaspr/dom.dart` here:


```suggestion
        details(
          [
            const summary(
              [
                MaterialIcon('chevron_right'),
                .text('On this page'),
              ],
            ),
            div(classes: 'toc-list', [toc.build()]),
          ],
        ),
```

sites/www/lib/styles/pages/_blog_page.scss
```diff
@@ -33,6 +33,61 @@
...
+      margin: 1.5rem 0;
+      background: var(--site-inset-bgColor);
+      border: 1px solid var(--site-inset-borderColor);
+      border-radius: var(--ui-border-radius-sm);
+
+      details {
+        padding: 1rem 1.25rem;
+      }
+
+      summary {
```
parlough, line 46: Can you update the summary text to use `var(--site-ui-fontFamily)` for its font family to match other UI elements on the blog?

sites/www/lib/styles/pages/_blog_page.scss
```diff
@@ -33,6 +33,61 @@
...
+      }
+
+      .toc-list {
+        margin: 0.75rem 0 0;
+
+        ul {
+          margin: 0;
+          padding-left: 1rem;
+          list-style: none;
+          border-left: 1px solid var(--site-outline-variant);
```
parlough, line 75: What do you think about adjusting layout to line this border up with the arrow, and the header entries with the summary? Sort of like this:

Screenshot of lining the border with the arrow

sites/www/lib/styles/pages/_blog_page.scss
```diff
@@ -33,6 +33,61 @@
...
+        padding: 1rem 1.25rem;
+      }
+
+      summary {
+        display: flex;
+        align-items: center;
+        gap: 0.25rem;
+        font-weight: var(--site-fontWeight-bold);
+        cursor: pointer;
+        list-style: none;
```
parlough, line 52: Add `user-select: none;` to the summary so that it's not unnecessarily selected when we open/close it.

```suggestion
        list-style: none;
        user-select: none;
```

packages/site_shared/lib/components/blog/blog_toc.dart
```diff
@@ -0,0 +1,56 @@
...
+/// before a table of contents is worth showing.
+const _minEntriesForToc = 2;
+
+/// Displays an automatically generated table of contents for the current
+/// page, derived from the `h2` and `h3` headings found in its rendered
+/// content.
+///
+/// Requires [TableOfContentsExtension] to be applied to the page so that
+/// its headings are available as a [TableOfContents] under the page's `toc`
+/// data.
```
parlough, line 22: Good idea calling this out! Thanks :D

packages/site_shared/lib/components/blog/blog_toc.dart
```diff
@@ -0,0 +1,56 @@
...
+      attributes: {'aria-label': 'Table of contents'},
+      [
+        Component.element(
+          tag: 'details',
+          children: [
+            const Component.element(
+              tag: 'summary',
+              children: [
+                MaterialIcon('chevron_right'),
+                Component.text('On this page'),
```
parlough: Let's go with "In this article" instead of "On this page" since this is for blog posts rather than pure technical documentation.

parlough (commented):
Thanks for this great addition @manuelzzz! Mostly looks ready to me, just some code cleanup and style suggestions to consider:

manuelzzz:
All the requested changes have been addressed. Everything should be ready for another review now.

parlough (changes requested):
Thanks for making those updates @manuelzzz! It seems some of the adjustments are still needed:

1. Adjusting the summary text "In this article"
2. Adjusting the summary font to use the `var(--site-ui-fontFamily)` font family instead.

If you think keeping them as is is better, that's fine too, just let me know what you think. Otherwise, looks good to me once those changes are made!

---

Note that I'm going to hold off on merging this PR for a few days as I'm planning to land a change to how we handle blog post hero images first. Don't worry though as I'll come back and land this once that is done. Thanks again for working on this and for your patience!

manuelzzz:
So, the corrections regarding the TOC title have been made and committed just now.

And about putting the PR on hold, don't worry about that. I'm just happy to be able to contribute to the project.

PR #13882, opened 2026-09-11
Add back-to-top button to blog posts

manuelzzz:
Adds a back-to-top button to long-form blog posts to make navigation easier.

* **Back-to-top button:** Adds `BackToTopButton` (`packages/site_shared/lib/components/common/client/back_to_top_button.dart`), a floating button that appears after the reader scrolls past the hero and smoothly scrolls back to the top when clicked. It reuses the existing `in-content` scroll-spy signal, so no additional scroll-tracking JavaScript is required.

The button is implemented as a shared `site_shared` component rather than directly in `sites/www` or `sites/docs`. For now, it is only wired into the blog layout, keeping it reusable for other long-form content in the future.

This is a companion to the table of contents PR (#13881). The changes are intentionally split so both PRs can be reviewed and landed independently; there is no code dependency between them.

https://github.com/user-attachments/assets/d8568351-40df-461d-b707-6fb5a854fc51


packages/site_shared/lib/components/common/client/back_to_top_button.dart
```diff
@@ -0,0 +1,31 @@
...
+  Component build(BuildContext _) => button(
+    classes: 'back-to-top',
+    attributes: {'aria-label': 'Back to top'},
+    events: {
+      'click': (_) {
+        web.window.scrollTo(web.ScrollToOptions(behavior: 'smooth', top: 0));
+      },
+    },
+    [const MaterialIcon('arrow_upward')],
+  );
```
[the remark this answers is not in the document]
manuelzzz: Just applied the fall back to `behavior: 'auto'` when `prefers-reduced-motion: reduce` is set, both for the click-triggered scroll and the show/hide transform/transition in CSS.

sites/www/lib/styles/pages/_blog_page.scss
```diff
@@ -106,3 +106,36 @@
...
+  color: var(--site-base-fgColor-alt);
+  background: var(--site-base-bgColor);
+  border: 1px solid var(--site-outline-variant);
+  border-radius: 50%;
+  box-shadow: var(--ui-drop-shadow-sm);
+  cursor: pointer;
+  opacity: 0;
+  transform: translateY(0.5rem);
+  transition: opacity 0.2s var(--ui-anim-func), transform 0.2s var(--ui-anim-func);
+  pointer-events: none;
```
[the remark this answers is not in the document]
manuelzzz: Seted the button's type explicitly via jaspr's `type: ButtonType.button` (the idiomatic equivalent of `type="button"` in this codebase — see `menu_toggle.dart`/`ide_explorer.dart` for the existing convention) instead of a raw `type` attribute.

sites/www/lib/styles/pages/_blog_page.scss
```diff
@@ -106,3 +106,41 @@
...
+  width: 3rem;
+  height: 3rem;
+  padding: 0;
+  color: var(--site-base-fgColor-alt);
+  background: var(--site-base-bgColor);
+  border: 1px solid var(--site-outline-variant);
+  border-radius: 50%;
+  box-shadow: var(--ui-drop-shadow-sm);
+  cursor: pointer;
+  opacity: 0;
```
parlough, line 127: Consider also setting `visibility: hidden;` here, then within `body.blog.in-content .back-to-top`, setting `visibility: visible`.

That way it's also hidden from the accessibility tree when appropriate.

packages/site_shared/lib/components/common/client/back_to_top_button.dart
```diff
@@ -0,0 +1,40 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:jaspr/dom.dart';
+import 'package:jaspr/jaspr.dart';
+import 'package:universal_web/web.dart' as web;
+
+import '../material_icon.dart';
+
+/// A floating button that scrolls the page back to the top when clicked.
```
parlough, line 11: Thanks for adding a helpful doc comment here!

packages/site_shared/lib/components/common/client/back_to_top_button.dart
```diff
@@ -0,0 +1,40 @@
...
+  const BackToTopButton({super.key});
+
+  @override
+  Component build(BuildContext _) => button(
+    type: ButtonType.button,
+    classes: 'back-to-top',
+    attributes: {'aria-label': 'Back to top'},
+    events: {
+      'click': (_) {
+        final prefersReducedMotion = web.window
```
parlough, line 29: Perhaps on click, we should also move the focus to the top of the page, so focus isn't lost for keyboard users.

One option is to update the `h1` in `blog_layout.dart` to have `tabindex="-1"` then on click, before scrolling, change the focus to there. Something like the following:

```dart
(web.document.getElementById('document-title') as web.HTMLElement?)
    ?.focus(web.FocusOptions(preventScroll: true));
```

What do you think?

parlough (commented):
Thanks for splitting this up! Mostly looks great, just two questions/suggestions around behavior for keyboard users:

packages/site_shared/lib/components/common/client/back_to_top_button.dart
manuelzzz: Sounds great! An interesting detail, I'll do that soon

manuelzzz:
All the requested changes have been made.

I was also thinking about opening an issue to turn this component into a more global one. I think a bunch of other pages could benefit from having this available as a shared component.

It might be a somewhat larger feature, but I'd be happy to work on it and, if needed, discuss the best approach for implementing it.

What do you think about this, @parlough?

parlough (approved):
Thanks for making those adjustments! Looks good to me.

parlough:
> I was also thinking about opening an issue to turn this component into a more global one. I think a bunch of other pages could benefit from having this available as a shared component.
>
> It might be a somewhat larger feature, but I'd be happy to work on it and, if needed, discuss the best approach for implementing it.

I can see it being a valuable addition on longer docs pages. If you're interested, feel free to open an issue with your thoughts and we can see what others think.

PR #13894, opened 2026-09-16
Update staging of sites to be more conditional

parlough:
Conditionally stage each site, similar to how we conditionally run the GitHub checks.

Since we use a single Cloud Build entrypoint for both sites, this needs to be done outside the config in our custom logic.


sfshaza2 (approved):
lgtm

PR #13895, opened 2026-09-16
[www] Use consistent z-index scale

parlough:
Updates flutter.dev to use a consistently defined z-index scale that matches the docs site, enabling sharing of more components and helping ensure consistency. It also fixes an issue where the blog's "back to top" button appeared over the mobile version of the navigation menu.


sfshaza2 (approved):
lgtm

